PR #130, opened 2015-02-11
Check for lengths on talmud texts as well

EliezerIsrael:


EliezerIsrael:
@blockspeiser - I wasn't sure how this snuck there.  It seems to be a no-brainer to check for length of Talmud as well.  Do you know why this is here? Any reason not to get rid of it?

blockspeiser:
Yeah, don't exactly why it isn't even attempted with Talmud.

The one caveat is that lengths for Talmud (or anything with Talmud address type) are expressed in terms of Amud (as they're stored in Mongo) and not in terms of Daf. Should be tested.

EliezerIsrael:
Good call.  The order of magnitude is right, but the check is off by 1 or 2 on each mesechet.  Working on it...

EliezerIsrael:
A few issues here:
- index_node.lengths had grown stale in a lot of places.  Added script for that.
- Now using AdressType classes so that Talmud addresses present correctly in errors.
- Added an offset of 2 for Bavli for those first two blank pages (may want to bake this into the Index, in the future.)

PR #151, opened 2015-07-15
checks for empty titles

elmehi:


blockspeiser:
Thanks! What's the reason for changing the default parameter of baselist to None, then changing it to [] if it's None? Wouldn't leaving the default at [] do the same?

EliezerIsrael:
Meir and I worked on this together. 

The empty list was not the cause of the issue we debugged here, but it is best practice to avoid empty lists as default values.  The issue is that the default value is evaluated once, and then reused for multiple function calls.  Since arrays are by reference, you could end up with a non-empty default value.

http://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

blockspeiser:
I did not know this. thanks!

PR #157, opened 2015-07-21
modified autocomplete to return only 15 titles (#3 in list of changes)

elmehi:
Please make sure I did this right. This is my first real pull request. :)

### DIFF
--- a/templates/js/headers.js
+++ b/templates/js/headers.js
@@ -569,7 +569,7 @@
 							return tag;
 						}
 					});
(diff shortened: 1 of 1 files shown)


blockspeiser:
Congrats on your first pull request! Looks good, seems performant. We increased the threshold to 30, just so that in the case where we have a lot of results the list spans off the screen (some not excessively long screens) so that we give more of the impression of abundance than comprehensiveness.

PR #162, opened 2015-08-03
API example with full navigation

dakom:


blockspeiser:
Thanks for this! I moved it into /static/examples/

PR #174, opened 2015-09-06
complex text export

nsantacruz:


EliezerIsrael:
Looks good.
A couple things -
1) Generally, we avoid 'from x import *'. It makes for name conflicts that are hard to debug.

The exception for us is the package sefaria.model, where we define the namespace specifically in /model/__init__.py. In general, anything you need from model.text should be exported via sefaria.model. 

2) are you using spaces, rather than tabs? Spaces is our standard. Hard for me to tell.

3) you restrict export to just Orot. Probably a leftover from testing.

Fix these up, and we'll merge it in.

betamidrash:
Yeah, those were pretty silly errors. I've pushed the changes. Tell if I'm using spaces now. You're right that it's hard to tell in PyCharm...

betamidrash:
also, it seems like I'm logged in as betamidrash now. That's interesting

PR #192, opened 2016-03-02
Update link to exported data

cmrberry:
The link on the current README to the exported data set points to a repo that's no longer there. Updating that link to point to `Sefaria/Sefaria-Project` instead.


blockspeiser:
thanks!

PR #220, opened 2016-05-29
remove temp code for backend migration from subsources

pzp1997:
Noticed this as I was working on sheets/views.py. Not sure what the exact implications of removing this snippet of code would be, but I noticed it is long past the removal date stated in the comment regarding it. Should the code be removed? Should the comment be updated to reflect its more permanent nature now? I believe @rneiss was the person who added it, so maybe it would be good for him to weigh-in on this.


rneiss:
Yes, code should be removed.

PR #226, opened 2016-06-27
added method to Ref to return a ranged segmented ref with tests

JonMosenkis:


EliezerIsrael:
Looks good.  
A few things:

- Add a test where the input is a range and the expected ending ref has segment > 1. 
  
- Add a test where input is a ref to a complex text and a test where ref is a commentary on complex text. Haggadah is decent candidate.  
  
- Do we expect it to work for refs to top level Jagged Array nodes, e.g. "Pesach Haggadah, Yachatz"? I'm guessing yes. Either way, we should document the limitations. 
    
- Do we expect this to work for refs to Schema nodes, e.g. "Pesach Haggadah, Magid"?   I'm guessing no. Either way, we should document the limitations. 
  
- Don't need the `abs()` on 2349, correct?
- `get_state_ja()` is an an expensive operation.  Best to move it out of the loop. 
- `temp_ref` needs a better name.  `current_ending_ref` perhaps?

JonMosenkis:
get_state_ja() needs to be in the loop because the size of the jagged array is needed to find the ending ref at each level.

EliezerIsrael:
`get_state_ja()` returns the same value on each call.  You can store it as a variable, and then call `sub_array_length()` on the stored value.

Issue #228, opened 2016-07-01
Source Sheet outsideText Hebrew is not displayed correctly


pzp1997:
When adding an outsideText in Hebrew, the alignment and direction of the text are displayed incorrectly. This is because it is not being wrapped in a `<div class="he"></div>`, which gives its enclosing text the styles: `direction: rtl; text-align: right;`. A quick fix would be to just append this tag to the outsideText when the user adds it to the sheet.

A better solution for the long run would probably be to create an `outsideHeText` and `outsideEnText` (like `outsideBiText`) source "type" (I use that word loosely) and then handle each one separately within `templates/sheet.html`. I believe that naively doing that, however, would involve quite a bit of work to ensure that nothing that relies on `outsideText` breaks. Additionally, legacy support would need to be maintained for the old `outsideText` attribute until all of the existing sheets are updated to fit the new model.

This isn't my main focus at the moment, but I'll see if I can implement some version of the quick fix in the morning.

PR #229, opened 2016-07-01
hotfix alignment and direction of an outsideText that is in Hebrew

pzp1997:
Closes issue #228.

### DIFF
--- a/static/js/sheets.js
+++ b/static/js/sheets.js
@@ -1327,8 +1327,9 @@ $(function() {
 			$("#addcustomTextDiv").on("click", ".button", function (e) {
 				var $target = $("#addInterface").prev(".sheetItem");
 				if ($(e.target).prev(".flexContainer").find(".contentToAdd:visible").length == 1) {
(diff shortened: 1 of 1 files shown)


rneiss:
This seems fine for adding -- but gets overridden immediately upon editing the text in ckeditor. You want to be adding the class to the parent here instead.

Best solution may be calling isHebrew() on the text before adding it in build source. This will be backwards compatible as well, though that function may be mucked up by all the extra inline styles markup from ckeditor if there's quite a bit of it.

https://github.com/Sefaria/Sefaria-Project/blob/fc788fc7038e3dd039234502fadace27b7fd451c/static/js/sheets.js#L2736

pzp1997:
Rewrote it to use your solution. I believe I've sufficiently dealt with your concern about the extra inline styles by calling the [`String.prototype.stripHtml`](https://github.com/Sefaria/Sefaria-Project/blob/master/static/js/util.js#L2680-L2684) method of `source.outsideText` to safely remove all of the HTML tags before passing it to `isHebrew`.

PR #235, opened 2016-07-12
Closes #233. Upload locally stored photos to Source Sheet

### LINKED ISSUE #233: Upload Photo to Source Sheets
Uploading photo from local file system to source sheet. 

I was interested in pasting an image from my file system and add it to a source sheet. I noticed one can only paste from a URL, Youtube video or MP3.

It would be great to be able to upload a photo locally.
If there is a way to do this already, feel free to close issue.

pzp1997:
I decided to take a stab at this issue tonight. My solution was to upload the image to Imgur using the [Imgur API](https://api.imgur.com/) and then, insert the direct link to the image that the API returns into [`inlineAddMediaInput`](https://github.com/Sefaria/Sefaria-Project/blob/master/templates/s2_sheets.html#L358) and trigger a click event on its respective `.button` using jQuery. I'll admit, it's a little bit hacky, but it was working great on JPGs, PNGs, and GIFs when I manually tested it. I would recommend cleaning up the CSS a little bit though. I was having some difficulty styling the `input[type=file]`.

### DIFF
--- a/static/css/sheets.css
+++ b/static/css/sheets.css
@@ -1688,7 +1688,7 @@ img {
 	display:none;
 }
 
-#inlineAdd {
+.inlineAddTextInput {
 	width: 320px;
 	height: 32px;
 	margin: 0;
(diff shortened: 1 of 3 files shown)


pzp1997:
I just cleaned up the CSS styles to make it fit in with the rest of the site. I think this is good to go now.

PR #246, opened 2016-09-09
Use labels on edit_text, for easier selecting

YairRand:
Use label elements instead of plain spans on the "Original Translation"/"Copied Text" options, so that the user can click anywhere on the option instead of just specifically on the radio.

### DIFF
--- a/static/css/reader.css
+++ b/static/css/reader.css
@@ -256,7 +256,7 @@ body.hebrew #about .he {
 	display: inline-block;
 }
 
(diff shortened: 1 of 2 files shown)


blockspeiser:
Sorry for the long delay on this PR. Change looks good after fixing the fontSize of labels to what it was previously. I've merged this into our `test` branch. Will be in test for a little while then merged into master and deployed. Thanks!

PR #257, opened 2016-10-19
Three Functions

stevekaplan123:
Fixed text_index_map and added all_segment_refs and all_section_refs to Ref class


sefaria/model/schema.py
```diff
@@ -361,16 +361,22 @@ def is_flat(self):
...
         """
-        callback(self, **kwargs)
+        if preorder:
+            callback(self, **kwargs)
         for child in self.children:
-            child.traverse_to_string(callback, **kwargs)
+            child.traverse_tree(callback, preorder, **kwargs)
+            if not preorder:
+                print child
+                callback(self, **kwargs)
```
EliezerIsrael: This will invoke `callback(self)` once for every child.

sefaria/model/schema.py
```diff
@@ -361,16 +361,22 @@ def is_flat(self):
...
         return not self.parent and not self.children
 
-    def traverse_tree(self, callback, **kwargs):
+
+    def traverse_tree(self, callback, preorder, **kwargs):
         """
         Traverse tree, invoking callback at each node, with kwargs as arguments
         :param callback:
         :param kwargs:
         :return:
```
EliezerIsrael: Document what `preorder` does.

sefaria/model/text.py
```diff
@@ -1395,19 +1395,26 @@ def version(self):
 
     def text_index_map(self,tokenizer):
```
EliezerIsrael: We need some tests for this, depth 1,2, & 3, with variations of spacing, paragraphs, etc.

sefaria/model/text.py
```diff
@@ -2782,6 +2789,48 @@ def subrefs(self, length):
...
 
+    def all_segment_refs(self):
+        seg_refs = []
+        if self.is_book_level():
+            return self.index.all_segment_refs()
+
+        sections = self.all_section_refs() if self.is_section_level() == False else [self]
+
+        for sec_ref in sections:
+            seg_refs += sec_ref.all_subrefs()
```
EliezerIsrael: This will return all *possible* segment refs, not all *populated* segment refs.  Is that the intention?  In any case, it should be documented.

EliezerIsrael: And it needs tests.

sefaria/model/text.py
```diff
@@ -2782,6 +2789,48 @@ def subrefs(self, length):
...
+                        "type": self.index.categories[0],
+                        "index_node": self.index_node,
+                        "sections": sections,
+                        "toSections": sections
+                        }
+                    )]
+        except Exception as e:
+            logger.warning(u"Failed to generate references for {}, section {}. {}".format(c.full_title("en"), ".".join([str(s) for s in sections]) if sections else "-", e.message))
+        return refs
+
```
EliezerIsrael: An easier and cleaner way to implement this method:

```
sub_ja = self.get_state_ja().subarray_with_ref(self)
return [self.subref(i+1) for k in sub_ja.non_empty_sections() for i in k]
```

sefaria/helper/schema.py
```diff
@@ -269,30 +270,45 @@ def change_node_title(snode, old_title, lang, new_title):
     pass
 
 
-def replaceBadNodeTitles(title, bad_char, good_char, lang):
+def change_char_node_titles(index_title, bad_char, good_char, lang):
+    '''
```
EliezerIsrael: We need tests for this function.  See the tests for other functions in this module.

PR #284, opened 2017-01-23
cascade without print statements

stevekaplan123:


sefaria/helper/schema.py
```diff
@@ -617,10 +618,37 @@ def construct_query(attribute, queries):
...
+        try:
+            return Ref(ref).index.title == title
+        except InputError:
+            return False
+
+    def rewriter(ref):
+        ref = Ref(ref)
+        for old_ref in mappings:
+            old_ref = Ref(old_ref)
+            if old_ref.contains(ref):
```
EliezerIsrael: I don't think this will correctly handle ranged references.  
If the original refs is of the form "Text Name 1:2-4" for example.

Some of these will be in one jagged array in the target schema, and thus convertible.
Some will bridge over two schema nodes in the target, and will need to be truncated (or deleted.)

PR #290, opened 2017-02-05
update export links to divide files for max git file size

herzberg:
Separates links into smaller files so that git will accept the files (it won't let files above 100MB).


EliezerIsrael:
@herzberg Please revise and resubmit, as per comment on companion PR.

PR #292, opened 2017-02-05
Export talmud converter

herzberg:


blockspeiser:
Splitting links into files looks good. 

I'm not sure if we should be converting the Talmud tags to <b> and <i> at this level though. Josh, can we be doing in the process that converts to andoird SQL instead?

This export is going to be the master copy of the reusable data for the world. If we think it's important not to call these <b> and <i> in our own data, presumably we want that to be the case for exported data too. @EliezerIsrael what do you think?

EliezerIsrael:
@herzberg The data now has <b> and <i> tags, instead of the <span class"..."> tags.  
Is there anything else in this PR that should be saved?  Want to revise and resubmit?

PR #306, opened 2017-02-27
new footnote i tag change and migrate simple to complex ranged refs

stevekaplan123:
--added to the "migrate simple to complex function" the ability to deal with ranged references
--also it can validate that mapping references do not overlap with one another
--finally added 'class' as an allowed attribute for i-tags for the new footnote feature


sefaria/helper/schema.py
```diff
@@ -653,35 +654,52 @@ def needs_rewrite(ref, *args):
...
+        for our_ref in [our_ref.starting_ref(), our_ref.ending_ref()]:
+            for map_ref in mappings:
+                map_ref = Ref(map_ref)
+                print "******"
+                print map_ref
+                print our_ref
+                try:
+                    terms = our_ref.in_terms_of(map_ref)
+                except:
+                    our_ref.in_terms_of(map_ref)
```
EliezerIsrael: This looks like a mistake.  
First - the except clause is over general - what exception are you expecting?  
Second, you're catching an exception and then calling the exact same line again - guaranteeing that you'll cause the exception to be raised upward - making the whole try/except redundant.

sefaria/helper/schema.py
```diff
@@ -653,35 +654,52 @@ def needs_rewrite(ref, *args):
...
+                if map_ref.contains(our_ref) and type(terms) is list:
```
EliezerIsrael: I'm having trouble following the new logic, below.  Can you document?

PR #327, opened 2017-05-02
Addition of alt-names

Cawhee:
Addition of alt-names used throughout current version of [Duties of The
Heart](https://www.sefaria.org/Duties_of_the_Heart)

Should resolve problems seen in #326


blockspeiser:
Thanks for this. I'm not going to accept this pull request (since the data won't actually flow from here) but I am going to leave it open for now so we can use it as a guide for someone on our team to add them to the right place.

EliezerIsrael:
I have added all of these except `Melachim`, which was too ambiguous without the 1/2 indicator. 
Thank you!

PR #328, opened 2017-05-07
Double links update and spreadsheet generator for QA

stevekaplan123:
gen_spreadsheet.py generates a CSV with information on a given commentator on all books they comment on -- for each book, the word count, link count per segment, and several other metrics to assist QA process.

### DIFF
--- a/scripts/double_links.py
+++ b/scripts/double_links.py
@@ -14,7 +14,14 @@ class DoubleLinks:
         self.general_auto_specific_manual = 0
         self.specific_auto_general_manual = 0
         self.both_manual = 0
-        self.bad_ids = {}
+        self.tanakh = 0
+        self.tanakh_only = False
+        self.other = 0
+        self.duplicate_ids = {}
+
+
+    def set_tanakh(self, bool):
+        self.tanakh_only = bool
 
 
     def get_doubles_for_delete(self, ref1, ref2):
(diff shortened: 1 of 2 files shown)


scripts/double_links.py
```diff
@@ -119,13 +119,59 @@ def delete_links_and_output_results(self):
...
+        ls = LinkSet()
+        link_refs = {}
+        count = 0
+        for l in ls:
+            refs = tuple(sorted(l.refs))
+            if refs in link_refs:
+                count += 1
+                if refs not in self.bad_ids:
+                    prev = link_refs[refs]
+                    self.bad_ids[refs] = [l._id, prev]
```
EliezerIsrael: Some links are more preferable than others.  If it has `source_text_oid`, it's preferable.  If not that, presence of `auto` or `generated_by` makes them preferable.

scripts/double_links.py
```diff
@@ -119,13 +119,59 @@ def delete_links_and_output_results(self):
...
+        total = 0
+        max_num = 0
+        which_one = []
+        deleted_num = 0
+        for each in self.bad_ids:
+            ids = self.bad_ids[each]
+            if len(ids) > max_num:
+                max_num = len(ids)
+                which_one = ids[0]
+            for id in ids[1:]:
```
EliezerIsrael: The slice here is misplaced.  All bad ids need to go.

scripts/double_links.py
```diff
@@ -152,7 +196,3 @@ def delete_links_and_output_results(self):
                 assert ref1.primary_category == ref2.primary_category == category
```
EliezerIsrael: These could also be candidates.

scripts/double_links.py
```diff
@@ -135,12 +181,10 @@ def delete_links_and_output_results(self):
         try:
             ref1 = Ref(ref1)
         except InputError:
```
EliezerIsrael: If a ref is bad, just delete the link object.

PR #337, opened 2017-06-28
makes lang parameter more resilient

nassan:
I had pulled down the repo to make a different pull request, and I noticed that the `lang` parameter being passed into `reader` was `None`, instead of what I think would be theoretically `'he'` or `'en'`. 
I don't know the code base that well, just dove into it today. Maybe this is what is supposed to be happening.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -60,6 +60,12 @@ logger = logging.getLogger(__name__)
 
 @ensure_csrf_cookie
 def reader(request, tref, lang=None, version=None):
+
+    # if lang parameter has not been set
+    # then pull it off of the request query parameters
+    if lang is None:
+        lang =  request.GET.get('lang', None)
+
     # Redirect to standard URLs
     def reader_redirect(uref, lang, version):
         url = "/" + uref


edamboritz:
Hi, thanks. Were you noticing a bug in the UI related to this? 
Otherwise, the code actually is as it should be. The lang parameter in that view function should indeed default to None.

nassan:
Well, I just wanted the title to reflect the language, so that when I share a link in Whatsapp with a Hebrew-Speaker the site preview will render the correct language. So I was trying to figure out a way to know within the backend what the language is set to

edamboritz:
Ah, I understand.  
Currently the web preview is almost always in English and is not governed by that parameter anyway. 
You'll also notice that when you browse to a text, that parameter is already being given a default value if not passed in explicitly. You can see it in the URL. Try browsing to https://www.sefaria.org/Genesis.1.2 as an example. 
We do however have in mind already to add proper Hebrew previews for Hebrew oriented links. 

Thanks a lot!

nassan:
Ah okay good! I share links in Whatsapp and Telegram with friends that only know Hebrew, and they always get upset when the site preview that pops up for them is english title and description. So I thought I would look into how hard it would be to change. Really glad to hear it is in the pipeline.  :)

edamboritz:
We appreciate the participation and keep on sharing! :-)

nassan:
How IS the language tracked on the backend, if that `lang`is supposed to be `None`?

edamboritz:
It's a little complicated, but I will attempt to clarify:
We have several parameters to indicate what language the user will see on screen when loading a text in the browser. 
The lang parameter you were attempting to edit serves as part of the url path that determines the version (of sometimes multiple choices available on Sefaria) of text to return to the user. It must be coupled with another parameter - `version` to choose a version title. Sefaria currently always loads both an English and a Hebrew version of the text (when available) to return to the browser. 
The parameter I mentioned earlier that the user can see in the URL query portion is one of a few mechanisms that determines which of the two versions that the browser loaded will actually show up on screen. It has 3 values: `he`, `en` and `bi` for bilingual text display (Of course, all of these can also be controlled from the language icon on the text itself).
We already have methods of determining the default values for any of these, so the user does not have to enter them repeatedly. 
Hope that explanation helps.

nassan:
Thanks :)

PR #352, opened 2017-09-10
all_subrefs() now returns only existent refs 

stevekaplan123:
all_subrefs() was returning non-existent refs along with the existent refs.  
For example, Ref("Tamid").all_subrefs return array started at "Tamid 1a" instead of "Tamid 25b" where it should start.


EliezerIsrael:
Need to separate out the various changes here in to separate pull requests.

PR #354, opened 2017-09-12
only activate lexicon box if selected text contains hebrew

### LINKED ISSUE #346: Dictionary lookup is called on English words
Double-clicking on a word in an English work or an English translation of a Hebrew work results in a dictionary lookup appearing, with no results (as the dictionary doesn't go in that direction).

alis0nc:
Should fix #346


blockspeiser:
Thank you @alis0nc!

We'll this is now in test, hopefully will deploy soon.

We've had some aspirations of incorporating an English dictionary, which but until we actually have data this is sensible patch.

PR #355, opened 2017-09-27
all_subrefs() now returns only existent refs

stevekaplan123:
A trello card mentions that Ref("Tamid").all_subrefs() didn't work as expected.
As of this commit, all_subrefs() still passes the old test in test_all_subrefs() and now passes two new tests I based on the bug mentioned in the trello card.


JonMosenkis:
I would like to just make a note that I'm not sure we should be changing the behavior of `all_subrefs`. Some cases may require getting a ref even if there isn't any text. An example would be a commentator that has no comments for a particular chapter. I would expect the results of `all_subrefs` to return the same result in both the commentator and the base text, but it seems that this would no longer be the case.

stevekaplan123:
I guess it depends on what is the intended usage of the function.  I now see that my code has a mistake anyway.

Issue #351, opened 2017-09-04
"Version Settings" and "File Format" aren't translated


YairRand:
In a work's Table of Contents, at the bottom of the "Versions" ( / "גרסאות" ) section, the texts "[Version Settings](https://github.com/Sefaria/Sefaria-Project/blob/585935056248a536cccc019d96361f201c906c84/static/js/ReaderTextTableOfContents.jsx#L199)" and "File Format" don't get translated to Hebrew when the user's site language is set to Hebrew.

Related issue: The hovertext for the search box is also always in English.

jaykaron:
Also, it seems "Version Settings" and "File Format" have already been translated.

PR #357, opened 2017-10-24
Added Hebrew Translation Tooltip to Search Box

jaykaron:
Unfortunately, this code has not been tested. Also, no promises that the hebrew is 100% accurate.
In response to issue #351.


edamboritz:
Hi,
Thanks for noticing this. 
We're currently using a slightly modified way of translating strings like these, so I'm not directly merging in your PR. 
https://github.com/Sefaria/Sefaria-Project/commit/3e76a621af98ebee01997a2da5ee7b1ffb03a93f

PR #359, opened 2017-11-09
I-tags now in English and Hebrew

stevekaplan123:


JonMosenkis:
I'm concerned that this works for the main reader panel, but may look weird in the sidebar. Is this code live somewhere that we can take a look?

stevekaplan123:
http://ste.sefaria.org/Rashi_on_Leviticus

EliezerIsrael:
It looks good for this case, where the comments don't have any numbering.  I'm wondering how it works in the case where comments are numbered / have unique identifiers.  

If that's not handled yet, let's evaluate the cost of doing so, and decide if we want to implement it at this time.  If it's not a lot of work, let's make this work for the general case.

PR #375, opened 2018-01-08
Or Neerav: Moves footnotes from Footnotes node into the main text

stevekaplan123:


EliezerIsrael:
@stevekaplan123 Looks like we still need to clean up the old footnotes on the bottom of sections.

PR #376, opened 2018-01-09
fix 9 yad ramah source sheets

stevekaplan123:


EliezerIsrael:
@stevekaplan123 You probably want to check in the data file as well.

PR #388, opened 2018-03-14
Test

shaneerosen:
address type changes to catch many more citations (changes mainly in Halakhah catching, Mishneh Torah, Tur and Shulchan Arukh)

### DIFF
--- a/scripts/change_address_types.py
+++ b/scripts/change_address_types.py
@@ -1,0 +1,47 @@
+# encoding=utf-8
+
+from sefaria.helper.schema import *
+
+def change_mishneh_torah():
+    '''
+    changes addressTypes of all Mishneh Torah indexs to ['Perek' , 'Halkhah']
+    :return:
+    '''
+
+    yad_list = library.get_indexes_in_category(u"Mishneh Torah")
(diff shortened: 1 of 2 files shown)


scripts/change_address_types.py
```diff
@@ -0,0 +1,47 @@
...
+    :return:
+    '''
+
+    yad_list = library.get_indexes_in_category(u"Mishneh Torah")
+    schema_yad_dict = {}
+    for title in yad_list:
+        schema_yad_dict[title] = library.get_schema_node(title)
+
+    for node in schema_yad_dict.values():
+        if len(node.sectionNames) != 2:
```
nsantacruz: what are you checking here? should depth one nodes get an address type of just 'Perek'? I don't know of examples

scripts/change_address_types.py
```diff
@@ -0,0 +1,47 @@
+# encoding=utf-8
+
+from sefaria.helper.schema import *
```
nsantacruz: In the future, it's better to explicitly say which functions you need because they can potentially overwrite variables in your script when you import with `*`

PR #408, opened 2018-05-24
Auto linking tests can pass now but find_broken_links must be run first

stevekaplan123:
Run these commands from Sefaria-Project root directory:
./run scripts/find_broken_links.py -e -a -d -m -r "Rashi on Genesis"
./run scripts/find_broken_links.py -e -a -d -m -r "Kos Shel Eliyahu on Pesach Haggadah"


sefaria/clean.py
```diff
@@ -1,6 +1,8 @@
 """
 Small utilities for fixing problems that occur in the DB.
 """
+import django
```
EliezerIsrael: We don't want to import these in a library.

PR #409, opened 2018-05-24
Address year

stevekaplan123:


static/js/ReaderTextTableOfContents.jsx
```diff
@@ -810,8 +810,13 @@ class JaggedArrayNodeSection extends Component {
         if (this.contentCountIsEmpty(this.props.contentCounts[i])) { continue; }
         if (this.props.addressTypes[0] === "Talmud") {
           var enSection = Sefaria.hebrew.intToDaf(i);
-          var heSection = Sefaria.hebrew.encodeHebrewDaf(enSection);
-        } else {
+          var heSection = Sefaria.hebrew.encodeHebrewDaf(section);
```
EliezerIsrael: Let's see if we can do this using the data from the API.  It may not be practical, but I'd like to avoid special casing here.

sefaria/model/schema.py
```diff
@@ -1588,6 +1588,25 @@ def toNumber(self, lang, s):
         elif lang == "he":
             return decode_hebrew_numeral(s)
 
+class AddressYear(AddressInteger):
+    """
+    :class: AddressYear stores Hebrew years as numbers, for example 778 for the year תשע״ח
+    To convert to Roman year, add 1240
+    """
+    @classmethod
```
EliezerIsrael: Shouldn't be a @classmethod.

PR #410, opened 2018-05-27
Auto linker default node

stevekaplan123:


sefaria/helper/link.py
```diff
@@ -99,7 +99,7 @@ def _save_link(self, tref, base_tref, **kwargs):
             else:
                 tracker.add(self._user, Link, nlink, **kwargs)
         except DuplicateRecordError as e:
-            pass
+            print e.message
```
EliezerIsrael: In production code, we use `logger.warning()` or `logger.error()`

EliezerIsrael: But I'm worried about this being too chatty and filling up the logs.  Duplicate links come up all the time.

PR #411, opened 2018-05-30
Adds Talmudic Relationships visualization

saengel:


reader/views.py
```diff
@@ -1743,6 +1743,9 @@ def visualize_links_through_rashi(request):
     json_file = "../static/files/torah_rashi_torah.json" if level == 1 else "../static/files/tanach_rashi_tanach.json"
     return render(request,'visualize_links_through_rashi.html', {"json_file": json_file})
 
+def talmudic_relationships(request):
+    json_file = "../static/files/talmudic_relationships_data.json"
```
EliezerIsrael: This data is available from our data store. 
Best practice would be to load it from there.  

See the classes in `sefaria/model/person.py` and the views `reader.views.person_index()` and `reader.views.person_page()`.

templates/talmudic_relationships.html
```diff
@@ -0,0 +1,733 @@
...
+
+                .attr("d", d3.svg.symbol()
+                    .size(function(d) {
+                        return Math.PI * Math.pow(size(d.size) || nominal_base_node_size, 2);
+                    })
+                    .type(function(d) {
+                        return d.type;
+                    }))
+
+                .style(tocolor, function(d) {
```
EliezerIsrael: This could be written in a more graceful way. 
Take a look at https://github.com/d3/d3-scale/blob/master/README.md

saengel:
Thank you so much for the encouragement and the feedback - going to look
through the notes asap :)

PR #412, opened 2018-05-31
Shenei Luchot HaBerit: fix source sheets' refs

stevekaplan123:
fixes source sheet refs based on match.csv


stevekaplan123:
Rename it and put inside scripts or put it inside code.

PR #414, opened 2018-06-08
Fix english talmud ref wrapping

wolfkevin:
See Berachot 1:5 in this [text](https://www.sefaria.org/Maaseh_Nissim_on_Pesach_Haggadah%2C_Magid%2C_Story_of_the_Five_Rabbis.2.1?ven=Rabbi_Mark_Greenspan&lang=en) and it incorrectly resolves to Talmud

Also other fixes from master
@EliezerIsrael

### DIFF
--- a/sefaria/model/__init__.py
+++ b/sefaria/model/__init__.py
@@ -10,8 +10,6 @@ symbols are then accessed directly as, e.g.:
 """
 
 import abstract
-import django
-django.setup()
 
 # not sure why we have to do this now - it wasn't previously required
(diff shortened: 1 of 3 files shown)


sefaria/model/__init__.py
```diff
@@ -10,6 +10,8 @@
 """
 
 import abstract
+import django
```
EliezerIsrael: Did you intend to put this here?  
I don't think that we want this at the model layer.

EliezerIsrael:
Looks Good!

PR #422, opened 2018-06-26
Not clean but displays all data, to be fine-tuned and moved

wolfkevin:
@EliezerIsrael 

don't like it and its not in the right place, but displays data (too sparsely) and don't want to get ahead of myself here before discussing a bit more


EliezerIsrael:
I'm merging now, since this helps development a lot.
Still - I think this logic belong on the `LexiconEntry` model, and not on the `DictionaryEntryNode`.

PR #429, opened 2018-07-19
Synthdict

wolfkevin:
@EliezerIsrael 
- add Jastrow ref type
- better formatting for headwords
- small segments trigger load


sefaria/model/lexicon.py
```diff
@@ -173,9 +174,17 @@ def get_sense(self, sense):
     
     def as_strings(self):
         new_content = u""
-
-        next_line = u', '.join([u'<strong dir="rtl">{}</strong>'.format(hw) for hw in [self.headword] + getattr(self, 'alt_headwords', [])])
-
+        next_line = u""
+        
+        for hw in [self.headword] + getattr(self, 'alt_headwords', []):
```
wolfkevin: possibly a cleaner way to do this, but old way wasn't loading e.g. http://localhost:8000/Jastrow%2C_%D7%90%D7%95%D6%BC%D7%91%D6%B0%D7%9C%D6%B8%D7%90_II?lang=en well

PR #431, opened 2018-08-30
Initial pass at add multi lookup

wolfkevin:
@EliezerIsrael


EliezerIsrael:
I took the intention here, and implemented it in a bit of a different way in the latest commit to this branch.  Good work figuring out this limitation.

PR #444, opened 2018-10-22
needs to be run both locally and on production/sandbox

stevekaplan123:
in order that spellings of names be consistent


EliezerIsrael:
@stevekaplan123 Some of these changes seem to be patching things that should be fixed systemically.  Mainly - why does Mesech Chochma use titles instead of terms for Parasha name?

PR #450, opened 2018-11-16
Try to fix () for sof & reish

wolfkevin:


wolfkevin:
@EliezerIsrael 

I put the `(sof|reish)` inside the `(?:to)` and converted it all to unicode so if there are unbalanced () issues it shouldn't be from this

PR #483, opened 2019-07-04
Added parashot sharedTitle checker to validate()

nmrapoport:
To prevent people from trying to add new indices to the Sefaria library that contain titles of Parashot not using sharedTitles in the future, I added a checker to the validate function that iterates through the indices' nodes' titles and raises an error if any of the titles are Parashot that aren't using sharedTitles.


nmrapoport:
Found bug. Will recommit and send new pull request.

PR #485, opened 2019-07-10
get_alt_struct_nodes, remove_shared_term, index validation

nmrapoport:
I added a function to return all the ArrayMap and JaggedArray nodes of a given index's altstructs (this is used in the replacer scripts in Sefaria-Data/sources/Scripts). I added a function to remove a sharedTerm from a node that was incorrectly inserted. I also fixed a bug in the parasha/sharedTerm section of the index validation.


nmrapoport:
I also refactored the get_alt_struct_nodes function to make it simpler and recursive.

nmrapoport:
I fixed a bug in get_alt_struct_nodes and I allowed for Onkelos indices to show aliyot headers and toggle the rendering of the aliyot headers.

PR #490, opened 2019-07-28
Setting #inlineTextPreview z-index to auto

### LINKED ISSUE #487: inlineTextPreview drawing over textBrowser
on the [new sheet page](https://www.sefaria.org/sheets/new), `#includeTextPreview` is drawing over `#textBrowser`:

bug

`z-index` for `#textBrowser` is set to `20` here:
https://github.com/Sefaria/Sefaria-Project/blob/94e8058322081a83993cbdc12e5db5734626a900/static/css/common.css#L433-L436

But `z-index` for `#inlineTextPreview` is set to `101` here:
https://github.com/Sefaria/Sefaria-Project/blob/94e8058322081a83993cbdc12e5db5734626a900/static/css/sheets.css#L429-L438

Should `z-index` for `#textBrowser` get bumped up to be higher than `101`?

orenfromberg:
Fixes #487 

This branch updates the z-index of `#inlineTextPreview` to auto so that the stacking level is the same as its parent and `#textBrowser` renders correctly over it.

Note: I only tested this fix with firefox dev tools

All feedback welcome, especially whether this approach makes sense or whether another approach would be appropriate.

### DIFF
--- a/static/css/sheets.css
+++ b/static/css/sheets.css
@@ -432,7 +432,7 @@ body:not(.user-is-tabbing) #sheet .cke_editable:focus {
 	background: white;
 	border: 1px solid #ddd;
 	border-radius: 3px;
-	z-index: 101;
+	z-index: auto;
 	width: 600px;
 	display: none;
 }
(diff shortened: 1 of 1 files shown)


orenfromberg:
closing to resubmit PR

PR #497, opened 2019-10-07
Add an application healthcheck API endpoint 

LorenzoFernando:


LorenzoFernando:
I've confirmed that this works locally.

PR #499, opened 2019-11-07
Update README.mkd

redapemusic35:
Delete extra “values” and “with”.


redapemusic35:
Not sure if this was a typo or if this fixes it.

PR #523, opened 2020-02-03
P3 topics

shaneerosen:
expandedRefs should be added automatically but the code is commented out because it uses Ref to get the info. 
RefTopicLink adds class type automatically 
validFrom and validTo contain valid topic slugs if exist in new TopicLinkType documents


sefaria/model/topic.py
```diff
@@ -189,6 +189,9 @@ class IntraTopicLink(abst.AbstractMongoRecord):
     optional_attrs = TopicLinkHelper.optional_attrs
     valid_links = []
 
+    def _normalize(self):
+        self["class"] = "intraTopic"
```
nsantacruz: I don't think you can use `[]` for `self` 
```suggestion
        setattr(self, "class", "intraTopic")
```

sefaria/model/topic.py
```diff
@@ -239,12 +242,17 @@ def _validate(self):
...
 class RefTopicLink(abst.AbstractMongoRecord):
     collection = TopicLinkHelper.collection
-    required_attrs = TopicLinkHelper.required_attrs + ['ref', 'expandedRefs', 'is_sheet']  # is_sheet attr is added automatically
+    required_attrs = TopicLinkHelper.required_attrs + ['ref', 'expandedRefs', 'is_sheet']  # is_sheet  and expandedRef attrs are defaulted automatically in normalize
     optional_attrs = TopicLinkHelper.optional_attrs + ['text']
 
     def _normalize(self):
         super(RefTopicLink, self)._normalize()
         self.is_sheet = bool(re.search("Sheet \d+$", self.ref))
+        self["class"] = "refTopic"
```
nsantacruz: same as above
```suggestion
        setattr(self, "class", "refTopic")
```

sefaria/model/topic.py
```diff
@@ -313,6 +321,19 @@ class TopicLinkType(abst.AbstractMongoRecord):
     related_type = 'related-to'
     isa_type = 'is-a'
 
+    def _validate(self):
+        super(TopicLinkType, self)._validate()
+        # Check that validFrom and validTo contain valid topic slugs if exist
+
+        validToList = self.validTo
```
nsantacruz: can be combined with line below

PR #524, opened 2020-02-06
resolve sheets race condition

JonMosenkis:
Added code to get a new sheet id when a second new sheet grabbed the same id. The issue will only come up if the sheet "id" index is set to unique (right now it's not).

This method involves an "infinite loop" so I just wanted to get another pair of eyes here. Also, Russel, I changed some code in rebuild_sheet_nodes that allows for a sheet not to have an id when this method is run. The only usage of sheet ids in this method is for printing error messages, so I imagine this not to be a critical issue. But I'm bringing your attention to this change nonetheless.

I've tested this locally. The development django server does allow for parallel processing of requests, so this race condition can come up under heavy load. I've confirmed that on occasion, a duplicate id will get created. If the id index is unique, mongo flags this successfully and throws an error. This code handles that error and generates a new id.

### DIFF
--- a/sefaria/sheets.py
+++ b/sefaria/sheets.py
@@ -435,7 +435,7 @@ def save_sheet(sheet, user_id, search_override=False, rebuild_nodes=False):
 				pass
 
 	else:
-		db.sheets.update({"id": sheet["id"]}, sheet, False, False)
+		db.sheets.update_one({"id": sheet["id"]}, sheet)
 
 	if "tags" in sheet:
 		update_sheet_tags(sheet["id"], sheet["tags"])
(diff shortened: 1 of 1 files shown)


sefaria/sheets.py
```diff
@@ -414,7 +423,19 @@ def save_sheet(sheet, user_id, search_override=False, rebuild_nodes=False):
...
+		while True:
+			try:
+				sheet["id"] = next_sheet_id()
+				db.sheets.insert_one(sheet)
+				break
+			except DuplicateKeyError:
+				pass
+
+	else:
+		db.sheets.update({"id": sheet["id"]}, sheet, False, False)
```
EliezerIsrael: `update` is deprecated.  
```suggestion
		db.sheets.update_one({"id": sheet["id"]}, sheet)
```

PR #532, opened 2020-03-19
Add an optional timezone parameter to /api/calendars

ronshapiro:
Discussions:
  - https://groups.google.com/forum/?fromgroups#!topic/sefaria-dev/r-MjlKAr8z8
  - https://groups.google.com/forum/?fromgroups#!topic/sefaria-dev/P31bOJ0LIcc\

I don't have a local dev environment set up, so I wasn't able to test things, but I tried within a repl to make sure that this was correct. If someone that does have a dev setup could validate, that would be great.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -11,6 +11,7 @@ from bson.json_util import dumps
 import socket
 import bleach
 from collections import OrderedDict
+import pytz
 
 from rest_framework.decorators import api_view
 from django.template.loader import render_to_string, get_template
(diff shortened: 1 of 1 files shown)


EliezerIsrael:
@ronshapiro Thanks for the patch!  It looks good to my eyes.  
I've ask one more person to give it a review and test.

reader/views.py
```diff
@@ -2199,13 +2200,23 @@ def calendars_api(request):
...
+
         try:
             year = int(request.GET.get("year", None))
             month = int(request.GET.get("month", None))
             day = int(request.GET.get("day", None))
-            datetimeobj = datetime.datetime(year, month, day)
+            datetimeobj = datetime.datetime(year, month, day, tzinfo=zone)
         except Exception as e:
-            datetimeobj = timezone.localtime(timezone.now())
+            datetimeobj = timezone.localtime(zone)
```
nsantacruz: Looks like the `timezone` param needs to be specified since it's the 2nd parameter
```suggestion
            datetimeobj = timezone.localtime(timezone=zone)
```

reader/views.py
```diff
@@ -2199,13 +2200,23 @@ def calendars_api(request):
...
         custom = request.GET.get("custom", None)
+
+        zone = request.GET.get("timezone", None)
+        if zone:
+            try:
+                zone = pytz.timezone(zone)
+            except pytz.exceptions.UnknownTimeZoneError as e:
+                return jsonResponse({"error": "Unknown 'timezone' value: '%s'." % zone})\
+        else:
+            zone = timezone.now()
```
nsantacruz: It seems like `zone` should be of type which is subclass of `tzinfo`. Here you're setting it to `datetime.datetime`. Pretty sure this is an error.

nsantacruz:
@ronshapiro hi! I've found 2 issues with this pull request. Can you fix those?

reader/views.py
ronshapiro: Hrm, yeah, you're right. I'd like to have the exception case that uses the localtime to also consider the timezone. I think that reverting that line, and then taking your suggestion above should do the trick

ronshapiro:
@nsantacruz thanks for the close eyes! I fixed those issues, and I fixed another below where I didn't properly propagate the timezone name to the response if it was specified as a parameter.

Do you prefer I squash my commits, or you'll do that automatically when merging?

nsantacruz:
@ronshapiro Looks great! Merged

ronshapiro:
Awesome! What's the best way to update the documentation? And how often do
you typically update your servers?

בתאריך יום א׳, 22 במרץ 2020, 14:43, מאת Noah Santacruz ‏<
notifications@github.com>:

PR #540, opened 2020-04-21
Remove seemingly unnecessary parameters

ronshapiro:
I'm not a python expert, but I believe these declarations are just using the same defaults that were already specified in the function declaration. It seems unnecessary and makes the code harder to read.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -1246,8 +1246,7 @@ def texts_api(request, tref):
             return text
 
         if not multiple or abs(multiple) == 1:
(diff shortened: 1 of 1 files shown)


JonMosenkis:
This is incorrect. The function definition defines defaults, the call to get_text does not use those defaults, but rather the parameters actually past by the user.

ronshapiro:
Are you sure? I don't see anywhere else that that function is used with other values (it's a local-function, so just those two calls).

It seems similar to doing this:

```
def foo(x):
    def _add(y, x=x):
        return y + x

    return _add(1)

print(foo(5))
```

PR #548, opened 2020-05-10
Cleanup: Inline conditional that is always true

ronshapiro:
This change removes an `if` statement that is always `True` (the `or True` clause causes the result to always be true, even if `not self._spanned_refs` is false.

The rest of the change is simply outdenting the code that was previously nested inside the `if`.

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -3603,53 +3603,51 @@ class Ref(object, metaclass=RefCacheType):
             [Ref('Shabbat 13b:3-50'), Ref('Shabbat 14a'), Ref('Shabbat 14b:1-3')]
 
         """
-        if not self._spanned_refs or True:
+        if self.index_node.depth == 1 or not self.is_spanning():
+            self._spanned_refs = [self]
 
-            if self.index_node.depth == 1 or not self.is_spanning():
-                self._spanned_refs = [self]
+        else:
+            start, end = self.sections[self.range_index()], self.toSections[self.range_index()]
+            ref_depth = len(self.sections)
+            to_ref_depth = len(self.toSections)
 
-            else:
-                start, end = self.sections[self.range_index()], self.toSections[self.range_index()]
-                ref_depth = len(self.sections)
-                to_ref_depth = len(self.toSections)
+            refs = []
+            for n in range(start, end + 1):
+                d = self._core_dict()
+                if n == start:
+                    d["toSections"] = self.sections[0:self.range_index() + 1]
 
-                refs = []
-                for n in range(start, end + 1):
-                    d = self._core_dict()
-                    if n == start:
-                        d["toSections"] = self.sections[0:self.range_index() + 1]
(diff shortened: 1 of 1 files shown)


JonMosenkis:
We're very appreciative of the initiative to clean up code, but I'm a little hesitant to accept a PR without a description. Please add:

1. A short description explaining what problem this PR is attempting to solve.
2. An outline of the test results. These should include both unit tests and a performance comparison.

Thanks!

ronshapiro:
@JonMosenkis I don't know how to run the tests, and I don't have a full development environment set up to run a performance comparison. But note that although the diff is ~40 lines, the only real change is removing the `if` statement at the top, and outdenting the rest of the code. Github gets a little confused with computing the diff because there are two lines that are identical, but one is indented one level deeper, so the diff considers that line to be unchanged.

EliezerIsrael:
Thanks for the clarification @ronshapiro.  I was scratching my head to figure out what was going on with the diff, at first glance.  Sound good to you @JonMosenkis?  You can go ahead an merge, from my POV.

JonMosenkis:
@ronshapiro Thanks for the clarification. @EliezerIsrael Yeah, going over the code, the `or True` statement seems to have been added to disable the `if` statement. I'm going to merge.

@ronshapiro Thanks again! Much appreciated!

JonMosenkis:
@ronshapiro @EliezerIsrael Travis is throwing failed tests here. I've reverted the merge

ronshapiro:
@JonMosenkis can you send a link to what's failing?

ronshapiro:
I don't see a passing Travis run in over a month...

sefaria/model/text.py
```diff
@@ -3603,53 +3603,51 @@ def split_spanning_ref(self):
...
-                    if n == start:
-                        d["toSections"] = self.sections[0:self.range_index() + 1]
+            refs = []
+            for n in range(start, end + 1):
+                d = self._core_dict()
+                if n == start:
+                    d["toSections"] = self.sections[0:self.range_index() + 1]
+
+                    for i in range(self.range_index() + 1, ref_depth):
+                        d["toSections"] += [self.get_state_ja().sub_array_length([s - 1 for s in d["toSections"][0:i]],until_lnonempty=True)]
```
EliezerIsrael, line 3621: The argument name was changed here.  That's causing test failures.

EliezerIsrael:
> I don't see a passing Travis run in over 2 months. And the failures from my commit seem to be similar to other failures from recent builds on Travis.

@ronshapiro  There are two test suites running on Travis.  The `pytest` suite of unit tests is dependable.  If it's got a red light, we're in trouble.  If that passes, Travis goes on to do front end tests w/ Selenium on Sauce Labs.  Those tests are terribly brittle, I'm sad to say.  

Not the best situation, w/ Travis.  But the `pytest` unit tests do need to pass.  (They're also pretty easy to run locally.  You can write to me directly if you want some help getting your env set up.  lev@sefaria.org)

PR #552, opened 2020-05-12
Run `npm audit fix`

ronshapiro:
When setting up my dev environment, I saw this in the logs after running `npm install`:

```
found 414 vulnerabilities (407 low, 5 moderate, 2 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```

I don't enough about most of these to know if this change is safe, but I figured I'd at least alert the team since you'll know better. I imagine this message doesn't come up often given that people don't set up their dev environment so often.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -224,9 +224,9 @@
       }
     },
     "acorn": {
-      "version": "6.3.0",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-6.3.0.tgz",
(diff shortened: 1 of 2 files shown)


JonMosenkis:
Thanks for raising awareness. Upgrading dependencies is a bit more involved than just changing some code in `package.json`, and can have unforeseen effects. 

I'm going to close this for now. For the future, I think opening a github _Issue_ would be better for cases where one is not sure if a change is safe.
I would like to express my gratitude in the input and feedback you've been sending.

PR #561, opened 2020-06-12
Propose that the gettext path is added to .bashrc

Krasen7:
In my opinion it is a good idea to propose to this contributors that work on macOS systems on which gettext won't run, to add the needed exports to their .bashrc.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -105,12 +105,11 @@ After installing the Python development libraries, run `pip install -r requireme
     brew install gettext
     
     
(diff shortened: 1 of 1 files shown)


nsantacruz:
@Krasen7 I agree the instructions can be clarified a bit. Can you instead edit this sentence to clarify where these commands should be put:


> On some macOS systems gettext will atill not run after installation and django manage.py makemessages will fail. In such a case, one easy solution is to run (replace x's with your gettext version number) before running functions that invoke gettext:

Krasen7:
Ok, I will just make another commit to patch-1, additionaly I have found a typo in 'atill', so i will fix that too :)

nsantacruz:
Thanks!

Krasen7:
No problem

PR #563, opened 2020-06-18
Update README.mkd

Krasen7:
Just replaced "We make a MongoDB dump of our database is available download." with "MongoDB dumps of our database are available to download.".


nsantacruz:
Thanks!

PR #574, opened 2020-07-21
Offer consistent size to virtual keyboard :onhover

yabbou:
If the size difference was intentional, perhaps reducing it from 1px to 0.6px could both retain the effect and not move around the keyboard when switching between rows.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -6178,7 +6178,7 @@ body #keyboardInputMaster tbody tr td table tbody tr td{
 body #keyboardInputMaster tbody tr td table tbody tr td:hover,
 body #keyboardInputMaster tbody tr td table tbody tr td:active,
 body #keyboardInputMaster tbody tr td table tbody tr td.pressed{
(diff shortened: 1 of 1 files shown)


EliezerIsrael:
Thanks for the pull request.
In the past, I have seen the keyboard shift around on hover, but I don't see that on the current site.  So you?  Can you detail under what conditions you see the bug?  I'd like to test the fix before merging.

yabbou:
For me, the selected key becomes slightly bigger than the others. 
On the home page and on the dictionary search. 
Microsoft Edge.

onboard

yabbou:
This issue seems to have been addressed for most of the keys from another pull request

nsantacruz:
I'm seeing all the keys remaining the same size onhover currently. @yabbou, are you still seeing any issues?

yabbou:
@nsantacruz Odd... last week the issue seemed resolved, but, for me, the issue has returned.

PR #584, opened 2020-08-07
Update StaticPages.jsx

chavaerica:
Update to contest landing page to allow for people to submit their completed project. Not sure I got the inline link right on the Hebrew for the Timeline section... Also updated the bottom bar. Text and button.


EliezerIsrael:
Looks good, @chavaerica.  Only thing is that for <a> links, we don't need <u>.  That should be handled by the styles.  I'll edit those out, and merge this into the main branch.

EliezerIsrael:
Sorry - the html got lost from my earlier comment.  I meant to write that we don't need `<u>` tags.

PR #585, opened 2020-08-09
Linker improvements

shaneerosen:
Three improvements to Linker:
Limit prefixes at the beginning of book name (ex: בשורות א ב will no longer link to Ruth 1 2)
A new parameter for parentheses by turning it on (parentheses = true) 'Linker' will link only to citations inside parentheses or brackets both in Hebrew and in English.
A New Parameter for quotation marks, by turning it on (quotationOnly=true) 'Linker' will link only to citations that the segment pointers in Hebrew are with quotations (Gershayim) if they are referred to with 2 or more letters. (ex: בראשית כ"א vs בראשית כא)


templates/js/linker.v2.js
```diff
@@ -352,6 +352,8 @@
         options.popupStyles = options.popupStyles || {};
         options.interfaceLang = options.interfaceLang || "english";
         options.contentLang = options.contentLang || "bilingual";
+        options.parentheses = options.parentheses || false;
```
nsantacruz, line 355: Let's rename this parameter `parenthesesOnly` to match `quotationsOnly`. I don't think it needs to be renamed in the python code but it should be renamed here so that people using the linker understand what it means more.

templates/js/linker.v2.js
```diff
@@ -360,7 +362,7 @@
         setupPopup(options, mode);
 
         ns.elems = document.querySelectorAll(selector);
-
+        ns.quotationOnly = options.quotationOnly;
```
nsantacruz, line 365: Should also add `ns.parenthesesOnly = options.parentheses` to avoid accessing `parentheses` through `ns.link.arguments[0].parentheses` on line 386

templates/js/linker.v2.js
```diff
@@ -409,32 +411,37 @@
...
-                        node.href = base_url + matched_ref;
-                        node.setAttribute('data-ref', matched_ref);
-                        node.setAttribute('aria-controls', 'sefaria-popup');
-                        node.textContent = portion.text;
-
-                        return node;
-                    },
+                            .replace(/[(){}[\]]+/g, "")
+                            .replace('\u05d3\u05e3', ""); // Filter out internal parenthesis todo: Don't break on parens in books names
+                            var matched_reg = new RegExp(`${book}\\s+([\u05d0-\u05ea]+?['\u05f3"\u05f4][\u05d0-\u05ea]*?(:|\\.)?(\\s|$)|[\u05d0-\u05ea](\\.|:)?([-\u2010-\u2015\u05be][\u05d0-\u05ea])?(\\s|$)|(\\d|\\.)+(a|b|:)?(\\s|[-\u2010-\u2015\u05be]|$))+`, 'g');
```
nsantacruz, line 424: these next few lines need comments

nsantacruz: Also, does this catch ranges in English?

templates/js/linker.v2.js
```diff
@@ -409,32 +411,37 @@
...
-                        node.className = "sefaria-ref";
-                        node.href = base_url + matched_ref;
-                        node.setAttribute('data-ref', matched_ref);
-                        node.setAttribute('aria-controls', 'sefaria-popup');
-                        node.textContent = portion.text;
-
-                        return node;
-                    },
+                            .replace(/[(){}[\]]+/g, "")
+                            .replace('\u05d3\u05e3', ""); // Filter out internal parenthesis todo: Don't break on parens in books names
```
nsantacruz, line 423: i'm a bit scared about this line. It can be added to the regex on line 424 as optional right after `${book}`

templates/js/linker.v2.js
shaneerosen: yes. It does catch English ranges, given that the ranges are in numbers (ex Job 3.1-2)

PR #605, opened 2020-12-01
fix footer_html when USE_NODE is False

grahame:
This PR fixes the following error when `USE_NODE` is `False` in settings:

```
UnboundLocalError at /
local variable 'lang' referenced before assignment
/www/sefaria/system/context_processors.py in footer_html, line 175
```


blockspeiser:
It turns out @edamboritz just made this exact fix with a refactor he just rolled out, but I'm accepting the pull request anyway so you can count among our contributors :).

PR #610, opened 2020-12-17
unclicking bookmark no longer sends you to bookmarked Ref page

stevekaplan123:


EliezerIsrael:
Thanks, @stevekaplan123.

This looks good to my eyes.  I dug in a bit, and I can't conceive of any reason why we'd want to perform the default link action on the click of a save star.  I'm going to merge this in.

PR #612, opened 2020-12-30
talmud refs without amud + relevant tests

stevekaplan123:


EliezerIsrael (commented):
Rather than adding this as another special case in the __init_tref() method of Ref, I'd love to see it in the methods of the AddressTalmud class in schema.py.

blockspeiser:
@EliezerIsrael Steve and I dug into this together - for this case it's hard to see how to get it down into the AddressTalmud without doing some deep refactoring. The issue is that this needs to affect the second part of the range of the ref, but those code paths currently only come up in __init_tref() when there is content after a hyphen in the ref, which there is not here. In the current structure, everything that comes out of the AddressTalmud only has the opportunity to have an effect on the `sections` component of the ref. 

I think for a similar reason this is why we end up with a special cased __parsed_talmud_range() in __init_tref().

Can you see a path that would get this there?

Short of that, this may just feel a lot cleaner if we pulled out a lot of the code into a helper function like __talmud_raf_lacks_amud()

nsantacruz:
It feels like we could a `lacks_amud` function to `AddressTalmud` and use the existing regex as a base for detecting if there's an amud. The amud regex [here](https://github.com/Sefaria/Sefaria-Project/blob/master/sefaria/model/schema.py#L1995-L1998) could be pulled out into a variable and used in that function.

blockspeiser:
@nsantacruz that's smart. It's still a special path in __init_tref() but all the core logic will live in one place in `AddressTalmud`, and there's won't be duplication of the regex. @EliezerIsrael what do you think?

EliezerIsrael:
I like the direction.  Ideally, think that the Ref class to not have special cases for Talmud, but rather use the methods of the Address classes  to know if special processing needs to happen and then to execute that processing.   According to that reasoning, `__parse_talmud_range` should also be pushed down into the Address class, and given a more general hook.

PR #614, opened 2021-01-10
Talmud refs without amud

stevekaplan123:


blockspeiser:
Just taking a quick look but I think this is looking good. Only note I would make: in __init_tref() instead of checking explicitly if the address type is "Talmud" we could instead check if the relevant class implements the method "parse_range_end()". Then there is no special casing of Talmud left in __init_tref() at all, and instead any time an Address class wants it's own logic here it can just implement its own "parse_range_end".

To do that in the scope of __init_tref you'll need a way to map the string of the address type to its actual class. You can do that with a static method on AddressType that takes a string and returns a class, basically the opposite of the method "toStrByAddressType" which it has already.

PR #619, opened 2021-01-27
5 bugs and backlog cards implemented

stevekaplan123:
1. Berakhot 2 -> Berakhot 2a-2b
2. Berakhot 2a-b -> Berakhot 2a-2b
3. Linker now picks up both ends of range Genesis 1-Genesis 3 and continues to catch Genesis 1-3 as one Ref
4. Also, both refs are picked up if the case is Genesis 1/Genesis 3 
5. Finally, refs are now sorted on save and ensure_indices will create a unique refs index on Links collection.  
In addition, this PR contains a script called delete_duplicates_and_sort_refs_field.py that will delete all duplicate links found in duplicates.csv, sort refs fields on all Link objects in the DB, and create the unique refs index on the Links collection.  However, this PR does not contain the file duplicates.csv because it is a fairly large file.  It can be found here: https://trello.com/c/rICgc7zi/785-ensure-link-refs-are-unique-in-mongo
duplicates.csv needs to be put in the scripts folder in Sefaria-Project for the code in delete_duplicates_and_sort_refs_field.py to run properly.


nsantacruz:
@stevekaplan123 realize this was merged which is great! really small point, why is `linker_test.html` loading linker.v1? shouldn't we be testing against v2?

blockspeiser:
@nsantacruz We need to be testing against both, since v1 is still powering MJL and getting a lot of traffic. The way I've been using that test page now usually involves tinkering with the setup of two switch between v1/v2 or change some of the invocation params on the linker to test specific behavior. So it may end up getting committed in different states.

PR #629, opened 2021-03-15
cancael button to null ref in alt structs

YishaiGlasner:


static/js/ReaderTextTableOfContents.jsx
```diff
@@ -917,6 +917,9 @@ class ArrayMapNode extends Component {
     if ("refs" in this.props.schema && this.props.schema.refs.length) {
       var sectionLinks = this.props.schema.refs.map(function(ref, i) {
         i += this.props.schema.offset || 0;
+        if (ref == "") {
```
nsantacruz, line 920:
```suggestion
        if (ref === "") {
```
Use === to make sure ref is actually `""` and not `null` or `false` or some other falsy value. See [this link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) for an explanation

static/js/ReaderTextTableOfContents.jsx
```diff
@@ -917,6 +917,9 @@ class ArrayMapNode extends Component {
...
+          return (null)
```
nsantacruz, line 921:
```suggestion
          return null;
```
No need for parentheses. And Javascript lines should end in `;` (although they don't throw an error if they don't have them, it's best practice to include them)

nsantacruz:
This looks great!

PR #631, opened 2021-03-16
Yoma 88a tests pass

stevekaplan123:


nsantacruz:
@stevekaplan123 this looks good. I pushed a change that switches from pulling up the versionState to using index_node.lengths. See the commit message for my reasoning.

I'm realizing the whole feature of `parse_range_end` is hard-coded for the AddressTalmud appearing in the first slot of sections. I'm seeing `0` indexes in many places that indicate this is hard-coded. We can technically have AddressTalmud at any position in sections. @EliezerIsrael Interested in your thoughts as to importance of generalizing this.

EliezerIsrael:
@nsantacruz @stevekaplan123  If I'm not wrong, this will fail to work for the Zohar, which has Talmud addresses in the second position.  It also does not have a "lengths" field, and the the lengths field does not handle lengths of second-level sections.  

We need to make this logic positionally aware, and have it fall back to version state queries when the length field is not available. 

We should start by writing failing tests for the Zohar cases, and then get them to pass.

PR #639, opened 2021-04-22
Added 'download MongoDB Database tools' to readme

yairm210:
As documented [here](https://docs.mongodb.com/database-tools/mongorestore/), from MongoDB 4.4 the database tools (including mongorestore) are versioned and installed separately from the MongoDB server, making their installation a necessary step for the import.

Also, since the import is slightly lengthier on Windows (download file - unzip *twice* - go to folder above dump/sefaria, run mongorestore there by adding the folder to path OR specifying the exact exe file location) I think it could be helpful to future Windows devs to state that. I feel that there are a lot of steps here that are assumed or glossed over, which can make it difficult for more junior devs to get started with it.


EliezerIsrael (approved):
Thank you.

PR #641, opened 2021-04-26
Type annotations

yairm210:
This is more for the discussion than for actual change.

As someone new to Sefaria, I find some of the code to be very abstruse.

A big part of this is the lack of typing in both the parameters that functions receive and their return values.

This PR is more of a question: are you open to adding type annotations to the code? They make the code much more immediately readable, especially if you're using a powerful (read: Jetbrains) IDE like Pycharm which understands, and can infer from, type annotations.

I'm starting with some obvious things, but if you're okay with this I hope to continue to some of the stranger functions.

### DIFF
--- a/sefaria/datatype/jagged_array.py
+++ b/sefaria/datatype/jagged_array.py
@@ -210,7 +210,7 @@ class JaggedArray(object):
                 return False
         return True
 
-    def is_empty(self, _cur=None):
+    def is_empty(self, _cur=None) -> bool:
         if _cur is None:
             return self.is_empty(_cur=self._store)
         if isinstance(_cur, list):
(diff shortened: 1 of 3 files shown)


EliezerIsrael:
We'd love to be making the code clearer and more usable for developers who would find it useful.  

Type annotations are pretty easy to agree to.  The only note of caution we have it with annotating using our own declared types.  We've found some cases where imports get hairy with those annotations.  But as long as everything still works, type annotations are only a positive.

yairm210:
Another question: I have a feeling that most of my PRs will be backend, and probably many will be Enhancements.

Would you prefer I add these as labels to the PRs?

PR #643, opened 2021-04-27
More typings. 

yairm210:
Some obvious, some (like Library._index_title_maps) required some research.


sefaria/model/text.py
```diff
@@ -1831,9 +1831,9 @@ def find_string(self, regex_str, cleaner=lambda x: x, strict=True):
         text_list = [x for x in self.ja().flatten_to_array() if len(x) > 0]
         if len(text_list) != len(ref_list):
             if strict:
-                raise ValueError("The number of refs doesn't match the number of starting words. len(refs)={} len(inds)={}".format(len(ref_list),len(ind_list)))
+                raise ValueError("The number of refs doesn't match the number of starting words. len(refs)={} len(inds)={}".format(len(ref_list),len(text_list)))
```
yairm210, line 1834: This was a broken reference - at some point in time someone changed "ind_list" to "text_list" above but didn't change these references to it.

nsantacruz:
I'm seeing some issues with some of the typing here. Specifically `list[int]` is not working. I'm getting the error: `TypeError: 'type' object is not subscriptable`

I think this SO answer is relevant for fixing the issue: https://stackoverflow.com/questions/63460126/typeerror-type-object-is-not-subscriptable-in-a-function-signature

yairm210:
Yeah I just noticed that as well 😓 
I'll make sure to run the server locally before I make future PRs to ensure that no obvious problems occur :)

PR #644, opened 2021-04-27
Resolved #541 - multi-page commentary references no longer split into individual letters at /api/texts

### LINKED ISSUE #541: Some comments in the /texts API are broken up letter by letter
https://www.sefaria.org/api/texts/Shabbat.51b?commentary=1 lists a Mesorat Hashas commentary for the ref `Shabbat 56b:20-57a:1` that has a `text` attribute with 575 items, and an `he` attribute with 90 items. Looking at them, it appears that something is causing them to break up each character into individual strings (except for the first `he` item, which has some HTML in it).

[Requesting the ref as a regular text](https://www.sefaria.org/Shabbat.56b.20-57a.1) shows the text as normal.

yairm210:
When you have references that span multiple pages, Sefaria attempts to connect the text in the pages together to a single list.
It goes over each page (wrapper.py:215, `for com_oref in con_orefs`) and for each tries to add it to a list.

For the first item in the list (which is usually all you have) it just uses that string (wrapper.py:259, `com[attr] = res`)

However on the SECOND one, it attempts to make a list that will include both.
:262, `com[attr] = [com[attr]]` works fine. However :263, `com[attr] += res` treats the res (which can be a single line of text) AS A LIST, and therefore adds all the elements of the list to the other list... since it's a string, that means the individual chars...

The solution is to check if it's a string or not.

However, I think the real source of the problem is the ambiguity in the JaggedArray.array() function.

:257 has `res = temp_nref_data['ja'].subarray(com_sections[1:], com_toSections[1:]).array()`. As a casual reader, I would expect that res is therefore an array (it has .array() at the end, after all). THIS IS A TRAP.

Would typing the .array() function to indicate it can return either an array or a string have helped? I don't know. But it might have avoided this edge case.

### DIFF
--- a/sefaria/client/wrapper.py
+++ b/sefaria/client/wrapper.py
@@ -250,7 +250,9 @@ def get_links(tref, with_text=True, with_sheet_links=False):
                             }
                     com_sections = [i - 1 for i in com_oref.sections]
                     com_toSections = [i - 1 for i in com_oref.toSections]
-                    for lang, (attr, versionAttr, licenseAttr, vtitleInHeAttr) in (("he", ("he","heVersionTitle","heLicense","heVersionTitleInHebrew")), ("en", ("text", "versionTitle","license","versionTitleInHebrew"))):
+                    for lang, (attr, versionAttr, licenseAttr, vtitleInHeAttr) in (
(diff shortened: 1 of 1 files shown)


nsantacruz:
This looks good to me. Good catch! I'll also close #541

PR #664, opened 2021-06-29
Talmud toggle punctuation

nissamai:
http://punctuationtoggle.cauldron.sefaria.org/


reader/views.py
```diff
@@ -224,6 +224,7 @@ def base_props(request):
             "layoutTanakh":  request.COOKIES.get("layoutTanakh", "segmented"),
             "aliyotTorah":   request.COOKIES.get("aliyotTorah", "aliyotOff"),
             "vowels":        request.COOKIES.get("vowels", "all"),
+            "punctuationTalmud": request.COOKIES.get("punctuationTalmud", "punctuationOn"),
```
blockspeiser, line 227: Minor style point: for readability the values of each of these keys were aligned with spaces. By adding a longer key, the other lines should all have spaces added so that each "request" lines of vertically.

static/js/ReaderApp.jsx
```diff
@@ -949,6 +949,7 @@ class ReaderApp extends Component {
         layoutTanakh:  "segmented",
         aliyotTorah:   "aliyotOff",
         vowels:        "all",
+        punctuationTalmud: "punctuationOn",
```
blockspeiser, line 952: Minor style point, same as above: for readability the values here can be aligned with spaces in each line.

static/js/ReaderPanel.jsx
```diff
@@ -1404,6 +1404,11 @@ class ReaderDisplayOptionsMenu extends Component {
     const hasEnglish = !!data.text.length;
     return !(hasHebrew && hasEnglish);
   }
+  renderPunctuationToggle() {
```
blockspeiser, line 1407: I think this is more clear as `shouldPunctuationToggleRender`, so it doesn't give the idea that function is performing the render.

static/js/ReaderPanel.jsx
```diff
@@ -1404,6 +1404,11 @@ class ReaderDisplayOptionsMenu extends Component {
...
+    if (this.props.currentData?.()?.primary_category === "Talmud" && (this.props.settings?.language === "hebrew" || this.props.settings?.language === "bilingual")) return true;
+    else return false;
```
blockspeiser, line 1409: We do want to use {} even on single line blocks.

PR #678, opened 2021-07-30
Chavruta iteration 2

ahoch10:
Fix extra comma issue in chavruta sources

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -234,8 +234,7 @@ class ReaderApp extends Component {
       else if (panel.filter.length !== 0) return panel.filter[0]
       else return
     }).filter(ele => ele).join(", ")
-    // console.log(currentlyReading)
-    // console.log(this.state.panels)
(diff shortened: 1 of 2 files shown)


static/js/ReaderApp.jsx
```diff
@@ -231,8 +231,10 @@ class ReaderApp extends Component {
 
     const currentlyReading = this.state.panels.map((panel) => {
       if (panel.mode === "Text") return panel.currentlyVisibleRef
-      else return panel.filter[0]
-    }).join(", ")
+      else if (panel.filter.length !== 0) return panel.filter[0]
```
blockspeiser, line 234: Minor code style point on these three lines: (1) We should always use semi-colons at the end of lines in JS. (2) We should always use {} around the bodies of if/else statements even if they are on one line. Leaving them out makes code more prone to error when somebody later on comes by to modify the code. @ahoch10 @rneiss

rneiss (approved):

PR #681, opened 2021-08-03
Sheet tagging admin

nissamai:
Sheet tagging administrative functionality for tagging + adding categories + "No Tags"

http://localhost:8000/admin/categorize-sheets (shows you newest sheet that is not categorized)
http://localhost:8000/admin/tag-sheets (shows you newest sheet that has no tags and is not tagged "no tags")

Your database will start with no categories, but if you add and save some, they will start showing up.

Question:
- categories_all = db.sheets.distinct("categories") is extremely slow. It seems (?) like I should potentially add an index or create a different collection with the sheet names or something a la relational database, but I'm not quite so good at mongo to know exactly what the most reasonable solution is.


sefaria/sheets.py
```diff
@@ -1202,3 +1202,19 @@ def change_tag(old_tag, new_tag_or_list):
 	for sheet in SheetSet({"tags": old_tag}):
 		sheet.tags = [tag for tag in sheet.tags if tag != old_tag] + new_tag_list
 		sheet.save()
+
+def get_sheet_categorization_info(does_not_contain):
```
blockspeiser, line 1206: I've been puzzling here about what does_not_contain means and what it does. Not clear to me! Either this needs a clarifying comment or a better name or both.

sefaria/sheets.py
```diff
@@ -1202,3 +1202,19 @@ def change_tag(old_tag, new_tag_or_list):
...
+	from pymongo import DESCENDING
+	if does_not_contain == "topics":
+		sheet = sheet_to_dict(db.sheets.find({"topics": {"$in": [None, []] }, "noTags": {"$in": [None, False]}, "status": "public"}).sort("id", DESCENDING)[0]) 
+	else: # categories
+		sheet = sheet_to_dict(db.sheets.find({"categories": {"$in": [None, []] }, "status": "public", "$where": "this.includedRefs.length != this.sources.length"}).sort("id", DESCENDING)[0])
+	
+	categories_all = db.sheets.distinct("categories") # this is slow; maybe add index or ...?
```
blockspeiser, line 1213: Yes, we should just add an index here. In the future we will certainly want to be querying sheets by category.

sefaria/views.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import io
 import os
+from sefaria.sheets import get_sheet_categorization_info
```
blockspeiser, line 4: We want to follow this code style for ordering imports: https://www.python.org/dev/peps/pep-0008/#imports

For us this means generally a group of system imports, a group of django imports, then a group of sefaria imports.

sefaria/views.py
```diff
@@ -59,7 +60,7 @@
 from sefaria.google_storage_manager import GoogleStorageManager
 
 
-from reader.views import render_template
+from reader.views import base_props, render_template, text_panels
```
blockspeiser, line 63: The fact that you need to import these here kinda says to me that they are in the wrong place in the first place.... (not sure what to do about that)

static/js/categorize_sheets.jsx
```diff
@@ -0,0 +1,207 @@
...
+            console.log(data);
+          }
+        }.bind(this)
+      });
+  }
+
+  getSheet() {
+    Sefaria.sheets.loadSheetByID(
+      this.state.sheetId,
+      function (x) {
```
blockspeiser, line 95: For style and readability we should default to arrow functions now.

blockspeiser (commented):
Mostly minor style points.

PR #708, opened 2021-11-07
Linker tracking

stevekaplan123:
This is actually code that I wrote and tested before I went on vacation in the summer.  The idea is that if a page had refs and is in our DB but then changed and no longer has refs, we would never know about it, so this code will track pages even if they don't have any refs, and then if the webpage already exists, the code on the backend will delete the page.   This code was put on hold because the major question surrounding it is if it will put to much of a load on our servers.  That could be a real downside and Brett and I discussed the possibility of putting this code up for just a few hours to see how much of an effect it would have.  However, there is a major upside -- it would lead to much better webpages data (and I believe is one half of the reason why going to a page does not always lead to lastUpdated being updated).  One thought I had that could mitigate the load on our servers is to use a random number generator and only track a page with no trackedMatches, let's say, 20% of the time.  What do you think?

### DIFF
--- a/sefaria/model/webpage.py
+++ b/sefaria/model/webpage.py
@@ -160,7 +160,7 @@ class WebPage(abst.AbstractMongoRecord):
             webpage = WebPage(data)
             existing = False
         webpage._normalize() # to remove bad refs, so pages with empty ref list aren't saved
-        if webpage.should_be_excluded():
+        if webpage.should_be_excluded() or len(data["refs"]) == 0:  # data["refs"] is empty when linker.js finds no refs on a page:
             if existing:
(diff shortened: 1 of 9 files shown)


nsantacruz:
The code looks find glancing at it, but I do share the concern that this could put too much load on our servers. I think it's probably ok to deploy this code for a few hours to see what the impact is. Another possibility is to hash the contents of pages and include this hash on an already existing request (e.g. for regexes). If the hash matches a page in our database, then the page hasn't changed since we last checked. This is complicated though, because many pages have dynamic content that changes daily. I'll discuss with Lev to see what he thinks.

PR #718, opened 2021-12-07
reject URL if it's bleached form is over 1,000 chars

stevekaplan123:


sefaria/model/webpage.py
```diff
@@ -129,7 +130,7 @@ def excluded_pages_url_regex():
     def excluded_pages_title_regex():
         bad_titles = [
             r"Page \d+ of \d+",  # Rabbi Sacks paged archives
-            r"^Page not found$",   # JTS 404 pages include links to content
+            r"Page [nN]{1}ot [fF]{1}ound$",  # 404 pages include links to content
```
nsantacruz, line 133:
```suggestion
            r"Page [nN]ot [fF]ound$",  # 404 pages include links to content
```
No need for `{1}`. brackets will default to only matching one character.

PR #723, opened 2021-12-23
Shorashim

YishaiGlasner:


sefaria/model/lexicon.py
```diff
@@ -141,8 +146,14 @@ def get_sense(self, sense):
         return text
 
     def headword_string(self):
-        return ', '.join(
-            ['<strong dir="rtl">{}</strong>'.format(hw) for hw in [self.headword] + getattr(self, 'alt_headwords', [])])
+        headwords = [self.headword] + getattr(self, 'alt_headwords', [])
+        if self.contents()['parent_lexicon_details']['version_lang'] == 'he':
```
nsantacruz, line 150: There seems to be a large inefficiency in `contents()` it loads the `Lexicon` from the database, but it's accessible by `self.parent.lexicon`. Use `self.parent.lexicon` instead of `contents()['parent_lexicon_details']` and can you fix `contents()` so that it uses `self.parent.lexicon`?

sefaria/model/lexicon.py
```diff
@@ -141,8 +146,14 @@ def get_sense(self, sense):
...
+            headwords = [f'<big>{hw}</big>' for hw in headwords]
```
nsantacruz, line 151: Why `<big>` for Hebrew?

sefaria/model/lexicon.py
```diff
@@ -141,8 +146,14 @@ def get_sense(self, sense):
...
+        string = ', '.join(
+            ['<strong dir="rtl">{}</strong>'.format(hw) for hw in headwords])
+        if self.contents()['parent_lexicon_details']['version_lang'] == 'he':
+            string += '\xa0\xa0'
```
nsantacruz, line 155: Why non-breaking space in Hebrew?

static/js/BookPage.jsx
```diff
@@ -813,6 +822,22 @@ ArrayMapNode.propTypes = {
...
   render() {
     if (this.props.schema.headwordMap) {
+      const dictionarySearch =
+              <DictionarySearch
+              lexiconName={this.props.lexiconName}
+              title={this.props.title}
+              showBaseText={this.props.showBaseText}
+              contextSelector=".bookPage"
+              currVersions={this.props.currVersions}/>;
+      const header = this.props.schema.title ? (
```
nsantacruz, line 832: The only thing that changes if there is a schema title is the text passed to `ContentText`. Best to calculate the text beforehand and then you don't need an if statement to create the header.

static/js/BookPage.jsx
```diff
@@ -539,7 +533,14 @@ class SchemaNode extends Component {
         );
       } else if (this.props.schema.nodeType === "DictionaryNode") {
         return (
-          <DictionaryNode schema={this.props.schema} />
+          <DictionaryNode
+              schema={this.props.schema}
+              lexiconName={this.props.lexiconName}
```
nsantacruz, line 538: Where does this prop come from? I'm not seeing it passed to `SchemaNode` but I might be missing something. And shouldn't `lexiconName` be on the `schema` prop already since this is a DictionaryNode?

sefaria/model/text.py
```diff
@@ -2032,7 +2032,7 @@ def __init__(self, oref, lang="en", vtitle=None, exclude_copyrighted=False, actu
         self.is_merged = False
         self.sources = []
 
-        if self.lang not in self._oref.index_node.supported_languages:
+        if self.lang != self._oref.index_node.parent.lexicon.version_lang:
```
nsantacruz, line 2035: `index_node` can be any `VirtualNode`, not just a `DictionaryEntryNode` (the only other example right now is `SheetNode`). I think instead we need a function on `VirtualNode` called `supports_language(lang)`. Then `DictionaryEntryNode` can use `parent.lexicon.version_lang` to make sure the languages match but `SheetNode` will always return `True`.

sefaria/model/schema.py
```diff
@@ -1479,10 +1479,15 @@ def first_child(self):
     def last_child(self):
         pass
 
+    def supports_language(self, lang):
```
nsantacruz, line 1482: This is a strange case because `DictionaryEntryNode` and `SheetNode` don't inherit from `VirtualNode`, but share properties. I think we need a `support_language(lang)` function on all three, each with its own implementation.

sefaria/model/text.py
```diff
@@ -2032,7 +2032,7 @@ def __init__(self, oref, lang="en", vtitle=None, exclude_copyrighted=False, actu
         self.is_merged = False
         self.sources = []
 
-        if self.lang not in self._oref.index_node.supported_languages:
+        if not self._oref.index_node.parent.supports_language(self.lang):
```
nsantacruz, line 2035: I believe we don't want `.parent` here. This should run directly on `index_node`.

sefaria/model/schema.py
```diff
@@ -1678,6 +1679,22 @@ def get_child_order(self, child):
...
+            "book": self.full_title("en"),
+            "primary_category": self.index.get_primary_category(),
+            "index_node": self,
+            "sections": [],
+            "toSections": []
+        }
+        return text.Ref(_obj=d)
+
+    def supports_language(self, lang):
+        return True if lang == self.lexicon.version_lang else False
```
nsantacruz, line 1696:
```suggestion
        return lang == self.lexicon.version_lang
```

PR #727, opened 2022-01-10
Fix 'copy link' button in share menu

ruebeckscube:
The 'copy link' button in the 'share' menu was jumping around when clicked, which then meant that the link wasn't actually getting copied because the mouse wasn't on the button when it was released.

To get to the button I'm referring to: Open Genesis, click a verse, scroll down in the sidebar to "share," click the copy link button next to the text box.


nsantacruz:
@ruebeckscube looks great! merging.

PR #741, opened 2022-01-20
graphql basic structure and beginning of index query

JonMosenkis:
Setting up a basic module for graphql queries. This looks pretty promising and very easy to enhance and expand in the future. I'd love to hear feedback!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -59,3 +59,4 @@ unicodecsv==0.14.1
 unidecode==1.1.1
 user-agents==2.2.0
 babel
+strawberry-graphql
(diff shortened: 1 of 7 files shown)


JonMosenkis:
This is still a work in progress, so I'd strongly recommend against merging at this time. But I would like periodic feedback if possible.

JonMosenkis:
I've been using the `graphql-request` library from npm to execute the graph querries. For example, to query indices I set up the file `graphql.mjs` as so:

```
import { request, gql } from 'graphql-request'

const query = gql`
  {
      Index(title: "Genesis") {
        title
        categories
        node {
          depth
          sectionNames
          titles {
            text
            primary
          }
        }
        altStructs
      }
  }
`

request('http://localhost:8000/api/graphql', query)
.then(data => {
  console.log(data);
})
.catch(err => console.error(err));
```

PR #757, opened 2022-02-14
Version links

stevekaplan123:
Version links now behave as expected, although a script on prod needs to be run.  One thing though: I'm not sure about the change I made to ReaderPanel, where I passed `replace` to showBaseText.  It seems like it should be passed but perhaps there was a reason why we didn't want to pass it?


nsantacruz:
@stevekaplan123 yeah, I think it's correct to pass replace to `showBaseText()`.

PR #779, opened 2022-03-14
fix(helm): add helm required labels

paddatrapper:


BrendanGalloway (approved):
nit: Configmap relocation should be a separate commit for clarity

paddatrapper:
Done

PR #780, opened 2022-03-14
feat(helm): add nodeAffinity

paddatrapper:


BrendanGalloway:
Is there a way to handle this via a helper?  All the pods have the same affinity requirements at the moment, so managing them in a single place would be preferred

paddatrapper:
It now uses a template

BrendanGalloway (approved):

PR #793, opened 2022-03-23
Divine names reader

rneiss:


EliezerIsrael:
Pulled, tested, and encountered an error.  Divine names displayed as `undefined`.

EliezerIsrael:
Display error on the ה׳ name. 
Screen Shot 2022-04-05 at 4 36 49 PM

---

I'm seeing that the results of a selection here are only saved when a user edits something else on the sheet.
Ideally, the change would be saved on selection.  Is that possible?

EliezerIsrael (commented):
All good!

PR #802, opened 2022-04-06
Rollout healthcheck fixes

edamboritz:


sites/s4d/urls.py
```diff
@@ -8,6 +8,8 @@
 site_urlpatterns = [
     url(r'^$', reader_views.home, name="home"),
     url(r'^healthz/?$', reader_views.application_health_api_nonlibrary),
+    url(r'^health-check/?$', reader_views.application_health_api_nonlibrary),
+    url(r'^healthz-rollout/?$', reader_views.rollout_health_api),
```
BrendanGalloway, line 12: I thought these were going to be added to some common location?  Not strictly necessary, but some way to ensure they are always present would be good, given the deployment system depends on them...

BrendanGalloway (approved):

PR #826, opened 2022-05-03
add ml model settings

nsantacruz:
Added new setting to localsettings.yaml in preparation for new feature launch.

### DIFF
--- a/helm-chart/templates/configmap/local-settings.yaml
+++ b/helm-chart/templates/configmap/local-settings.yaml
@@ -225,6 +225,16 @@ data:
 
     MOBILE_APP_KEY = os.getenv("MOBILE_APP_KEY")
 
+    RAW_REF_MODEL_BY_LANG_FILEPATH = {
+        "en": None,
+        "he": None
+    }
+
+    RAW_REF_PART_MODEL_BY_LANG_FILEPATH = {
+        "en": None,
+        "he": None
+    }
+
     LOGGING = {
         'version': 1,
         'disable_existing_loggers': False,


helm-chart/templates/configmap/local-settings.yaml
```diff
@@ -225,6 +225,16 @@ data:
 
     MOBILE_APP_KEY = os.getenv("MOBILE_APP_KEY")
 
+    RAW_REF_MODEL_BY_LANG_FILEPATH = {
+        "en": None,
+        "he": None
```
BrendanGalloway, line 230: this should be `json.loads(os.getenv("RAW_REF_MODEL_BY_LANG_FILEPATH"))` to allow the value to be overwritten in different environments.

Then at the start of the configmap you need to add a new key:
`RAW_REF_MODEL_BY_LANG_FILEPATH: {{ .Values.localSettings.RAW_REF_MODEL_BY_LANGUAGE | toJson | quote }}` so that the override can be injected in the pod environment.

And then in values.yaml you would have:
```
localsettings:
  RAW_REF_MODEL_BY_LANG_FILEPATH: {"en": null, "he": null}
```
as the sane defaults that are always passed in (and that can be overridden in specific environments to look different)

You'll need to repeat the above for RAW_REF_PART_MODEL

BrendanGalloway (changes requested):

helm-chart/templates/configmap/local-settings.yaml
BrendanGalloway: This will always be null in web, we will create a new workload where this feature is enabled, with it's own localsettings

PR #830, opened 2022-05-05
ci: deploy helm chart to github pages

paddatrapper:
This does not handle any tagging or respond to tags (that will be a future improvement), nor does it handle semantic versioning and automatic update of the application version when the main project does releases. I intend to work on that over the coming week

PR #832, opened 2022-05-09
ci: add pipeline to publish helm charts

BrendanGalloway:
Created github pages branch using: https://gist.github.com/ramnathv/2227408

@edamboritz @EliezerIsrael if theres a need for consistent branding for all aspects of Sefaria you can set up themes: https://docs.github.com/en/pages/quickstart
You probably also want to check that any other github pages settings are to your preference.

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -1,0 +1,59 @@
+name: Helm Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+
+  changes:
+    permissions:
+      contents: read  # for dorny/paths-filter to fetch a list of changed files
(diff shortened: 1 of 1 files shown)


paddatrapper:
@BrendanGalloway can you compare to #830?

BrendanGalloway:
@paddatrapper Apologies, didn't realise you had found the time to send that up.  I think this pipeline is a bit more robust at the moment since it should autopublish on chart changes and handle the versioning itself (albeit with fairly simple logic rather than semantic from what I can see)

edamboritz:
@BrendanGalloway @EliezerIsrael Selected one of the default themes for now, that looked like the only relevant customization in the Github Pages settings.

PR #850, opened 2022-05-19
ci: remove explicit chart version

BrendanGalloway:


paddatrapper:
This will cause the lint to fail and any installation directly from Git to fail too

PR #872, opened 2022-06-06
feat: add manifests for linker service

BrendanGalloway:


helm-chart/sefaria-project/templates/configmap/nginx.yaml
```diff
@@ -122,6 +128,12 @@ data:
           access_log off;
           proxy_pass https://storage.googleapis.com/sefaria-sitemaps$request_uri;
         }
+
+        {{- if .Values.linker.enabled }}
+        location /linker/ {
```
BrendanGalloway, line 133: /api/find-refs

helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
```diff
@@ -211,15 +210,9 @@ data:
 
     MOBILE_APP_KEY = os.getenv("MOBILE_APP_KEY")
 
-    RAW_REF_MODEL_BY_LANG_FILEPATH = {
-        "en": None,
-        "he": None
-    }
+    RAW_REF_MODEL_BY_LANG_FILEPATH = json.loads(os.getenv("RAW_REF_MODEL_BY_LANG_FILEPATH", '{ "en": None, "he": None }'))
```
nsantacruz, line 213: Default should use `null` not `None`.

BrendanGalloway: yes, thanks for the catch

helm-chart/sefaria-project/templates/rollout/linker.yaml
```diff
@@ -0,0 +1,191 @@
...
+          - name: NODEJS_HOST
+            value: "node-{{ .Values.deployEnv }}-{{ .Release.Revision }}"
+          - name: VARNISH_HOST
+            value: "varnish-{{ .Values.deployEnv }}-{{ .Release.Revision }}"
+          - name: HELM_REVISION
+            value: "{{ .Release.Revision }}"
+          - name: ENABLE_LINKER
+            value: "True"
+          - name: DISABLE_AUTOCOMPLETER
+            value: "True"
```
nsantacruz, line 90: From what I understand, `local-settings-file.yaml` will run `bool("True")` which will be `True`. but `bool("False")` is also `True`. So not currently an issue but confusing for the future if someone changes this to `"False"`.

helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
```diff
@@ -179,11 +179,10 @@ data:
 
     # Turns off search autocomplete suggestions, which are reinitialized on every server reload
     # which can be annoying for local development.
-    DISABLE_AUTOCOMPLETER = False
-    ENABLE_LINKER = False
+    DISABLE_AUTOCOMPLETER = bool(os.getenv("DISABLE_AUTOCOMPLETER", False))
```
nsantacruz, line 182: Do we want the default value here to be `False`? Web pods, which are arguably the standard use-case, will set this to `True`.

nsantacruz:
@BrendanGalloway added some comments. I wasn't able to see where you set the environment variables `LINKER_ENABLED` and `DISABLE_AUTOCOMPLETER` for the web pods (I see where you're doing it for the linker pod).

BrendanGalloway:
> @BrendanGalloway added some comments. I wasn't able to see where you set the environment variables `LINKER_ENABLED` and `DISABLE_AUTOCOMPLETER` for the web pods (I see where you're doing it for the linker pod).

I have not set them for web, it should just fall through to the default instead.

helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
BrendanGalloway: I thought web pods generally have the autocompleter enabled?  So it should not be disabled?

helm-chart/sefaria-project/templates/rollout/linker.yaml
BrendanGalloway: That was not my intention.  Do you know a good way to handle this?  Doing a string compare instead of a parse seems a bit clunky but I'm not sure there's a better option?

nsantacruz: I would use `"0"` and `"1"` and then in then in the reading on the envvars use `bool(int(os.getenv(...)))`. It's not the prettiest but works

nsantacruz:
> > @BrendanGalloway added some comments. I wasn't able to see where you set the environment variables `LINKER_ENABLED` and `DISABLE_AUTOCOMPLETER` for the web pods (I see where you're doing it for the linker pod).
> 
> I have not set them for web, it should just fall through to the default instead.

Makes sense

helm-chart/sefaria-project/templates/rollout/linker.yaml
BrendanGalloway: I see there are some other places where we have used a string compare - for consistency I will do so here too

helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
```diff
@@ -179,11 +179,10 @@ data:
 
     # Turns off search autocomplete suggestions, which are reinitialized on every server reload
     # which can be annoying for local development.
-    DISABLE_AUTOCOMPLETER = False
-    ENABLE_LINKER = False
+    DISABLE_AUTOCOMPLETER = os.getenv("DISABLE_AUTOCOMPLETER", "false") == "true"
+    ENABLE_LINKER = os.getenv("ENABLE_LINKER", "false").lower == "true"
```
nsantacruz, line 183: @BrendanGalloway I think you meant `.lower()` with parentheses. Also, should you use `.lower()` on the line above as well?

helm-chart/sefaria-project/templates/rollout/linker.yaml
```diff
@@ -0,0 +1,191 @@
...
+          - name: REDIS_HOST
+            value: "redis-{{ .Values.deployEnv }}"
+          - name: NODEJS_HOST
+            value: "node-{{ .Values.deployEnv }}-{{ .Release.Revision }}"
+          - name: VARNISH_HOST
+            value: "varnish-{{ .Values.deployEnv }}-{{ .Release.Revision }}"
+          - name: HELM_REVISION
+            value: "{{ .Release.Revision }}"
+          - name: ENABLE_LINKER
+            value: "True"
```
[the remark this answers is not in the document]
nsantacruz: missed that! all good.

PR #873, opened 2022-06-06
stub for redirecting find-refs api to linker

nsantacruz:
@BrendanGalloway here's a stub for the redirect to the linker pod for the relevant API call. We need to define an environment variable for the location of the linker pod. I left that as a placeholder for now.


nsantacruz:
@BrendanGalloway great!

PR #944, opened 2022-07-28
feat: Search in book

rneiss:
Adds the ability to search in a book directly from the connection panel.


nissamai (changes requested):
Comments on trello card!

nissamai (approved):

PR #953, opened 2022-08-10
Maintain scroll position in sidebar

EliezerIsrael:
This feature aims to maintain the user's scroll position in lists of commentators.  The user will scroll down to e.g. Rashbam, click into Rashbam, and then click back.  The position in the list of commentators will be retained. 

Similarly the position in the list of Websites will be maintained as the user clicks into the webpage list and then back. 

The scroll position in sheets is maintained as well, but the use case is more limited.  Clicking on a sheet closes the sidebar, and the scroll state is lost.  But if a user scrolls through sheets then navigates up to resources and back to sheets, the position is maintained. 

Review of this feature should pay attention to any impact on user history collection.  There was a pre-existing scroll event on the connections panel that was used to see what texts were in the viewport for the sake of collecting history.  

There is also a top level scroll event in effect on the .content div, which is a generic function for handling <back> actions.  That shouldn't be impacted by this.

### DIFF
--- a/static/js/ConnectionFilters.jsx
+++ b/static/js/ConnectionFilters.jsx
@@ -20,7 +20,7 @@ class CategoryFilter extends Component {
     }
   }
   render() {
-    const filterSuffix = this.props.category  == "Quoting Commentary" ? "Quoting" : null;
+    const filterSuffix = this.props.category  === "Quoting Commentary" ? "Quoting" : null;
     const textMissingDescription = null; //"missing description"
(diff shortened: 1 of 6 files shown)


EliezerIsrael:
There's also some no-op code style fixes in here.  Sorry for the clutter.  The commits are split though - only the first commit has the code cleanup.

nissamai (approved):

EliezerIsrael:
After review, spec for this was adjusted.  Rather than remembering the scroll position for every sidebar view, we will remember only the scroll position for the immediate parent of the current view.

PR #1012, opened 2022-09-14
Bdb

YishaiGlasner:


static/js/LexiconBox.jsx
```diff
@@ -303,18 +342,21 @@ class LexiconEntry extends Component {
       </span>);
     }
 
-    var entryHeadHtml = (<span className="headline" dir="ltr">
-      {headwords
+    var headwordString = RegExp(/^BDB.*?Dict/).test(entry['parent_lexicon']) ? this.bdbHeadwordString(entry) : headwords
```
nsantacruz, line 345: Can you write a function to calculate the default headword string?

nsantacruz: Also, can you calculate a variable `isBDB` instead of calculating inline?

PR #1022, opened 2022-09-21
ci(deploy): abstract out image names for contextus fork

BrendanGalloway:
@EliezerIsrael We will need IMAGE_NAME secret set with the appropriate values for each project in the corresponding repos before this merges, or deploy will break


EliezerIsrael (approved):
Added IMAGE_NAME.

PR #1035, opened 2022-10-03
Use enumerate instead of range(len(...)

yuvalabou:
Use enumerate instead of range(len(...) where applicable


nsantacruz:
Given that this needs to unpack a tuple, I actually prefer `range(len())`. They're both using iterators so I imagine they're similar speeds. @edamboritz what do you think?

nsantacruz:
@yuvalabou after discussing with @edamboritz, we believe this change is unnecessary. It doesn't offer a significant speed boost (unclear if there's any speed boost actually) and it's arguable whether `enumerate()` is more pythonic or not.

Closing for now but happy to hear if we missed anything.

PR #1038, opened 2022-10-05
Don't use bare except E722

yuvalabou:
The overall exception handling in Sefaria is a mishmash of either incomplete or not up to standards.
I will try my best to improve that.

[x] E722
[] Format the exceptions string
[] Remove or improve exceptions that now we just ignore.
[] Use logger instead of print

Maybe -
[] Sefaria exception classes?


nsantacruz:
@yuvalabou this looks like it will be helpful. I see the PR is a draft. Can we set up a time to discuss the other goals of this PR? I would like to coordinate what the best approach is. You can respond to the email I sent you to set up a time.

PR #1045, opened 2022-10-23
Move all import to toplevel in ./scripts

yuvalabou:
Move all imports to toplevel in Sefaria ./scripts.
This reduces lines and duplicated imports as well as code readability overall.
This is also part of the PEP8 Standard (C0415)


nsantacruz:
Thanks! In general, moving imports to the top of the file may cause circular imports so we need to be careful with this. However, no files should be importing files in scripts.

PR #1046, opened 2022-10-23
Move all import to toplevel in ./reader

yuvalabou:
Move all imports to toplevel in Sefaria ./reader.
This reduces lines and duplicated imports as well as code readability overall.
This is also part of the PEP8 Standard (C0415)


nsantacruz:
@yuvalabou tests pass with this change. merging.

PR #1062, opened 2022-10-27
Don't allow more than one statement on a single line

yuvalabou:
Indenting the blocks to make the code mor readable and clean.
[PEP8 ](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/multiple-statements.html)


yuvalabou:
Is this still planned?

PR #1065, opened 2022-11-02
Autocompleter

stevekaplan123:


EliezerIsrael:
Purported Scope:

* Topic Autocompleter 
  * Frontend in TopicSearch.jsx / Misc.jsx
  * Backend in alongside other autocompleters in the Library
* Ref autocompleter is used in source sheets - it was refactored here
* Topic autocompleter is used in topic search - which shows up in connections panel, for connecting a ref to a topic, 
* On topic edit - don't run `admin/reset/toc`, but instead reset topic cache from within the view
* Add Hebrew fields in topic editor

EliezerIsrael (approved):

PR #1094, opened 2022-11-10
Scrollposition

EliezerIsrael:
This is a refactor of the PR originally in https://github.com/Sefaria/Sefaria-Project/pull/953

Behavior was changed to only record the scroll position on the most recently visited sidebar view.  Scroll position resets when the user navigates to Resources.  

Scroll position is recorded on connections lists (text categories), webpages site list, and sheets list.  

Because sheet open actions close the sidebar, the impact of this feature is minimal in the sheets case.


nissamai (approved):
Approved, sent comments via Slack.

PR #1109, opened 2022-11-16
Update webpage.py

orenfromberg:
This change request addresses the following warning by checking for value instead of reference equality:
```
/www/sefaria/model/webpage.py:131: SyntaxWarning: "is" with a literal. Did you mean "=="?
   if len(bad_urls) is 0:
```


nsantacruz:
@orenfromberg thank you for the PR! This looks like a good change.

PR #1115, opened 2022-11-17
feat(linker): enable linker on production

nsantacruz:
I think these are the right changes to enable linker but would appreciate review.

### DIFF
--- a/helm-chart/sefaria-project/values.yaml
+++ b/helm-chart/sefaria-project/values.yaml
@@ -105,7 +105,7 @@ web:
       ref:
 
 linker:
-  enabled: false
+  enabled: true
   # key-pair values to load into web pod environment.  Takes precedence over global localsettings
   model_paths:
     # en: future_path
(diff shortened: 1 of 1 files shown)


helm-chart/sefaria-project/values.yaml
```diff
@@ -105,7 +105,7 @@ web:
       ref:
 
 linker:
-  enabled: false
```
BrendanGalloway, line 108: The intention is not to enable the linker as default behaviour, which is what this does.  The production specific values are in build/ci/production-values.yaml. I need to add a couple extra lines to the deploy script to populate the deploy tag for the linker too, I will add enabling the linker to that change

BrendanGalloway (changes requested):

helm-chart/sefaria-project/values.yaml
nsantacruz: Got it. Sorry for the mix-up. We still need the change to the linker image though below, right? that should be default for all usages of linker.

PR #1119, opened 2022-11-18
ci: remove selenium tests

BrendanGalloway:


EliezerIsrael:
@BrendanGalloway Looks good.  Just to be sure - we still need all of this sandbox setup for the pytest suite, yes?  Are there any aspects of the sandbox setup that were just for Selenium tests?

EliezerIsrael (approved):

PR #1120, opened 2022-11-18
feat: create Dockerfile for standalone development database container

BrendanGalloway:
This ended up being a bit more difficult that I originally thought because of some fine grained stuff about how docker build works and what the mongo image does upstream.  Essentially docker is told to ignore anything that ends up in /data/db because it's expected that the container will be connected to some form of persistent volume at that location, because generally DB contents are explicitly not supposed to be in an image.  I worked around this by shifting the whole DB to a new location.  There's a couple caveats now though:

If a dev wants to run this image with any funny commandline overrides to mongo (eg hosting it  on a different port), they'll also need to know to point mongo at the new database location.
Devs would need to know they need to take snapshots of the image if they want to keep db changes between restarts of the container ( note these can increase in size pretty rapidly)

As an alternative, the solution to this out the box dev experience would be a script that starts the upstream mongo container with a persistent local volume configured and does a restore into it from the latest dump?  Maybe not as 'in a can' convenient, but might make for a more amenable  workflow for development actions?

In the meantime can you build the dockerfile and run the resultant image with -p 27017, then connect to it and check it has all the content youexpect?

### DIFF
--- a/.github/workflows/weekly.yaml
+++ b/.github/workflows/weekly.yaml
@@ -5,5 +5,43 @@ on:
     - cron: '0 0 * 6 *'
 
 jobs:
-
+  build-mongo:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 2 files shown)


paddatrapper (approved):

EliezerIsrael:
I built this and ran it w/ ports mapped.  Worked perfectly.

PR #1217, opened 2023-01-23
helm(feat): deploy standalone dev mongo

paddatrapper:
~Still need to override the standard configuration when this is enabled~ (done)

### DIFF
--- a/helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
+++ b/helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml
@@ -51,13 +51,20 @@ data:
 
     MONGO_HOST = os.getenv("MONGO_HOST")
     MONGO_PORT = 27017
-    {{- if eq .Values.sandbox "true" }}
+    {{- if .Values.standaloneMongo.enabled }}
+    SEFARIA_DB = "sefaria"
+    {{- else if eq .Values.sandbox "true" }}
     SEFARIA_DB = os.getenv("SEFARIA_DB") + "-" + {{ .Values.deployEnv | quote }}
     {{- else }}
     SEFARIA_DB = os.getenv("SEFARIA_DB")
     {{- end }}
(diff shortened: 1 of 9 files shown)


helm-chart/sefaria-project/templates/service/mono.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if eq .Values.standaloneMongo.enabled true }}
```
BrendanGalloway, line 1: filename is mono no mongo?

BrendanGalloway (changes requested):
We also need a way to override the localsettings input releated to:

```
    MONGO_HOST = os.getenv("MONGO_HOST")
    MONGO_PORT = 27017
    {{- if eq .Values.sandbox "true" }}
    SEFARIA_DB = os.getenv("SEFARIA_DB") + "-" + {{ .Values.deployEnv | quote }}
    {{- else }}
    SEFARIA_DB = os.getenv("SEFARIA_DB")
    {{- end }}
    SEFARIA_DB_USER = os.getenv("SEFARIA_DB_USER")
    SEFARIA_DB_PASSWORD = os.getenv("SEFARIA_DB_PASSWORD")
    APSCHEDULER_NAME = os.getenv("APSCHEDULER_NAME")```
    
    to point to the standalone instance and the db that is already present in it...

File: helm-chart/sefaria-project/templates/configmap/local-settings-file.yaml [hunk 2]
```diff
@@ -51,13 +51,20 @@ data:
 
     MONGO_HOST = os.getenv("MONGO_HOST")
```
BrendanGalloway, line 52: We need to force mongo host to be `mongo-standalone-{{ .Values.deployEnv }}`

BrendanGalloway (changes requested):

BrendanGalloway (approved):

PR #1239, opened 2023-02-07
fix: log real IP in nginx logs

EliezerIsrael:


BrendanGalloway (approved):

EliezerIsrael:
Underscores in name are giving helm job indigestion.

PR #1360, opened 2023-05-09
Enable display of images in version-texts

yonadavGit:


yonadavGit:
I replaced "ContentText" element in places where an actual version text is loaded with a new element called "VersionContent". This element in turn performs the logic of deciding if the current version-segment is an image or not, and returns one of 2 new elemets - VersionImage or VersionText accordingly.

PR #1370, opened 2023-05-16
making search result sources highlightable, only source-title clickable

yonadavGit:


edamboritz (approved):
This PR should be squashed

PR #1377, opened 2023-05-22
Bug: fix copy past for non html elements

akiva10b:
Bug: handleCopyEvent event is only meant to handle copy of HTML elements, when no HTML element is present the copied text is lost.

Fix checks if string is HTML element, and if it isn't we allow the browser to continue as usual


edamboritz (approved):

jatkinson72 (commented):
Trust

PR #1383, opened 2023-05-23
Fix copy past

akiva10b:


static/js/ReaderApp.jsx
```diff
@@ -1856,6 +1856,10 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
 
   handleCopyEvent(e) {
     // Custom processing of Copy/Paste
+    if (e.srcElement.tagName === "INPUT") {
```
rneiss, line 1859: The issue at the core here I think is that `window.getSelection()` only works with selections in text nodes or elements, and therefore an `input` element doesn't qualify (It's probably also the same for say `textarea` and maybe some other places.) 

I think the cleaner thing here is just to see if the value of `textOnly` down below is an empty string, and if so to return.

akiva10b: that should work just fine, the core issue of course is that we are making exceptions to a rule that itself should be an exception.

akiva10b: An issue that can arise here if we implement the function on an input is that the empty string we get from the function isn't actually empty, it is a whitespace of ' ' - this means that even if we trim the text or parse it to check that everything is a whitespace, we can still have a future addition of "copied from sefaria" that would break the exception

rneiss: Perhaps then we should use `select()` or `setSelectionRange()`  -- although as I mentioned on Slack, I believe that the rangeCount in this situation defaults to 0 which means that we can simply just append a return on the else

```
    if (selection.rangeCount) {
      ....
    }
else {
   return
}
```

If you want to structure it so that we put the rule first then probably something like the following is best --

```
    const selection = document.getSelection()
    if (!selection.rangeCount) {
      return
   }

   else {

    const closestReaderPanel = this.state.panels.length > 0 && e.target.closest('.readerPanel') || null
    let textOnly = selection.toString();
    let html = textOnly;
    let selectedEls;
   
    [... All the rest of the code currently in selection.rangeCount...]

}

akiva10b: `selection.rangeCount` is still 1 when getting the data from an input, any other ideas of something generic? or will we stoop to putting in exceptions for the input/textarea? I am not familiar enough with other use-cases to know if this is hurting us elsewhere
Screenshot 2023-05-24 at 9 11 03

rneiss:
Screenshot from 2023-05-24 07-39-11

This solves the issue in Chrome. 

`getSelection()` still doesn't select the content in FF however. This is a known issue: https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection#related_objects

rneiss (approved):
I think the approach I outlined in the last comment is better. But this doesn’t break anything and I know product wants to keep this moving.

Issue #1387, opened 2023-05-29
Select version history with Bulktext API


ronshapiro:
I'd love a way to select the version of texts selected with the /api/bulktext endpoint, e.g. pin a specific tanakh translation, similar to the `ven` and `vhe` parameters in `/api/texts`.

I see the TextChunk class has a `vtitle` parameter - perhaps that can be leveraged?

I am fine if the API is strict - i.e. if I have tanakh sources and other sources that don't have the same version title, I can split them into separate requests. But I would ideally like to be able to group sources of similar types so that I can minimize requests; the `/api/texts` API already gives the flexibility but I'm requesting too many texts for that to be practical.

PR #1388, opened 2023-05-29
Support selecting versions in /api/bulktext. 

ronshapiro:
Fixes https://github.com/Sefaria/Sefaria-Project/issues/1387

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -44,7 +44,7 @@ from sefaria.model.collection import CollectionSet
 from sefaria.export import export_all as start_export_all
 from sefaria.datatype.jagged_array import JaggedTextArray
 # noinspection PyUnresolvedReferences
-from sefaria.system.exceptions import InputError
+from sefaria.system.exceptions import InputError, NoVersionFoundError
 from sefaria.system.database import db
 from sefaria.system.decorators import catch_error_as_http
 from sefaria.utils.hebrew import is_hebrew, strip_nikkud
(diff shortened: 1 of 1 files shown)


sefaria/views.py
```diff
@@ -379,8 +381,13 @@ def bundle_many_texts(refs, useTextFamily=False, as_sized_string=False, min_char
                     'url': oref.url()
                 }
             else:
-                he_tc = model.TextChunk(oref, "he", actual_lang=translation_language_preference)
-                en_tc = model.TextChunk(oref, "en", actual_lang=translation_language_preference)
+                he_tc = model.TextChunk(oref, "he", actual_lang=translation_language_preference, vtitle=hebrew_version)
+                en_tc = model.TextChunk(oref, "en", actual_lang=translation_language_preference, vtitle=english_version)
+                if hebrew_version and he_tc.is_empty():
+                  raise ValueError(f"{oref.normal()} does not have the Hebrew version: {hebrew_version}")
```
ronshapiro, line 387: Is there a better error type to raise here so that the user sees the error?

ronshapiro: Done :)

edamboritz:
I'm inclined to approve this because it seems to be a small enough change that will bring benefit to API users. 

That being said, the bulktext api endpoint suffers from the same problems as the current main texts api endpoint, which is that it makes some outdated assumptions about versions. 

It was also not designed primarily to be used too often, outside of the few places inside Sefaria it's currently used, as caching its responses is difficult by nature.

edamboritz (approved):

ronshapiro:
I know not to expect much from the bulktext API. This felt like a conveniently small change.

I'm indifferent to caching - the requests I make have no need to be cached. They should be quite unique.

ronshapiro:
Thanks! How often are releases so I know when it's in prod?

בתאריך יום ה׳, 1 ביוני 2023, 12:11, מאת Ephraim Damboritz ‏<
***@***.***>:

PR #1392, opened 2023-05-30
static(modals): shavuot modal down

shaneerosen:
Take down Shavuot modal 2023


static/mobile/message-he.json
relyks, line 1: Possible the same issue

relyks (changes requested):
Should this be changed?

relyks (approved):

PR #1403, opened 2023-06-04
fix(ref test): skip Zohar tests (that are for Talmud at second place …

YishaiGlasner:
…indexes which we currently have none), change Zohar refs in other tests.


akiva10b:
opened a new branch instead

PR #1407, opened 2023-06-05
Playwright integration

akiva10b:


.github/workflows/continuous.yaml
```diff
@@ -266,6 +266,79 @@ jobs:
...
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 30
+      - name: Authenticate GHA Runner to GCP
+        uses: google-github-actions/setup-gcloud@v0
+        with:
+          service_account_key: ${{ secrets.GCS_BUILD_UPLOAD_KEY }}
+          project_id: ${{ secrets.BUILD_UPLOAD_PROJECT }}
+        if: ${{ always() }}
```
BrendanGalloway, line 341: Rather make sure that the runner JWT has the appropriate set of permissions to perform the requisite actions?  If this pipeline needs a different set of permissions to the regular github actions that you don't want to have stray into other jobs, then we should set up a second WIF binding that is only accessible in this workflow

BrendanGalloway (changes requested):

.github/workflows/integration-testing.yaml
```diff
@@ -1,7 +1,8 @@
 name: Run Integration Tests
 on:
-  workflow_dispatch: {}
-  merge_group: {}
+  push:
+    branches:
+      - master
```
BrendanGalloway, line 5: This will only run after a merge has completed as written.  That is, the tests will only highlight failures on code that is already in master...  The original trigger intentions were:
- workflow dispatch to allow engineers to trigger on demand tests on a branch
- merge_group to run the tests after merge approval, but before merge completes.

.github/workflows/integration-testing.yaml
```diff
@@ -188,7 +189,33 @@ jobs:
...
+      - name: Set SANDBOX_URL
+        run: echo "SANDBOX_URL=http://nginx-integration-${{ needs.variables.outputs.commit }}.${{ secrets.DEV_SANDBOX_NAMESPACE }}/texts" >> $GITHUB_ENV
+      
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
```
BrendanGalloway, line 202: This is a little bit old - github is in the process of upgrading all runners to node 20 by default at the moment, actions that expect node to be present will start to expect that soon.  It might also be worth baking node into the runner image, save some time during testing

.github/workflows/integration-testing.yaml
```diff
@@ -188,7 +189,33 @@ jobs:
...
         run: while [[ $( curl -s -o /dev/null -w "%{http_code}" "nginx-integration-${{ needs.variables.outputs.commit }}.${{secrets.DEV_SANDBOX_NAMESPACE}}/texts") != "200" ]]; do sleep 10; done
         timeout-minutes: 5
-      - name: Run tests
-        run: if [[ $( curl -s -o /dev/null -w "%{http_code}" "nginx-integration-${{ needs.variables.outputs.commit }}.${{secrets.DEV_SANDBOX_NAMESPACE}}/texts") != "200" ]]; then exit 1; fi
+
+      - name: Set SANDBOX_URL
+        run: echo "SANDBOX_URL=http://nginx-integration-${{ needs.variables.outputs.commit }}.${{ secrets.DEV_SANDBOX_NAMESPACE }}/texts" >> $GITHUB_ENV
+      
+      - name: Checkout Code
+        uses: actions/checkout@v2
```
BrendanGalloway, line 197: THe code is already checked out, not sure why it needs to be checked out again

package.json
```diff
@@ -67,6 +67,7 @@
     "jquery.scrollto": "^2.1.2",
     "js-cookie": "^2.2.1",
     "patch-package": "^6.2.2",
+    "playwright": "^1.36.2",
```
BrendanGalloway, line 70: I believe the packages.json can be setup to indicate certain packages are dev/ci only so they don't get installed in the deploy node image?

.github/workflows/integration-testing.yaml
akiva10b: ah yes, that is a better flow, Ill revert it back

.github/workflows/integration-testing.yaml
akiva10b: makes sense - what do we need to do in order to get it into the image? that would save us a lot of time

.github/workflows/integration-testing.yaml
BrendanGalloway: Great - note that you will need to enable merge queues on this repo.  I can't remember if that has been done already @edamboritz

BrendanGalloway (dismissed):

BrendanGalloway (approved):

PR #1416, opened 2023-06-07
Crm 6084 - mark delete

nissamai:
Adds functionality to mark users in CRM when deleted.

https://app.shortcut.com/sefariatest-1/story/6084/

### DIFF
--- a/sefaria/helper/crm/salesforce.py
+++ b/sefaria/helper/crm/salesforce.py
@@ -110,7 +110,9 @@ class SalesforceConnectionManager(CrmConnectionManager):
         except:
             # log
             return False
-        return res
+
+    def mark_for_review_in_crm(self, crm_id):
+        return self.mark_as_spam_in_crm(crm_id)
 
     def find_crm_id(self, email=None):
         if email:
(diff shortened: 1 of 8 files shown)


sefaria/helper/crm/salesforce.py
```diff
@@ -110,7 +110,9 @@ def mark_as_spam_in_crm(self, uid):
         except:
             # log
             return False
-        return res
+
```
shaneerosen, line 113: Do we want to add the logging here?

shaneerosen (commented):
How did you test this code?

sefaria/utils/user.py
```diff
@@ -18,6 +22,13 @@ def delete_user_account(uid, confirm=True):
             print("Canceled.")
             return
 
+    try:
```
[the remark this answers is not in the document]
nissamai: I *think* that return will only get hit if the user doesn't type in "DELETE" when confirm==True? (the lines with the if block somehow get cut out of the github changes view)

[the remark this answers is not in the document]
shaneerosen: yes! git misled me! I was surprised, now it makes sense. sorry!

sefaria/helper/crm/salesforce.py
shaneerosen: no. I actually like it better in user.py, but then let's delete that comment as well. and now that line 28 &30 are actually running things make more sense.

nissamai: Ok, removed all the log comments! Thanks shanee!

shaneerosen (approved):

PR #1430, opened 2023-06-12
Chore add commit msg hook

edamboritz:


.githooks/commit-msg
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+# This regular expression represents the conventional commit format, including
+# support for custom semantic commits message parts, e.g., 'feat(custom-prefix-123): ...'.
+CONVENTIONAL_COMMIT_REGEX="^((build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test|static|helm)((\([a-zA-Z0-9\-_]*\))|(:))?(\!)?: .+)|(Merge .+)|(Revert .+)$"
```
BrendanGalloway, line 5: Not sure why you have an inner 'scope or semicolon' check? Should just be scope?
if you want to enforce scoping, you need to remove the ? from after the scope or semicolon check

edamboritz: Ah, my idea was initially to not enforce scope? But you do have a point.

BrendanGalloway: up to you whether scope isencouraged or mandatory

BrendanGalloway (approved):

akiva10b (approved):

PR #1435, opened 2023-06-14
Api versions v3

YishaiGlasner:


sefaria/urls.py
```diff
@@ -13,6 +13,7 @@
 import sourcesheets.views as sheets_views
 import sefaria.gauth.views as gauth_views
 import django.contrib.auth.views as django_auth_views
+import api.views as v3_views
```
edamboritz, line 16: I think this is going to get confusing later on. We really should not be naming things inside the code as "v3". Rather I suggest something more sustainable and straightforward like "api_views"

api/api_errors.py
```diff
@@ -0,0 +1,35 @@
+import django
+django.setup()
+from sefaria.model import *
+from typing import List
+
+
+class APIError():
```
edamboritz, line 7: The way this has been implemented is great. Very orderly and clear. 
That being said, I have a concern we might be giving the impression that this is used similarly to an exception, which it is not. The naming scheme here may cause some confusion. Suggest we change all errors to "response"? 

Also The subclasses all seem to be unique to the texts api, so I would suggest we further group them under a parent like that. Otherwise in the future this might become more and more confusing for incoming developers to use.

sefaria/urls.py
YishaiGlasner: why not v3? it's officialy v3, so I think we can name it like this. on the other hand, it's not the only api views we'll have, so api_views seems confusing to me. as long as we use api from more than one version I think naming it after its version is the most informative.

api/api_errors.py
YishaiGlasner: you're right, but we're using them to flag some 'error', and Response does not reflect it. also response is a 'saved' term for the whole http response.
i'm thinking aloud - we're using it for flagging that some (or all) of the request includes things we're missing. so maybe APIMissingMessage or APILackMessage? @edamboritz @akiva10b what do you think?

api/helper.py
```diff
@@ -0,0 +1,14 @@
+from typing import List
+
+def split_at_pipe_with_default(string: str, list_length: int, defaults: List[str]) -> List[str]:
```
edamboritz, line 3: This function look like it handles text api urls parameters? 
I think we should name it according to waht it does, in a way that makes it simple and clear to understand. The current name is perhaps too specific, to the point that its not clear where the function fits in the overall design. 

The variables inside the function can be called something more specific than "substrings" to explain what they are more clearly. 

I'm also not sure why its not grouped with the rest of the Texts API specific code? 
I like that we have a module for helper functions, but if its not reusable code, we might want to consider saving a helper module for functions such as those. 

Also see my lower comments on using class based Django views.

YishaiGlasner: i'm changing the names
it is not reusable code, for we have just one endpoint in v3 for now. when we'll have more. we'll probably use pipes.
i see no comment on using Django (catually i see just 3 comments)

api/api_errors.py
akiva10b: we spoke about this earlier as well, is a bad ref or no version really an "error", it it an HTTP 404 error or just an empty response?
This is an area with tradeoffs and I would vote for whatever jives best with what devs are familiar with from API documentation. For instance, in https://www.weatherapi.com/docs/ which is somewhat similar to us in the sense that you can ask for data for multiple parts and only receive some, "errors" only refer to something malfunctioning in the API or the constraints around it (expires key, bad json formatting etc.)
The unifying metric here is "unavailableData" so I would name the class UnavailableDataResponse and so on for the rest of the classes NoVersionResponse... @YishaiGlasner what do you think?

akiva10b: it likely also makes sense to put that data in a "unavailable_data" parameter in the API (we are putting it in "errors" now I think which is not 100% correct)

YishaiGlasner: i think respone is not the right word here. it's a part of the response. UnavailabeDataMessage?

PR #1437, opened 2023-06-14
fix(Trello Job): inc memory limit

stevekaplan123:


BrendanGalloway (approved):

BrendanGalloway (changes requested):
@stevekaplan123 changes to the helm chart need to be committed with `helm(): ` otherwise the versioning system won't be able to parse that helm-specific changes have occured

BrendanGalloway (approved):

PR #1457, opened 2023-06-20
fix: Force a fix of previously reverted code

edamboritz:


sefaria/utils/util.py
```diff
@@ -458,6 +458,7 @@ def truncate_string(string, min_length, max_length):
         return string
     for break_char in ".;, ":
         # enumerate all places where this char is in segment
+        #Note: this code was reverted to because the change that was made to it affected the related api whule fetching webpages.
```
nsantacruz, line 461:
```suggestion
        #Note: this code was reverted to because the change that was made to it affected the related api while fetching webpages. 
```

nsantacruz (approved):

PR #1470, opened 2023-06-22
feat(error tracking): sentry integration

akiva10b:


templates/base.html
```diff
@@ -171,6 +171,17 @@
 
 
     {% autoescape off %}
+
```
edamboritz, line 174: I have some concerns about the way are are passing the data between the server and client here, and that it doesnt match our latest practices, but this should work, at least initially.

sefaria/settings.py
```diff
@@ -316,7 +316,7 @@
 # GLOBAL_INTERRUPTING_MESSAGE = None
 
 
-
+SENTRY_DSN = os.getenv("SENTRY_DSN", None)
```
edamboritz, line 319: I believe this shoud be placed in the local settings file as templated by the helm chart? 
This would make it consistent with how we manage other similar variables and make the management of environment specific customization easier.

edamboritz (changes requested):
Looks good overall!

a few notes

sefaria/local_settings_example.py
```diff
@@ -309,3 +309,6 @@
     wrapper_class=structlog.stdlib.BoundLogger,
     cache_logger_on_first_use=True,
 )
+
+SENTRY_DSN = os.getenv("SENTRY_DSN", None)
+CLIENT_SENTRY_DSN = os.getenv("CLIENT_SENTRY_DSN", None)
```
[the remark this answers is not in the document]
edamboritz: Also add the other values noted above

sefaria/settings_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+        # If there is no exception, drop the event
+        if 'exception' not in event:
+            return None
+
+        return event
+
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        environment=os.getenv("SENTRY_ENVIRONMENT", "unknown"),
```
edamboritz, line 24: Here we should be referencing the variables out of local settings, not the envvars directly. Parsing them from envvars should be done in uniformity with the rest, in local_settings

sefaria/settings_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        integrations=[DjangoIntegration()],
+        traces_sample_rate=1.0,
+        send_default_pii=False,
+        before_send=before_send,
+        max_breadcrumbs=30,
+        release=os.getenv("SENTRY_CODE_VERSION", "unknown"),
```
edamboritz, line 30: Same comment as above

static/js/client.jsx
```diff
@@ -5,9 +5,29 @@ import React from 'react';
...
 
 $(function() {
+  // Initialize Sentry, sentryDSN is defined in base.html
+  if (sentryDSN) {
+    Sentry.init({
+      dsn: sentryDSN,
+      integrations: [
+        new Sentry.BrowserTracing({
+          // Set `tracePropagationTargets` to control for which URLs distributed tracing should be enabled
+          tracePropagationTargets: ["localhost", /^https:\/\/sefaria\.org\/api/],
```
edamboritz, line 19: I'm not sure what this does, but should the host here also be customiseable  for use in non prod environments?

edamboritz (commented):
Looks good! 
I think we should jsut patch up the comments I made about variable placement and where they get read from the env

static/js/client.jsx
akiva10b: this looks at client side API requests and captures failures, the idea of the regex is to lower the endpoints we follow that. However I see in their docs that this feature is deprecated so I will remove the tracePropagationTargets

sefaria/settings_utils.py
akiva10b: this function is being implemented on init in the settings.py file, but I do hear what you are saying - so I expanded the vars it will accept

helm-chart/sefaria-project/templates/configmap/local-settings.yaml
```diff
@@ -24,3 +24,7 @@ data:
   SERVER_EMAIL: "{{ .Values.localSettings.SERVER_EMAIL }}"
   MULTISERVER_ENABLED: "{{ .Values.localSettings.MULTISERVER_ENABLED }}"
   CRM_TYPE: "{{ .Values.localSettings.CRM_TYPE }}"
+  SENTRY_ENVIRONMENT: {{ .Values.deployEnv | quote }}
+  SENTRY_CODE_VERSION: {{ .Values.web.containerImage.tag }}
+  SENTRY_DSN: {{ .Values.localSettings.SENTRY_DSN | quote }}
```
edamboritz, line 29: Are these values meant to be secrets? Because if so, I don't think they are defined correctly here.

sefaria/settings.py
```diff
@@ -330,7 +330,9 @@
 
 
 if SENTRY_DSN:
-    init_sentry(SENTRY_DSN)
+    SENTRY_CODE_VERSION = os.getenv("SENTRY_CODE_VERSION", "unknown")
+    SENTRY_ENVIRONMENT = os.getenv("SENTRY_ENVIRONMENT", "unknown")
```
edamboritz, line 334: We still are left with a problem here where this is still creating a new pattern where local developers are required to define environment variables, something not required before. 

Perhaps this entire function call should be in local_settings rather than settings (since it is something that changes in between environments)

edamboritz (approved):

PR #1474, opened 2023-06-27
Remove profile messaging

saengel:
Remove the ability for users to message each other from one profile to another.


scripts/purge_messages_db.py
```diff
@@ -0,0 +1,18 @@
...
+# In cleaning out old Messages code, this purges any remaining messages in the database
+import django
+django.setup()
+from sefaria.model import NotificationSet
+
+def purge_messages():
+    count = 0
+    ns = NotificationSet({'type': 'message'})
+    print(f"There are {len(ns)} messages in the database which are about to be purged")
+    for message in ns:
```
edamboritz, line 11: I dont think its critical here, but in general there is a slightly more performant way to delete a MongoSet of records: 
See `AbstractMongoSet.delete(bulk_delete=True)`

edamboritz (commented):
The removal of code here looks like it was done thoughtfully and thoroughly! 

I think I may have caught a few places that seem to me like they also should be removed? 
Could be there was a reason they were left so will just point them out in case: 

The component `MessageModal` in Misc.jsx and its associated functionality and imports (UserProfile.jsx and UpdatesPanel.jsx as well)

Also some additional sections of code in UserProfile.jsx? (noted below)
https://github.com/Sefaria/Sefaria-Project/pull/1474/files#diff-c6b1a9749b69e2c9142df231d1de6d239bc89758b22f9037247c9b85227f3f8eR15
https://github.com/Sefaria/Sefaria-Project/pull/1474/files#diff-c6b1a9749b69e2c9142df231d1de6d239bc89758b22f9037247c9b85227f3f8eR54
https://github.com/Sefaria/Sefaria-Project/pull/1474/files#diff-c6b1a9749b69e2c9142df231d1de6d239bc89758b22f9037247c9b85227f3f8eR331
https://github.com/Sefaria/Sefaria-Project/pull/1474/files#diff-c6b1a9749b69e2c9142df231d1de6d239bc89758b22f9037247c9b85227f3f8eR358
https://github.com/Sefaria/Sefaria-Project/pull/1474/files#diff-c6b1a9749b69e2c9142df231d1de6d239bc89758b22f9037247c9b85227f3f8eR445

scripts/purge_messages_db.py
saengel: Okay great, thank you for pointing that out @edamboritz - very helpful.

edamboritz (approved):

akiva10b (approved):

PR #1495, opened 2023-07-05
fix(admin): don't autocomplete form to delete user accounts [sc-18897]

nissamai:


templates/registration/delete_user_account.html
```diff
@@ -10,7 +10,7 @@
         <h1>
             <span class="int-en">Delete User</span>
         </h1>
-        <form method="post" if="delete-form" action="{% url "delete/user-account" %}">
+        <form method="post" autocomplete="off" if="delete-form" action="{% url "delete/user-account" %}">
             {% csrf_token %}
             {{ form.as_p }}
```
relyks, line 15: If there were input elements generated here, the same attribute you added might be needed

relyks (approved):

PR #1502, opened 2023-07-11
Gematria

stevekaplan123:


static/js/sefaria/hebrew.js
```diff
@@ -20,19 +21,109 @@ class Hebrew {
...
     }
 
-    var values = this.hebrewNumerals;
+    const values = this.hebrewNumerals;
 
-    var heb = "";
+    let heb = "";
     if (n >= 100) {
-      var hundreds = n - (n % 100);
+      const hundreds = n - (n % 100);
```
akiva10b, line 126: love the refactor!

static/js/sefaria/hebrew.js
```diff
@@ -20,19 +21,109 @@ class Hebrew {
...
+
+	breakIntMagnitudes(15000)
+	[10000, 5000, 0, 0, 0]
+     */
+    if (!!start) {
+      if (!(start % 10 === 0 || start === 1)) {
+        throw new TypeError(`Argument 'start' must be 1 or divisible by 10, ${start} provided.`);
+      }
+    } else {
+      start = Math.pow(10, Math.floor(Math.log10(n)));
```
akiva10b, line 47: it seems like start can have a default of 10 instead of null and that way this line can just be taken care of in the `const thisIntMagnitude = Math.floor(n / start) * start;`
Am I wrong?

stevekaplan123: I'm not sure what you mean.  If 'n' is 1250, start should be 1000, and if 'n' is 120, 'start' should be 100.  Why should it default to 10?

akiva10b (approved):

PR #1503, opened 2023-07-11
feat(topic editor): decouple components and make them reusable

akiva10b:


static/js/Misc.jsx
```diff
@@ -1164,18 +1180,21 @@ const ReorderEditorWrapper = ({toggle, type, data}) => {
     is a dictionary of the topic whose sources can be accessed via its `refs` field.
      */
     const reorderingSources = data.length !== 0;
-    const filterAndSortRefs = (refs) => {
+    const _filterAndSortRefs = (refs) => {
+        if (!refs) {   
+            return [];
```
akiva10b, line 1185: based on the code `_filterAndSortRefs(data.refs?.about?.refs)` a ref can be inputted as undefined so the check should happen on the function

static/js/Misc.jsx
```diff
@@ -1203,45 +1222,64 @@ const ReorderEditorWrapper = ({toggle, type, data}) => {
...
       return <EditTextInfo initTitle={data} close={toggle}/>;
     case "sources":
         const [topicSlug, refData] = data;
         return <SourceEditor topic={topicSlug} origData={refData} close={toggle}/>;
     case "cats":
-        return getEditorForExistingCategory();
+        return <EditorForExistingCategory toggle={toggleFunction} data={dataObject} />;
     case "topics":
-        return getEditorForExistingTopic();
+        return <EditorForExistingTopic toggle={toggleFunction} data={dataObject} />;
```
akiva10b, line 1282: it seemed that we could stay consistent here in what CategoryEditorWrapper returns, if we are expecting a functional component, it is cleaner to have that be the case across all conditions

static/js/Misc.jsx
```diff
@@ -1203,45 +1222,64 @@ const ReorderEditorWrapper = ({toggle, type, data}) => {
...
-        origHe: tocObject.heCategory,
-        origDesc,
-        origCategoryDesc,
-        isPrimary: tocObject.isPrimary
-      };
-      return <CategoryEditor origData={origData} close={toggle} origPath={data.slice(0, -1)}/>;
-  }
+const EditorForExistingTopic = ({ toggle, data }) => {
+  const initCatSlug = TopicToCategorySlug(data);
+  const origData = {
```
akiva10b, line 1227: combined the variable declarations

stevekaplan123 (approved):

PR #1529, opened 2023-07-23
Feature/sc 19098/images in text

stevekaplan123:


static/js/sefaria/util.js
```diff
@@ -163,6 +163,12 @@ class Util {
                                now.getUTCHours(), now.getUTCMinutes(), now.getUTCSeconds());
       return Math.round(nowUTC/1000);
     }
+    static _stripImgs(s) {
```
nsantacruz, line 166: Since this function is used publically, can you remove the leading underscore?

static/js/sefaria/sefaria.js
```diff
@@ -1926,7 +1932,7 @@ _media: {},
     // Used when isACaseVariant() is true to prepare the alternative
     return data["completions"][0] + query.slice(data["completions"][0].length);
   },
-  makeSegments: function(data, withContext) {
+  makeSegments: function(data, withContext, sheets=false) {
```
nsantacruz, line 1935: I would not strip images in this function since it adds unnecessary responsibility to it. Instead, let's create another function that strips images from segments by looping over the segments created from `makeSegments()`. That should encapsulate the image stripping logic and make it easy to modify once we have more clarity on how images should display.

nsantacruz (dismissed):

nsantacruz (approved):

PR #1540, opened 2023-07-25
Feature/sc 19305/fix find refs api to find english refs if

rneiss:


nsantacruz (dismissed):

nsantacruz (dismissed):

edamboritz (approved):

sefaria/utils/hebrew.py
```diff
@@ -420,12 +420,25 @@ def strip_nikkud(rawString):
...
 any_english = regex.compile(r"[a-zA-Z]")
 
-def is_hebrew(s, heb_only=False):
-	if not heb_only and any_hebrew.search(s):
-		return True
-	elif heb_only and any_hebrew.search(s) and not any_english.search(s):
-		return True
-	return False
+
+def has_hebrew(s):
```
akiva10b, line 424: can we add types to all params return -> typing? it will allow for intenseness in the IDEs to know to warn when the return param is used incorrectly

sefaria/utils/tests/hebrew_test.py
```diff
@@ -71,9 +71,26 @@ def test_strip_nikkud(self):
...
+    def test_is_all_hebrew(self):
+        assert h.is_all_hebrew("גגגדגחח")
+        assert not h.is_all_hebrew("שלום world")
+
+    def test_is_mostly_hebrew(self):
+        assert h.is_mostly_hebrew("שלום עולם")
+        assert not h.is_mostly_hebrew("שלום world")  # exactly one less than half
+        assert not h.is_mostly_hebrew("שלוגם world")  # exactly half
+        assert h.is_mostly_hebrew("שלוגם word")  # exactly one more than half
+
```
akiva10b, line 93: would an empty string return an error or Hebrew?

PR #1543, opened 2023-07-26
Api v3

YishaiGlasner:
new branch for only api


api/views.py
```diff
@@ -0,0 +1,29 @@
...
+    405 - unsuppored method
+    200 - any other case. when requested version doesn't exist the returned object will include the message
+    """
+    try:
+        oref = Ref.instantiate_ref_with_legacy_parse_fallback(tref)
+    except Exception as e:
+        return jsonResponse({'error': getattr(e, 'message', str(e))}, status=400)
+    if oref.is_empty():
+        return jsonResponse({'error': f'We have no text for {oref}.'}, status=400)
+    cb = request.GET.get("callback", None)
```
edamboritz, line 20: I dont think we need these kind of things anymore

api/views.py
edamboritz, line 1: This code is generally speaking great, its doing the outline of the logic of a GET to the texts api and delegating all other work. 

Given the structure of texts_api.py though, I'd suggest using Django Class Based Views to group this with its other API logic supporting functions .
https://docs.djangoproject.com/en/1.11/topics/class-based-views/intro/

api/texts_api.py
edamboritz, line 1: This file is ding a great job of breaking pieces of logic into separate functions and even classes. 

 We should be looking more carefully into the proper separation of concerns here and what utility each component is serving. Meaning, that we need to make sure that we have components that each take care of a distinct task. This helps break down how we want the classes and functions to work in terms of shared functionality, in terms of grouping functions for the same task and also naming. 
 
1. In order to also service future support of fetching text for a server side rendered view that is outside of the API module (its in the main Django app) we need a component who's sole job it is to fetch data from the database. The component/class would do this based on a  input list of tuples of language and version title. We could think of this as a server side TextManager. A component like this would need to live, not in the api submodule, but rather in a place like sefaria.helper or even sefaria.model. The TextManger can have defaults that fetch the default "source" and "translation" versions if nothing explicit is passed in. 
The important things is, the TextManager does not care about API query parameters and it is unaware of them. 

The parsing of those is left to - 

2. A component of class (basically a Class based API view) who's function it is to parse the API url and query parameters, and translate those into the proper inputs for the TextManager and also handle returning the correct API response object. 

For example, a function to handle errors and format a response to the client is great, but should probably not be in the same class as the shared logic to pull data from the database.

api/texts_api.py
```diff
@@ -0,0 +1,170 @@
...
+        no vtitle - the version with the max priority attr of the specified language
+        all - all versions of the specified language
+    representing_string is the original string that came from an API call
+    """
+
+    def __init__(self, lang: str, vtitle: str, representing_string=''):
+        self.lang = lang
+        self.vtitle = vtitle
+        self.representing_string = representing_string
+        if not self.representing_string:
```
edamboritz, line 26: I think making this an input is really confusing. If we only need this string for one error message later, it should be taken care of in that context and not here.

api/texts_api.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        version_params_list = []
+        for params_string in version_params:
+            lang, vtitle = split_query_param_and_add_defaults(params_string, 2, [''])
+            vtitle = vtitle.replace('_', ' ')
+            version_params = VersionsParams(lang, vtitle, params_string)
+            if version_params not in version_params_list:
+                version_params_list.append(version_params)
+        return version_params_list
+
```
edamboritz, line 47: A function for parsing the version param itself should sit here under parse_api_params, but both of them should be in a class that encapsulates the API logic.

api/texts_api.py
```diff
@@ -0,0 +1,170 @@
...
+
+class TextsForClientHandler():
```
edamboritz, line 49: The doc string and the name of the class mean very different things to a reader reading this code.

edamboritz: This class should be refactored to the pieces that hadnel either the API logic or the text fetching logic (which should not be in the API submodule)

sefaria/model/text.py
edamboritz, line 1: I think the diff here is a little difficult to read and we should have a verbal conversation discussing the changes?

edamboritz (changes requested):
Overall, great work! 
We have a lot to fine tune and clean up but really great approach to this major refactor.

sefaria/model/text_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        fields.remove('chapter') # not metadata
+        self.return_obj = {
+            'versions': [],
+            'missings': [],
+            'available_langs': sorted({v.actualLanguage for v in self.all_versions}),
+            'available_versions': [{f: getattr(v, f, "") for f in fields} for v in self.all_versions]
+        }
+
+    def _append_version(self, version):
+        #TODO part of this function duplicate the functionality of Ref.versionlist(). maybe we should mvoe it to Version
```
edamboritz, line 32: 👍

sefaria/model/text_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        fields = Version.optional_attrs + Version.required_attrs
+        for attr in ['chapter', 'title', 'language']:
+            fields.remove(attr)
+        version_details = {f: getattr(version, f, "") for f in fields}
+        text_range = TextRange(self.oref, version.actualLanguage, version.versionTitle, self.fill_in_missing_segments)
+
+        if self.fill_in_missing_segments:
+            # we need a new VersionSet of only the relevant versions for merging. copy should be better than calling for mongo
+            relevant_versions = copy.copy(self.all_versions)
+            relevant_versions.remove(lambda v: v.actualLanguage != version.actualLanguage)
```
edamboritz, line 42: Do we test anywhere that the merge functionality in TextRange results in the same version being merged as the list of versions created here for the metadata of the merge?

sefaria/model/text_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        text_range.versions = relevant_versions
+        version_details['text'] = text_range.text
+
+        sources = getattr(text_range, 'sources', None)
+        if sources is not None:
+            version_details['sources'] = sources
+
+        if self.oref.is_book_level():
+            first_section_ref = version.first_section_ref() or version.get_index().nodes.first_leaf().first_section_ref()
+            version_details['firstSectionRef'] = first_section_ref.normal()
```
edamboritz, line 55: is firstSectionRef not present if not book level?

sefaria/model/text.py
```diff
@@ -4442,6 +4515,9 @@ def part_projection(self):
         # todo: reimplement w/ aggregation pipeline (see above)
         # todo: special case string 0?
 
+        if self.index_node.is_virtual:
```
edamboritz, line 4518: where are we handling virtual nodes?

sefaria/model/text.py
```diff
@@ -1686,6 +1739,76 @@ def __call__(cls, *args, **kwargs):
...
+            subtext = reduce(lambda x, _: x[-1], range(s), text)
+            del subtext[section:]
+        for s, section in enumerate(self.oref.sections[1:], 1):
+            subtext = reduce(lambda x, _: x[0], range(s), text)
+            del subtext[:section-1]
+        matching_sections = itertools.takewhile(lambda pair: pair[0] == pair[1], zip(self.oref.sections, self.oref.toSections))
+        redundant_depth = len(list(matching_sections))
+        return reduce(lambda x, _: x[0], range(redundant_depth), text)
+
+    @property
```
edamboritz, line 1797: This is great, will allow us the flexibility to return a flattened version of the text as well in the future

sefaria/model/text.py
```diff
@@ -1686,6 +1739,76 @@ def __call__(cls, *args, **kwargs):
...
+        self.merge_versions = merge_versions
+        self._versions = []
+        self._text = None
+        self.sources = None
+
+    @property
+    def versions(self):
+        if self._versions == []:
+            condition_query = self.oref.condition_query(self.lang) if self.merge_versions else \
+                {'title': self.oref.index.title, 'actualLanguage': self.lang, 'versionTitle': self.vtitle}
```
edamboritz, line 1762: I may be missing something but where is the lang param being correctly funneled to the correct database field (i.e. not original "language" that is only he/en)

edamboritz (commented):
This is overall terrific work! 
I like all the various python class syntax you made use of, the code is clean, pretty clear and straight forward. 

I think we still need to address: 
Consistency with database fields and what exactly is being queried 
"Virtual Texts" and any other edge cases of texts that the old API supported? 
Validating backwards compatibility with old API

sefaria/model/text_manager.py
```diff
@@ -0,0 +1,142 @@
...
+    BASE = 'base'
+    SOURCE = 'source'
+    TRANSLATION = 'translation'
+
+    def __init__(self, oref: Ref, versions_params: List[List[str]], fill_in_missing_segments=True):
+        self.versions_params = versions_params
+        self.oref = oref
+        self.fill_in_missing_segments = fill_in_missing_segments
+        self.handled_version_params = []
+        self.all_versions = self.oref.versionset()
```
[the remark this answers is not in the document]
YishaiGlasner: We are calling all versions (with part projection of the ref) once for any request. That indeed means we're getting surplus data, which is burden especially in texts that have many versions (which I guess are the poplar and most qureied texts).
The other oprion is calling mongo more than once in every request. Akiva suggested that calling mongo is the bottleneck, so getting surplus data is preferable.

sefaria/model/text_manager.py
YishaiGlasner: we have a test called test_fill_in_missing_segments. we didn't change the method of the merging itself which happens on VersionSet.merge.
If you have an idea for further testing, it will be great, please be more specific.

sefaria/model/text_manager.py
YishaiGlasner: this is what happems in the api/texts. I'm not sure why (indeed I think there is no reason), but I just did the same.

sefaria/model/text.py
YishaiGlasner: in the text property of TextRange. i'm adding a test for that.

sefaria/model/text.py
YishaiGlasner: here - it uses 'actualLanguage' and not the he/en 'language'

edamboritz (approved):

PR #1550, opened 2023-07-31
fix: pull prod images from GAR instead of GCR

dabasvibhor:


.github/workflows/production-deploy.yaml
```diff
@@ -164,7 +164,7 @@ jobs:
           context: .
           push: true
           build-args: |
-            SRC_IMG=gcr.io/${{ secrets.PROD_GKE_PROJECT }}/${{ secrets.IMAGE_NAME }}-web:${{ needs.workflow-check.outputs.ref }}
+            SRC_IMG=us-east1-docker.pkg/${{ secrets.PROD_GKE_PROJECT }}/containers/${{ secrets.IMAGE_NAME }}-web:${{ needs.workflow-check.outputs.ref }}
```
BrendanGalloway, line 167: was this not previously update for dual push?  I thought all the deploys were already GAR and GCR... For now, let's not remove the GCR push until after we've deployed using GCR a couple times.

BrendanGalloway (changes requested):

.github/workflows/production-deploy.yaml
dabasvibhor: It is a dual push. `tags` has both images from the previous `meta` step. 
This change is a dockerfile argument which will make it pull the gar image(instead of gcr)

BrendanGalloway: oh right, this is the build arg

BrendanGalloway (approved):

PR #1551, opened 2023-07-31
Source edit bug

stevekaplan123:


sefaria/helper/topic.py
```diff
@@ -107,6 +107,56 @@ def group_links_by_type(link_class, links, annotate_links, group_related):
...
+
+def update_data_source_in_link(curr_link, new_link, data_source):
+    data_source_obj = library.get_topic_data_source(data_source)
+    curr_link['dataSources'][data_source] = data_source_obj.displayName
+    del new_link['dataSource']
+
+def update_learning_team_data(curr_link, new_link, data_source):
+    # in case the new_link was created by the learning team, we want to use ref of learning team link
+    # in the case when both links are from the learning team, use whichever ref covers a smaller range
+    if data_source != 'learning-team':
```
akiva10b, line 131: we seem to be making a check in code for permissions. I understand that this is happening globally which isn't a great way to verify permissions. Lets rename the functions based on what the function does as opposed to who is doing the operation, and then have a decorator or some other check that verifies the person doing the operation is allowed to. The issue with "update_learning_team_data" is that I don't know what that function does, I only know who is supposed to be doing it

PR #1558, opened 2023-08-01
refactor(text cache): new class for caching text.

YishaiGlasner:


static/js/sefaria/textCache.js
```diff
@@ -0,0 +1,75 @@
...
+        if (!this.hasOwnProperty(ref)) {
+            this[ref] = new SegmentVersions;
+        }
+        this[ref].setVersion(language, versionTitle, version);
+    }
+    setSectionVersion(ref, language, versionTitle, version) {
+        if (!this.hasOwnProperty(ref)) {
+            this[ref] = new SectionVersions;
+        }
+        this[ref].addVersion(language, versionTitle, version);
```
akiva10b, line 70: where is `addVersion` defined?

static/js/sefaria/textCache.js
```diff
@@ -0,0 +1,75 @@
+class SegmentVersionsFofLanguage {
+    setVersion(versionTitle, version) {
+        this[versionTitle] = version;
```
akiva10b, line 3: `this` refers to the object it belongs to. But it is not recommended to use it as a storage for data, as it can lead to conflicts. Instead, consider using a separate storage property such as a Map or an Object.

static/js/sefaria/textCache.js
```diff
@@ -0,0 +1,75 @@
...
+        return this.segments.map((segment) => {
+            segment.getVersion(language, versionTitle);
+        });
+    }
+}
+
+
+export default class TextCache {
+    setSegmentVersion(ref, language, versionTitle, version) {
+        if (!this.hasOwnProperty(ref)) {
```
akiva10b, line 61: In general It's better to use `in` instead of `hasOwnProperty`: `hasOwnProperty` will only return true if the key is on the object itself, but will return false if the key is in the prototype. The `in` operator will return true if the property is on the object itself or anywhere in its prototype chain. However, in your case, you're working with instances where prototype properties would not likely be a concern, but it's still a good practice.

static/js/sefaria/textCache.js
```diff
@@ -0,0 +1,75 @@
+class SegmentVersionsFofLanguage {
```
akiva10b, line 1: typo, should be SegmentVersionsForLanguage but perhaps even better would be"
`SegmentLanguageVersions`

static/js/sefaria/textCache.js
akiva10b: so something similar to what you did at the bottom with segments:

```
class SegmentLanguageVersions {
    constructor() {
        this.versions = {};
    }
    
    setVersion(versionTitle, version) {
        this.versions[versionTitle] = version;
    }
    
    updateVersion(segmentVersionsFofLanguage) {
        Object.keys(segmentVersionsFofLanguage).forEach((versionTitle) => {
            this.setVersion(versionTitle, segmentVersionsFofLanguage[versionTitle]);
        });
    }
    
    getVersion(versionTitle) {
        return this.versions[versionTitle];
    }
}
```

static/js/sefaria/textCache.js
```diff
@@ -0,0 +1,75 @@
...
+        this[ref].setVersion(language, versionTitle, version);
+    }
+    setSectionVersion(ref, language, versionTitle, version) {
+        if (!this.hasOwnProperty(ref)) {
+            this[ref] = new SectionVersions;
+        }
+        this[ref].addVersion(language, versionTitle, version);
+    }
+    getVersion(ref, language, versionTitle) {
+        return this[ref].getVersion(language, versionTitle);
```
akiva10b, line 73: I imagine we aren't always sure the ref is in cache, so you might want to add a check with a question mark

`return this.cache[ref]?.getVersion(language, versionTitle);`

PR #1562, opened 2023-08-02
chore: [sc-19062] Create URL at sefaria.org/gala that goes to some to be determined classy page

rneiss:
Story details: https://app.shortcut.com/sefaria/story/19062

### DIFF
--- a/sites/sefaria/urls.py
+++ b/sites/sefaria/urls.py
@@ -84,7 +84,7 @@ site_urlpatterns += [
     url(r'^contribute/?$', lambda x: HttpResponseRedirect('https://github.com/Sefaria/Sefaria-Project/wiki/Guide-to-Contributing')),
     url(r'^faq/?$', lambda x: HttpResponseRedirect('/collections/sefaria-faqs' if x.interfaceLang == 'english' else '/collections/%D7%A9%D7%90%D7%9C%D7%95%D7%AA-%D7%A0%D7%A4%D7%95%D7%A6%D7%95%D7%AA-%D7%91%D7%A1%D7%A4%D7%A8%D7%99%D7%90')),
     url(r'^help/?$', lambda x: HttpResponseRedirect('/collections/sefaria-faqs' if x.interfaceLang == 'english' else '/collections/%D7%A9%D7%90%D7%9C%D7%95%D7%AA-%D7%A0%D7%A4%D7%95%D7%A6%D7%95%D7%AA-%D7%91%D7%A1%D7%A4%D7%A8%D7%99%D7%90')),
-    url(r'^gala/?$', lambda x: HttpResponseRedirect('https://www.501auctions.com/sefaria')),
+    url(r'^gala/?$', lambda x: HttpResponseRedirect('https://donate.sefaria.org/event/sefarias-10-year-anniversary-gala/e486954')),
     url(r'^jfn?$', lambda x: HttpResponseRedirect('https://www.sefaria.org/sheets/60494')),
     url(r'^[nN]echama/?', lambda x: HttpResponseRedirect("/collections/גיליונות-נחמה")),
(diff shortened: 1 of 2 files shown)


templates/messages/2023-08-01-tu-b-av.html
```diff
@@ -70,7 +70,7 @@
           Show your love for Sefaria this Tu B’Av!
         </span>
         <span class="int-he">
-          !הביעו את אהבתכם לספריא לרגל ט"ו באב
+הביעו את אהבתכם לספריא לרגל ט"ו באב!
```
relyks, line 73: Is the punctuation in the right location here?

sites/sefaria/urls.py
```diff
@@ -84,7 +84,7 @@
     url(r'^contribute/?$', lambda x: HttpResponseRedirect('https://github.com/Sefaria/Sefaria-Project/wiki/Guide-to-Contributing')),
     url(r'^faq/?$', lambda x: HttpResponseRedirect('/collections/sefaria-faqs' if x.interfaceLang == 'english' else '/collections/%D7%A9%D7%90%D7%9C%D7%95%D7%AA-%D7%A0%D7%A4%D7%95%D7%A6%D7%95%D7%AA-%D7%91%D7%A1%D7%A4%D7%A8%D7%99%D7%90')),
     url(r'^help/?$', lambda x: HttpResponseRedirect('/collections/sefaria-faqs' if x.interfaceLang == 'english' else '/collections/%D7%A9%D7%90%D7%9C%D7%95%D7%AA-%D7%A0%D7%A4%D7%95%D7%A6%D7%95%D7%AA-%D7%91%D7%A1%D7%A4%D7%A8%D7%99%D7%90')),
-    url(r'^gala/?$', lambda x: HttpResponseRedirect('https://www.501auctions.com/sefaria')),
+    url(r'^gala/?$', lambda x: HttpResponseRedirect('https://donate.sefaria.org/event/sefarias-10-year-anniversary-gala/e486954')),
```
relyks, line 87: Are you sure this going to be the right link? There currently isn't anything at the URL

relyks (changes requested):
I'm just trying to be thorough lol

templates/messages/2023-08-01-tu-b-av.html
rneiss: Shockingly it is. 

Here's the same thing if you convert the string to JS numeric strings:

`\u{5D4}\u{5D1}\u{5D9}\u{5E2}\u{5D5} \u{5D0}\u{5EA} \u{5D0}\u{5D4}\u{5D1}\u{5EA}\u{5DB}\u{5DD} \u{5DC}\u{5E1}\u{5E4}\u{5E8}\u{5D9}\u{5D0} \u{5DC}\u{5E8}\u{5D2}\u{5DC} \u{5D8}"\u{5D5} \u{5D1}\u{5D0}\u{5D1}!`

One of the most challenging things about dealing w/ rtl content is that often punctuation appears to be incorrectly placed when it is, or incorrectly placed when it's not.

sites/sefaria/urls.py
rneiss: Great catch, but it is indeed the correct link as shared by Annie yesterday. It is currently off-line as they make the last minute changes, but this URL is an orphan one on the site and only is navigated to via comms so the current effective 404 won't be an issue as it will ameliorated on the Classy side before the emails go out.

templates/messages/2023-08-01-tu-b-av.html
relyks: Yeah, I don't think GitHub has properly implemented dealing with Hebrew... lol

sites/sefaria/urls.py
relyks: All good, just checking :)

relyks (approved):
All good 😄

PR #1566, opened 2023-08-06
fix(pytest): fix broken pytest after changing index error to log

akiva10b:


sefaria/utils/util.py
```diff
@@ -611,3 +612,20 @@ def decorated_function(*args, **kwargs):
             return return_value
         return decorated_function
     return argumented_decorator
+
+
+def conditional_graceful_exception(logger=None, logLevel="exception", return_value=[], exception_type=Exception):
```
nsantacruz, line 617: This function could use documentation and typing for the parameters

nsantacruz (approved):

PR #1589, opened 2023-08-14
fix(Sidebar Descriptions): use ContentText instead of InterfaceText s…

stevekaplan123:
…o that content language is used instead of interface language

### DIFF
--- a/static/js/ContentText.jsx
+++ b/static/js/ContentText.jsx
@@ -6,6 +6,7 @@ const ContentText = (props) => {
    /* Renders content language throughout the site (content that comes from the database and is not interface language).
    * Gets the active content language from Context and returns only the appropriate child(ren) for given language
    * text {{text: object}} a dictionary {en: "some text", he: "some translated text"} to use for each language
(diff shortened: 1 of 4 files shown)


static/js/ContentText.jsx
```diff
@@ -53,11 +54,12 @@ const VersionImageSpan = ({lang, content, languageToFilter, imageLoadCallback})
...
 };
 
-const _filterContentTextByLang = ({text, html, overrideLanguage, defaultToInterfaceOnBilingual=false, bilingualOrder = null}) => {
+const _filterContentTextByLang = ({text, html, markdown, overrideLanguage, defaultToInterfaceOnBilingual=false, bilingualOrder = null}) => {
   /**
    See ContentText for documentation
    */
-  const contentVariable  = html ? html : text;
+  const contentVariable = html ?
+                          html : markdown ? markdown : text;  // assumption is `markdown` or `html` are preferred over `text` if they are present
```
akiva10b, line 62: can this be:
`const contentVariable = html || markdown || text;`

seems cleaner and achieves the same outcome

akiva10b (approved):

PR #1590, opened 2023-08-14
ci: clean alpha chart release channel name

BrendanGalloway:


helm-chart/release-rules.sh
```diff
@@ -38,17 +38,18 @@ plugins:
     - "successComment": false
 EOF
 export branch=$(git branch --show-current)
+export channel=$(echo $branch | awk '{print tolower($0)}' | sed 's|.*/\([^/]*\)/.*|\1|; t; s|.*|\0|' | sed 's/[^a-z0-9\.\-]//g')
```
akiva10b, line 41: overall looks good, I am not sure that `t; s|.*|\0|` will actually do anything as s|.*|\0| seems like it is replacing the data with itself, but I could be missing something

akiva10b (approved):

helm-chart/release-rules.sh
BrendanGalloway: Doing a little bit more testing you may be correct, at least for linux.  I think there was something about how sed works in mac that meant we needed to add the extra handling to keep the branchname unchanged if it did not match the initial pattern.  There was a lot of iteration initially to get a sed instruction that could parse the branch name correctly for all the different dev environments.  This snippet is used in multiple places, so for now let's keep it consistent and I can go through and change it everywhere if the second path is not needed.

akiva10b (approved):

PR #1595, opened 2023-08-15
fix(collections): add rout to history on click

akiva10b:


static/js/Misc.jsx
```diff
@@ -1604,6 +1604,9 @@ const SheetListing = ({
       handleSheetClick(e, sheet, null, connectedRefs);
       e.preventDefault();
     }
+    if (e.currentTarget.target === "_self") {
+      history.pushState(null, null, sheet.sheetUrl);
+    }
```
akiva10b, line 1609: I am not sure if this solves the core issue or will break other parts of the system. I don't completely understand why we aren't currently adding to the history stack on-click. So I hope a reviewer can point me in the right direction

rneiss: I'm not sure that this breaks anything (though we only generally push history state in ReaderApp.jsx) but it does not solve the underlying issue outlined on the card i.e. 

```Enter to s collections like (hok le Israel) https://www.sefaria.org.il/collections/%D7%97%D7%A7-%D7%9C%D7%99%D7%A9%D7%A8%D7%90%D7%9C?tab=sheets

pick a day
click back

expected to go back to the tab=sheets of the parashah that was picked earlier

what happens
routed to a general page with all the parashot```

I think the solution that the card advocates for is actually the incorrect one, what I think should happen instead is that when one clicks on a link in the content section of a collection which leads one to a filtered list of sheets by the clicked tag, that the URL in the browser at that moment should change.  I'm assuming the logic should be more or less around here: https://github.com/Sefaria/Sefaria-Project/blob/master/static/js/ReaderApp.jsx#L509

PR #1599, opened 2023-08-17
chore: [sc-19894] Allow HTML in dedications in TOC

rneiss:
Story details: https://app.shortcut.com/sefaria/story/19894

### DIFF
--- a/static/js/TextsPage.jsx
+++ b/static/js/TextsPage.jsx
@@ -171,8 +171,8 @@ const Dedication = () => {
         dedicationData && (dedicationData.en || dedicationData.he) ?
         <div className="dedication">
           <span>
-              <span className="int-en">{dedicationData?.en}</span>
-              <span className="int-he">{dedicationData?.he}</span>
+              <span className="int-en" dangerouslySetInnerHTML={ {__html: dedicationData?.en} }></span>
+              <span className="int-he" dangerouslySetInnerHTML={ {__html: dedicationData?.he} }></span>
           </span>
         </div>
         : null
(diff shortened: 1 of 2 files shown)


static/js/TextsPage.jsx
```diff
@@ -171,8 +171,8 @@ const Dedication = () => {
         dedicationData && (dedicationData.en || dedicationData.he) ?
         <div className="dedication">
           <span>
-              <span className="int-en">{dedicationData?.en}</span>
-              <span className="int-he">{dedicationData?.he}</span>
+              <span className="int-en" dangerouslySetInnerHTML={ {__html: dedicationData?.en} }></span>
```
nissamai, line 174: can we at least run a function to sanitize the dedication data before sticking it here?

nissamai (changes requested):

static/js/TextsPage.jsx
rneiss: I think this is 100% the right approach. I'm trying to hit a Monday deadline that was tossed on us this AM, but maybe I can harness the new pathway that exists for using Markdown in topic page descriptions. Gonna give that a shot.

Thanks for keeping me honest.

static/js/TextsPage.jsx
```diff
@@ -171,8 +171,7 @@ const Dedication = () => {
         dedicationData && (dedicationData.en || dedicationData.he) ?
         <div className="dedication">
           <span>
-              <span className="int-en" dangerouslySetInnerHTML={ {__html: dedicationData?.en} }></span>
-              <span className="int-he" dangerouslySetInnerHTML={ {__html: dedicationData?.he} }></span>
+              <InterfaceText markdown={{en: dedicationData?.en, he: dedicationData?.he}}/>
```
rneiss, line 174: @nissamai this turned out to be way easier than I expected. Also, much cleaner in the end. Again -- thanks.

nissamai (approved):

PR #1610, opened 2023-08-27
Feature/sc 20203/internal tooling for learning team

stevekaplan123:


sefaria/helper/topic.py
```diff
@@ -1037,6 +1039,39 @@ def topic_change_category(topic_obj, new_category, old_category="", rebuild=Fals
...
+    process_topic_place_change(topic_obj, data)
+    topic_obj = update_author_era(topic_obj, data, dataSource='learning-team-editing-tool')
+
+    return topic_obj
+
+def update_author_era(topic_obj, data, dataSource='learning-team-editing-tool'):
+    for k in ["birthYear", "deathYear"]:
+        year = data.get(k, '')  # set to None if blank, otherwise validate number
+        if year != '':
+            assert int(year), f"'{k}' must be a number but is {year}"
```
akiva10b, line 1067: the year being an integer should be validated at the API level even before this function and return a 400 with an explanation of the error. That being said, this is usually done in a cleaner way which we are not set up for that. So best to put in a comment alongside a type check at the API level

sefaria/helper/descriptions.py
```diff
@@ -12,6 +12,38 @@
...
+        "fromTopic": topic.slug,
+        "linkType": "is-a",
+        "dataSource": "sefaria",
+        "generatedBy": "update_authors_data"
+    })
+    itl.save()
+
 
 
 def update_authors_data():
```
akiva10b, line 49: This function seems to have the same name as another function imported from topics, consider a different name for readability

sefaria/helper/topic.py
```diff
@@ -1051,11 +1086,9 @@ def update_topic(topic_obj, **kwargs):
...
     orig_slug = topic_obj.slug
 
-    if 'title' in kwargs and kwargs['title'] != topic_obj.get_primary_title('en'):
-        topic_obj.add_title(kwargs['title'], 'en', True, True)
-
-    if 'heTitle' in kwargs and kwargs['heTitle'] != topic_obj.get_primary_title('he'):
-        topic_obj.add_title(kwargs['heTitle'], 'he', True, True)
+    update_topic_titles(topic_obj, kwargs)
+    if kwargs.get('catSlug') == 'authors':
+        topic_obj = update_authors_data(topic_obj, kwargs)
```
akiva10b, line 1091: if update_authors_data failed and update_topic_titles passed, would it create a mess in our data? It is difficult to account for every failure, but one thing we can account for is serializing and validating data before it gets to the functions. All the data that we expect to go to internal functions should be validated at the API level, and once it is clean and clear to go we can be much more assured that it won't fail

sefaria/model/place.py
```diff
@@ -46,8 +46,24 @@ def primary_name(self, lang=None):
 
     def secondary_names(self, lang=None):
         return self.name_group.secondary_titles(lang)
-
-    ###
+    
+    @classmethod 
+    def create_new_place(cls, en, he=None):
```
akiva10b, line 51: I see cls isn't used in the entire class method, perhaps replace `p = Place().load({'key': en})` with `p = cls().load({'key': en})` so that it justifies this being a class method. Also, once that us the case, this kind of function can be abstracted to be used on multiple objects

sefaria/model/place.py
```diff
@@ -46,8 +46,24 @@ def primary_name(self, lang=None):
...
+        p = Place({'key': en})
+        p.name_group.add_title(en, 'en', True, True)
+        if he:
+            p.name_group.add_title(he, 'he', True, True)
+        p.city_to_coordinates(en)
+        p.save()
+        return p
+
+    def city_to_coordinates(self, city):
+        geolocator = Nominatim(user_agent='hello@sefaria.org')
```
akiva10b, line 64: very good thinking out of the box! using Nominatim for finding the coords is great, what were they doing till now? does that data not come in in other ways?

static/js/TopicEditor.jsx
```diff
@@ -81,52 +90,77 @@ const TopicEditor = ({origData, onCreateSuccess, close, origWasCat}) => {
...
          requestWithCallBack({url, data: postCategoryData, setSavingStatus, redirect: () => window.location.href = "/topics"});
     }
-    const saveTopic = function () {
-        toggle();
-        let url = "";
+    const authorItems = (type) => {
+        switch(type) {
+            case 'labels':
+                return Sefaria._siteSettings.TORAH_SPECIFIC ?
+                            ["English Alternate Titles", "Hebrew Alternate Titles", "Birth Place", "Hebrew Birth Place", "Birth Year", "Place of Death", "Hebrew Place of Death", "Death Year", "Era"]
```
akiva10b, line 97: Please put the lists in names variables (its very hard to understand why the different lists right now)

akiva10b (commented):
Overall this was done really quickly! and with a lot of creativity. Some general points:
1. There should be unit tests checking all the functions that update data as this is our most prized possession!
2. All data that arrives from a form should be validated before it reaches internal functions and return a 400 if there is something wrong with it. A lot of the data validation happens on the functions itself (which is also great) but the endpoint doesn't stop the request right away meaning that other operations can take place before the call fails.
3. Please use variables for strings that are used multiple times, for instance lang == 'he' should be lang == HE
4. Please add typing to the parameters of the functions

static/js/TopicEditor.jsx
```diff
@@ -81,52 +90,68 @@ const TopicEditor = ({origData, onCreateSuccess, close, origWasCat}) => {
...
             "heTitle": data.heTitle};
         if (isCategory) {
             postData = {...postData, "catDescription": {"en": data.enCategoryDescription, "he": data.heCategoryDescription}};
         }
-        postData.category = data.catSlug;
-
-        if (isNew) {
-          url = "/api/topic/new";
+        if (data?.era && Sefaria.util.inArray(data.era, Sefaria._eras) === -1) {
+            delete postData.era;
```
akiva10b, line 101: in order to clean data before sending it out to the server, best to check that data has changed and then add a key to the post object. That way if a user wants to remove data or set it to an empty string it is possible. This can also be done generically and not on a specific element such as "era"

akiva10b (approved):
Looks great, there are some issue we discussed in the previous comment that should be addressed in the future

akiva10b (approved):

PR #1615, opened 2023-08-29
feat(Scroll Links): inline links can now be scrollLink or refLink

stevekaplan123:
allow for book to contain links to the same book and upon clicking, scroll to the new ref instead of opening new panel

### DIFF
--- a/static/js/TextRange.jsx
+++ b/static/js/TextRange.jsx
@@ -375,6 +375,7 @@ class TextRange extends Component {
             formatEnAsPoetry={formatEnAsPoetry}
             formatHeAsPoetry={formatHeAsPoetry}
             placeSegmentNumbers={this.placeSegmentNumbers}
+            navigatePanel={this.props.navigatePanel}
           />
         </span>
       );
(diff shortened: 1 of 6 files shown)


static/js/TextRange.jsx
```diff
@@ -492,21 +494,38 @@ class TextSegment extends Component {
       this.props.unsetTextHighlight();
     }
   }
+  handleRefLinkClick(refLink, event) {
+    // two options: either open a new panel or if refLink's class is 'scrollLink' and the new ref is in the same book, scroll in same panel to new location
+    event.preventDefault();
+    let newRef = Sefaria.humanRef(refLink.attr("data-ref"));
+    const newBook = Sefaria.parseRef(newRef)?.book;
+    const currBook = Sefaria.parseRef(this.props.sref)?.book;
+    const typeRefLink = refLink.attr('class');
```
akiva10b, line 503: why are we using a class instead of a data attribute? might be correct but still unclear

akiva10b: I see, this could be error prone as a future dev may not be aware that the class also has functionality associated with the classname, so a change to a style name would break functionality. Instead, lets use a data tag called something like data-scroll-link which a future dev will know likely carries functionality with it

akiva10b (approved):

PR #1627, opened 2023-09-07
feat(Search Results): Normalize ref

stevekaplan123:
Normalize ref if Sefaria.parseRef can't find if the index

### DIFF
--- a/static/js/SearchTextResult.jsx
+++ b/static/js/SearchTextResult.jsx
@@ -14,17 +14,8 @@ class SearchTextResult extends Component {
         super(props);
 
         this.state = {
-            duplicatesShown: false,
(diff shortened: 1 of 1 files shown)


static/js/SearchTextResult.jsx
```diff
@@ -14,9 +14,18 @@ class SearchTextResult extends Component {
         super(props);
 
         this.state = {
-            duplicatesShown: false
+            duplicatesShown: false,
+            ref: this.props.data._source.ref
         };
     }
+    componentDidMount() {
```
akiva10b, line 21: this should probably be done onclick and not onmount

akiva10b (approved):

PR #1636, opened 2023-09-14
Banners (and modals) wouldn't go away but now they will

relyks:
## Bug Report

A user reported being unable to get rid of the current deployed banner when they closed it out. They continued to see the banner after interacting and making repeated visits to the Sheets Editor page.

## Bug Sleuthing Investigation
The bug occurred in the following scenarios:
* Interacting with a banner (hitting the close button or donate button) and then going to a static page with ReaderApp in header-only mode
* Interacting with a banner and then starting a new session (closing tab or window)

**Discovered problem:** the entry in local storage to mark whether a banner or modal had been interacted with was being cleared on every page load or session even in the usual case where the current banner or modal hadn't changed.

## Solution (Extermination)
If the current modal or banner is found to be the same as the eligible one returned from Strapi, the marked entry in local storage for that modal or banner should not be cleared, so that its interacted status stays across sessions. Only old entries for modals or banners in local storage that are not valid should be removed. A condition to not consider the current modal or banner was added during the clearing process.

### DIFF
--- a/static/js/context.js
+++ b/static/js/context.js
@@ -194,12 +194,15 @@ function StrapiDataProvider({ children }) {
             let modals = result.data?.modals?.data;
             let banners = result.data?.banners?.data;
 
-            let removeKeysFromLocalStorageWithPrefix = (prefix) => {
+            let removeContentKeysFromLocalStorage = ({ prefix = "", except = "" } = {}) => {
               let keysToRemove = [];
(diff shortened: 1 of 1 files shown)


rneiss (approved):
Nice bug hunting. Really clear PR, good comments in code. Great work all around.

PR #1640, opened 2023-09-18
feat: [sc-20316] Inform users they can choose alternative translations

nsantacruz:
Story details: https://app.shortcut.com/sefaria/story/20316

Add TransCallToActionBanner which informs users they can change translations.


static/js/ReaderPanel.jsx
```diff
@@ -1382,7 +1383,15 @@ class ReaderControls extends Component {
...
           <DisplaySettingsButton onClick={this.props.openDisplaySettings} />
         </div>);
-    let transLangPrefSuggBann = hideHeader || connectionsHeader ? null : <TranslationLanguagePreferenceSuggestionBanner setTranslationLanguagePreference={this.props.setTranslationLanguagePreference} />;
+    const openTranslations = this.props.openConnectionsPanel.bind(null, [this.props.currentRef], null, {"connectionsMode": "Translations"});
+    const transCallToActionApplies = () => Sefaria.transCallToActionApplies(this.props.currentBook(), this.props.settings.language);
+    let banner = (hideHeader || connectionsHeader) ? null : (
+        <TextColumnBannerChooser
+            setTranslationLanguagePreference={this.props.setTranslationLanguagePreference}
+            openTranslations={openTranslations}
+            transCallToActionApplies={transCallToActionApplies}
```
edamboritz, line 1392: Its more props to pass, but I's prefer passing in the book and language rather than passing in the callback, it seems less dependent that way to me?

static/js/TextColumnBanner.jsx
```diff
@@ -0,0 +1,132 @@
...
+import $  from './sefaria/sefariaJquery';
+import {
+    CloseButton, InterfaceText, EnglishText, HebrewText
+} from './Misc';
+
+const cookie = Sefaria._inBrowser ? $.cookie : Sefaria.util.cookie;
+const { translation_language_preference_suggestion } = Sefaria;
+
+export const TextColumnBannerChooser = ({ setTranslationLanguagePreference, openTranslations, transCallToActionApplies }) => {
+    const [bannerAccepted, setBannerAccepted] = useState(false);
```
edamboritz, line 12: I'm not sure i understand the logic here. accepted only seems to be used for one kind of banner and it seems to make sense to use more generally if its out here in the choose component?

edamboritz: This works, but if more banners get added in the future, I'm concerned the order of operations here is going to get more and more murky and we have states that are only relevant for one sub component

edamboritz (commented):
Great work! 
I noted some small things below

static/js/ReaderPanel.jsx
nsantacruz: It is actually less coupled to the ReaderControls component when you pass a callback. I think it's likely the business logic of when it will apply will require different pieces of data and it felt less random than passing seemingly unrelated props (current book and language). But if you still think it's better, happy to change it. It's not clear to me my way is better, but I wasn't sure if you understood my considerations.

static/js/sefaria/sefaria.js
```diff
@@ -1046,6 +1046,15 @@ Sefaria = extend(Sefaria, {
       }
       return ref ? this.getRefFromCache(ref) : null;
   },
+  transCallToActionApplies: (book, textLanguage) => {
```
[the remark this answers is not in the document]
nsantacruz: The idea was that we tend to put functions that purely deal with data (as opposed to rendering) is sefaria.js. When we need to take into account existing translations, I imagine it would either need to do an API call or check the versions cache for some new data we're sending.

static/js/ReaderPanel.jsx
edamboritz: I understand. My main issue is that we now have a function called "transCallToActionApplies" sitting completely out of context from the one component its really relevant to. Maybe renaming it?
Also, if it stays here it should also be moved out of the render? Though I can understand more why its there, since its technically part of the render, but still.

edamboritz (approved):

PR #1642, opened 2023-09-19
Book descriptions

yonadavGit:


static/js/TopicPage.jsx
```diff
@@ -351,13 +352,30 @@ const TopicHeader = ({ topic, topicData, multiPanel, isCat, setNavTopic, openDis
         <div>
           <div className="sectionTitleText authorIndexTitle"><InterfaceText>Works on Sefaria</InterfaceText></div>
           <div className="authorIndexList">
-            {topicData.indexes.map(({text, url}) => <SimpleLinkedBlock key={url} {...text} url={url} classes="authorIndex" />)}
+            {topicData.indexes.map(({url, title, description}) => <AuthorDisplayItem key={url} url={url} enTitle={title.en} heTitle={title.he} enDesc={description.en} heDesc={description.he}/>)}
```
nsantacruz, line 355: It would be cleaner to pass `title` and `description` as opposed to `enTitle` `heTitle` etc because you need them as an object within `AuthorDisplayItem`

static/js/TopicPage.jsx
```diff
@@ -351,13 +352,30 @@ const TopicHeader = ({ topic, topicData, multiPanel, isCat, setNavTopic, openDis
...
           </div>
         </div>
        : null}
     </div>
 );}
 
+const AuthorDisplayItem = ({url, enTitle, heTitle, enDesc, heDesc}) => {
```
nsantacruz, line 362: This name could use work. It doesn't explain what it is rendering.

static/js/TopicPage.jsx
```diff
@@ -351,13 +352,30 @@ const TopicHeader = ({ topic, topicData, multiPanel, isCat, setNavTopic, openDis
...
+  const classes = classNames({ navBlockTitle: 1 });
```
nsantacruz, line 363: `classNames` is useful when there are classes that need to change based on a variable. Since `navBlockTitle` is constant, you can add it directly to the `className` prop of the `<a>` tag.

static/css/s2.css
```diff
@@ -2468,6 +2468,10 @@ input.resourcesLink {
   padding-bottom: 12px;
   border-top: 1px solid #EDEDEC;
 }
+.navBlockAuthorPage {
```
nsantacruz, line 2471: this class seems to replace `.authorIndex` but it doesn't have all the same styles. Is there a reason?

static/js/TopicPage.jsx
```diff
@@ -351,13 +352,30 @@ const TopicHeader = ({ topic, topicData, multiPanel, isCat, setNavTopic, openDis
...
+  return (
+    <div className="navBlockAuthorPage" >
```
nsantacruz, line 365: This code is unfortunately duplicated in many places in our code with slight variations. I think it's out of scope for this project but something to be careful about.

PR #1651, opened 2023-09-21
feat: Remove version notes from translation selection sidebar.  Keep …

EliezerIsrael:
…them in about-this-text and book page.

### DIFF
--- a/static/js/TranslationsBox.jsx
+++ b/static/js/TranslationsBox.jsx
@@ -89,6 +89,7 @@ class TranslationsBox extends Component {
                 openVersionInSidebar={this.openVersionInSidebar}
                 viewExtendedNotes={this.props.viewExtendedNotes}
                 inTranslationBox={true}
+                showNotes={false}
             />
           </>
       );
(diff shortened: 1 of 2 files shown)


static/js/VersionBlock.jsx
```diff
@@ -171,6 +171,9 @@ class VersionBlock extends Component {
     }
   }
   makeVersionNotes(){
+    if (!this.props.showNotes) {
+      return null;
+    }
```
akiva10b, line 176: Do we also want to have `vnotes && <span className="" dangerouslySetInnerHTML={ {__html: vnotes} } />`
as to not render that element at all in line 301?

akiva10b (approved):

static/js/TranslationsBox.jsx
```diff
@@ -89,6 +89,7 @@ class TranslationsBox extends Component {
                 openVersionInSidebar={this.openVersionInSidebar}
                 viewExtendedNotes={this.props.viewExtendedNotes}
                 inTranslationBox={true}
+                showNotes={false}
```
akiva10b, line 92: will this ever be true?

PR #1657, opened 2023-09-28
Fix learning team links

stevekaplan123:
1.  When merging sources to display in a topic page, any source added by the learning team should be treated separately.  I added checks [here](https://github.com/Sefaria/Sefaria-Project/compare/fix_learning_team_links?expand=1#diff-19364da9061bcb429f3f3c5f63aeab5541e461154bd59e61719b2cd68f631356R173) and [here](https://github.com/Sefaria/Sefaria-Project/compare/fix_learning_team_links?expand=1#diff-19364da9061bcb429f3f3c5f63aeab5541e461154bd59e61719b2cd68f631356R185-R186)
2. I removed the decorator function which was interfering in the merging of two sources where the new one isn't a learning team source.  We wanted to use the source with the smaller expandedRefs but that was not possible because of the decorator.  And it's irrelevant now that learning team links are not merged with others.
3. I also fixed a bug when deleting sources if the source had no curatedPrimacy [here](https://github.com/Sefaria/Sefaria-Project/compare/fix_learning_team_links?expand=1#diff-19364da9061bcb429f3f3c5f63aeab5541e461154bd59e61719b2cd68f631356L1253)
4. Finally, I fixed a minor bug [here](https://github.com/Sefaria/Sefaria-Project/compare/fix_learning_team_links?expand=1#diff-475168e098f3c1facb482e69846fc2ec04b59599ed36e103ae73276f2c93def0R25) involving clicking the down arrow on the last item in the list of sources.

### DIFF
--- a/sefaria/helper/topic.py
+++ b/sefaria/helper/topic.py
@@ -114,7 +114,7 @@ def merge_props_for_similar_refs(curr_link, new_link):
     # as well as datasource and descriptions of all the similar refs
     data_source = new_link.get('dataSource', None)
     if data_source:
-        curr_link = update_refs(curr_link, new_link, data_source)
+        curr_link = update_refs(curr_link, new_link)
         curr_link = update_data_source_in_link(curr_link, new_link, data_source)
 
     if not curr_link['is_sheet']:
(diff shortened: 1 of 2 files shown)


edamboritz (approved):
The PR itself is written very nicely, describing the issue and what was done to address. 

I'm not really versed in this area of code, so its hard for me to verify the actual logic, and it looks to me as if the entire code of around editing topics can be cleaned up and clarified greatly, but that would be out of scope for this PR.

PR #1658, opened 2023-09-28
Steinsaltz portal

nsantacruz:
Main changes
- add descriptions to topic author pages
- add portal data model with rendering on topic pages
- refactor newsletter JS component so that it can connect to different newsletter APIs


sefaria/urls.py
```diff
@@ -385,7 +390,8 @@
 
 # Email Subscribe
 urlpatterns += [
-    url(r'^api/subscribe/(?P<email>.+)$', sefaria_views.subscribe),
+    url(r'^api/subscribe/steinsaltz/(?P<email>.+)$', sefaria_views.subscribe_steinsaltz_newsletter_view),
```
edamboritz, line 393: I think we can probably generalize this. I dont think we need to have a dedicated view specifically for Steinsaltz.  We can have a general view for a foreign newsletter and perform the routing based on parameter to the correct logic if needed.

sefaria/views.py
```diff
@@ -170,28 +170,62 @@ def accounts(request):
...
+    """
+    API for subscribing to mailing lists, in `lists` url param.
+    Currently active lists are:
+    "Announcements_General", "Announcements_General_Hebrew", "Announcements_Edu", "Announcements_Edu_Hebrew"
+    """
+    body = json.loads(request.body)
+    language = body.get("language", "")
+    educator = body.get("educator", False)
+    crm_mediator = CrmMediator()
+    return crm_mediator.subscribe_to_lists(email, first_name, last_name, educator=educator, lang=language)
```
edamboritz, line 209: I'm not sure why this removes the (not so much) more informative logging + json response there was before?

sefaria/model/abstract.py
```diff
@@ -637,3 +643,71 @@ def foo(obj, **kwargs):
             rec.save()
 
     return foo
+
+class SchemaValidationException(Exception):
```
edamboritz, line 647: Why are these here and not in the exceptions module?

sefaria/model/portal.py
```diff
@@ -0,0 +1,121 @@
...
+from urllib.parse import urlparse
+from . import abstract as abst
+import urllib.parse
+from django.core.validators import URLValidator
+from django.core.exceptions import ValidationError
+import structlog
+logger = structlog.get_logger(__name__)
+
+
+class InvalidURLException(Exception):
```
edamboritz, line 11: Also why not in exceptions module?

sefaria/model/portal.py
```diff
@@ -0,0 +1,121 @@
...
+            "title": ({"en": (str, "required"), "he": (str, "required")}, "required"),
+            "title_url": (str, "optional"),
+            "description": ({"en": (str, "required"), "he": (str, "required")}, "optional"),
+            "api_schema": ({"http_method": (str, "required"),
+                            "payload": ({"first_name_key": (str, "optional"), "last_name_key": (str, "optional"), "email_key": (str, "optional")}, "optional")}
+                           , "optional")
+        }
+
+        if hasattr(self, "about"):
+            abst.validate_dictionary(self.about, about_schema)
```
edamboritz, line 95: Why make them a module level method if you just need to use these validators as inherited methods?

sefaria/model/topic.py
```diff
@@ -69,6 +71,8 @@ def _validate(self):
         if getattr(self, 'subclass', False):
             assert self.subclass in self.subclass_map, f"Field `subclass` set to {self.subclass} which is not one of the valid subclass keys in `Topic.subclass_map`. Valid keys are {', '.join(self.subclass_map.keys())}"
 
+        Portal.validate_slug_exists(self.portal_slug)
```
edamboritz, line 74: Should this not only trigger if there _is_ a portal slug?

edamboritz (approved):
This looks great! Do take note of a few minor comments though

sefaria/model/abstract.py
```diff
@@ -637,3 +643,71 @@ def foo(obj, **kwargs):
...
+        super().__init__(self.message)
+
+class SchemaInvalidKeyException(Exception):
+    def __init__(self, key):
+        self.key = key
+        self.message = f"Invalid key '{key}' found in data dictionary."
+        super().__init__(self.message)
+
+
+def validate_dictionary(data, schema):
```
nsantacruz, line 666: @edamboritz I moved this to a new file called `validators.py`

sefaria/model/portal.py
nsantacruz: moved

sefaria/model/portal.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class InvalidURLException(Exception):
+    def __init__(self, url):
+        self.url = url
+        self.message = f"'{url}' is not a valid URL."
+        super().__init__(self.message)
+
+
+def validate_url(url):
```
nsantacruz, line 18: @edamboritz I moved this to a new file called `validators.py`

sefaria/model/portal.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class InvalidHTTPMethodException(Exception):
+    def __init__(self, method):
+        self.method = method
+        self.message = f"'{method}' is not a valid HTTP API method."
+        super().__init__(self.message)
+
+
+def validate_http_method(method):
```
nsantacruz, line 37: @edamboritz I moved this to a new file called `validators.py`

sefaria/model/portal.py
nsantacruz: This method feels generic enough that it can apply to other classes besides AbstractMongoRecord.

sefaria/views.py
nsantacruz: It doesn't, the logging now happens in `generic_subscribe_to_newsletter_api()`

sefaria/urls.py
nsantacruz: Did you have in mind that also the Sefaria newsletter would be generalized so there's only one subscribe view?

edamboritz: Not necessarily, but it can be (i.e. waht the view defaults to if no parameter is passed in)

sefaria/model/portal.py
edamboritz: But then it doesnt get hooked into the normal flow of the validate methods on the object?

edamboritz (approved):

PR #1667, opened 2023-10-17
fix(Index Save): validate compDate and pubDate are lists of integers

stevekaplan123:
I added this to prevent us from adding compDate and pubDate values that are not lists of integers as someone did recently change compDate and pubDate of two texts to integers seemingly via the API.  (It isn't possible to make this mistake using the admin tool UI so I assume the only possibility is the API).

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -687,11 +687,11 @@ class Index(abst.AbstractMongoRecord, AbstractIndex):
         if not Category().load({"path": self.categories}):
             raise InputError("You must create category {} before adding texts to it.".format("/".join(self.categories)))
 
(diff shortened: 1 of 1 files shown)


sefaria/model/text.py
```diff
@@ -687,6 +687,12 @@ def _validate(self):
         if not Category().load({"path": self.categories}):
             raise InputError("You must create category {} before adding texts to it.".format("/".join(self.categories)))
 
+        for dateKey in ['compDate', 'pubDate']:
```
nsantacruz, line 690: change `dateKey` to `date_key` to match Python styling

nsantacruz (approved):

PR #1678, opened 2023-10-24
Gershayim

YishaiGlasner:
should fix this issue - https://app.shortcut.com/sefaria/story/760/exact-term-not-recognized-in-search - and similar issues (for example רלב״ג קהלת which is ref and not author).
in fact there are 2 issues:
1. the auto completion - that's fixed on server side.
2. the search itself (when pressing enter) - that's fixed on client side.

### DIFF
--- a/sefaria/model/autospell.py
+++ b/sefaria/model/autospell.py
@@ -29,14 +29,13 @@ letter_scope = "\u05b0\u05b1\u05b2\u05b3\u05b4\u05b5\u05b6\u05b7\u05b8\u05b9\u05
             + "\u05c1\u05c2" \
             + "\u05d0\u05d1\u05d2\u05d3\u05d4\u05d5\u05d6\u05d7\u05d8\u05d9\u05da\u05db\u05dc\u05dd\u05de\u05df" \
             + "\u05e0\u05e1\u05e2\u05e3\u05e4\u05e5\u05e6\u05e7\u05e8\u05e9\u05ea" \
-            + "\u05f3\u05f4" \
             + "\u200e\u200f\u2013\u201c\u201d\ufeff" \
             + " Iabcdefghijklmnopqrstuvwxyz1234567890[]`:;.-,*$()'&?/\""
 
 
 def normalizer(lang):
     if lang == "he":
-        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)]).replace('"', '״')
+        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)])
     return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in str.lower(x)])
 
 
--- a/sefaria/model/autospell.py
+++ b/sefaria/model/autospell.py
@@ -36,7 +36,7 @@ letter_scope = "\u05b0\u05b1\u05b2\u05b3\u05b4\u05b5\u05b6\u05b7\u05b8\u05b9\u05
 
 def normalizer(lang):
     if lang == "he":
-        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)])
+        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)]).replace('"', '״')
     return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in str.lower(x)])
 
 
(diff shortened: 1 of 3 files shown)


sefaria/model/autospell.py
```diff
@@ -36,7 +36,7 @@
 
 def normalizer(lang):
     if lang == "he":
-        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)])
+        return lambda x: "".join([c if c in letter_scope else unidecode(c) for c in hebrew.normalize_final_letters_in_str(x)]).replace('"', '״')
```
nsantacruz, line 39: unidecode will convert gershayim to ASCII quotes. We're already using it here. All we need to do is remove gershayim from `letter_scope`.

See this:
```python
In [3]: from unidecode import unidecode

In [4]: unidecode('״')
Out[4]: '"'

```

nsantacruz: Note, in your code, the standard quotation mark is gershayim and ASCII is converted to gershayim. If we use unidecode, the standard quotation mark will be ASCII.

static/js/Header.jsx
```diff
@@ -310,6 +310,14 @@ class SearchBar extends Component {
           this.submitSearch(Sefaria.repairCaseVariant(query, d));
           return;
         }
+        if (Sefaria.isGershayimVariant(query, d)) {
+          if (query.includes('"')) {
```
nsantacruz, line 314: Why is this `if` necessary? We already know it's a gershayim variant. Shouldn't we just replace all gershayim with ASCII quotes?

static/js/sefaria/sefaria.js
```diff
@@ -1943,6 +1943,17 @@ _media: {},
...
           data["completions"][0] != query.slice(0, data["completions"][0].length))
   },
+  isGershayimVariant: function(query, data) {
+    function replace_gershayim_by_merkhaot(string) {
+        return string.replace('"', '״');
+    }
+    return (
+      !(data["is_ref"]) &&
+      data.completions &&
+      !data.completions.includes(query) &&
```
nsantacruz, line 1953: previous line is redundant. this line can be `!data.completions?.includes(query)`

static/js/sefaria/sefaria.js
```diff
@@ -1943,6 +1943,17 @@ _media: {},
...
+      data.completions.map(x => replace_gershayim_by_merkhaot(x)).includes(replace_gershayim_by_merkhaot(query))
```
nsantacruz, line 1954: I don't think `map()` will work in this case. It returns an array but you want a boolean. You're also checking `includes()` but don't you want to check for exact equality?

static/js/sefaria/sefaria.js
```diff
@@ -1943,6 +1943,17 @@ _media: {},
           data["completions"][0].toLowerCase().replace('״','"') == query.slice(0, data["completions"][0].length).toLowerCase().replace('״','"') &&
           data["completions"][0] != query.slice(0, data["completions"][0].length))
   },
+  isGershayimVariant: function(query, data) {
+    function replace_gershayim_by_merkhaot(string) {
```
nsantacruz, line 1947: I don't understand this function name. Should it be `replace_quotes_with_gershayim()`? (also, note the order of the replace will need to change if we switch to using unidecode).

static/js/Header.jsx
YishaiGlasner: how do we know that?

static/js/sefaria/sefaria.js
YishaiGlasner: we're replacing all the gershayim in all completions, and checking if the new array includes the quiry string (also after replacing)

static/js/sefaria/sefaria.js
YishaiGlasner: the order doesn't matter. we're not changing the server response. just how it compares query to data (i'll change the name)

static/js/sefaria/sefaria.js
YishaiGlasner: but using 'some' will be better

nsantacruz: ah! I misunderstood this. That makes sense.

static/js/Header.jsx
nsantacruz: Ah! now I understand the issue. Can you wrap this if statement in a function similar to `repairCaseVariant`?

static/js/Header.jsx
```diff
@@ -310,12 +310,10 @@ class SearchBar extends Component {
...
         }
-        if (Sefaria.isGershayimVariant(query, d)) {
-          if (query.includes('"')) {
-            this.submitSearch(query.replace('"', '״'));
-          } else {
-            this.submitSearch(query.replace('״', '"'));
-          }
+        const repairedQuery = Sefaria.repairGershayimVariant(query, d);
+        if (repairedQuery !== query) {
+          console.log(query, d);
```
nsantacruz, line 315: Can you remove this console.log?

nsantacruz (dismissed):

nsantacruz (approved):

PR #1679, opened 2023-10-26
Banners don't jumble up the mobile navigation menu anymore!

relyks:
A simple property change was made to ensure the navigation menu on mobile web is displayed properly when a banner is displayed.

In collaboration with @yonadavGit :)

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -606,7 +606,7 @@ input.noselect {
   font-size: 18px;
 }
 .mobileNavMenu {
-  position: fixed;
+  position: absolute;
   height: calc(100% - 60px);
   box-sizing: border-box;
   top: 60px;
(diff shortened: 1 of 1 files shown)


edamboritz (changes requested):
I'm not sure this solves the issue. Looking at some static pages (about, team etc) this fix actually doesnt allow the menu to show at all. 

Is there a fix we can do on the css of the banner, rather than the menu that can fix this?

edamboritz:
I've added two commits: 
The first fixes the issue where the `position:absolute`d nav manu would not have height because percentage of the containing ekement became meaningless, so changing to vh (viewport height units) fixed. 
The second is a fix that would not work for Firefox until later this year but seemed better to me in terms of maintaining the original intent - adding special css with the :has psuedo class that can target the nav menu if the banner element is present beofre it, and adds its height to the height and offest calculations done for the nav menu

nsantacruz (dismissed):

edamboritz (approved):

nsantacruz (approved):

PR #1680, opened 2023-10-29
Use Strapi for Teams page content

relyks:


relyks:
@edamboritz I just want to note here after you suggested it the other day. A team member can be made inactive (and not deleted) by unpublishing their entry in Strapi. The entry will stay there but won't show up in the JSON returned by API request

static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,249 @@ const StaticHR = () =>
 const ConditionalLink = ({ link, children }) =>
   link ? <a href={link} target="_blank">{children}</a> : children;
 
+/*
+* Team Page
+*/
+
+
+
+const partition = (arr, prop) =>
```
edamboritz, line 2645: Its not clear what this function is doing. A line of documentation might be helpful. Also, can we move this into the component its being used with?

edamboritz (approved):

edamboritz (dismissed):

static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,251 @@ const StaticHR = () =>
...
+                <BoardMember key={boardMember.id} boardMember={boardMember} />
+            ))}
+            {regularBoardMembers.sort(byLastName()).map((boardMember) => (
+                <BoardMember key={boardMember.id} boardMember={boardMember} />
+            ))}
+        </>
+    );
+};
+
+const fetchTeamMembersJSON = async () => {
```
edamboritz, line 2746: Please move this function into the component it is called in?

static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,251 @@ const StaticHR = () =>
...
+    }
+};
+
+const TeamMembersPage = memo(() => {
+    const [ordinaryTeamMembers, setOrdinaryTeamMembers] = useState([]);
+    const [teamBoardMembers, setTeamBoardMembers] = useState([]);
+    const [error, setError] = useState(null);
+
+    useEffect(() => {
+        const loadTeamMembers = async () => {
```
edamboritz, line 2806: We discussed not defining functions where they are called. Plese move this function definition into the components scope and not the hook's scope

edamboritz (approved):

PR #1684, opened 2023-10-31
fix: Performance improvement of Ref.contains

EliezerIsrael:
If `other` Ref is a whole book, and `self` begins anywhere after first segment, `contains` is trivially false and we can avoid expensive code.

### DIFF
--- a/sefaria/model/tests/ref_test.py
+++ b/sefaria/model/tests/ref_test.py
@@ -744,6 +744,7 @@ class Test_comparisons(object):
         assert not Ref("Exodus 6:2").contains(Ref("Exodus"))
         assert not Ref("Exodus 6:2-12").contains(Ref("Exodus"))
 
+        assert Ref("Leviticus").contains(Ref("Leviticus"))
         assert Ref("Leviticus").contains(Ref("Leviticus 1:1-27.34"))
         assert Ref("Leviticus").contains(Ref("Leviticus 1-27"))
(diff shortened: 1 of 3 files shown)


edamboritz (approved):
Amazing!!

sefaria/model/tests/ref_test.py
```diff
@@ -744,6 +744,17 @@ def test_contains(self):
         assert not Ref("Exodus 6:2").contains(Ref("Exodus"))
         assert not Ref("Exodus 6:2-12").contains(Ref("Exodus"))
 
+        assert Ref("Leviticus").contains(Ref("Leviticus 1:1-27.34"))
+        assert Ref("Leviticus").contains(Ref("Leviticus 1-27"))
+        assert Ref("Leviticus 1:1-27.34").contains(Ref("Leviticus"))
+        assert not Ref("Leviticus 1:1-27.30").contains(Ref("Leviticus"))
+        assert not Ref("Leviticus 1:2-27.30").contains(Ref("Leviticus"))
+        assert not Ref("Leviticus 2:2-27.30").contains(Ref("Leviticus"))
```
nsantacruz, line 752: @EliezerIsrael consider adding `assert Ref("Leviticus").contains(Ref("Leviticus"))`

EliezerIsrael: @nsantacruz  Added

PR #1688, opened 2023-11-01
Use Strapi for Jobs page

relyks:
This changes the Jobs page to use Strapi
* A JobsPage component was created. Jobs are grouped and listed by department.
* A [new content type](https://github.com/Sefaria/Sefaria-Strapi/commit/532cf729d1973b2fec20a93f5829f650819e5a43) was created in the Strapi instance for job postings. Job postings currently do not support being localized.


static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,222 @@ const StaticHR = () =>
...
+                    <h2 className="anchorable">{department}</h2>
+                </header>
+                <JobsListForDepartment key={department} jobsList={departmentJobPostings} />
+            </section>
+        );
+    };
+    
+    // Show all the job postings, but first group them by department and render each department separately
+    const JobPostings = ({ jobPostings }) => {
+        const groupedJobPostings = jobPostings.reduce((jobPostingsGroupedByDepartment, jobPosting) => {
```
edamboritz, line 2717: I think we can do this bucketing already when retrieving the data initially, no? Seems a shame for there to be a data manipulation step in this component?

static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,222 @@ const StaticHR = () =>
...
+                        ספריא איננה מחפשת כעת עובדים חדשים.
+                        עקבו אחרינו ב<a target="_blank" href="http://www.facebook.com/sefaria.org" >פייסבוק</a>&nbsp;
+                        כדי להשאר מעודכנים במשרות עתידיות.
+                    </span>
+                </p>
+            </div>
+        );
+    };
+
+    const [jobPostings, setJobPostings] = useState([]);
```
edamboritz, line 2759: The state and hooks are generally defined at the top of the component, for readability.

edamboritz (changes requested):
The code here is generally great! 
That being said there are several patterns at use that I think we can improve in terms of code quality and best practices: 
1. React components should not be defined inside other components, but rather along side them. This is an ineffective way to "scope" components as they wind up being re-created. 
2. Similarly, hooks should not be defining functions but rather calling other functions already defined on the component or other scope. 
3.

edamboritz (dismissed):

static/js/StaticPages.jsx
```diff
@@ -2636,6 +2636,224 @@ const StaticHR = () =>
...
+        throw error;
+    }
+};
+
+const JobsPage = memo(() => {
+    const [groupedJobPostings, setGroupedJobPostings] = useState({});
+    const [error, setError] = useState(null);
+
+    useEffect(() => {
+        const loadJobPostings = async () => {
```
edamboritz, line 2801: This function should be defined outside of the hook

edamboritz (dismissed):

edamboritz (approved):

PR #1689, opened 2023-11-02
Tineperiod strings

YishaiGlasner:


sefaria/model/topic.py
```diff
@@ -438,26 +438,36 @@ def contents(self, **kwargs):
     
     # A person may have an era, a generation, or a specific birth and death years, which each may be approximate.
     # They may also have none of these...
-    def most_accurate_time_period(self) -> Optional[TimePeriod]:
+    def _most_accurate_period(self, obj) -> Optional[LifePeriod]:
```
nsantacruz, line 441: Can you change `obj` to `TimePeriodClass: Type[TimePeriod]`. The `Type[TimePeriod]` will make it clear that the parameter needs to be of type `TimePeriod`. You also need to add this import: `from typing import Type`.

sefaria/model/topic.py
```diff
@@ -438,26 +438,36 @@ def contents(self, **kwargs):
...
         elif self.get_property("generation"):
-            return TimePeriod().load({"symbol": self.get_property("generation")})
+            return obj().load({"symbol": self.get_property("generation")})
         elif self.get_property("era"):
-            return TimePeriod().load({"symbol": self.get_property("era")})
+            return obj().load({"symbol": self.get_property("era")})
         else:
             return None
 
+    def most_accurate_time_period(self):
```
nsantacruz, line 466: Can you add documentation for these two functions. They're obviously simple, but I don't think it will be obvious when to use each one.

sefaria/model/topic.py
```diff
@@ -438,26 +439,42 @@ def contents(self, **kwargs):
...
+            return time_period_class().load({"symbol": self.get_property("generation")})
         elif self.get_property("era"):
-            return TimePeriod().load({"symbol": self.get_property("era")})
+            return time_period_class().load({"symbol": self.get_property("era")})
         else:
             return None
 
+    def most_accurate_time_period(self):
+        '''
+        :return: most accurate period as TimePeriod (used when a person's LifePeriod is using as a a general TimePeriod)
```
nsantacruz, line 469:
```suggestion
        :return: most accurate period as TimePeriod (used when a person's LifePeriod should be formatted like a general TimePeriod)
```

sefaria/model/topic.py
```diff
@@ -438,26 +439,42 @@ def contents(self, **kwargs):
...
+        '''
+        return self._most_accurate_period(TimePeriod)
+
+    def most_accurate_life_period(self):
+        '''
+        :return: most accurate period as LifePeriod. currently the difference from TimePeriod is only the string
```
nsantacruz, line 475:
```suggestion
        :return: most accurate period as LifePeriod. currently the only difference from TimePeriod is the way the time period is formatted as a string.
```

nsantacruz (approved):

PR #1691, opened 2023-11-05
Chore/sc 21398/fix failing topic tests

stevekaplan123:
The main problem was in [update topic titles](https://github.com/Sefaria/Sefaria-Project/compare/chore/sc-21398/fix-failing-topic-tests?expand=1#diff-19364da9061bcb429f3f3c5f63aeab5541e461154bd59e61719b2cd68f631356L1045).  The order in which I was removing old titles and adding new ones was leaving the old one cached.  I also added a test for alt titles and hebrew titles in the two tests below and created [distinct new titles](https://github.com/Sefaria/Sefaria-Project/compare/chore/sc-21398/fix-failing-topic-tests?expand=1#diff-3eb576d5ed14f2bd2ce8aa0c80864a7f72f02c2ed94cbe4005879679af587534R86) so that all the nodes don't end up with the same name.


sefaria/helper/topic.py
```diff
@@ -1043,14 +1043,16 @@ def topic_change_category(topic_obj, new_category, old_category="", rebuild=Fals
     return topic_obj
 
 def update_topic_titles(topic_obj, data):
-    for lang in ['en', 'he']:
```
nsantacruz, line 1046: This is feedback on the function in general and not on the changes made. `topic_obj` seems like it should be renamed to `topic` which seems more standard. It is preferable not to pass a parameter called `data` which is confusing to know what's in it. Instead, pass the contents of `data` as kwargs (i.e. `title`, `heTitle`, `altTitles`) and optional kwargs can default to `None` (it seems `altTitles` is optional). When calling `update_topic_titles()` you can call it using `**` like this `update_topic_titles(topic, **data)` which will pass the keys of `data` as kwargs.

stevekaplan123: @nsantacruz [New test here](https://github.com/Sefaria/Sefaria-Project/pull/1691/commits/a005786221e3b63c984e988b7ff5f97e69564e09#diff-3eb576d5ed14f2bd2ce8aa0c80864a7f72f02c2ed94cbe4005879679af587534R94)
[Changed remove title here](https://github.com/Sefaria/Sefaria-Project/pull/1691/commits/a005786221e3b63c984e988b7ff5f97e69564e09#diff-ba861216d2529388aee5b8984af25f2eb0eacedee84c915c33d2952ead8ebb33R128) 
Changed the code back so that I [remove titles first](https://github.com/Sefaria/Sefaria-Project/pull/1691/commits/a005786221e3b63c984e988b7ff5f97e69564e09#diff-19364da9061bcb429f3f3c5f63aeab5541e461154bd59e61719b2cd68f631356R1048) because now it's possible

sefaria/model/schema.py
```diff
@@ -125,6 +125,9 @@ def secondary_titles(self, lang=None):
         return [t for t in self.all_titles(lang) if t != self.primary_title(lang)]
 
     def remove_title(self, text, lang):
+        is_primary = len([t for t in self.titles if not (t["lang"] == lang and t["text"] == text and t.get('primary'))])
```
nsantacruz, line 128: shouldn't you remove `not`? wouldn't that locate a primary title?

sefaria/helper/topic.py
```diff
@@ -1042,81 +1042,79 @@ def topic_change_category(topic_obj, new_category, old_category="", rebuild=Fals
...
-    if not hasattr(topic_obj, 'properties'):
-        topic_obj.properties = {}
-    process_topic_place_change(topic_obj, data)
-    topic_obj = update_author_era(topic_obj, data, dataSource=dataSource)
-
-    return topic_obj
+    if not hasattr(topic, 'properties'):
+        topic.properties = {}
+    process_topic_place_change(topic, **kwargs)
+    return update_author_era(topic, **kwargs, dataSource=dataSource)
```
nsantacruz, line 1062: you can't use `**` before named parameters. move to be the last parameter

sefaria/helper/topic.py
```diff
@@ -1042,81 +1042,79 @@ def topic_change_category(topic_obj, new_category, old_category="", rebuild=Fals
...
-        for title in data['altTitles'][lang]:
-            topic_obj.add_title(title, lang)
-
-    topic_obj.add_title(data['title'], 'en', True, True)
-    topic_obj.add_title(data['heTitle'], 'he', True, True)
-    return topic_obj
-
-
-def update_authors_place_and_time(topic_obj, data, dataSource='learning-team-editing-tool'):
+def update_topic_titles(topic, **kwargs):
```
nsantacruz, line 1045: My idea of using `**data` here was that so that we could use explicit named parameters here which makes the function easier to understand. However, in many of the other cases in this PR where kwargs are used it seems fine because we don't know necessarily what the keys of the kwargs will be.

In summary: use kwargs when the keys are dynamic. Use named parameters when they're static so users can more easily understand how to call the function.

sefaria/helper/topic.py
nsantacruz: This is great! Good job on the tests.

nsantacruz (approved):

PR #1692, opened 2023-11-05
chore: Update VersionState Docs

EliezerIsrael:
chore: Update VersionState Docs

### DIFF
--- a/sefaria/model/version_state.py
+++ b/sefaria/model/version_state.py
@@ -30,25 +30,28 @@ class VersionState(abst.AbstractMongoRecord, AbstractSchemaContent):
     The `content` attribute is a dictionary which is the root of a tree, mirroring the shape of a Version, where the leaf nodes of the tree are dictionaries with a shape like the following::
         {
             "_en": {
-                "availableTexts":
-                "availableCounts":
-                "percentAvailable":
-                "textComplete":
-                'completenessPercent':
-                'percentAvailableInvalid':
-                'sparseness':  # was isSparse
+                "availableTexts":  Mask of what texts are available in this language.  Boolean values (0 or 1) in the shape of the JaggedArray
+                "availableCounts":  Array, with length == depth of the node.  Each element is the number of available elements at that depth.  e.g [chapters, verses]
+                "percentAvailable":  Percent of this text available in this language TODO: Only used on the dashboard. Remove?
+                'percentAvailableInvalid':  Boolean. Whether the value of "percentAvailable" can be trusted.  TODO: Only used on the dashboard. Remove?
+                "textComplete":  Boolean. Whether the text is complete in this language. TODO: Not used outside of this file. Should be removed.
+                'completenessPercent':  Percent of this text complete in this language TODO: Not used outside of this file. Should be removed.
+                'sparseness': Legacy - present on some records, but no longer in code TODO: remove
             }
-            "_he": ...
+            "_he": {...} # same keys as _en
             "_all" {
-                "availableTexts":
+                "availableTexts": Mask what texts are available in this text overall.  Boolean values (0 or 1) in the shape of the JaggedArray
                 "shape":
-                    For depth 1: Integer - length
-                    For depth 2: List of chapter lengths
-                    For depth 3: List of list of chapter lengths?
+                    For depth 1: Integer -length
+                    For depth 2: List of section lengths
+                    For depth 3: List of list of section lengths
             }
         }
 
-    For example, the `content` attribute for `Pesach Haggadah` will be a dictionary with keys: "Kadesh", "Urchatz", "Karpas" ... each with a value of a dictionary like the above.  They key "Magid" has a value of dictionary with keys "Ha Lachma Anya", etc.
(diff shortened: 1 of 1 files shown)


sefaria/model/version_state.py
```diff
@@ -20,73 +20,42 @@
...
+                'percentAvailableInvalid':
+                'sparseness':  # was isSparse
+            }
+            "_he": ...
+            "_all" {
+                "availableTexts":
+                "shape":
+                    For depth 1: Integer - length
+                    For depth 2: List of chapter lengths
+                    For depth 3: List of list of chapter lengths?
```
saengel, line 47: Is there a way we can make the comment about a `depth 3` instance definitive, instead of leaving it with a question mark?

sefaria/model/version_state.py
```diff
@@ -20,73 +20,42 @@
...
+            "_en": {
+                "availableTexts":
+                "availableCounts":
+                "percentAvailable":
+                "textComplete":
+                'completenessPercent':
+                'percentAvailableInvalid':
+                'sparseness':  # was isSparse
+            }
+            "_he": ...
```
saengel, line 41: Is `_he` dot-dot-dotted because it mimics the exact keys of the `_en` dictionary? If so, can we explicitly say that?

sefaria/model/version_state.py
```diff
@@ -20,73 +20,42 @@
...
+            "_all" {
+                "availableTexts":
+                "shape":
+                    For depth 1: Integer - length
+                    For depth 2: List of chapter lengths
+                    For depth 3: List of list of chapter lengths?
+            }
+        }
+
+    For example, the `content` attribute for `Pesach Haggadah` will be a dictionary with keys: "Kadesh", "Urchatz", "Karpas" ... each with a value of a dictionary like the above.  They key "Magid" has a value of dictionary with keys "Ha Lachma Anya", etc.
```
saengel, line 51: Can we give a simpler text as an example as well? The structure of the `Pesach Haggadah` is very unique, so I think having a simpler text as well as the `Pesach Haggadah` will be more instructive.

sefaria/model/version_state.py
```diff
@@ -20,73 +20,42 @@
...
 
 
 class VersionState(abst.AbstractMongoRecord, AbstractSchemaContent):
     """
     This model overrides default init/load/save behavior, since there is one and only one VersionState record for each Index record.
+
+    The `content` attribute is a dictionary which is the root of a tree, mirroring the shape of a Version, where the leaf nodes of the tree are dictionaries with a shape like the following::
+        {
+            "_en": {
+                "availableTexts":
```
saengel, line 33: Can we provide an explanation for each of these keys? Some of them are more self-evident than others. A brief line about each would be helpful.

saengel (commented):
Hi @EliezerIsrael - thank you for updating these docs! 

I just ran into a bug with `vstate` last week, so this was very instructive - and great timing for me to see. 

I left a couple of questions, but my larger question is just understanding a bit more of the context for the update. Does this have anything to do with some of the formatting we had been talking about for docstrings - or is this an unrelated PR to update docs? I assumed the latter, and therefore didn't write anything having to do with docstring style. Let me know - and I can give style feedback as well. 

Thanks!

sefaria/model/version_state.py
EliezerIsrael: You caught me.  That was there for years, and I just left it.  I verified; it's a list of list of integers.

EliezerIsrael:
Thanks for the feedback, @saengel.   Yes - this is just refining old documentation - content focused; not related to the format question.

sefaria/model/version_state.py
```diff
@@ -20,78 +20,50 @@
...
+                    For depth 1: Integer -length
+                    For depth 2: List of section lengths
+                    For depth 3: List of list of section lengths
+            }
+        }
+
+    For example:
+    - the `content` attribute for a simple text like `Genesis` will be a dictionary with keys "_en", "_he", and "_all", as above.
+    - the `content` attribute for `Pesach Haggadah` will be a dictionary with keys: "Kadesh", "Urchatz", "Karpas" ... each with a value of a dictionary like the above.
+        The key "Magid" has a value of dictionary with keys "Ha Lachma Anya", etc.
```
saengel, line 54: A thought about this line for clarity:

`The key "Magid" has a value of a dictionary, where each key is a different sub-section of Magid. The value for each key is a dictionary as detailed above, specific to each sub-section. So for example, one key will be "Ha Lachma Anya" and the value will be a dictionary, like the above, specific to the details of "Ha Lachma Anya". `

Perhaps also it would be worth adding a line that seems to encompass the overarching rule: `Every JaggedArrayNode has a corresponding vstate dictionary. So for complex texts, each leaf node (and leaf nodes by definition must be JaggedArrayNodes) has this corresponding dictionary`

EliezerIsrael: Incorporated in latest commit.

saengel (approved):
Looks great!

PR #1697, opened 2023-11-06
Period year

YishaiGlasner:
this PR should solve this problem in reindex ElasticSearch - https://sefaria.slack.com/archives/C1DAVPYJZ/p1699190830515399
it adds a new function to TimePeriod - determine_year_estimate, and uses it in page ranking rather than using the start attribute which is not a required attribute.

### DIFF
--- a/sefaria/model/timeperiod.py
+++ b/sefaria/model/timeperiod.py
@@ -225,6 +225,16 @@ class TimePeriod(abst.AbstractMongoRecord):
             else:
                 return topic.Topic({"properties.generation.value": self.symbol})
 
+    def determine_year_estimate(self):
+        start = getattr(self, 'start', None)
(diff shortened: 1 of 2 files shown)


sefaria/model/timeperiod.py
```diff
@@ -225,6 +225,16 @@ def get_people_in_generation(self, include_doubles = True):
             else:
                 return topic.Topic({"properties.generation.value": self.symbol})
 
+    def determine_year_estimate(self):
+        start = getattr(self, 'start', None)
+        end = getattr(self, 'end', None)
+        if start != None and end != None:
```
nsantacruz, line 231: Should be `is not None`

nsantacruz (approved):

PR #1704, opened 2023-11-08
[sc-20618] Topic Images Backend Work & CSS Refactor

saengel:
## Summary:
This PR is the part two of adding image functionality to the following six topic pages: Rosh Hashanah, Yom Kippur, The Four Species, Sukkot, Simchat Torah, and Shabbat. [The first PR](https://github.com/Sefaria/Sefaria-Project/pull/1654) added the functionality to the frontend, this second PR handles the backend, enabling image storage in our database. Additionally, this PR cleans up some of the CSS, to allow for more generic and reusable React components. 

## Code Changes:

**Backend**:
- An option `image` attribute was added to the Topic model in `topic.py`. Additionally, a check was added to the `_validate()` function to make sure an images saved to the DB are properly stored in GCP. This check relies on a new Exception class, `ExternalImageError` which was added to `exceptions.py`. 
- The `add_image_to_topic()` function was added to `helper/topic.py`. This function assists with data migration of existing images _already stored in GCP with the proper configuration_ to our topics in the database. 

**Frontend**: 
- The `<TopicImage />` component was refactored. The more general image with caption component was extracted into `<TopicWithCaption />` in `Misc.jsx`, with the CSS classes for the generalized component renamed to reflect the generalized nature of the component. The `<TopicImage /> component inside of `TopicsPage.jsx` wraps topic page specific CSS around the `<TopicWithCaption />`
- The hardcoded map of images was removed (since now we can pull our images from the database), and the `<TopicImage />` component was adjusted in order to rely on data from the topics API call.


static/js/Misc.jsx
```diff
@@ -3300,6 +3300,17 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
     )
 }
 
+const ImageWithCaption = ({photoLink, enCaption, heCaption }) => {
```
nsantacruz, line 3303: I would change the caption props to be `caption` that is an object with keys `en` and `he` to match the way we generally pass bilingual data.

static/js/TopicPage.jsx
```diff
@@ -742,11 +717,8 @@ const TopicImage = ({photoLink, enCaption, heCaption }) => {
   
   return (
     <div class="topicImage">
-        <img class="topicPhoto" src={photoLink}/>
-        <div class="topicImageCaption"> 
-          <InterfaceText text={{en:enCaption, he:heCaption}}  />
-        </div>
-      </div>);
+      <ImageWithCaption photoLink={photoLink} enCaption={enCaption} heCaption={heCaption} />
```
nsantacruz, line 720: same as above. I would make the prop `caption`.

static/js/Misc.jsx
saengel: I totally agree, that makes a lot more sense. Thanks Noah!

sefaria/model/topic.py
```diff
@@ -69,6 +70,11 @@ def _validate(self):
         super(Topic, self)._validate()
         if getattr(self, 'subclass', False):
             assert self.subclass in self.subclass_map, f"Field `subclass` set to {self.subclass} which is not one of the valid subclass keys in `Topic.subclass_map`. Valid keys are {', '.join(self.subclass_map.keys())}"
+        if getattr(self, 'image', False):
```
edamboritz, line 73: Its possible this code isnt doing what you expect it to do. 
Youre using the index() method to test for the existence of a substring and expecting a custom exception to be thrown. 
However, according to the docs, a `ValueError` is what is thrown in such a case. 
1. To correct this block of code, you'd need to catch that error and then raise the custom exception. That a lot of lines of code.  I'd suggest re-writing to use `in` or `find()` so you dont have to handle the ValueError explicitly. 
2. Is there code outside this function that is catching the custom exception for it to be useful? (for example, the Topic.save() method?)

sefaria/model/topic.py
```diff
@@ -44,7 +44,8 @@ class Topic(abst.SluggedAbstractMongoRecord, AbstractTitledObject):
         'good_to_promote',
         'description_published',  # bool to keep track of which descriptions we've vetted
         'isAmbiguous',  # True if topic primary title can refer to multiple other topics
-        "data_source"  #any topic edited manually should display automatically in the TOC and this flag ensures this
+        "data_source",  #any topic edited manually should display automatically in the TOC and this flag ensures this
+        'image'
```
edamboritz, line 48: We are starting use of a tool to validate attributes that are dictionaries and need to have their structure validated. Would be good to use here as well.

static/js/TopicPage.jsx
```diff
@@ -498,11 +471,12 @@ const TopicPage = ({
       onClickFilterIndex = displayTabs.length - 1;
     }
     const classStr = classNames({topicPanel: 1, readerNavMenu: 1});
+
     return <div className={classStr}>
         <div className="content noOverflowX" ref={scrollableElement}>
             <div className="columnLayout">
                <div className="mainColumn storyFeedInner">
-                    <TopicHeader topic={topic} topicData={topicData} topicTitle={topicTitle} multiPanel={multiPanel} setNavTopic={setNavTopic} openSearch={openSearch} openDisplaySettings={openDisplaySettings} />
+                    <TopicHeader topic={topic} topicData={topicData} topicTitle={topicTitle} multiPanel={multiPanel} setNavTopic={setNavTopic} openSearch={openSearch} openDisplaySettings={openDisplaySettings} topicImage={topicImage} />
```
edamboritz, line 479: Why does topic image need to be passed in explicitly and not as part of the larger topicData here?

static/js/TopicPage.jsx
```diff
@@ -633,7 +608,7 @@ TopicLink.propTypes = {
 };
 
 
-const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, setNavTopic, timePeriod, properties, topicTitle, multiPanel }) => {
+const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, setNavTopic, timePeriod, properties, topicTitle, multiPanel, topicImage }) => {
```
edamboritz, line 611: Same as above, though that pattern seems to predate this PR...

static/js/TopicPage.jsx
saengel: I originally tried getting it from topicData, but for some reason topicData seems to be filtered at some point along the road. It was easier to capture the specific data I needed and pass it through.

sefaria/model/topic.py
saengel: Ah OK thanks Ephraim, I will revisit this :)

saengel: Hi Ephraim - pushing a commit to fix this now, and to prefer the use of `in` in the context of an `assert`. In my prior conversation with @nsantacruz I recall him mentioning possibly us preferring the custom exceptions over `assert` statements, so I just want to loop him in here to check and clarify. Thanks!

static/js/TopicPage.jsx
saengel: Following up here - as per our discussion, we are considering leaving this as a tech-debt chore.

sefaria/model/topic.py
saengel: A further update (based also on our conversation) -- removed any need for an `assert` by using the new validation pattern.

sefaria/model/topic.py
saengel: Great, pushed! Let me know what you think :)

scripts/topic_data_migration.py
```diff
@@ -0,0 +1,43 @@
+import django
```
nsantacruz, line 1: We don't add one-time scripts to the repo anymore. This script should be removed and instead manually copied to the server that will run it.

sefaria/helper/topic.py
```diff
@@ -1261,4 +1261,23 @@ def delete_ref_topic_link(tref, to_topic, link_type, lang):
...
+    """
+    A function to add an image to a Topic in the database. Helper for data migration.
+    This function queries the desired Topic, adds the image data, and then saves.
+    :param topic_slug String: A valid slug for a Topic
+    :param image_uri String: The URI of the image stored in the GCP images bucket, in the topics subdirectory.
+                             NOTE: Incorrectly stored, or external images, will not pass validation for save
+    :param en_caption String: The English caption for a Topic image
+    :param he_caption String: The Hebrew caption for a Topic image
+    """
+    topic = Topic().load({'slug': topic_slug})
```
nsantacruz, line 1277: Clearer to use `Topic.init(topic_slug)`. `.init()` is a function that applies to any object with a slug (these objects all inherit from `SluggedAbstractMongoRecord`)

sefaria/model/topic.py
```diff
@@ -71,9 +100,9 @@ def _validate(self):
         super(Topic, self)._validate()
         if getattr(self, 'subclass', False):
             assert self.subclass in self.subclass_map, f"Field `subclass` set to {self.subclass} which is not one of the valid subclass keys in `Topic.subclass_map`. Valid keys are {', '.join(self.subclass_map.keys())}"
-
-        if getattr(self, 'portal_slug', None):
```
nsantacruz, line 75: Why were these lines deleted?

saengel: I mistakenly committed when I copied over the code from the portal to `topic.py` for the sake of mimicking the pattern - needed to delete those lines.

scripts/topic_data_migration.py
saengel: Got it! Thanks @nsantacruz :)

sefaria/helper/topic.py
saengel: Okay, sounds good! Thank you!

edamboritz (approved):
Great work!

PR #1715, opened 2023-11-15
Clarify text settings

yonadavGit:
- more conventional A (or ALPEH in hebrew site) icons in header for accessing text settings, instead of the a-aleph that many users seem to miss.
- verse numbers are darker (#000)
- bookmark and A button are darker (#666666)


static/js/Misc.jsx
```diff
@@ -1343,9 +1343,17 @@ class CloseButton extends Component {
 class DisplaySettingsButton extends Component {
   render() {
     var style = this.props.placeholder ? {visibility: "hidden"} : {};
-    var icon = Sefaria._siteSettings.TORAH_SPECIFIC ?
-      <img src="/static/img/ayealeph.svg" alt="Toggle Reader Menu Display Settings" style={style} /> :
-      <span className="textIcon">Aa</span>;
+    var icon;
```
nsantacruz, line 1346: Use `let` instead of `var`. `var` is from older versions of JS and is no longer very relevant to use.

nsantacruz (approved):

PR #1729, opened 2023-11-28
feat(text helper): function for modifying texts all over texts collec…

YishaiGlasner:
…tion.

### DIFF
--- a/sefaria/helper/text.py
+++ b/sefaria/helper/text.py
@@ -1,5 +1,9 @@
 # encoding=utf-8
+import unicodecsv as csv
+import io
 import re
+
+import pymongo
 
 from sefaria.model import *
 from sefaria.system.database import db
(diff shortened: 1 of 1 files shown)


sefaria/helper/text.py
```diff
@@ -1,6 +1,10 @@
 # encoding=utf-8
+import unicodecsv as csv
```
nsantacruz, line 2: In Python 3, `import csv` is enough. `csv` handles unicode.

nsantacruz (approved):

sefaria/helper/text.py
YishaiGlasner: but the lines:
    output = io.BytesIO()
    report = csv.writer(output)
won't work in the new csv module, and they were used here 3 times

PR #1735, opened 2023-12-04
Feature/sc 21570/staff who edit topic pages can upload an

stevekaplan123:
This PR allows for admins to upload and remove topic pictures.  The main frontend change is this new component [PictureUploader](https://github.com/Sefaria/Sefaria-Project/compare/master...feature/sc-21570/staff-who-edit-topic-pages-can-upload-an#diff-074d6375b4ee43ce060b48019510214b7aa2c3a3e1f9189d4dd23d5e46fc782eR1580-R1657)
I used this component in the TopicEditor and pass it down to the generic component AdminEditor that all admin tools use.  The PictureUploader can upload, remove, and show images.  

The [core backend change](https://github.com/Sefaria/Sefaria-Project/compare/master...feature/sc-21570/staff-who-edit-topic-pages-can-upload-an#diff-26c40bdd3eeaef442340754126fdbcb1ed9e451eb3bb31c02cb78053f5702eaaR3566-R3601) allows for uploading and deleting of images.  If the user uploads a picture and then uploads another picture, `old_filename` gets passed and is [deleted when uploading the new picture.](https://github.com/Sefaria/Sefaria-Project/pull/1735/files#diff-26c40bdd3eeaef442340754126fdbcb1ed9e451eb3bb31c02cb78053f5702eaaR3593). The PictureUploader component uses `old_filename` to know URL to display and passes this along to the API when a new picture is uploaded.

The uploading/removing of a picture takes place independently of saving the topic which could lead to problems, so here I [don't allow the user to upload/remove picture and then click "Cancel" thereby not saving the changed picture to the topic.](https://github.com/Sefaria/Sefaria-Project/compare/master...feature/sc-21570/staff-who-edit-topic-pages-can-upload-an#diff-19b9d117defe6ff554421255c9ada09189fc0c3bcd2be0f30095311b20a958ccR42-R49)  If the user uploads/removes a picture, they must save their changes.

I also had to add a variable TOPICS_BUCKET since this bucket is currently not accessed in code.


reader/views.py
```diff
@@ -3560,6 +3562,43 @@ def profile_follow_api(request, ftype, slug):
     return jsonResponse({"error": "Unsupported HTTP method."})
 
 
+@catch_error_as_json
+def topic_delete_photo(request, file):
```
edamboritz, line 3566: In the future, we should definitely look to implement more restful api endpoints that handle both POST and DELETE under the same API

reader/views.py
```diff
@@ -3560,6 +3562,43 @@ def profile_follow_api(request, ftype, slug):
...
+    import uuid
+    import base64
+    if not request.user.is_authenticated:
+        return jsonResponse({"error": _("You must be logged in to update a topic photo.")})
+    bucket_name = GoogleStorageManager.TOPICS_BUCKET
+    file = request.POST.get('file')
+    old_filename = request.POST.get('old_filename')  # delete file from google storage if there is one there
+    if old_filename:
+        old_filename = f"topics/{old_filename.split('/')[-1]}"
+    if request.method == "DELETE":
```
edamboritz, line 3590: What is this DELETE doing versus the separate api endpoint?

sefaria/urls.py
```diff
@@ -102,7 +102,10 @@
     url(r'^topics/?$', reader_views.topics_page),
     url(r'^topics/b/(?P<topic>.+)$', reader_views.topic_page_b),
     url(r'^topics/(?P<topic>.+)$', reader_views.topic_page),
-    url(r'^api/topic/completion/(?P<topic>.+)', reader_views.topic_completion_api)
+    url(r'^api/topic/completion/(?P<topic>.+)', reader_views.topic_completion_api),
+    url(r'^api/topics/upload-image$', reader_views.topic_upload_photo),
```
edamboritz, line 106: I would change this to be api/topics/image

reader/views.py
edamboritz: It seems, from the client code below, that this api view is never called with the DELETE http verb. 
Therefore it doesnt make sense to be checking for it here in the view. 

That being said, the functionality inside `topic_delete_photo()` _should_ be placed in this same API view as a separate path for when the api is called with DELETE - which is when the user wants to completely delete a topic photo, not when they want to replace it during editing - that part is implicitly handled in the code for the upload already.

sefaria/helper/topic.py
```diff
@@ -1105,6 +1105,13 @@ def update_topic(topic, **kwargs):
     if "description" in kwargs or "categoryDescription" in kwargs:
         topic.change_description(kwargs.get("description", None), kwargs.get("categoryDescription", None))
 
+    if "image" in kwargs:
```
edamboritz, line 1108: As we discussed verbally, Might be worth adding a line of inline comment explaining that this is to validate against orphaned captions.

sefaria/urls.py
edamboritz: and also consolidate the api endpoints into one, as discussed.

static/js/Misc.jsx
```diff
@@ -1575,6 +1577,84 @@ FollowButton.propTypes = {
 
 };
 
+const PictureUploader = ({callback, old_filename, caption}) => {
```
edamboritz, line 1580: Worth noting that if we stipulate a caption as a prop to this component, then its not really a generic picture uploader tool, but one that is specific to topics.

static/js/Misc.jsx
```diff
@@ -1575,6 +1577,84 @@ FollowButton.propTypes = {
...
+     */
+    const fileInput = useRef(null);
+
+    const uploadImage = function(imageData, type="POST") {
+      const formData = new FormData();
+      formData.append('file', imageData.replace(/data:image\/(jpe?g|png|gif);base64,/, ""));
+      if (old_filename !== "") {
+        formData.append('old_filename', old_filename);
+      }
+      $.ajax({
```
edamboritz, line 1592: please use vanilla js here and not jQuery

edamboritz (changes requested):
Really good work! 
See some notes below. 

It occurs to me that in the future we will want to think about the overall architecture both of the client components of the Admin Tools and their backend interaction - but that is for another time. 

One thing we should consider is that its currently possible for a staff memeber to upload a picture (and have it actually upload to the server) but the saving process of the topic never happens so that image gets lost and never connected to the topic in the database (whether internet connection issues, back button etc). 
We should consider either always saving the topic when an image uploads or alternatively, delaying the actual call to upload the image so its bundled with the image save.

edamboritz (approved):

PR #1740, opened 2023-12-05
Trends Cronjob & General Cronjob House Cleaning

edamboritz:
This PR simply adds another cronjob in the helm chart. 

It adds a manifest file for the cronjob `trends.yaml` and a corresponding script `recalculate_trends.py`
It adds a variable for enabling it in the values file and enables it where other cronjobs are enabled. 

This PR also moves any python scripts that are used by cron jobs into their own dedicated directory, separate from other scripts that are legacy in the project code. To do so we also refactor all other cronjob manifests to point to the new script locations.


rneiss (dismissed):

BrendanGalloway (approved):
pytest is failing, but helm changes look correct

PR #1743, opened 2023-12-13
Helperbugs

stevekaplan123:
I want to re-structure Guide for the Perplexed and several commentaries on it.  The script for this is here: https://github.com/Sefaria/Sefaria-Data/blob/master/sources/Content_Quality/guide%20%2B%20comm.py
Its use is `./run "guide + comm.py" "Guide for the Perplexed"`
and then the same usage but with "Crescas on Guide for the Perplexed", "Efodi on Guide for the Perplexed" and "Shem Tov on Guide for the Perplexed".  This will need to be run on prod therefore multiple times.
I made a few changes to allow for the re-structuring of Guide for the Perplexed.
1. `exact_match`: I added this parameter to `change_parent` so that links to "X on Guide for the Perplexed" are not modified when re-structuring "Guide for the Perplexed."  This is crucial because otherwise it would generate invalid refs.  When my script is used again to re-structure commentaries on Guide for the Perplexed it then will modify the links with refs "X on Guide for the Perplexed".
2. `ignore_cascade`: I added this parameter to `change_node_title` because I wanted to cascade differently than this function allowed.
3. `force_update`: The most significant change is this one which modifies actual Sefaria Project/model code.  After changing the parent, the new_normal_form was not updating properly.  Calling node.ref() would give me the old title of the node, so I needed to run `new_normal_form = node.ref(force_update=True).normal()` so I had to change `node.ref()` itself


sefaria/helper/schema.py
```diff
@@ -338,17 +342,17 @@ def refresh_version_state(title):
...
     If the title changing is the primary english title, cascades to all of the impacted objects
     :param snode:
     :param old_title:
     :param lang:
     :param new_title:
+    :param ignore_cascade:
     :return:
     """
-
     def rewriter(string):
```
edamboritz, line 356: Really what we want here is a refactor that makes rewriter param a callback thats passed in, instead of hard coded to one piece of logic?

sefaria/helper/schema.py
```diff
@@ -306,13 +307,16 @@ def change_parent(node, new_parent, place=0):
...
     node.parent = new_parent
-    new_normal_form = node.ref().normal()
+    new_normal_form = node.ref(force_update=True).normal()
 
     index.save(override_dependencies=True)
     library.rebuild()
 
     for link in linkset:
-        link.refs = [ref.replace(old_normal_form, new_normal_form) for ref in link.refs]
+        if exact_match:
```
edamboritz, line 316: Seems there's a broader issue here of asking why link refs that simply contain the original title are all indiscriminately re-written?

sefaria/model/schema.py
```diff
@@ -1374,11 +1374,11 @@ def version_address(self):
         """
         return self.address()[1:]
 
-    def ref(self):
+    def ref(self, force_update=False):
         from . import text
         d = {
             "index": self.index,
-            "book": self.full_title("en"),
+            "book": self.full_title("en", force_update=force_update),
```
edamboritz, line 1381: The change here is fine in the current context, that being said - the proper thing to do here would be to address the underlying cache issue.

edamboritz (approved):
This looks good overall to me, though we should at least create tickets fr the raises review issues, if not dealt with at this time. 
Furthermore we should verify that there are enough tests around these functions to make sure that this doesnt break any existing functionality.

PR #1744, opened 2023-12-14
Updated 990 for Ways To Give Page

relyks:
The most recent 990 form was added and the links on the English and Hebrew pages were updated to point to that new PDF file.


shaneerosen (approved):

relyks:
@edamboritz PyTest isn't passing on this 🤷

PR #1748, opened 2023-12-19
Allow "Chair" as a title for "Chairman"

relyks:
Currently, in the Board Members section of the Team page, the "chairman" is listed first and then the "co-founders" are listed next before the other board members. Marketing and Communications has requested that the title for Sefaria's current chairman to be "Chair" instead. 

This change was necessary to keep the chair first instead of being included with the other board members. This PR changes to allow for different variations to be used for those individuals' titles while still preserving the same ordering.

Should a more complex and custom ordering be needed in the future, each content entry for a team board member will need to be given a specified type instead of using their title as a workaround to determine their type.

### DIFF
--- a/static/js/StaticPages.jsx
+++ b/static/js/StaticPages.jsx
@@ -2712,26 +2712,23 @@ const BoardMember = ({ boardMember }) => (
 );
 
 const BoardMembers = ({ boardMembers }) => {
-    let chairmanBoardMember;
-    const chairmanIndex = boardMembers.findIndex(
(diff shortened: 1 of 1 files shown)


edamboritz:
May I suggest a different approach here? 
Or rather, one you already did yourself? 

Lets have the **_data_** reflect with a flag or field the seniority level of certain board members, but leave the string describing their title as just that - descriptive. 
Then if you use the data to order the members and just output the title as is, it can be changed as needed without code changes.

PR #1754, opened 2023-12-24
Api final

YishaiGlasner:
doc strings
change 400 to 404
refactor TextRange - change versions to be a private attribue with a pu bublic setter


sefaria/model/text.py
```diff
@@ -1334,12 +1334,12 @@ class Version(AbstractTextRecord, abst.AbstractMongoRecord, AbstractSchemaConten
...
         "hasManuallyWrappedRefs",  # true for texts where refs were manually wrapped in a-tags. no need to run linker at run-time.
-        "actualLanguage",
-        'languageFamilyName',
-        "isBaseText",
-        'isSource',
-        'isPrimary',
-        'direction', # 'rtl' or 'ltr'
+        "actualLanguage",  # ISO language code
+        'languageFamilyName',  # full name of the language, but without specificity (for Judeo Arabic actualLanguage=jrb, languageFamilyName=arabic
+        "isBaseText",  # should be deprecated (needs some changes on client side)
```
edamboritz, line 1339: can probs remove this line now

edamboritz (dismissed):

sefaria/model/text.py
YishaiGlasner: no. we should deploy, then remove it from data, and then we can remove this line (and say everyone to remove it from data, doing it on shmuels' cauldron, etc.)

edamboritz: Sure

edamboritz (approved):

PR #1759, opened 2023-12-26
fix(Topics): display without links in the TOC but with links on topic pages

stevekaplan123:
We don't want to display links in markdown text in all places on the site, specifically in the topic TOC.  However, we still want other markdown tags to show, such as italics.  Therefore, I added a field to InterfaceText `disallowedMarkdownElements` which is a list of the elements you don't want to show in markdown.  In this case in the change I made in TopicPage.jsx, I pass ['a'] and links do not show up but italics does show up.

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -60,7 +60,7 @@ const __filterChildrenByLanguage = (children, language) => {
   return newChildren;
 };
 
-const InterfaceText = ({text, html, markdown, children, context}) => {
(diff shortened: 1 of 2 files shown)


edamboritz (approved):
We are overloading InterfaceText with too much functionality, its getting out of hand. 
It will become more and more difficult to use in isolation. 
Why was the makdown rendering incorporated into it in the first place and not the other way around? 
InterfaceText should be used as a low level component... 

Also please note that our usage of the library seems to be out of date: https://www.npmjs.com/package/react-markdown?activeTab=versions 

I'm approving this for now but there needs to be a serious refactor conversation here.

PR #1762, opened 2024-01-02
fix(Topics): topics shown in sidebar did not display markdown properly

stevekaplan123:
Markdown text currently shows in the sidebar instead of being rendered as markdown.  This is because ContentText never was fully implemented to work with markdown.  ContentText previously received markdown as a prop but did nothing with it.  It now passes it down to ContentSpan which renders markdown appropriately.


edamboritz (approved):
See same notes here: https://github.com/Sefaria/Sefaria-Project/pull/1759#pullrequestreview-1804110958

PR #1763, opened 2024-01-03
Add a new redirect for tracking Munich promotional presentation

relyks:
@EliezerIsrael is presenting at a talk in Munich and would like to provide a shortened user-friendly URL to facilitate donations (`sefaria.org/give`). The URL provided has a `mu` GET parameter to track the campaign.  

### Why, Sky?
* `sefaria.org/give` was chosen over `give.sefaria.org` due to simplicity. With the latter, an update to a CNAME record on CloudFlare would need to be made, a separate redirect in the Nginx configuration would need to be added, and a helm chart update would need to happen. CNAME records are unable to resolve to URLs with paths or GET parameters, so the task at hand can not solely be handled through DNS.
* A custom redirect method was also added, so that the path `sefaria.org/give/X` can redirect to the donation campaign URL with the GET parameter set to `X`. This will allow the use of future redirect URLs without making more code changes.

### Future considerations
If this becomes a regular request to have other user-friendly links that redirect to specific campaigns (besides ones with `give` as the path), a separate utility or admin panel plugin should be created to save desirable URL paths and their counterpart donation URLs to the database. A URL handler could check if there are any special donation paths available. However, caching for those URLs would be necessary to prevent slow URL resolution from making a call to the database on every request.


EliezerIsrael:
@relyks I think we should add a generic redirect like sefaria.org/give/X , where X is any campaign ID, that redirects to https://donate.sefaria.org/give/550774/#!/donation/checkout?c_src=X.  That would be more future-proof, and excuse us from doing one-off redirects.

EliezerIsrael (changes requested):
@relyks I think we should add a generic redirect like sefaria.org/give/X , where X is any campaign ID, that redirects to https://donate.sefaria.org/give/550774/#!/donation/checkout?c_src=X. That would be more future-proof, and excuse us from doing one-off redirects.

relyks:
> @relyks I think we should add a generic redirect like sefaria.org/give/X , where X is any campaign ID, that redirects to https://donate.sefaria.org/give/550774/#!/donation/checkout?c_src=X. That would be more future-proof, and excuse us from doing one-off redirects.

I agree. I implemented what you suggested :+1:

edamboritz (approved):

PR #1770, opened 2024-01-09
Feature/sc 21691/recently viewed texts

stevekaplan123:
This PR adds a feature whereby the /texts page lists the 3 recently viewed history items using the RecentlyViewed component.  Server side rendering is used so that when `texts_list` is called on the backend, a new function `get_user_history_props` is called which loads up the first 20 history items and these items get stored in the Sefaria.userHistory cache.  Then, the RecentlyViewed component filters the 20 history items, looking for the first 3 unique history items (i.e. it won't show the same book twice).  If server side rendering failed and the `Sefaria.userHistory` cache is empty, the `RecentlyViewed` component can load the first 20 history items by calling a new function `Sefaria.loadUserHistory`.  Finally, from the Recently Viewed component, there is a link "All history" to take the user to the history page, and no history items need to be loaded in the history page as the first 20 history items are already in the cache.


static/js/NavSidebar.jsx
```diff
@@ -85,6 +87,59 @@ const TitledText = ({enTitle, heTitle, enText, heText}) => {
...
+        Sefaria.userHistory.items.forEach(x => {
+        if (!booksFound.includes(x.book) && x.book !== "Sheet") {
+           booksFound.push(x.book);
+           itemsToShow.push(x);
+        }});
+
+        itemsToShow = itemsToShow.slice(0, 3).map((x, i) => {
+            return <li><a href={x.ref} onClick={() => trackItem(x.ref, i+1)}>{Sefaria.interfaceLang === "hebrew" ? x.he_ref : x.ref}</a></li>;
+         });
+        setRecentlyViewedItems(itemsToShow);
```
edamboritz, line 114: I'd rather you set state with data, not html. then use said state later on in the render process. 
also, if its being repeated it might be wroth createing a small component to encapsulate the render of the item

static/js/NavSidebar.jsx
```diff
@@ -85,6 +87,59 @@ const TitledText = ({enTitle, heTitle, enText, heText}) => {
...
+const RecentlyViewed = ({toggleSignUpModal, mobile}) => {
+   const [recentlyViewedItems, setRecentlyViewedItems] = useState(null);
+   const handleAllHistory = (e) => {
+    if (!Sefaria._uid) {
+      e.preventDefault();
+      toggleSignUpModal(SignUpModalKind.ViewHistory);
+    }
+    gtag('event', 'recently_viewed', {link_type: 'all_history', logged_in: !!Sefaria._uid});
+   }
+   const trackItem = (ref, pos) => {
```
edamboritz, line 99: Whats the relationship to the tracking code above this? 
i.e. why are there two?

static/js/sefaria/strings.js
```diff
@@ -453,6 +454,10 @@ const Strings = {
     "Alternate Hebrew Titles": "חלופות כותר עבריות",
     "Alternate English Titles": "חלופות כותר לועזיות",
     "Optional": "אופציונלי",
+    "Recently viewed": "נצפו לאחרונה",
+    "All history >": "כל ההסטוריה >",
```
edamboritz, line 458: the chevron should be styled in css not text

static/js/NavSidebar.jsx
```diff
@@ -85,6 +87,59 @@ const TitledText = ({enTitle, heTitle, enText, heText}) => {
   </Module>
 };
 
+const RecentlyViewed = ({toggleSignUpModal, mobile}) => {
+   const [recentlyViewedItems, setRecentlyViewedItems] = useState(null);
+   const handleAllHistory = (e) => {
+    if (!Sefaria._uid) {
+      e.preventDefault();
+      toggleSignUpModal(SignUpModalKind.ViewHistory);
```
edamboritz, line 95: I think perhaps this needs to be handled outside the component, i'm not sure I love that this is being passed into it.

edamboritz (commented):
This is really clean and well done code. 
Glad to see earlier feedback was incorporated. 
Note a few comments below.

static/js/NavSidebar.jsx
stevekaplan123: In my most recent push, I fixed this by setting state with data rather than HTML.  I created a small component RecentlyViewedList as well so as not to duplicate code.

static/js/NavSidebar.jsx
stevekaplan123: trackItem tracks each ref in the list, whereas the gtag in handleAllHistory tracks clicks of the "All History" link.

static/js/TextsPage.jsx
```diff
@@ -108,6 +109,7 @@ const TextsPage = ({categories, settings, setCategories, onCompareBack, openSear
             { about }
             { dedication }
             { libraryMessage }
+            { !multiPanel && <RecentlyViewed toggleSignUpModal={toggleSignUpModal} mobile={true}/>}
```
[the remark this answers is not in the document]
stevekaplan123: I could implement another component RecentlyViewedMobile but it would require taking 2 functions out of RecentlyViewed component and having them be standalone functions that are shared.  Is this preferable over the current approach?

static/js/NavSidebar.jsx
stevekaplan123: I don't see how to handle it outside the component.  Within the component is a link to the history page and if one is logged out when they click to go to their history page, we want to prompt them to sign up/sign in.  I can't think of another approach.  What do you think?

edamboritz (approved):

PR #1780, opened 2024-01-18
fix: Fix recursion call in `TreeNode.traverse_tree`.  

EliezerIsrael:
Also add docs to `Topic.merge`


saengel (approved):
Looks great and fixes the bug!

PR #1792, opened 2024-01-30
fix(api): patches a bug where setting a url param to false or 0 would…

rneiss:
… return true at certain API endpoints.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -48,7 +48,7 @@ class Text(View):
         if not versions_params:
             versions_params = ['primary']
         versions_params = [self.split_piped_params(param_str) for param_str in versions_params]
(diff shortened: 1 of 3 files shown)


edamboritz (approved):
This looks good! 
Have we made sure that client side code using these APIs in their bugged out state doesnt need to be corrected to match?

rneiss:
> This looks good! Have we made sure that client side code using these APIs in their bugged out state doesnt need to be corrected to match? From looking at the code superficially it doesn't look like anything would break, but just making sure

Yep. In each of these cases the client code always sent a `1` as the value or didn't include the attribute at all thus resulting in the proper falsey response.

PR #1827, opened 2024-03-05
Feature/sc 22700/super about page

stevekaplan123:
The "super" about page refers to the fact that many pages are now presented together through the use of a sidebar: about, team, jobs, supporters, testimonials metrics, updates, terms of use, and privacy policy pages are part of the "super" about page.  The pages are still technically separate from each other, but the user is presented with a sidebar on each of these pages with links to the other pages.  The sidebar is implemented in the HTML page _sidebar.html.  On mobile the design is a bit different.  There isn't a sidebar on each page in mobile.  Instead, when you click 'about sefaria' in the mobile nav menu, you will be brought to a page (/mobile-about-menu) that simply displays the sidebar menu with links to the about, team, jobs, supporters, testimonials metrics, updates, terms of use, and privacy policy pages.  

All of the pages are implemented as separate Django templates.  Most of the code of this PR is found in additions to static.css, although I had to add a few lines of CSS to each Django template in order that each of the pages in the super about page would use flexbox without applying flexbox to all static pages.  

/updates used to be an API integrated into the React app as a possible value for menuOpen in the ReaderPanel.  This is no longer the case as /updates is implemented via a Django templates updates.html.  This Django template includes the React component UpdatesPanel.  I had to change some of the JavaScript code in the scroll handler of the UpdatesPanel to get it to continue working with the sidebar being present on the page.  Previously, in the updates page, `.readerNavMenu` was absolute and the scroll handler was tied to the child, `.content`, of `.readerNavMenu`.  However, this absolute positioning was interfering in the flexbox css used to create the sidebar.  I changed UpdatesPanel so that the `.readerNavMenu` is not positioned absolute on that page, and the scroll handler is tied to the window instead of `.readerNavMenu`'s child `.content`


static/js/UpdatesPanel.jsx
```diff
@@ -27,15 +26,23 @@ class UpdatesPanel extends Component {
...
     this.getMoreNotifications();
   }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.handleScroll);
+  }
   handleScroll() {
     if (this.state.loadedToEnd || this.state.loading) { return; }
-    var $scrollable = $(ReactDOM.findDOMNode(this)).find(".content");
+    var $scrollable = $(window);
```
edamboritz, line 40: This is still jQuery? Seems a shame to leave it...

templates/static/team.html
```diff
@@ -6,6 +6,14 @@
 
 {% block description %}{% trans "Sefaria's team is working to create the Jewish future." %}{% endblock %}
 
+{% block css %}
+    #staticContentWrapper #content {
+      display: flex;
+      flex-direction: row;
+    }
+
+{% endblock %}
```
akiva10b, line 15: See if you can use a subclass of interface-{lang} to get it to work instead of clock css

would end up something like:

style.css:
```css
      .interface-he {
        direction: rtl;
        text-align: right;
      }
      .interface-he .flex {
        display: flex;
        flex-direction: row-reverse;
      }

``` 

base.html
```html
    {% if not html %}
    <div id="staticContentWrapper">
        <div id="content" class="interface-he">
            <div class="flex"></div>
        </div>
    </div>
```

akiva10b (approved):

templates/static/team.html
edamboritz: setting proper direction here should also let you not bother with setting flex direction separately fro hebrew interface?

edamboritz (approved):
I have left a few non critical comments but looks fine. I'm not sure i'd make the same css choices but if they work thats good.

PR #1830, opened 2024-03-07
feat: add helm config for tasks pods

BrendanGalloway:


helm-chart/sefaria-project/templates/_helpers.tpl
```diff
@@ -175,3 +175,13 @@ preferredDuringSchedulingIgnoredDuringExecution:
             - "true"
 {{- end }}
 {{- end }}
+
+{{/*
+Setup complete tasks queue info
+*/}}
+{{- define "sefaria.tasks.internalQueues" }}
```
nsantacruz, line 182: @BrendanGalloway can you explain what this code is doing?

helm-chart/sefaria-project/values.yaml
```diff
@@ -235,6 +235,34 @@ varnish:
     image: ibmcom/varnish-metrics-exporter
     tag: 0.32.0
 
+tasks:
+  enabled: false
+  resources:
+    requests:
+      memory: "100Mi"
```
nsantacruz, line 242: @BrendanGalloway these resources may be too low but we'll see once the pod comes up.

BrendanGalloway: ack, just put in random values for now

helm-chart/sefaria-project/templates/_helpers.tpl
BrendanGalloway: The queues.tasks value should be derived from the environment to avoid collisions.  Ideally a user shouldn't be required to make sure the .tasks.queues.tasks value and the .deployEnv value align, so this sets up an internal variable with pre-build queue info, which is then merged with the value provided by the user before being passed to the app.  See the configmap CELERY_QUEUES value and the tasks pod args for usage

nsantacruz (approved):

PR #1831, opened 2024-03-07
fix(Commentary Redirects): Ramban on Genesis, Introduction

stevekaplan123:
Ramban on Genesis, Introduction should not redirect to Genesis, Introduction.  Instead clicking on it from the TOC should send you to Ramban on Genesis, Introduction.  I also caught a potential other bug whereby the parsedRef cache was being modified via the variable 'commRef' whereas the copy of the object 'commRef', 'commRefCopy' should be modified instead.

My reason for putting the check in `isCommentaryRefWithBaseText` because it will prevent any attempt to convert "Ramban on Genesis, Introduction" since `openPanelAt` directly calls `isCommentaryRefWithBaseText` to determine whether to convert the ref.  I could've put this check in the actual conversion code (i.e. sefaria.js's `convertCommentaryRefToBaseRef`) and then had it fail but I think this would overly complicate the code.

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -2206,7 +2206,20 @@ _media: {},
       let refToCheck = Array.isArray(ref) ? ref[0] : ref;
       const parsedRef = Sefaria.parseRef(refToCheck);
       const book = Sefaria.index(parsedRef.index);
-      return this.isCommentaryWithBaseText(book);
+      if (!this.isCommentaryWithBaseText(book)) {
+          return false;
+      }
+
+      // by this point, we know the book is in the right form, but we still need to check that the ref is in the right form
+      // The ref 'Ramban on Genesis, Introduction 1' shouldn't generate "Genesis, Introduction 1"
+      // this can be tested by modifying the ref and then calling parseRef on the modified ref
+      const parsedRefCopy = Object.create(parsedRef);
+      const baseText = book.base_text_titles[0];
+      parsedRefCopy.ref = parsedRefCopy.ref.replace(book.title, baseText);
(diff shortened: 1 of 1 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -2206,7 +2206,20 @@ _media: {},
       let refToCheck = Array.isArray(ref) ? ref[0] : ref;
       const parsedRef = Sefaria.parseRef(refToCheck);
       const book = Sefaria.index(parsedRef.index);
-      return this.isCommentaryWithBaseText(book);
+      if (!this.isCommentaryWithBaseText(book)) {
+          return false;
+      }
+
+      // by this point, we know the book is in the right form, but we still need to check that the ref is in the right form
```
akiva10b, line 2213: could this logic be useful elsewhere as a helper function? It would also be cleaner to have a function name describing what this code is doing (isValideRef or something of the like)

static/js/sefaria/sefaria.js
```diff
@@ -2206,7 +2206,20 @@ _media: {},
...
+      // The ref 'Ramban on Genesis, Introduction 1' shouldn't generate "Genesis, Introduction 1"
+      // this can be tested by modifying the ref and then calling parseRef on the modified ref
+      const parsedRefCopy = Object.create(parsedRef);
+      const baseText = book.base_text_titles[0];
+      parsedRefCopy.ref = parsedRefCopy.ref.replace(book.title, baseText);
+      if (Sefaria.parseRef(parsedRefCopy.ref).error) {
```
akiva10b, line 2219: this could probably be:
`return !!Sefaria.parseRef(parsedRefCopy.ref).error`
If I understand this correctly.

PR #1838, opened 2024-03-12
New autocomplete

yonadavGit:


static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+
+    useEffect(() => {
+      showVirtualKeyboardIcon(false); // Initially hide the virtual keyboard icon
+    }, []);
+   const { onBlur, onKeyDown, ...otherDownShiftProps } = getInputProps();
+
+   const clearSearchBox = function () {
+     getInputProps().onChange({ target: { value: '' } });
+  }
+   const submitSearch = (query) => {
```
nsantacruz, line 455: This function is long and I think it would be better if it was pulled out of the InputBox component. The function can return an object of the form:

```
{ type: str, id: str }
```

Where `type` is either `Ref`, `Topic`, `Person`, `Collection`, `TocCategory` or `Search`
`id` is the unique ID of the search result. For `Ref`, that's the ref itself. For `Topic` that's the `topic_slug`, for `Search` that's the query itself. For all other types it's `key`

I believe this way we can separate the state management from the search logic.

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+          onBlur={blurSearch}
+          maxLength={75}
+          title={Sefaria._("Search for Texts or Keywords Here")}
+          {...otherDownShiftProps}
+        />
+      </div>
+    );
+  };
+const SuggestionsDispatcher = ({ suggestions, getItemProps, highlightedIndex}) => {
+  function groupByType(seggestedItems) {
```
nsantacruz, line 591: this function can be pulled out of this component

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+    return result;
+}
+
+    let groupedSuggestions = groupByType(suggestions);
+    let universalIndex = 0;
+
+    return (
+        <>
+            {groupedSuggestions.map((object, index) => {
+                const UniversalInitialIndexForGroup = universalIndex;
```
nsantacruz, line 619: I think this variable should be called `initialIndexForGroup`

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+                universalIndex += object.items.length;
+                return (
+                    <SuggestionsGroup
+                        getItemProps={getItemProps}
+                        highlightedIndex={highlightedIndex}
+                        key={index}
+                        suggestions={object.items}
+                        universalInitialIndex={UniversalInitialIndexForGroup}
```
nsantacruz, line 627: I think this prop should be called `initialIndexForGroup`

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+    return (
+        <>
+            {groupedSuggestions.map((object, index) => {
+                const UniversalInitialIndexForGroup = universalIndex;
+                universalIndex += object.items.length;
+                return (
+                    <SuggestionsGroup
+                        getItemProps={getItemProps}
+                        highlightedIndex={highlightedIndex}
+                        key={index}
```
nsantacruz, line 625: The `key` should be the type of the group, not the index. Using the index is considered an anti-pattern in React.

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+            })}
+        </>
+    );
+};
+
+ const Autocomplete = ({onRefClick, showSearch, openTopic, openURL}) => {
+  const [suggestions, setSuggestions] = useState([]);
+  const searchOverridePre = Sefaria._('Search for') +': "';
+  const searchOverridePost = '"';
+  const getURLForObject = function(type, key) {
```
nsantacruz, line 667: This function should be pulled out of the component

static/js/Header.jsx
```diff
@@ -393,6 +401,346 @@ SearchBar.propTypes = {
...
+            showSearch={showSearch}
+            openTopic={openTopic}
+            openURL={openURL}
+            suggestions={suggestions}
+            hideHebrewKeyboard={false}
+            highlightedIndex={highlightedIndex}
+      />
+      <ul
+        {...getMenuProps()}
+        style={{ position: 'absolute', top: '100%', left: 0, zIndex: 999 }}>
```
nsantacruz, line 735: we use CSS, not react styles. Please refactor.

static/js/Autocomplete.jsx
```diff
@@ -0,0 +1,377 @@
...
+      } else if (d["type"] === "Person" || d["type"] === "Collection" || d["type"] === "TocCategory") {
+        return {'type': d["type"], 'id': d["key"]};
+      } else {
+        return {'type': "Search", 'id': query};
+      }
+    });
+};
+
+
+const SearchSuggestion = ({ itemKey, itemType, itemLabel, itemUrl, itemPic }) => {
```
nsantacruz, line 88: Remove `item` from the prefix of these props. It should be understood that they refer to an item based on the context.

nsantacruz: `itemKey` seems to be unused

static/js/Autocomplete.jsx
```diff
@@ -94,40 +120,55 @@ const SearchInputBox = ({getInputProps, suggestions, highlightedIndex,
...
-          _submitSearch(repairedCaseVariant);
-          return;
-        }
-        const repairedQuery = Sefaria.repairGershayimVariant(query, d);
-        if (repairedQuery !== query) {
-          _submitSearch(repairedQuery);
-          return;
-        }
+      getQueryObj(query).then(queryObj => {
+          console.log(queryObj)
```
nsantacruz, line 123: Remove console log

static/js/Autocomplete.jsx
```diff
@@ -94,40 +120,55 @@ const SearchInputBox = ({getInputProps, suggestions, highlightedIndex,
...
-        if (repairedCaseVariant !== query) {
-          _submitSearch(repairedCaseVariant);
-          return;
-        }
-        const repairedQuery = Sefaria.repairGershayimVariant(query, d);
-        if (repairedQuery !== query) {
-          _submitSearch(repairedQuery);
-          return;
-        }
+      getQueryObj(query).then(queryObj => {
```
nsantacruz, line 122: you can deconstruct the queryObj here instead of on line 125 and 126 like this `{type: queryType, id: queryId}`

static/js/Autocomplete.jsx
```diff
@@ -94,40 +120,55 @@ const SearchInputBox = ({getInputProps, suggestions, highlightedIndex,
...
-        } else {
-          Sefaria.track.event("Search", "Search Box Search", query);
-          _showSearch(query);
-        }
-      });
-  }
+
+      }
+      )
+    //   if (d["is_ref"]) {
```
nsantacruz: Remove commented code

static/js/Autocomplete.jsx
```diff
@@ -13,7 +13,18 @@ const _type_icon_map = {
   "ref": "iconmonstr-book-15.svg",
   "search": "iconmonstr-magnifier-2.svg",
   "Term": "iconmonstr-script-2.svg",
-  "User": "iconmonstr-user-2%20%281%29.svg"
+  "User": "profile.svg"
+};
+const type_title_map = {
```
nsantacruz, line 18: This map should be unnecessary. It can just contain a mapping from type to the English. The Hebrew can be translated via `<InterfaceText>` using the following syntax `<IntefaceText>{en}</InterfaceText>`. This will pull out the correct Hebrew from `strings.js`. Confirm that the correct translation for each of these categories exists in `strings.js`.

static/js/Autocomplete.jsx
```diff
@@ -85,90 +133,84 @@ const getQueryObj = (query) => {
 };
 
 
-const SearchSuggestion = ({ type, label, url, pic }) => {
-
+const SearchSuggestion = ({ value, type, label, url, pic,
```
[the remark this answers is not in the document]
nsantacruz: Let's meet to think of ways to make this code better. It's currently a bit spaghetti-ish.

static/js/Autocomplete.jsx
```diff
@@ -244,7 +288,8 @@ const SearchInputBox = ({getInputProps, suggestions, highlightedIndex,
       </div>
     );
   };
-const SuggestionsDispatcher = ({ suggestions, getItemProps, highlightedIndex}) => {
+const SuggestionsDispatcher = ({ suggestions, getItemProps, highlightedIndex,
+                                            _submitSearch, _redirectToObject}) => {
```
nsantacruz, line 293: what does it mean to have an underscore here? normally that means private but it's not private if it's a prop.

static/js/Autocomplete.jsx
```diff
@@ -271,41 +319,65 @@ const SuggestionsDispatcher = ({ suggestions, getItemProps, highlightedIndex}) =
...
+    const type = suggestions[0].type;
+    const enTitle = type_title_map[type]?.en;
+    const heTitle = type_title_map[type]?.he;
 
     return (
-        <>
+        <div className={"search-group-suggestions"}>
+
+            {(type != 'search') &&
+            <div className={'type-title'}><InterfaceText text={{en:enTitle, he:heTitle}} /></div>
```
nsantacruz: It looks like you can pass `type_title_map[type]` directly to the `text` prop

static/js/Autocomplete.jsx
```diff
@@ -0,0 +1,512 @@
...
+          else {
+              Sefaria.track.event("Search", "Search Box Search", queryId);
+              _showSearch(queryId);
+              _clearSearchBox();
+          }
+      }
+      )
+    };
+
+  const _showSearch = (query) => {
```
nsantacruz, line 454: rename to showSearchWrapper

static/css/s2.css
```diff
@@ -4522,6 +4522,172 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
...
+}
+.singlePanel .autocomplete-dropdown {
+  width: 100vw;
+  position: fixed;
+  top: 120px;
+  inset-inline-start: 0;
+  padding-bottom: 10px;
+}
+.search-group-suggestions{
+  border-bottom-color: rgb(204, 204, 204);
```
nsantacruz, line 4546: Our style is to combine all `border-bottom` styles into one style. Use GPT to help you. This applies to `padding-left`, `padding-right` etc. which can all be combined into `padding` and `border-top-x` which appears below.

static/js/Header.jsx
```diff
@@ -459,14 +214,14 @@ const MobileNavMenu = ({onRefClick, showSearch, openTopic, openURL, close, visib
...
-          onNavigate={close}
-          fullWidth={true}
-          hideHebrewKeyboard={true} />
+        <Autocomplete
+            onRefClick={onRefClick}
+            showSearch={showSearch}
+            openTopic={openTopic}
+            openURL={openURL}
+            onNavigate={close}
+            hideHebrewKeyboard={true}
```
nsantacruz, line 223: Remind me why this can't be done with CSS-only?

static/js/Header.jsx
```diff
@@ -132,266 +136,17 @@ Header.propTypes = {
...
-          const $widget = $("ul.ui-autocomplete");
-          $(".readerApp > .header").append($widget);
-      },
-      select: ( event, ui ) => {
-        event.preventDefault();
-
-        if (this.catchSearchOverride(ui.item.label)) {
-          return false;
-        }
+// SearchBar.propTypes = {
```
nsantacruz, line 140: Remove commented out code

static/js/Autocomplete.jsx
```diff
@@ -0,0 +1,489 @@
...
+    seggestedItems.forEach(item => {
+        if (!groupedItems[item.type]) {
+            groupedItems[item.type] = [];
+        }
+        groupedItems[item.type].push(item);
+    });
+
+    //Convert into a datastructure like this: [{"type": name,
+    //                                         "items" : [item1, item2]}]
+    const result = Object.keys(groupedItems).map(type => ({
```
nsantacruz, line 52: Simply return the result here without creating a variable

static/js/Autocomplete.jsx
```diff
@@ -0,0 +1,489 @@
...
+  "TocCategory": "Categories",
+  "PersonTopic": "Topics",
+  "Topic": "Topics",
+  "ref": "Books",
+  "search": "",
+  "Term": "Terms",
+  "User": "Users"
+};
+
+const type_icon = function(itemType, itemPic) {
```
nsantacruz, line 31: Use the syntax `function type_icon(itemType, itemPic)` which is used directly below.

static/js/Autocomplete.jsx
```diff
@@ -0,0 +1,489 @@
...
+const SuggestionsDispatcher = ({ suggestions, getItemProps, highlightedIndex,
+                                            submitSearch, redirectToObject}) => {
+
+    let groupedSuggestions = groupByType(suggestions);
+    let universalIndex = 0;
+
+    return (
+        <>
+            {groupedSuggestions.map((object, index) => {
+                const InitialIndexForGroup = universalIndex;
```
nsantacruz, line 271: make this variable lower case. In React, upper case variables are meant for components generally

nsantacruz (approved):

PR #1847, opened 2024-03-19
Bug/sc 25982/index editor doesn t work with jerusalem

stevekaplan123:
The Index Editor did not work with any Jerusalem Talmud text as well as some other texts because of two things: 1) the Index Editor never passed a 'schema' value to the API and 2) we have very old code (mostly from 2014) that, if no schema is passed, makes up a schema based on reasonable defaults.  However, those reasonable defaults do not work in all cases.  For example, it assumes Talmud texts' addressTypes are ["Daf", "Integer"] which they are not in the case of Jerusalem Talmud texts.  My code makes two changes.  First it removes in sefaria/model/text.py the 10 year old code because it is unnecessary.  When Content Engineers work, they provide a 'schema' in the post_index function.    Why was this 10 year old code there in the first place?  The comment there explains:            
           # Data is being loaded from dict in old format, rewrite to new format
            # Assumption is that d has a complete title collection
But we no longer use this old format where we have indices without schemas so I removed the code.

I also made changes to the POST Index API:
The core change is here -- If 'schema' is not passed to the POST Index API, check whether the book already exists.  If the book doesn't exist, return an error.  If the book already exists, just use the schema that it already has.  Therefore, there's no reason to create a new schema as either a content engineer provides one or it already exists.
```
if 'schema' not in j and book:
        # if book already exists and no schema provided, use old schema
        j['schema'] = book.contents()['schema']
```

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -34,7 +34,7 @@ from bson.objectid import ObjectId
 
 from sefaria.model import *
 from sefaria.google_storage_manager import GoogleStorageManager
-from sefaria.model.user_profile import UserProfile, user_link, user_started_text, public_user_data, UserWrapper
+from sefaria.model.user_profile import UserProfile, user_link, public_user_data, UserWrapper
 from sefaria.model.collection import CollectionSet
 from sefaria.model.webpage import get_webpages_for_ref
 from sefaria.model.media import get_media_for_ref
@@ -1253,7 +1253,7 @@ def edit_text_info(request, title=None, new_title=None):
         # Edit Existing
         title = title.replace("_", " ")
         i = library.get_index(title)
-        if not (request.user.is_staff or user_started_text(request.user.id, title)):
+        if not request.user.is_staff:
             return render_template(request,'static/generic.html', None, {
                 "title": "Permission Denied",
                 "content": "The Text Info for %s is locked.<br><br>Please email hello@sefaria.org if you believe edits are needed." % title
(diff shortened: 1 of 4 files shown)


reader/views.py
```diff
@@ -1678,15 +1687,9 @@ def index_api(request, title, raw=False):
...
-        else:
-            title = j.get("oldTitle", j.get("title"))
-            try:
-                library.get_index(title)  # getting the index just to tell if it exists
-                # Only allow staff and the person who submitted a text to edit
-                if not request.user.is_staff and not user_started_text(request.user.id, title):
-                   return jsonResponse({"error": "{} is protected from change.<br/><br/>See a mistake?<br/>Email hello@sefaria.org.".format(title)})
-            except BookNameError:
-                pass  # if this is a new text, allow any logged in user to submit
+        elif book and not request.user.is_staff and not user_started_text(request.user.id, title):
```
akiva10b, line 1690: @nsantacruz @edamboritz do you guys know the history of `user_started_text` based on Steves understanding and myself this check seems unnecessary.

reader/views.py
```diff
@@ -1678,15 +1687,9 @@ def index_api(request, title, raw=False):
...
+            # Only allow staff and the person who submitted a text to edit
+            return jsonResponse({"error": "{} is protected from change.<br/><br/>See a mistake?<br/>Email hello@sefaria.org.".format(title)})
         @csrf_protect
```
akiva10b, line 1693: this seems like weak protection, as any user signed in can hit this api, this should be verified @stevekaplan123

reader/views.py
```diff
@@ -1678,15 +1687,9 @@ def index_api(request, title, raw=False):
             if not apikey:
```
akiva10b, line 1687: Seems like there should be a decorator @api_protect that covers this.

reader/views.py
```diff
@@ -1658,42 +1658,53 @@ def index_api(request, title, raw=False):
...
 
-        func = tracker.update if request.GET.get("update", False) else tracker.add
-        j = json.loads(request.POST.get("json"))
-        if not j:
-            return jsonResponse({"error": "Missing 'json' parameter in post data."})
-        j["title"] = title.replace("_", " ")
+        USER_TYPE = None  # indicates whether a Content Engineer or Admin is modifying the Index
+        CONTENT_TYPE = "CONTENT"
+        ADMIN_TYPE = "ADMIN"
+        UID = None
```
akiva10b, line 1667: only constants should be upper case, variables should be lower case

reader/views.py
```diff
@@ -1658,42 +1658,53 @@ def index_api(request, title, raw=False):
...
+            return jsonResponse({"error": "Missing 'json' parameter in post data."})
+
+        if USER_TYPE == CONTENT_TYPE:
+            return index_post(request, UID, "API")
+        elif USER_TYPE == ADMIN_TYPE:
+            if 'schema' not in j:
+                # if book already exists and no schema provided, use old schema
+                j["title"] = title.replace("_", " ")
+                title = j.get("oldTitle", j.get("title"))
+                try:
```
akiva10b, line 1701: throw error if book does not exists

PR #1852, opened 2024-03-20
Fillin

YishaiGlasner:


api/tests.py
```diff
@@ -179,6 +179,12 @@ def test_api_get_text_no_version(self):
         self.assertEqual(data['warnings'][0]['hebrew|Kishkoosh']['message'],
                          'We do not have version named Kishkoosh with language hebrew for The Book of Maccabees I 1')
 
+    def test_api_get_text_no_version_fill_in(self):
```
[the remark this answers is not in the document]
akiva10b: are we also testing when its merged?

api/views.py
```diff
@@ -54,5 +53,5 @@ def get(self, request, *args, **kwargs):
             return jsonResponse({'error': f'return_format should be one of those formats: {self.RETURN_FORMATS}.'}, status=400)
         text_manager = TextRequestAdapter(self.oref, versions_params, fill_in_missing_segments, return_format)
         data = text_manager.get_versions_for_query()
-        data = self._handle_warnings(data)
+        self._handle_warnings(data)
```
akiva10b, line 56: @edamboritz what is your general opinion here? I personally prefer manipulating class bound params, and returning scoped params. But that both are logical, we should have a consistent pattern. @YishaiGlasner can you explain why you went with manipulating instead of returning `data`?

YishaiGlasner: the data is changed. i see no reason to return and reasign it.

edamboritz: I think the previous way of returning the data here was better as well?

YishaiGlasner: The standard in python is modify or return, but not both. All the methods on mutables returns none.
If you see return you may expect it not to modify the object.

edamboritz: But thats not really relevant to this case. You are manipulating `data` which is not a property of the class being discussed...

akiva10b: While it is true that technically "data" itself is being manipulated, so technically
data = self._handle_warnings(data)
is like saying
data = data
The paradigm is a bit odd. You have `data` which is a free variable in a `class method` being manipulated by a another `class method`. Returning the data makes it look a little clearer. That being said, now that I look at what is happening in `_handle_warnings` I see that this so more suitable for a `dataclass` method on data itself, or treating `data` as a class variable.

A classic example for something like this would be "sort", when you run `data.sort()` or `data.extend()` (on a list) you don't need to run `data = data.sort()` because the method is bound to the `dataclass` - it seems that `data` itself can have a method `data.handle_warnings()` which would be bound do a data class that would work nicely

edamboritz: While in principal I agree, it seems a bit overkill for this particular case... no? 
Implementing a data class should be done with intention and not just so we can shift a function around?

YishaiGlasner: I too think it's overkill, but i think returning is misleading, because it's cause you to think that this function does not modifying (like sort - it sin't that you don't need to run data = data.sort(), but that if you would do that you will have None in your data).
But I agree that the function name should reflect that. Something like _add_warnings (or _add_warnings_to_data, but the shorter seems to me to be clear enough).

akiva10b: I can live with _add_warnings - clear enough for me

PR #1858, opened 2024-03-28
docs: Correct list of values for api/v3/texts return_format

edamboritz:
This PR is a suggestion on a fix to supply the v3 Texts API parameter: return_fromat with the correct list of possible values. 

It adds an enum to the schema of that param.

### DIFF
--- a/docs/openAPI.json
+++ b/docs/openAPI.json
@@ -992,16 +992,16 @@
                         }
                     },
                     "name": "return_format",
-                    "description": "This query parameter formats the text that will be returned",
+                    "description": "This parameter formats the text that will be returned from the `v3` `texts/` API. It can have one of four values:\n1. `text_only` - This strips footnotes, inline references and all html tags from the returned text\n2. `strip_only_footnotes` - This only strips the footnotes and commentator <i> tags without stripping the other HTML tags.This is useful for the native app, where we do not display footnotes.\n3. `wrap_all_entities` - This wraps the HTML for links and topic links\n4. `default` - This returns the basic text as it’s saved in Sefaria’s DB\n",
                     "schema": {
-                        "description": "This parameter formats the text that will be returned from the `v3` `texts/` API. It can have one of four values:\n1. `text_only` - This strips footnotes, inline references and all html tags from the returned text\n2. `strip_only_footnotes` - This only strips the footnotes and commentator <i> tags without stripping the other HTML tags.This is useful for the native app, where we do not display footnotes.\n3. `wrap_all_entities` - This wraps the HTML for links and topic links\n4. `default` - This returns the basic text as it’s saved in Sefaria’s DB\n",
-                        "type": "string",
                         "enum": [
-                            'default', 
-                            'wrap_all_entities', 
-                            'strip_only_footnotes', 
-                            'text_only'
-                        ]
+                            "default", 
+                            "wrap_all_entities", 
+                            "strip_only_footnotes", 
+                            "text_only"
+                        ],
+                        "type": "string",
+                        "default": "default"
                     },
                     "in": "query"
                 }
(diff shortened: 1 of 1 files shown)


docs/openAPI.json
```diff
@@ -9007,10 +9014,6 @@
                     }
                 }
             },
-            "return_format": {
```
saengel, line 9010: I would recommend keeping this in the description anyways, might be helpful to have an elaboration of the possible values

docs/openAPI.json
```diff
@@ -994,7 +994,14 @@
                     "name": "return_format",
                     "description": "This query parameter formats the text that will be returned",
                     "schema": {
-                        "$ref": "#/components/schemas/return_format"
+                        "description": "This parameter formats the text that will be returned from the `v3` `texts/` API. It can have one of four values:\n1. `text_only` - This strips footnotes, inline references and all html tags from the returned text\n2. `strip_only_footnotes` - This only strips the footnotes and commentator <i> tags without stripping the other HTML tags.This is useful for the native app, where we do not display footnotes.\n3. `wrap_all_entities` - This wraps the HTML for links and topic links\n4. `default` - This returns the basic text as it’s saved in Sefaria’s DB\n",
```
saengel, line 997: Fantastic, thank you @edamboritz

docs/openAPI.json
```diff
@@ -994,7 +994,14 @@
...
+                        "type": "string",
```
saengel, line 998: In order for this to render as expected (i.e. a dropdown for all enum values) we need to format this as follows:

```{
                    "examples": {
                        "Text Only": {
                            "value": "text_only"
                        },
                        "Wrap All Entities": {
                            "value": "wrap_all_entities"
                        }
                    },
                    "name": "return_format",
                    "description": "This parameter formats the text that will be returned from the `v3` `texts/` API. It can have one of four values:\n1. `text_only` - This strips footnotes, inline references and all html tags from the returned text\n2. `strip_only_footnotes` - This only strips the footnotes and commentator <i> tags without stripping the other HTML tags.This is useful for the native app, where we do not display footnotes.\n3. `wrap_all_entities` - This wraps the HTML for links and topic links\n4. `default` - This returns the basic text as it’s saved in Sefaria’s DB\n",
                    "schema": {
                        "enum": [
                            "default", 
                            "wrap_all_entities", 
                            "strip_only_footnotes", 
                            "text_only"
                        ],
                        "type": "string"
                    },
                    "in": "query"
                }
```


The main differences are:
1. The description outside the schema (noted in an early comment as well, apologies for the redundancy)
2. The `type` moved into the `enum` field but outside the array (i.e. to indicate that the overall type of this param is an `enum` of `string`s)

saengel (changes requested):
Thank you so much @edamboritz - really appreciate you taking care of this. 

It's almost there, if you see the comments you'll see a few changes we need to make in order to get this to render in the most helpful way possible for our users. Feel free to let me know if you have any questions!

docs/openAPI.json
edamboritz: Ok, 
1. Description was indeed in the wrong place by accident, but that means we have two description fields where one (coming from the defunct reusable ref) is much more detailed than the other. 
2. I'm not sure I understand, I followed https://swagger.io/docs/specification/data-models/enums/ and it clearly shows the type inside the schema as a sibling to the enum itself?

saengel: 1. Got it!
2. Hmmm, when I read this I see the `type` field nested inside the `enum` field, not adjacent. Maybe I'm misreading the yaml? Happy to go over together if that's better so we can get clarity. Thanks for following the spec so closely. 

Thanks!

saengel (approved):

PR #1864, opened 2024-04-04
Add reference to 2023 annual report & change URL path to be dynamic

relyks:
## Context
Every year, a new annual impact report is created and links need to be changed in more than one place when the report is ready for the Sefarian world abroad. 

## Solution
To handle this scenario and minimize the amount of code changes, the path for the annual report will now be `/annualreport` in links instead of `/annualreport/<year>/`. The default path will now refer to the most recent report. However, the ability to refer to specific years by the path will still be allowed and to ensure backwards compatibility with previously given out links as was previously done.

## Changes
* Footer link to annual report is now `/annualreport`
* Link on _Ways to Give_ page is now `/annualreport`
* The endpoint function for that path now handles both URL patterns `/annualreport` and  `/annualreport/<year>/`
* The function determines which is the most recent year to use for the default case
* An entry for the 2023 annual report was added to an Issuu reader

### DIFF
--- a/sites/sefaria/urls.py
+++ b/sites/sefaria/urls.py
@@ -102,5 +102,5 @@ site_urlpatterns +=[
     url(r'^ideasforteaching/?$',lambda x: HttpResponseRedirect(STATIC_URL + 'files/Sefaria_Teacher_Generated_Ideas_for_Your_Classroom.pdf')),
     url(r'^strategicplan/?$',lambda x: HttpResponseRedirect(STATIC_URL + 'files/Sefaria_Strategic_Plan.pdf')),
     url(r'^annualreport2021?$', lambda x: HttpResponseRedirect('/annualreport/2021')), # Added for backwards compatability for old links that might still point to this
-    url(r'^annualreport/(?P<report_year>\d+)$', reader_views.annual_report),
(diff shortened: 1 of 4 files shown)


sites/sefaria/urls.py
```diff
@@ -102,5 +102,5 @@
     url(r'^ideasforteaching/?$',lambda x: HttpResponseRedirect(STATIC_URL + 'files/Sefaria_Teacher_Generated_Ideas_for_Your_Classroom.pdf')),
     url(r'^strategicplan/?$',lambda x: HttpResponseRedirect(STATIC_URL + 'files/Sefaria_Strategic_Plan.pdf')),
     url(r'^annualreport2021?$', lambda x: HttpResponseRedirect('/annualreport/2021')), # Added for backwards compatability for old links that might still point to this
-    url(r'^annualreport/(?P<report_year>\d+)$', reader_views.annual_report),
+    url(r'^annualreport(/(?P<report_year>\d+)/?|/?)$', reader_views.annual_report),
```
relyks, line 105: I think there might be a more simplified way of doing this, but I couldn't figure it out. It works for `/annualreport`, `/annualreport/`, `/annualreport/2023/`, and `/annualreport/2023/`

edamboritz (approved):

PR #1866, opened 2024-04-06
Sacks books layout

yonadavGit:


sefaria/model/topic.py
```diff
@@ -580,6 +596,11 @@ def index_is_commentary(index):
                 for temp_index in temp_indexes:
                     index_or_cat_list += [(temp_index, None, None)]
                 continue
+            if not collective_title and is_prefix(best_base_cat_path, all_best_paths):
```
nsantacruz, line 599: move this if statement into previous if statement

nsantacruz: also move this if statement
```python
            if len(temp_indexes) == 1:
                index_or_cat_list += [(temp_indexes[0], None, None)]
                continue
```

nsantacruz: also break up the huge if statement say each disjunctive (OR) is on its own line

nsantacruz (approved):

PR #1877, opened 2024-04-11
Temp give reindex admin access

nsantacruz:


build/ci/production-values.yaml
```diff
@@ -165,7 +165,10 @@ nginx:
   containerImage:
     imageRegistry:
     tag:
-  SEARCH_HOST: elasticsearch-es-http.elasticsearch.svc
+  SEARCH_HOST: "searchfix.cauldron.sefaria.org"
+  SEARCH_PORT: "443"
+  SEARCH_PATH: "api/search"
+  SEARCH_SSL_ENABLE: true
   disableScraping: false
```
BrendanGalloway, line 172: future, breaking:

```
elasticsearch:
  host:
  port:
  path:
  ssl-enable:
```

build/ci/production-values.yaml
```diff
@@ -194,7 +197,7 @@ cronJobs:
   dailyEmailNotifications:
     enabled: true
   indexFromQueue:
-    enabled: true
+    enabled: false
```
BrendanGalloway, line 200: This needs to be turned back on?

sefaria/helper/llm/topic_prompt.py
```diff
@@ -86,7 +86,7 @@ def _make_llm_topic(sefaria_topic: Topic) -> LLMTopic:
     )
 
 
-def _make_topic_prompt_source(oref: Ref, context: str) -> TopicPromptSource:
```
BrendanGalloway, line 89: Should this be part of this change?

BrendanGalloway (changes requested):

build/ci/production-values.yaml
edamboritz: After the main index job completes, yes

sefaria/helper/llm/topic_prompt.py
nsantacruz: you're right. I'll remove.

build/ci/production-values.yaml
BrendanGalloway: ah, thought this was the job that has the temporary override

PR #1889, opened 2024-05-06
Bug/sc 25982/index editor doesn t work with jerusalem

stevekaplan123:
This PR does 2 things:
1) General refactor of the Index API:  Removed the unnecesssary function `user_started_text` and generally re-organized the Index API into several nested functions.

2) Fixed the Index Editor: Jerusalem Talmud indices, some Bavli indices (i.e. Avot DeRabbi Natan), and Mishnah commentaries did not work well with the Index Editor because using the Index Editor with these texts changed their `addressTypes` and sometimes added `checkFirst` to the Index.  Several changes prevent that from happening.  If the Index is not new, `addressTypes` shouldn't change at all now, and since we have the entire Bavli, we have no reason to add any more texts with `checkFirst` so those lines were removed.


reader/views.py
```diff
@@ -1647,73 +1647,66 @@ def text_titles_api(request):
...
-        def protected_index_post(request):
-            return jsonResponse(
-                func(request.user.id, Index, j, raw=raw).contents(raw=raw)
-            )
+            key = request.POST.get("apikey")
+            if key:
+                apikey = db.apikeys.find_one({"key": key})
+                if apikey:
+                    return CONTENT_TYPE, apikey["uid"]
+            return jsonResponse({"error": "Authentication failed. Must be staff or provide a valid API key."})
```
akiva10b, line 1688: you will get "not enough vars to unpack" if return error - test return error

akiva10b (approved):

PR #1897, opened 2024-05-19
Feature/sc 25577/tool request edit collective title

stevekaplan123:
This PR adds the ability to add three new properties to indices via the Index Editor: dependence, collective_title, and base_text_titles.  This new feature can be seen on this cauldron: https://index-editor2.cauldron.sefaria.org/texts

'Dependence' is set via a dropdown menu.  If the text is a dependent text, then it's possible to set the collective title and base text titles, but if the text is not set as a dependent text, it is not possible to set the collective title or base text titles.  In case we are dealing with a dependent text, it is possible to create a new collective title, which is the same thing as creating a new term.  When the admin saves the book, the front-end validates the collective title through the function validateCollectiveTitle.  This function behaves differently depending on the admin's choice to create a new collective title or not.  If the admin chose to create a new collective title, this function posts to the Terms API a new term; if the admin chose not to create a new collective title, then validateCollectiveTitle calls a GET request to the Terms API to confirm it already exists.  'Base text titles' are also validated on the front end by using Sefaria.index to check that each entry is actually a book.
On the backend, there is no extra validation but there is normalization.  On Index save, _normalize() removes the fields 'dependence', 'collective_title', and/or 'base_text_titles' from the index object if these are set as empty strings by the Index Editor.  This is necessary because empty string values for these fields are the only way of getting rid of these fields from an Index object.


edamboritz (approved):
I think this PR suffers from the same unfortunate readability and component complexity that most of this tools suite suffers from, but correcting that is out of scope. 

For the task at hand, it's good.

PR #1899, opened 2024-05-21
Make the hey replacement have a hebrew geresh instead of a single quote

wolfkevin:
This should help make the english hey replacement better so that it reads properly

Screenshot 2024-05-21 at 21 31 49

Screenshot 2024-05-21 at 21 31 39


akiva10b (approved):
@wolfkevin Thank you for fixing this!

PR #1900, opened 2024-05-21
Guides

EliezerIsrael:


static/js/GuideBox.jsx
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Renders the summary prompts in a box.
+ * @param {Object} props.prompt - The summary prompt object.
+ * @param {Function} props.onClick - The function to handle click events on prompts.
+ * @returns {JSX.Element} The SummaryBox component.
+ */
+const SummaryBox = ({ prompt, onClick }) => {
+  const title = (commentaryRef) => {
+    const index = Sefaria.parseRef(commentaryRef)?.index;
+    return index ? index.replace(/ on Pesach Haggadah$/, "") : commentaryRef;
```
akiva10b, line 51: @EliezerIsrael this was done bespokely for the Haggadah project. It solved an issue with showing the names of the authors properly. This is clearly not a generic solution and no longer applicable, should we determine a new product definition for what we show here (for instance, show the name of the index no matter what) or have something more generic (like use something other than the index to get the correct data)

static/js/GuideBox.jsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  /**
+   * Overrides the setState method to add the current state to the history before updating it.
+   * @param {Object} newState - The new state object.
+   */
+  setState(newState) {
+    this.stateHistory.push(this.state);
+    super.setState(newState);
```
akiva10b, line 105: I override setState to keep a history of the state and send `popState` to the parent component so that it can control the level of the child component generically.

static/js/ConnectionsPanelHeader.jsx
```diff
@@ -49,7 +49,25 @@ class ConnectionsPanelHeader extends Component {
       /** TODO: fix for interfacetext */
     const previousMode = this.getPreviousMode();
     let title;
-    if (this.props.connectionsMode === "Resources") {
+    const backButtonSettings = this.props.backButtonSettings;
+    if (backButtonSettings) {
```
akiva10b, line 53: @edamboritz I have added a new way for a child component to articulate the back button in the sidebar. This is a generic back button which accepts `backButtonSettings` and based on them will display text in the sidebar and control what happens in the child component. In this case I am using `popState` from the child component described later on and send that to the back button in onClick. Would love comments on this

static/js/GuideBox.jsx
EliezerIsrael: Fixed, and added a :"todo" for generic replacement logic

static/css/s2.css
```diff
@@ -6914,6 +6938,17 @@ But not to use a display block directive that might break continuous mode for ot
   align-self: flex-start;
 }
 
+.toolsButtonContainer.blue .experimentLabel {
+  --english-font: var(--english-sans-serif-font-family);
+  color: #ccc;
+  font-size: 14px;
+  border: 1px solid #CCCCCC;
```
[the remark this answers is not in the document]
EliezerIsrael: Updated

static/js/ConnectionsPanel.jsx
```diff
@@ -1322,7 +1336,7 @@ AdvancedToolsList.propTypes = {
 
 const ToolsButton = ({ en, he, onClick, urlConnectionsMode = null, icon, image,
                        count = null, control = "interface", typeface = "system", alwaysShow = false,
-                       secondaryHe, secondaryEn, greyColor=false }) => {
+                       secondaryHe, secondaryEn, greyColor=false, blueBackground=false, experiment=false }) => {
```
akiva10b, line 1339: Will require refactor

akiva10b (approved):

PR #1903, opened 2024-05-27
Remove mq collisions

BrendanGalloway:


.github/workflows/continuous.yaml
```diff
@@ -30,26 +30,20 @@ jobs:
...
+        uses: docker/setup-buildx-action@v3
       - id: auth
         name: Authenticate to Google Cloud
-        uses: google-github-actions/auth@v1
+        uses: google-github-actions/auth@v2
         with:
           token_format: 'access_token'
           workload_identity_provider: 'projects/${{ secrets.DEV_GKE_PROJECT_ID}}/locations/global/workloadIdentityPools/github/providers/github'
           service_account: '${{ secrets.DEV_GKE_SA }}'
-      - name: Login to GCR
```
edamboritz, line 45: I take it this step is no longer needed (and in other removed places)?

.github/workflows/helm.yaml
```diff
@@ -43,8 +43,6 @@ jobs:
         with:
           # Fetch entire history. Required for chart-releaser; see https://github.com/helm/chart-releaser-action/issues/13#issuecomment-602063896
           fetch-depth: 0
-      - name: fetch head of branch to ensure semantic can run
```
edamboritz, line 46: Is this step also not needed

.github/workflows/continuous.yaml
BrendanGalloway: yes, it's a login to the old container registry which we've replaced with the artifact registry - we removed the push to this registry some time ago, I see the login was not taken out at the same time

.github/workflows/helm.yaml
BrendanGalloway: hmm, this should not have been part of this change, will revert

edamboritz (approved):

PR #1912, opened 2024-06-02
Add tooltip to reader options icon

yonadavGit:
At the top of the reader is a toolbar including an A/Aleph icon which opens the "Reader Options" control. The task here is to add a fast-acting tooltip which is triggered by hover on the icon.
We have this functionality today right next to the Reader Options icon - the bookmark icon has a fast-acting tooltip and we can simply re-use whatever code powers that.. changing the message of course.
Tooltip EN - Text display options
Tooltip HE - אפשרויות תצוגת טקסט

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -1350,7 +1350,7 @@ class DisplaySettingsButton extends Component {
     let style = this.props.placeholder ? {visibility: "hidden"} : {};
     let icon;
     const altText = Sefaria._('Text display options')
-    const classes = classNames({readerOptionsTooltip: 1, "tooltip-toggle": true});
+    const classes = "readerOptionsTooltip tooltip-toggle";
 
     if (Sefaria._siteSettings.TORAH_SPECIFIC) {
       icon =
@@ -1374,9 +1374,8 @@ class DisplaySettingsButton extends Component {
                 onKeyPress={function(e) {e.charCode == 13 ? this.props.onClick(e):null}.bind(this)}>
                 {icon}
               </a>
-          </ToolTipped>
-
-              );
+            </ToolTipped>
+            );
   }
 }
 DisplaySettingsButton.propTypes = {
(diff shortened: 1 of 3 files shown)


static/js/Misc.jsx
```diff
@@ -1349,6 +1349,8 @@ class DisplaySettingsButton extends Component {
   render() {
     let style = this.props.placeholder ? {visibility: "hidden"} : {};
     let icon;
+    const altText = Sefaria._('Text display options')
+    const classes = classNames({readerOptionsTooltip: 1, "tooltip-toggle": true});
```
nsantacruz, line 1353: There's no reason to use `classNames` if both class names are always true. Just make it a static string.

static/js/Misc.jsx
```diff
@@ -1359,17 +1361,22 @@ class DisplaySettingsButton extends Component {
...
-              role="button"
-              aria-haspopup="true"
-              aria-label="Toggle Reader Menu Display Settings"
-              style={style}
-              onClick={this.props.onClick}
-              onKeyPress={function(e) {e.charCode == 13 ? this.props.onClick(e):null}.bind(this)}>
-              {icon}
-            </a>);
+    return (
+            <ToolTipped {...{ altText, classes}}>
```
nsantacruz, line 1365: This is nice syntax!

static/js/Misc.jsx
```diff
@@ -1359,17 +1361,22 @@ class DisplaySettingsButton extends Component {
...
+                aria-haspopup="true"
+                aria-label="Toggle Reader Menu Display Settings"
+                style={style}
+                onClick={this.props.onClick}
+                onKeyPress={function(e) {e.charCode == 13 ? this.props.onClick(e):null}.bind(this)}>
+                {icon}
+              </a>
+          </ToolTipped>
+
+              );
```
nsantacruz, line 1379: Fix the indentation here

nsantacruz (approved):

PR #1921, opened 2024-06-10
fix(search bar): enable search with input from virtual keyboard

yonadavGit:
This PR fixes synchronicity between virtual keyboard and search bar

### DIFF
--- a/static/js/Autocomplete.jsx
+++ b/static/js/Autocomplete.jsx
@@ -174,7 +174,7 @@ const EntitySearchSuggestion = ({label, onClick, type, url, ...props}) => {
     );
 }
 
-const SearchInputBox = ({getInputProps, suggestions, highlightedIndex, hideHebrewKeyboard,
+const SearchInputBox = ({getInputProps, suggestions, highlightedIndex, hideHebrewKeyboard, setInputValue,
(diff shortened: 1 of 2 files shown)


static/js/Autocomplete.jsx
```diff
@@ -174,31 +174,37 @@ const EntitySearchSuggestion = ({label, onClick, type, url, ...props}) => {
     );
 }
 
-const SearchInputBox = ({getInputProps, suggestions, highlightedIndex, hideHebrewKeyboard,
+const SearchInputBox = ({getInputProps, suggestions, highlightedIndex, hideHebrewKeyboard, setInputValue,
                         setSearchFocused, searchFocused,
                             submitSearch, redirectToObject}) => {
 
+    const getInputValue = () =>{
+        return otherDownShiftProps.value || document.getElementsByClassName('keyboardInput')[0].value;
```
nsantacruz, line 182: this function should use `getVirtualKeyboardInputValue()`

nsantacruz: Also, this is very dangerous since you're getting the first instance of a CSS class. If this class is used by some other element this could break. Can you come up with a more specific CSS selector?

nsantacruz (approved):

PR #1925, opened 2024-06-13
Fix topic ordering issues

nsantacruz:
This fixes a few issues with inconsistent sorting found by the learning team:

- If a source appears multiple times on the page, editing will prefer the one with the highest curatedPrimacy
- Server-side and client-side relevance sorting are now (hopefully) identical
- Don't filter sources that don't have text in `interfaceLanguage` in the admin reorder view since this doesn't match how sources are sorted in the client-side code.


EliezerIsrael (changes requested):
@nsantacruz There's a good bit of opaque logic here, with conditionals based on data with unclear semantics.  Since you've had to suss out the logic, can I ask you to add some documentation to the code?

PR #1929, opened 2024-06-18
Only edit learning team links

nsantacruz:
To temporarily solve admin issues with topic page editing, we will only allow editing of learning-team links


edamboritz (approved):
This is probably creating more edge cases in solving the one its intended to solve, however...

edamboritz (approved):

PR #1936, opened 2024-06-24
Module switcher

saengel:
## Description
Module switcher first pass for modularization

## Code changes:
1. `s2.css` - Styling for the new `<DropdownMenu />` generic component, and all of its sub-components. Small shifts to existing code for `.interfaceLinks-options` to enable compatibility with the existing language menu dropdown and the new generic dropdown, mostly with using flex for text alignment & switching between languages, as opposed to setting `text-align`
2. `static/js/common/DropdownMenu.jsx` - A new file containing a generic dropdown menu (to be reused in the future) that's able to render children as a composable component. Heavily based on the existing code for `<InterfaceLanguageMenu />` in `Misc.jsx`. The subcomponents are re-usable as well. 
3. `static/js/Header.jsx` - Added the module switcher next to the language toggle, with the implementation configuration of the composable components in such a way to meet the specs of this specific project. 
4. Added the specific icons needed for the specific implementation of the module switcher. 

## Notes:
- Once sheets vs library internal state is confirmed, the logic around opening links will have to be slightly modified to enable conditional opening in a new tab.


edamboritz (commented):
This is great work

As discussed - 
1. Lets make the css for the generic component its own. As in separate it out from existing css and give distinct class names rather than ones that existed on the menu used for inspiration
2. Lets move out the code for the module switcher into its own component and use that inside the Header render
3. We can use above new component to set things things like width that should be flexible on the generic component

saengel:
Thank you for the helpful feedback @edamboritz - I just finished pushing changes which address all of the above. Let me know what you think now :)

edamboritz (approved):

PR #1940, opened 2024-06-27
Feature/sc 28125/backend for sheets with ref

stevekaplan123:
For /api/sheets/ref, passed parameter include_collections to `get_sheets_for_ref`, which in turn calls `get_collections_with_sheets`.  This function returns every public collection that has a sheet in `sheet_ids` so that `get_sheets_for_ref` returns sheets with a collections field.  Instead of including the whole collection, I include the name and slug only.  I also added a 'dateCreated' field since on the front end, in the "Sheets With" page, each sheet entry displays a date.


sefaria/sheets.py
nsantacruz, line 1: Tabs throughout this PR seem way too big

sefaria/sheets.py
```diff
@@ -874,6 +889,9 @@ def get_sheets_for_ref(tref, uid=None, in_collection=None):
 				"category":        "Sheets", # ditto
 				"type":            "sheet", # ditto
 			}
+			if include_collections:
+				sheet_data["collections"] = sheet_id_to_collection[sheet["id"]]
+				sheet_data["dateCreated"] = sheet["dateCreated"]
```
nsantacruz, line 894: why is dateCreated part of this if statement?

sourcesheets/views.py
```diff
@@ -1020,7 +1020,7 @@ def sheets_by_ref_api(request, ref):
     """
     API to get public sheets by ref.
     """
-    return jsonResponse(get_sheets_for_ref(ref))
+    return jsonResponse(get_sheets_for_ref(ref, include_collections=True))
```
nsantacruz, line 1023: I think this API should take a URL parameter that indicates whether it should include collections. For boolean parameters, our standard is to use 0 to mean false and 1 to mean true. Here's an example ```bool(int(request.GET.get("annotate", 0)))```

sefaria/sheets.py
stevekaplan123: I'm pretty sure github is formatting it strangely.  The tabs don't look weird on my screen

sefaria/sheets.py
stevekaplan123: I wasn't sure where to put it.  It's supposed to show up next to the user profile in SearchSheetResult now.  It's not necessary for the related API to have access to dateCreated, so I didn't add dateCreated to the big "sheet_data = {...}" but it is necessary for the sheets ref API to have access to dateCreated.  But it is definitely confusing to set dateCreated only when include_collections is True, so maybe dateCreated should just be set in the original "sheet_data = {...}".  What do you think?

PR #1942, opened 2024-06-27
Feature/sc 28260/re factor sheetsearchresult

stevekaplan123:
Re-factored SearchSheetResult so that it receives props `metadata` and `snippet`.  Also modified the way it displays in line with the modularization design by re-arranging the order of the components, removing the ColorBarBox, and adding a date.

### DIFF
--- a/static/js/SearchSheetResult.jsx
+++ b/static/js/SearchSheetResult.jsx
@@ -12,7 +12,7 @@ import {
 
 class SearchSheetResult extends Component {
     handleSheetClick(e) {
-      var s = this.props.data._source;
+      const s = this.props.metadata;
       if (this.props.onResultClick) {
         e.preventDefault()
         this.props.onResultClick("Sheet " + s.sheetId);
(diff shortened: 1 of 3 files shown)


static/js/SearchSheetResult.jsx
```diff
@@ -12,72 +12,77 @@ import {
...
-    get_snippet_markup(data) {
-      let snippet = data.highlight.content.join("..."); // data.highlight ? data.highlight.content.join("...") : s.content;
-      snippet = snippet.replace(/^[ .,;:!-)\]]+/, "");
+    get_snippet_markup() {
+      const snippet = this.props.snippet.replace(/^[ .,;:!-)\]]+/, "");
       const lang = Sefaria.hebrew.isHebrew(snippet) ? "he" : "en";
       return { markup: {__html: snippet}, lang };
     }
+    formatDate(dateString) {
+        const date = new Date(dateString);
```
nsantacruz, line 32: Is this code from GPT? I'd never seen this class before but it looks very helpful. I would remove the comments since they basically just say what is already obvious from the code

static/js/SearchSheetResult.jsx
```diff
@@ -12,72 +12,77 @@ import {
...
         var clean_title = $("<span>" + s.title + "</span>").text();
         var href = "/sheets/" + s.sheetId;
-        const snippetMarkup = this.get_snippet_markup(data);
-        const snippetClasses = classNames({snippet: 1, en: snippetMarkup.lang == "en", he: snippetMarkup.lang == "he"});
+        const snippetMarkup = this.get_snippet_markup();
+        const snippetClasses = classNames({snippet: 1, en: snippetMarkup.lang === "en", he: snippetMarkup.lang === "he"});
         const ownerIsHe = Sefaria.hebrew.isHebrew(s.owner_name);
         const titleIsHe = Sefaria.hebrew.isHebrew(clean_title);
-        const tags = s.tags && s.tags.length ? Sefaria.util.zip(s.tags, s.topic_slugs, s.topics_he) : [];
+        const date = this.formatDate(this.props.metadata.dateCreated);
```
nsantacruz, line 46: I would change this variable name to something like dateString or formattedDate.

nsantacruz (approved):

PR #1946, opened 2024-07-01
Sidebar Footer

saengel:
## Description
First pass at design and implementation for the new sidebar footer. The "stickiness" will be addressed in a separate PR. 

## File Changes
1. `s2.css` - Styling for the `footerContainer` flex container, and the links inside. 
2. `static/js/NavSidebar.jsx` - Adding a function to return the data, and a functional component to weave into the sidebar for the footer. 
3. `static/js/TextsPage.jsx` - Incorporating the footer module. 

## Notes:
- Which other pages do we want to include this on? 
- This PR does not include other forthcoming work relating to the footer


edamboritz (approved):

edamboritz (approved):
We need to have a mechanic whereby the footer is included automatically everywhere its needed. 
This means either including it in the bottom of the NavSidebar render or designing a new interim component that sits between NavSidebar and the modules that can be used for all the templates where there is a footer expected.

PR #1956, opened 2024-07-08
Footer retention on Scroll

saengel:
## Description
Add "stickiness" to new footer, as well as position the component to render every time the sidebar renders. 

## Code Changes
1. `static/css/s2.css` - Add the `stickySidebarFooter` class to trigger "stickiness" for the footer within the sidebar. 
2.  `static/js/NavSidebar.jsx` - General render for the footer, add wrapper div with `stickySidebarFooter` class. 

## Notes:
- This is a branch off [feature/sc-26817/design-and-implement-new-sidebar-footer](https://github.com/Sefaria/Sefaria-Project/tree/feature/sc-26817/design-and-implement-new-sidebar-footer), and should be merged into that, and then into modularization when ready.


static/js/TopicPage.jsx
```diff
@@ -607,6 +608,7 @@ const TopicPage = ({
                         multiPanel={multiPanel}
                         topicImage={topicImage}
                     />
+                    <SidebarFooter />
```
edamboritz, line 611: Why is this here explicitly if we are making the component generic?

static/js/NavSidebar.jsx
```diff
@@ -267,19 +271,21 @@ const getSidebarFooterData = () => [{'he': 'אודות','en': 'About', 'url': 'w
...
-      <div className="footerContainer">
-        {data.map(footerLink => 
-          <a href={footerLink.url}>
-            <InterfaceText text={{'en': footerLink.en, 'he': footerLink.he}}  />
-          </a>
-        )}
-      </div>
-    </Module>
+    <div className = "stickySidebarFooter">
+      <Module>
```
edamboritz, line 278: I dont think this should be wrapped as a <Module> anymore

static/js/NavSidebar.jsx
```diff
@@ -8,14 +8,18 @@ import { Promotions } from './Promotions'
...
-  return <div className="navSidebar sans-serif">
+  return (
+    <div className="navSidebar sans-serif">
     {modules.map((m, i) =>
       <Modules
         type={m.type}
         props={m.props || {}}
         key={i} />
     )}
+    <SidebarFooter />
```
edamboritz, line 19: Do we want to allow for the possibility that someone chooses to render the NavSidebar in a non default way without a footer?

static/css/s2.css
```diff
@@ -13161,6 +13161,11 @@ span.ref-link-color-3 {color: blue}
   padding: 3px 10px;
 }
 
+.stickySidebarFooter {
```
edamboritz, line 13164: Where are we sticking this to with top:0?

edamboritz (commented):
Great for adding tests! 

I'm not entirely sure I understand the various changes here and overall approach though. 
Its a good move to add the footer as a fixture of the NavSidebar, though we may want to allow some flexibility about it. 
But the goal here is to not have to render the footer explicitly in each page component so the fact that it is there ion the TopicsPage is somewhat confusing

static/js/NavSidebar.jsx
saengel: Can you clarify? It might be by mistake the tag you were referencing got deleted from the message. Thanks!

static/js/NavSidebar.jsx
saengel: Hmm possibly, what's the best way to do that? Control via props?

static/css/s2.css
saengel: It's how I managed to get the stickiness working... based on an article I can no longer find... unfortunately. Would you suggest another approach?

saengel: Oh, I'm sorry - I misread the `where` as a `why`. My bad! 

We are sticking it to the top of the sidebar nav when you scroll down, as per the specs.

edamboritz (approved):

PR #1968, opened 2024-07-16
Fix "flash" of "no jobs" on Jobs page

relyks:
## Description
The jobs page shows that there are "no jobs" while the data call to Strapi is being made. On slow internet connections, the job listings can take a short while to load, while the user may be given the impression that there are no jobs and leave without waiting. This PR fixes that to show that the page is in a loading state until it's received the result of the call. 

## Code Changes
* A new state variable was added to keep tracking of the loading state for the network call. Once the call is received, the status of whether there are job listings available becomes known.
* The status of whether there are jobs listing available was also refactored to variable

### DIFF
--- a/static/js/StaticPages.jsx
+++ b/static/js/StaticPages.jsx
@@ -3021,6 +3021,7 @@ const NoJobsNotice = () => {
 const JobsPage = memo(() => {
     const [groupedJobPostings, setGroupedJobPostings] = useState({});
     const [error, setError] = useState(null);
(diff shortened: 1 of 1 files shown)


edamboritz (commented):
We already have loading elements defined elsewhere that you can use. 
Otherwise looks good

relyks:
> We already have loading elements defined elsewhere that you can use.

Added them

edamboritz (approved):

PR #1971, opened 2024-07-17
Feature/sc 28105/re factor sheetsearchfilters to work for

stevekaplan123:
## Description
This PR implements the feature "SheetsWithRef" whereby clicking "Sheets" in the reader sidebar now opens a new tab with a view similar to SearchPage to display all sheets that are found with the ref from the reader.  There are other branches that addressed other issues.  For example, this PR doesn't get rid of the distinction totally between sheetSearchState and textSearchState -- another branch does that.  This PR doesn't have the sheetsWithRef URL params working -- another branch does that.  Finally, there is a search filtering bug I found during the execution of this project, and I fixed that in another branch as well.  In this PR:  

## Code Changes
Back-end --
1. get_sheets_for_ref -- It can now include in each sheet result, the public collections that have at least one of those sheets.  It also has a boolean value that allows the inclusion of the first comment of the sheet.  The reason for this is in case there is no sheet summary, Mickey asked to include the 150 characters of the first comment of the sheet.
2. sheetsWithRef route implemented to accept the possibilityt that the ref is a range and also to accept filters and sorting parameters

Front-end -- 
There are two major front-end code changes: 1) Re-factors SearchResultList so that ElasticSearchQuerier handles querying and SearchResultList handles rendering the data.  2) A new component SheetsWithRefPage acts in a parallel way to ElasticSearchQuerier by loading up the sheets data per given ref, and SearchResultList handles rendering the data.

SheetsWithRefPage works as follows:
1) in useEffect, it calls the sheets ref API and processes the results in handleSheetsLoad, creating FilterNodes from the results, and setting the searchState's availableFilters to those filterNodes.  It also sets origAvailableFilters to searchState.availableFilters, so that no matter what filters are used on the SheetsWithRefPage, the component still has access to the original available filters.
2) Every time the component is rendered, the following code is called, which will be explained step-by-step.
```
    updateOrigAvailableFilters();
    let sortedSheets = [...sheets];
    sortedSheets = applyFilters(sortedSheets);
    sortedSheets = applySortOption(sortedSheets);
    sortedSheets = normalizeSheetsMetaData(sortedSheets);
```
A. `updateOrigAvailableFilters()` -- This takes the prop SearchState and looks at the selected attribute of the availableFilters in the SearchState.  It updates origAvailableFilters 'select' data to keep it in sync.
B.  `applyFilters` --  This does two main things.  First of all, if there are any appliedFilters, it applies those filters to the sheets that are passed down to SearchResultList.  Second of all, it updates the topics/collections filters with the new docCounts now that the sheets have been filtered.  This change is applied to the SearchState prop, but origAvailableFilters is used (cloned actually) so that we have access to all availableFilters.  If the docCounts have now changed (because the appliedFilters are different than they were in the last render), it uses the function `checkForRegisteringAvailableFilters` to update the ReaderApp SearchState with the new availableFilters.
C. `applySortOption` -- This simply applies the current sort option -- 'views' or 'dateCreated' on the sheets.
D. `normalizeSheetsMetaData` -- This normalization function is necessary because ElasticSearchQuerier formats data differently than the sheets ref API and so normalization is necessary before passing the data down to SearchResultList.

Two other front-end changes:
1.  Remove old way of clicking on "Sheets" in the sidebar that showed MySheetsList and PublicSheetsList
2.  SearchSheetResult was re-factored so that it can be used the same way by search and sheetsWithRef functionality.  It also was re-designed based on the figma.


static/js/ElasticSearchQuerier.jsx
```diff
@@ -0,0 +1,376 @@
...
+            heTitle: d.name,
+            url: "/collections/" + collection.key,
+            topicCat: "Collections",
+            heTopicCat: Sefaria.hebrewTranslation("Collections"),
+            enDesc: d.description,
+            heDesc: d.description,
+            numSheets: d.sheets.length
+        }
+    }
+    async _executeTopicQuery() {
```
nsantacruz, line 144: Move all of the topic / ref / category getting functions to own class in this file. Mark them all as private except `_executeTopicQuery()`.

static/js/sefaria/sefaria.js
```diff
@@ -2779,6 +2781,55 @@ _media: {},
...
+      }
+
+      let filters = {'topics': [], 'collections': []};
+      sheets.forEach(sheet => {
+        Object.keys(filters).forEach(itemsType => {
+            let slugsFound = [];  // keep track of topic slugs or collection slugs in this sheet
+            sheet[[itemsType]]?.forEach(item => {
+              if (!slugsFound.includes(item.slug)) { // we don't want to increase docCount when one sheet already
+                                                    // has a topic/collection with the same slug as the current topic/collection
+                let filter = filters[[itemsType]].find(f => f.aggKey === item.slug);
```
nsantacruz, line 2820: why the double brackets?

static/js/sefaria/sefaria.js
```diff
@@ -2779,6 +2781,55 @@ _media: {},
...
+              [title, heTitle] = [item.name, item.name];
+          }
+          return {
+              title, heTitle,
+              docCount: 0, aggKey: item.slug,
+              selected: 0, aggType: type,
+          };
+      }
+
+      let filters = {'topics': [], 'collections': []};
```
nsantacruz, line 2813: Maybe `filters` should be an object with keys `${slug}|${type}`

static/js/sefaria/sefaria.js
```diff
@@ -2779,6 +2781,55 @@ _media: {},
...
+      sheets.forEach(sheet => {
+        Object.keys(filters).forEach(itemsType => {
+            let slugsFound = [];  // keep track of topic slugs or collection slugs in this sheet
```
nsantacruz, line 2816: change to Set()

static/js/SidebarSearch.jsx
```diff
@@ -2,13 +2,14 @@ import { useState, useEffect } from "react";
...
 import SearchState from './sefaria/searchState';
-import SearchResultList  from './SearchResultList';
+import {SearchResultList}  from './SearchResultList';
 import DictionarySearch  from './DictionarySearch';
 import classNames from 'classnames';
 
 import {
   SearchButton,
 } from './Misc';
+import {ElasticSearchQuerier} from "./ElasticSearchQuerier";
```
nsantacruz, line 12: Move before Misc import

sourcesheets/views.py
```diff
@@ -1020,8 +1022,28 @@ def sheets_by_ref_api(request, ref):
...
+        "initialSearchType": "sheet",
+        "initialTextSearchField": search_params["textField"],
+        "initialSheetSearchFilters": search_params["sheetFilters"],
+        "initialSheetSearchFilterAggTypes": search_params["sheetFilterAggTypes"],
+        "initialSheetSearchSortType": search_params["sheetSort"]
+    }
+    he_tref = Ref(tref).he_normal()
+    normal_ref = tref if request.interfaceLang == "english" else he_tref
+    title = _(f"Sheets with ")+normal_ref+_(" on Sefaria")
+    props["sheetsWithRef"] = {"en": tref, "he": he_tref}
```
nsantacruz, line 1045: Why isn't title enough? I think we should try to avoid sending this prop if possible.

stevekaplan123: If I just send the title, how would we generate the URL when making a history state?
https://github.com/Sefaria/Sefaria-Project/blob/feature/sc-28105/re-factor-sheetsearchfilters-to-work-for/static/js/ReaderApp.jsx#L459
To generate the URL, we need access to the English ref.  If it's not passed as a prop, wouldn't we need to access window.location to know what ref we're currently on?  Also, if the hebrew ref isn't available, how would we generate the title for the history state?  We could rely on document.title.

nsantacruz: Ok, these are good points. Let's think about it like this: Currently this page can only load on fresh load as far as I understand but theoretically it could be accessed in the app. How would the app know this information if you navigate in-app? You would need state on ReaderApp presumably.

static/js/sefaria/sefaria.js
```diff
@@ -2779,6 +2781,56 @@ _media: {},
     return Sefaria.topic_toc.filter(x => x.slug == slug).length > 0;
   },
   sheets: {
+    getSheetsByRef: function(srefs, callback) {
+        return Sefaria._cachedApiPromise({
+          url: `${Sefaria.apiHost}/api/sheets/ref/${srefs}?include_collections=1&include_first_comment=1`,
```
nsantacruz, line 2786: Remove include_first_comment which I believe is no longer used.

static/js/sefaria/sefaria.js
```diff
@@ -2779,6 +2781,56 @@ _media: {},
...
+          key: srefs,
```
nsantacruz, line 2787: Include `include_collections` on the key so that these results remain separate from other results that don't have collections.

static/js/Sheets/SheetsWithRefPage.jsx
```diff
@@ -0,0 +1,191 @@
...
+            if (selected && selected !== Boolean(availableFilter.selected)) {
+                availableFilter.setSelected(true);
+            } else if (selected !== Boolean(availableFilter.selected)) {
+                availableFilter.setUnselected(true);
+            }
+        })
+    }
+
+    useEffect(() => {
+      // 'collections' won't be present if the related API set _sheetsByRef,
```
nsantacruz, line 152: This breaks encapsulation with the cache. Instead store results for this page using a separate key (see my comments on `Sefaria.getSheetsByRef()`. This useEffect should not access the cache at all.

static/js/Sheets/SheetsWithRefPage.jsx
```diff
@@ -0,0 +1,191 @@
...
+                newAvailableFilters[i].docCount = slugs[newFilter.aggKey];
+            }
+        })
+    }
+    const getSheetSlugs = (type, sheet) => {
+        const items = type === 'topics_en' ? sheet.topics : sheet.collections;
+        return items.map(x => x.slug);
+    }
+    const applyFiltersToSheets = (sheets) => {
+        searchState.appliedFilters.forEach((appliedFilter, i) => {
```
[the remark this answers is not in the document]
stevekaplan123: Yeah it works, unless I'm misunderstanding the question.  You can select multiple filters and it seems to filter properly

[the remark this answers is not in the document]
stevekaplan123: For applied filters "Shabbat" and "Education", it will only display sheets that have BOTH of these filters on.  That seems to be the way the filters work for searching for sheets, so I copied that behavior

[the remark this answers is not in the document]
nsantacruz: Ah, you're right. This logic looks sound.

nsantacruz (approved):

PR #1972, opened 2024-07-17
Improve error messaging and remove 500 errors

saengel:
## Description
Improve error messaging and remove 500 errors on the texts API (v1 and v2)

## Code Changes
1. `sefaria/model/text.py` - Write more robust error message to be more helpful for our third party users. 
2. `api/views.py` - Properly catch error for Book Level Complex Texts, and report helpful error message.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -56,13 +56,18 @@ class Text(View):
             return jsonResponse({'error': f'return_format should be one of those formats: {self.RETURN_FORMATS}.'}, status=400)
         text_manager = TextRequestAdapter(self.oref, versions_params, fill_in_missing_segments, return_format)
 
+
         try:
-            # For a SchemaNode, data is an error which handle_warnings doesn't handle (or even get triggered?)
-            # How to trigger a 400 appropriate error with an appropriate message?
+            # Todo - Maybe this error should be inside the get_versions_for_query() fxn?
             data = text_manager.get_versions_for_query()
-            data = self._handle_warnings(data)
-            return jsonResponse(data)
-        except InputError as e:
-            raise InputError(e)
+        except Exception as e:
+            # Todo - which 400 code exactly to pass? Will have to check the message of the error to
+            # make sure you're sending the right response to the user. 
+            return jsonResponse({'error': "Please pass a more specific Ref"}, status=400)
+
+        data = self._handle_warnings(data)
+        return jsonResponse(data)
+
+
 
 
--- a/api/views.py
+++ b/api/views.py
@@ -62,7 +62,7 @@ class Text(View):
             data = text_manager.get_versions_for_query()
         except Exception as e:
             # Todo - which 400 code exactly to pass? Will have to check the message of the error to
-            # make sure you're sending the right response to the user. 
+            # make sure you're sending the right response to the user.
             return jsonResponse({'error': "Please pass a more specific Ref"}, status=400)
 
         data = self._handle_warnings(data)
--- a/api/views.py
+++ b/api/views.py
@@ -58,13 +58,11 @@ class Text(View):
 
 
         try:
-            # Todo - Maybe this error should be inside the get_versions_for_query() fxn?
             data = text_manager.get_versions_for_query()
-        except Exception as e:
-            # Todo - which 400 code exactly to pass? Will have to check the message of the error to
-            # make sure you're sending the right response to the user.
-            return jsonResponse({'error': "Please pass a more specific Ref"}, status=400)
-
+        except InputError as e:
+            # Todo - which 400 code exactly to pass?
+            if str(e) == "Can not get TextRange at this level, please provide a more precise reference":
+                return jsonResponse({'error': "Please pass a more specific ref for this book, and try again. The ref you passed is a \'complex\' book-level ref. We only support book-level refs in cases of texts with a 'simple' structure. To learn more about the structure of a text on Sefaria, see: https://developers.sefaria.org/docs/the-schema-of-a-simple-text"}, status=400)
         data = self._handle_warnings(data)
         return jsonResponse(data)
 
--- a/api/views.py
+++ b/api/views.py
@@ -6,7 +6,6 @@ from django.views import View
 from .api_warnings import *
 
 
-
 class Text(View):
 
     RETURN_FORMATS = ['default', 'wrap_all_entities', 'text_only', 'strip_only_footnotes']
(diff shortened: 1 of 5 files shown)


sefaria/system/decorators.py
```diff
@@ -42,7 +42,10 @@ def wrapper(*args, **kwargs):
         except exps.InputError as e:
             logger.warning("An exception occurred processing request for '{}' while running {}. Caught as JSON".format(args[0].path, func.__name__), exc_info=True)
             request = args[0]
-            return jsonResponse({"error": str(e)}, callback=request.GET.get("callback", None))
```
edamboritz, line 45: I think this should be in the view, its too specific for a generic decorator

reader/views.py
```diff
@@ -1447,8 +1447,12 @@ def _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=comment
             return text
 
         if not multiple or abs(multiple) == 1:
-            text = _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=commentary, context=context, pad=pad,
-                             alts=alts, wrapLinks=wrapLinks, layer_name=layer_name)
+            try:
+                text = _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=commentary, context=context, pad=pad,
+                                 alts=alts, wrapLinks=wrapLinks, layer_name=layer_name)
+            except Exception as e:
```
edamboritz, line 1453: IIRC we discussed that we need to raise a proper error rather than do string comparison on the error message?

api/views.py
```diff
@@ -53,6 +55,16 @@ def get(self, request, *args, **kwargs):
         if return_format not in self.RETURN_FORMATS:
             return jsonResponse({'error': f'return_format should be one of those formats: {self.RETURN_FORMATS}.'}, status=400)
         text_manager = TextRequestAdapter(self.oref, versions_params, fill_in_missing_segments, return_format)
-        data = text_manager.get_versions_for_query()
+
+
+        try:
+            data = text_manager.get_versions_for_query()
+        except InputError as e:
```
edamboritz, line 62: see below?

reader/views.py
saengel: Yes, I recall discussing that - and recall vaguely there being a reason we stuck with this. What would you recommend at this point?

edamboritz (approved):

reader/views.py
```diff
@@ -1447,8 +1447,12 @@ def _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=comment
             return text
 
         if not multiple or abs(multiple) == 1:
-            text = _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=commentary, context=context, pad=pad,
-                             alts=alts, wrapLinks=wrapLinks, layer_name=layer_name)
+            try:
```
edamboritz, line 1450: This looks like a pre existing issue but why are we returning an http response from this lower level function?

api/views.py
```diff
@@ -53,6 +55,20 @@ def get(self, request, *args, **kwargs):
...
         text_manager = TextRequestAdapter(self.oref, versions_params, fill_in_missing_segments, return_format)
-        data = text_manager.get_versions_for_query()
-        data = self._handle_warnings(data)
+
+
+        try:
+            data = text_manager.get_versions_for_query()
+            data = self._handle_warnings(data)
+
+        except Exception as e:
```
yitzhakc, line 64: This will cause _all exceptions_ to be `caught` and resolved, ultimately returning on line 68.
Do we want to `re-raise` the other (non `InputError`) exceptions?

saengel: Good point. We definitely do catch a lot of errors further up, but let's try to be robust. @edamboritz - what would you recommend?

api/views.py
```diff
@@ -53,6 +55,19 @@ def get(self, request, *args, **kwargs):
         if return_format not in self.RETURN_FORMATS:
             return jsonResponse({'error': f'return_format should be one of those formats: {self.RETURN_FORMATS}.'}, status=400)
         text_manager = TextRequestAdapter(self.oref, versions_params, fill_in_missing_segments, return_format)
-        data = text_manager.get_versions_for_query()
-        data = self._handle_warnings(data)
+
```
yitzhakc, line 58: There are a lot of empty lines in this file, let's try and clean it up a bit :-) 
(Remove lines 9, 59, 68-73. 63 and 66 are up to you)

api/views.py
yitzhakc: Looks good! I think you can resolve this now :-)

api/views.py
```diff
@@ -53,6 +55,19 @@ def get(self, request, *args, **kwargs):
...
+
+        try:
+            data = text_manager.get_versions_for_query()
+            data = self._handle_warnings(data)
+
+        except Exception as e:
+            return jsonResponse({'error': str(e)})
```
yitzhakc, line 65: Is there a reason you don't set the status code to an error (i.e. 400)?

sefaria/model/text.py
```diff
@@ -3990,8 +3990,10 @@ def padded_ref(self):
             except AttributeError:  # This is a schema node, try to get a default child
                 if self.has_default_child():
                     return self.default_child_ref().padded_ref()
+                elif self.is_book_level():
+                    raise ComplexBookLevelRefError(book_ref=self.normal())
                 else:
-                    raise InputError("Can not pad a schema node ref")
+                    raise InputError("Cannot pad a schema node ref.")
```
yitzhakc, line 3996: I see you added a period here, but not on line 2427. Was that on purpose? I assume we don't have a clear convention on that? (yet ;-) )

sefaria/system/exceptions.py
```diff
@@ -99,3 +99,13 @@ def __init__(self, method):
         self.method = method
         self.message = f"'{method}' is not a valid HTTP API method."
         super().__init__(self.message)
+
```
yitzhakc, line 102: You should have two lines before a new class (per [PEP8](https://peps.python.org/pep-0008/#blank-lines)

api/views.py
saengel: Oversight on my end, thanks for catching!

sefaria/model/text.py
saengel: Yes, I believe it was intentional, felt like the sentence needed a period (and aligns with the other grammar fix 😄 ). Let me add on line 2427, good catch!

sefaria/system/exceptions.py
saengel: Thank you!

yitzhakc (approved):

PR #1981, opened 2024-07-28
fix(Category API): allow for json or body in POST

stevekaplan123:
## Description
Previously, the Category API assumed a 'json' parameter being passed in request.POST, but adminEditorApiRequest in sefaria.js passes a POST body instead.  Therefore, the Category Editor stopped working.

## Code Changes
I simply allowed for both options  -- 'json' in request.POST or a body -- in case someone wants to edit the category not using the Admin Editor.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -2438,10 +2438,7 @@ def category_api(request, path=None):
         else:
             return jsonResponse({"error": "Only Sefaria Moderators can add or delete categories."})
 
-        if "json" in request.POST:
-            j = request.POST.get("json")
-        else:
-            j = request.body
+        j = request.body
         if not j:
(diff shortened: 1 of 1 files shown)


reader/views.py
```diff
@@ -2438,9 +2438,12 @@ def _internal_do_post(request, update, cat, uid, **kwargs):
         else:
             return jsonResponse({"error": "Only Sefaria Moderators can add or delete categories."})
 
-        j = request.POST.get("json")
+        if "json" in request.POST:
```
nsantacruz, line 2441: Is there any code that still uses the `json` key? If not, let's remove it because we want to be moving to use `request.body` instead of the `json` key.

nsantacruz (approved):

PR #1982, opened 2024-07-28
Feature/sc 28127/frontend for sheets with ref

stevekaplan123:
## Description
This eliminates `textSearchState` and `sheetSearchState` and replaces it with `searchState`.  In addition, it contains a small bug fix which was introduced in the Sheets With Ref PR.  This bug occurs when a user who is on a sheets-with-ref page then tries to search the site.  The user will still be seeing the sheets-with-ref applied filters.  This isn't right as someone who tries to search the site should not be using any applied filters from a sheets-with-ref page.  This bug occurs because both use cases share the same searchState type -- 'sheet'.  The searchState filters are now reset with the function resetSearchFilters by the return function of useEffect in SheetsWithRefPage.jsx.

## Code Changes
1. `searchState` does contain a variable `type` 'sheet' or 'text' which determines what metadata the `searchState` uses.   I do not pass down the type as a separate variable as it can just be accessed via `searchState`.
2. 'makeURL' still takes a prefix of 't' or 's' and the URL still contains prefixes of 't' or 's' in the URL params.  
3. I changed `get_search_params` on the backend which behaves differently for sheet and text.
4. Core functions in ReaderApp were changed so that they no longer take a type ('sheet' vs 'text') -- such as updateAvailableFilters, _getSearchState, updateSearchState, and updateQuery.
5.  I left this as a draft PR because of two lines in ReaderApp that I wasn't sure about:
 `     const searchType = props.initialSearchType || 'text';  // should really be set by URL such as sheets.sefaria.org or www.sefaria.org`

and

`      searchState:             state.searchState             || new SearchState({ type: this.props.initialSearchType || 'text' }),`

and 
```
    const searchState =  hasSearchState  ? this.state.panels[0].searchState.update({ filtersValid: false })
        : new SearchState({ type: this.props.initialSearchType || 'text'});
```


reader/views.py
```diff
@@ -785,24 +785,26 @@ def get_filters(prefix, filter_type):
...
         "query": urllib.parse.unquote(get_dict.get(get_param("q", i), "")),
         "tab": urllib.parse.unquote(get_dict.get(get_param("tab", i), "text")),
-        "textField": ("naive_lemmatizer" if get_dict.get(get_param("tvar", i)) == "1" else "exact") if get_dict.get(get_param("tvar", i)) else "",
-        "textSort": get_dict.get(get_param("tsort", i), None),
-        "textFilters": text_filters,
-        "textFilterAggTypes": [None for _ in text_filters],  # currently unused. just needs to be equal len as text_filters
-        "sheetSort": get_dict.get(get_param("ssort", i), None),
-        "sheetFilters": sheet_filters,
-        "sheetFilterAggTypes": sheet_agg_types,
+        "field": ("naive_lemmatizer" if get_dict.get(get_param("tvar", i)) == "1" else "exact") if get_dict.get(get_param("tvar", i)) else "",
```
nsantacruz, line 804: field should be `None` for sheets so that the default value is chosen.

nsantacruz (approved):

PR #1989, opened 2024-07-30
Disable Strapi for Banners, Modals, Sidebar Ads

relyks:
## Description
Strapi is being disabled during the trial of using Unbounce for banners and modals. Code changes were minimally made to prevent having to comment out or directly remove code. Sidebar Ads will be inaccessible during this trial. Modals will still be accessible by the mobile application.

## Code Changes
* The Banner and InterruptingMessage components were removed
* The StrapiDataContext provider was removed
* Prevent the execution of Strapi API calls by adding a `false` boolean to the conditionals

### DIFF
--- a/static/js/Promotions.jsx
+++ b/static/js/Promotions.jsx
@@ -11,7 +11,8 @@ const Promotions = () => {
   const context = useContext(AdContext);
   const strapi = useContext(StrapiDataContext);
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


edamboritz (approved):
The disabled things here might be overkill but ok

relyks:
> The disabled things here might be overkill but ok

I think so too, but I wanted to make sure no API calls were made and any code that's not necessary to not be evaluated

PR #1990, opened 2024-07-30
Add Unbounce

relyks:
## Description
The Unbounce embed script has been added. Banners and Modals will be done through Unbounce for the trial's duration.


edamboritz (commented):
Possibly this should actually be at the bottom of the <body>?

PR #1991, opened 2024-07-31
fix(Topic Pages): getBulkText bug that led it to generate long URLs

stevekaplan123:
## Description
The URLs were over 3800 characters because the pipe character gets encoded with 3 characters: "%7C", but the comparison treated the pipe character as one character.

## Code Changes
Instead of comparing the length of `${hostStr}${last}|${encodedRef}${paramStr}` to 3800, it's necessary to encode `${hostStr}${last}|${ref}${paramStr}` and compare that to 3800 instead.

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -479,10 +479,9 @@ Sefaria = extend(Sefaria, {
     let refStrs = [""];
     refs.map(ref => {
       let last = refStrs[refStrs.length-1];
-      const encodedRef = encodeURIComponent(ref);
-      const encodedFullURL = encodeURIComponent(`${hostStr}${last}|${ref}${paramStr}`);
-      if (encodedFullURL.length > MAX_URL_LENGTH) {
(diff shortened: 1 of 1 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -479,9 +479,10 @@ Sefaria = extend(Sefaria, {
     let refStrs = [""];
     refs.map(ref => {
       let last = refStrs[refStrs.length-1];
-      const encodedRef = encodeURIComponent(ref)
-      if (`${hostStr}${last}|${encodedRef}${paramStr}`.length > MAX_URL_LENGTH) {
-        refStrs.push(encodedRef)
+      const encodedRef = encodeURIComponent(ref);
+      const encodedFullURL = encodeURIComponent(`${hostStr}${last}|${ref}${paramStr}`);
```
nsantacruz, line 483: the only thing that needs to be encoded here is pipe if I understand correctly (and we need to use encodedRef instead of ref).

PR #1995, opened 2024-08-08
fix(Category Editor): do not look for books that are not in the category being reordered

stevekaplan123:
## Description
We sometimes have categories with a book inside the category with the same title. For example, the category "Zohar" has the book "Zohar" inside of it.  The Category Editor doesn't work properly in this case.  

## Code Changes
The Category Editor previously would look to re-order a book without checking that its path is inside the category being re-ordered.  This was a reasonable assumption to make in most situations, but in the above case, this means that the book "Zohar" could get re-ordered when the intention was to re-order the category "Zohar".  I fixed this by only looking for books to re-order that are actually inside the category being re-ordered.  In the code, this means an Index is only loaded when its `categories` property is identical to the `cat_path` being re-ordered.

### DIFF
--- a/sefaria/helper/category.py
+++ b/sefaria/helper/category.py
@@ -186,16 +186,15 @@ def update_order_of_category_children(cat, uid, subcategoriesAndBooks):
     results = []
     for subcategoryOrBook in subcategoriesAndBooks:
         order += 5
-        try:
-            obj = Index().load({'title': subcategoryOrBook, 'categories': cat_path})
-            assert obj is not None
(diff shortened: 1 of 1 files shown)


edamboritz (changes requested):
Asserts are not a recommended approach to catching errors, prolific in our code though they may be. There are many pitfalls and non standard behaviour that may arise. 
I would recommend/ask replacing them with an appropriate exception.

edamboritz (approved):

PR #2000, opened 2024-08-18
User Dropdown for Logged out Library

saengel:
## Description
Dropdown menu for modularized library, specifically for logged out users. 

## Code Changes
The following changes were made to the files below:
1. `static/css/s2.css` - Some minor padding shifts, adding styling for the language toggle with a pseudo-class for the dot separator. 
2. `static/icons/logged_out.svg` - The new logged out icon 
3. `static/js/common/DropdownMenu.jsx` - Improving the generic `<DropdownMenu />` component to enhance reusability (i.e. making the icon for the dropdown customizable, and control over whether or not to open links in new tabs). 
4. `static/js/Header.jsx` - Adding the new logged out component (`<LoggedOutMenu />`), implementing the shifts to the generic component mentioned above in the module switcher.


static/css/s2.css
```diff
@@ -13349,6 +13350,38 @@ span.ref-link-color-3 {color: blue}
...
+  height: 23px;
+  padding-bottom: 15px;
+}
+
+.languageDot{
+  font-size: 25px;
+  padding: 10px 2px;
+}
+
+.englishLanguageButton::after {
```
edamboritz, line 13381: I think you can actually make this a bit more general, not needing to create a special class for the enlish link

edamboritz (approved):

static/css/s2.css
saengel: Good thinking! 

I did this because of annoying spacing issues with r/l padding on English vs Hebrew buttons. Is that OK in this case? Happy to show you if it's clearer that way :)

PR #2003, opened 2024-08-19
Bug/sc 28927/bugs on the jerusalem topic page

yonadavGit:
## Description
This PR rewrites the algorithm that partitions the refs list in getBulkTest into separate requests, such that the questions marks in the refs will be converted to a valid URL encoding 
## Code Changes
added partitionArrayForURL to be used from getBulkText in Sefaria.js

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -463,7 +463,7 @@ Sefaria = extend(Sefaria, {
         });
   },
   _bulkTexts: {},
-  partitionArray: function partitionArrayForURL(urlMaxLength, arr, dividerToken) {
+  partitionArrayForURL: function(urlMaxLength, arr, dividerToken) {
     const result = [];
     const dividerTokenLength = encodeURIComponent(dividerToken);
     let currentPartition = [];
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -274,6 +274,12 @@ Sefaria = extend(Sefaria, {
     let index = Sefaria.index(pRef.index);
     return index && index.categories ? index.categories : [];
   },
+  refIndexTitle: function(ref) {
+    let pRef = Sefaria.parseRef(ref);
+    if ("error" in pRef) { return null; }
+    let index = Sefaria.index(pRef.index);
+    return index?.title
+  },
   sectionRef: function(ref, deriveIfNotFound=false) {
     // Returns the section level ref for `ref` or null if no data is available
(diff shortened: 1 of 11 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -469,31 +469,64 @@ Sefaria = extend(Sefaria, {
         });
   },
   _bulkTexts: {},
+  partitionArrayForURL: function(urlMaxLength, arr, dividerToken) {
```
edamboritz, line 472: I would switch the positional arguments here arr and urlMaxLength, its more customary for the main input to be the first param.

static/js/sefaria/sefaria.js
```diff
@@ -469,31 +469,64 @@ Sefaria = extend(Sefaria, {
...
-      let last = refStrs[refStrs.length-1];
-      const encodedFullURL = encodeURI(`${hostStr}${last}|${ref}${paramStr}`);
-      if (encodedFullURL.length > MAX_URL_LENGTH) {
-        refStrs.push(ref)
-      } else {
-        refStrs[refStrs.length-1] += last.length ? `|${ref}` : ref;
-      }
-    });
+    const limit = MAX_URL_LENGTH-(hostStr+paramStr).length-ASSUMED_HOSTNAME_LENGTH_BOUND
+    const refsSubArrays = this.partitionArrayForURL(limit, refs, '|');
```
edamboritz, line 525: The function above is great! 
But i would like to separate the concerns of encoding the refs for the URI vs partitioning them. 
The former always needs to happen, the latter may one day have a different solution and I'd like to be able to just remove the call to the partition logic and the rest of this function proceeds as is.

static/js/sefaria/sefaria.js
```diff
@@ -469,31 +469,64 @@ Sefaria = extend(Sefaria, {
...
+    const refStrs = refsSubArrays.map(refsSubArray => refsSubArray.join('|'));
+
     let promises = refStrs.map(refStr => this._cachedApiPromise({
-      url: encodeURI(`${hostStr}${refStr}${paramStr}`),
+      url: `${hostStr}${encodeURIComponent(refStr)}${paramStr}`,
```
edamboritz, line 529: Why are we encoding this again? Because of the pipes?

static/js/sefaria/sefaria.js
edamboritz: nm this is already fine

edamboritz (approved):

PR #2006, opened 2024-08-21
fix(Connections Panel): only open Commentary in the Connections Panel…

stevekaplan123:
## Description
This problem was originally discovered because when one clicked on a Haggadah commentary on a topics page, the commentary didn't load properly in the side panel.  But it turned out to be a wider problem also affecting depth 2 Talmud commentaries.  This PR changes the way commentaries can be opened in the sidebar.  It only opens a commentary in the side panel if depth of commentary ref is >= 3 or the commentary ref is opened from book TOC.  The reasoning for this is as follows: If the commentary ref is opened from the book TOC, such as "Rashi on Genesis 1:2", we know exactly what ref to show in the main panel, Genesis 1:2, and we know what comments to show -- Rashi on Genesis 1:2.  In the case of opening a commentary ref from a search page or topic page where the commentary ref has a depth of 3, such as "Rashi on Genesis 1:1:1" it is also clear what to do -- show the "Rashi on Genesis 1:1" section in the main panel and the comments in the side panel.  However, if a depth 2 commentary ref such as "Rashi on Genesis 1:5" appeared on a topic page, we wouldn't want to show "Rashi on Genesis 1" in the main panel and all the comments for "Rashi on Genesis 1" as it would be too many comments to be useful.  So it's not clear what to do in that situation.  Complicating things are Talmud commentaries that are depth 2, instead of depth 3, where we have no way of knowing what comments to show for a given ref, since such Talmud commentaries are not one_to_one mappings of the depth 2 Talmud.  And complicating things further are complex texts like Haggadah commentaries where we don't even know what the depth of a given node is.  Therefore, it makes it simpler to require that a commentary ref be of at least depth 3 since then we know how to alter the ref to get to the base ref by simply going one level up.  If the user was on a book TOC for a commentary, we don't need to require depth 3 because we don't have to alter the ref. "Rashi on Genesis **1:6**" simply becomes "Genesis **1:6**".

## Code Changes
From the book TOC, I pass a variable, openCommentarySidePanel, to showBaseText, that ultimately gets processed in the Sefaria module's isCommentaryRefWithBaseText function with the simple check: if openCommentarySidePanel is false (i.e. we were not on the book TOC, but on a search or topic page), and the depth of the ref is < 3, return false so that the commentary ref will be rendered in the main panel.

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -1434,7 +1434,8 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     this.state.panels = []; // temporarily clear panels directly in state, set properly with setState in openPanelAt
     this.openPanelAt(0, ref, currVersions, options);
   }
-  openPanelAt(n, ref, currVersions, options, replace, convertCommentaryRefToBaseRef=true, replaceHistory=false, saveLastPlace=true) {
+  openPanelAt(n, ref, currVersions, options, replace, convertCommentaryRefToBaseRef=true,
+              replaceHistory=false, saveLastPlace=true, openCommentarySidePanel=false) {
     /* Open a new panel or replace existing panel. If book level, Open book toc
     * @param {int} n: Open new panel after `n` with the new ref
     * @param {string} ref: ref to use for new panel.  `ref` can refer to book, actual ref, or sheet.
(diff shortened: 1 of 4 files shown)


static/js/ReaderApp.jsx
```diff
@@ -1434,7 +1434,8 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     this.state.panels = []; // temporarily clear panels directly in state, set properly with setState in openPanelAt
     this.openPanelAt(0, ref, currVersions, options);
   }
-  openPanelAt(n, ref, currVersions, options, replace, convertCommentaryRefToBaseRef=true, replaceHistory=false, saveLastPlace=true) {
+  openPanelAt(n, ref, currVersions, options, replace, convertCommentaryRefToBaseRef=true,
+              replaceHistory=false, saveLastPlace=true, openCommentarySidePanel=false) {
```
nsantacruz, line 1438: I think the way this parameter is currently called is confusing and the docs for it are also confusing because it's stated in the negative and there are multiple if statements in the description of the parameter.

Maybe instead call this `forceOpenCommentaryPanel` and default is `false`. Docs can say something like `when true, this will open the commentary side panel with this ref in it regardless of the refs depth. When false, will only open side panel if ref is depth 3 or greater`.

nsantacruz (approved):

PR #2021, opened 2024-09-04
Update drop drown for logged in desktop (library)

saengel:
## Description
Updating the user drop down for library logged-in view. 

## Code Changes
1. In `static/css/s2.css` - CSS shifts to get everything working, tested to make sure it doesn't break anything. Commenting out the chevron dropdown (see note below)
2. `static/icons/logged_out.svg` - The icon for logged out users. 
3.  `static/js/Header.jsx` - The new logged in dropdown component (`<LoggedInDropdown />`), rendering it. Fixes also to the module switcher to jive with the adjustment to `<DropdownMenu />`
4. `static/js/common/DropdownMenu.jsx` - Adjusting the dropdown menu component to flexibly allow opening in new tabs vs same window, and accepting an icon component instead of a url for more flexibility. 

## Notes
- A subtle shift was introduced to the `<DropdownMenu />` component, accepting the icon image (from which the dropdown 'hangs') as a component instead of a url string. This allows for greater flexibility when rendering components like the initials icon. We will need to likely shift some small pieces in other PRs for everything to merge seamlessly, but it should be fairly easy. 
- The hanging chevron was commented out, seems to no longer be in spec. Will return if needed.


static/js/Header.jsx
```diff
@@ -18,22 +18,74 @@ import {Autocomplete} from './Autocomplete'
...
   return (
-    <DropdownMenu>
-    <DropdownMenuItem url={'/'}>
+    <DropdownMenu menuIconComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
+    <DropdownMenuItem url={'/'} newTab={true}>
       <DropdownMenuItemWithIcon icon={'/static/icons/library_icon.svg'} textEn={'Library'} textHe={'ספריה'} />
     </DropdownMenuItem>
     <DropdownMenuSeparator />
-    <DropdownMenuItem url={'//sheets.sefaria.org'}>
+    <DropdownMenuItem url={'//sheets.sefaria.org'} newTab={true}>
```
edamboritz, line 80: What are these double slashes?

static/js/common/DropdownMenu.jsx
```diff
@@ -40,7 +45,7 @@ const DropdownMenuItemWithIcon = ({icon, textEn, textHe}) => {
   );
 }
 
-const DropdownMenu = ({children}) => {
+const DropdownMenu = ({children, menuIconComponent}) => {
```
edamboritz, line 48: What are you passing in here? another react component as a prop? thats generally not considered a good practice...

static/icons/logged_out.svg
edamboritz, line 1: Why is this icon in this PR if the PR is for logged in menu?

static/js/common/DropdownMenu.jsx
edamboritz: If you need to be able to insert a dynamic component the way to do it is composability, as we have done with the rest of the menu. 
Further break down the top components so you can render the menu icon component inline as children. 
For example 

```
  <DropdownMenu>
     <ProfilePic>
     <DropDownMenuContents>
        <DropDownMenuItem>
        .....
     </DropDownMenuContents>
 </DropdownMenu>
 ````

File: static/js/Header.jsx (hunk 1 above)
Comment (saengel) [reply static/js/Header.jsx:80]: I don't recall, likely an error! Will remove.

File: static/icons/logged_out.svg
Comment (saengel) [reply static/icons/logged_out.svg:1]: Mistake, taking out. Thanks for catching :)

File: static/js/common/DropdownMenu.jsx (hunk 2 above)
Comment (saengel) [reply static/js/common/DropdownMenu.jsx:48]: Thanks for pointing this out! I didn't realize, and really appreciate learning about better React practices.

I will push back a drop - in this case, I needed parity with the code that renders the "initials" when a profile image is not present, and needed to structure the code this way iirc to enable that in both PRs. Happy to discuss more, and hear your insights as to if there's a better and more proper React way of doing this. Thanks!

File: static/js/Header.jsx (hunk 1 above)
Comment (saengel) [reply static/js/Header.jsx:80]: Might be `/sheets`, confirm with SK.

File: static/js/Header.jsx [hunk 3]
```diff
@@ -18,22 +18,74 @@ import {Autocomplete} from './Autocomplete'
...
       <DropdownMenuItemWithIcon icon={'/static/icons/sheets_icon.svg'} textEn={'Sheets'} textHe={'דפים'}/>
     </DropdownMenuItem>
     <DropdownMenuSeparator />
-    <DropdownMenuItem url={'//developers.sefaria.org'}>
+    <DropdownMenuItem url={'//developers.sefaria.org'} newTab={true}>
```
saengel, line 84: Switch to full URL

edamboritz (approved):

PR #2026, opened 2024-09-05
Rtl client

akiva10b:
## Description

## Code Changes

## Notes


static/js/common/DropdownMenu.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+const useIsMenuOpen = (context) => {
+    //React must have the same hooks any time, so we should have context and state any way.
+    //For that we also define FallbackContext. It's defined outside for not be created in any use but only once.
+    const contextValue = useContext(context || FallbackContext);
+    const [stateIsMenuOpen, stateSetIsMenuOpen] = useState(false);
+    return contextValue ? [contextValue.isMenuOpen, contextValue.setIsMenuOpen] : [stateIsMenuOpen, stateSetIsMenuOpen];
+}
+
+const DropdownMenu = ({buttonContent, menu, context=null}) => {
```
YishaiGlasner, line 15: should have children

static/js/TextRange.jsx
```diff
@@ -1,41 +1,43 @@
...
-    window.addEventListener('resize', this.handleResize);
+    this.setState({isMounted: true});
+    this.setData()
+    window.addEventListener('resize', this.conditionalPlaceSegmentNumbers);
   }
   componentWillUnmount() {
-    this._isMounted = false;
-    window.removeEventListener('resize', this.handleResize);
+    this.setState({isMounted: false});
+    window.removeEventListener('resize', this.conditionalPlaceSegmentNumbers);
```
akiva10b, line 31: verify that changes don't stem from changing cache

static/js/TextRange.jsx
```diff
@@ -605,6 +576,12 @@ class TextSegment extends Component {
     return(textArray)
   }
 
+  isRtl(language) {
```
akiva10b, line 579: requires doc-string to explain what it si used for

static/js/TextRange.jsx
```diff
@@ -605,6 +576,12 @@ class TextSegment extends Component {
...
+    const rtlOnly = (!this.props.en && this.props?.primaryDirection === 'rtl') || (!this.props.he && this.props?.translationDirection === 'rtl');
+    const ltrOnly = (!this.props.en && this.props?.primaryDirection === 'ltr') || (!this.props.he && this.props?.translationDirection === 'ltr');
```
akiva10b, line 581: we should likely use the data instead of the props - this is something we should look over all of TextRange to fix

static/js/common/DropdownMenu.jsx
```diff
@@ -0,0 +1,55 @@
...
+const DropdownMenu = ({children, buttonContent}) => {
+    /**
+     * buttonContent is a React component for the opening/closing of a button.
+     * the menu will be closed in click anywhere except in an element with classname 'preventClosing'.
+     * this class is using useRef for open/close rather than useState, for changing state triggers re-rendering of the
+     * component and all its children, so when clicking on children their onClick won't be executed.
+     */
+
+    const dropdownLinksClass = 'dropdownLinks-menu';
+    const isOpenRef = useRef(null);
```
akiva10b, line 14: this is actually referring to the dropdown elm. Considure a name like "dropdownRef". In line 26 you can have a function `isOpen(dropdownRef)` which returns `dropdownRef.current?.className?.includes('open')` for something that is very readable

static/js/VersionsTextList.jsx
```diff
@@ -0,0 +1,103 @@
...
+    }, [vFilter]);
+
+
+    const getSectionRef = () => {
+        const ref = srefs[0]; // TODO account for selections spanning sections
+        return Sefaria.sectionRef(ref) || ref;
+    };
+
+    const getVersion = () => {
+        return Object.values(versionLangMap).flat().filter(
```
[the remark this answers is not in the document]
YishaiGlasner: using find is better, and i think with that we don't need doc string

static/js/ReaderPanel.jsx
```diff
@@ -41,21 +41,35 @@ import {
...
+      ...this.clonePanel(props.initialState),
+      initialAnalyticsTracked: false,
+      width: this.props.multiPanel ? 1000 : 500, // Assume we're in a small panel not using multipanel
+      backButtonSettings: null,
+      data: null,
+    };
     this.sheetRef = React.createRef();
     this.readerContentRef = React.createRef();
   }
+  conditionalSetTextData() {
```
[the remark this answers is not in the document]
YishaiGlasner: i'm not sure what this comment neams. the code here uses Sefaria.getTextFromCurrVersions, i.e. the cache.
but when thinking about it, now that the data is in the context, maybe it will be better to get it from there (but we are using the data on componentDidUpdate, so if we want to use context we'll have to use getSnapshotBeforeUpdate for storing the prev context as snapshot).

static/js/TextRange.jsx
YishaiGlasner: here it's textSegment, not textRange, and i think it's reasonable it will get it as props (i've removed this whole function that was doing nothing).

static/js/TextRange.jsx
YishaiGlasner: i don't see the connection to cache. there are 2 things here: 1. changing this._isMounted to isMounted that's stored in state, which i think is more reasonable. 2. changing the name of the function that place segments

static/js/ReaderDisplayOptionsMenu.jsx
```diff
@@ -0,0 +1,135 @@
...
+    const isSidePanel = !['Text', 'Sheet'].includes(panelMode);
+    const showLangaugeToggle = () => {
+      if (panelMode === 'Sheet') { return false; }
+      if (Sefaria._siteSettings.TORAH_SPECIFIC) return true;
+
+      const hasHebrew = !!textsData.he.length;
+      const hasEnglish = !!textsData.text.length;
+      return !(hasHebrew && hasEnglish);
+    };
+    const showPrimary = language !== 'english';
```
akiva10b, line 22: current: change from magic strings

static/js/ReaderDisplayOptionsMenu.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const borderLine = <div className="text-menu-border"/>;
+
+    const showLayoutsToggle = () => {
+        if ((panelMode === 'Sheet' && Sefaria.interfaceLang === 'hebrew') || //sheets in hebrew interface are hebrew
+        (width <= 600 && language === 'bilingual')) { //no loyout for mobile biilingual
+            return false;
+        }
+        return true;
+    }
```
akiva10b, line 37: current: why not simplify
```
return ((panelMode === 'Sheet' && Sefaria.interfaceLang === 'hebrew') || (width <= 600 && language === 'bilingual'))
```

static/js/ReaderDisplayOptionsMenu.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const hasAliyot = () => {
+        let booksWithAliyot = ["Genesis", "Exodus", "Leviticus", "Numbers", "Deuteronomy", "Onkelos Genesis", "Onkelos Exodus", "Onkelos Leviticus", "Onkelos Numbers", "Onkelos Deuteronomy"];
+        return booksWithAliyot.includes(textsData?.book);
+    };
+    const showAliyotToggle = () => {
+        return hasAliyot() && panelMode !== "Sheet";
```
akiva10b, line 44: current: lots of magic strings all over the place

static/js/LayoutButtons.jsx
```diff
@@ -0,0 +1,56 @@
...
+        : (primaryDir !== translationDir || panelMode === 'Sheet') ? 'mixed' //two texts with different directions
+            : (primaryDir === 'rtl') ? 'bi-rtl' //two rtl texts
+                : 'bi-ltr'; //two ltr texts
+}
+
+const LayoutButtons = () => {
+    const {language, textsData, setOption, layout, panelMode} = useContext(ReaderPanelContext);
+    const layoutState = calculateLayoutState();
+
+    const getPath = (layoutOption) => {
```
akiva10b, line 20: current: use a single patterns for functions, either in the closure or outside of it. @edamboritz do you have a recommendation of a pattern here?

akiva10b: ideally have distributed functions and call them into the functional component

static/js/ConnectionsPanel.jsx
```diff
@@ -232,51 +232,29 @@ class ConnectionsPanel extends Component {
...
-          en: ((this.props.masterPanelLanguage !== "hebrew" && !!data.text.length) || (this.props.masterPanelLanguage === "hebrew" && !data.he.length)) ? this.getVersionFromData(data, "en") : null,
-          he: ((this.props.masterPanelLanguage !== "english" && !!data.he.length) || (this.props.masterPanelLanguage === "english" && !data.text.length)) ? this.getVersionFromData(data, "he") : null,
-        },
-        mainVersionLanguage: currentLanguage,
-        sectionRef: data.sectionRef,
+        currObjectVersions: { en: null, he: null },
       });
+    }
+    const [primary, translation] = Sefaria.getPrimaryAndTranslationFromVersions(data.versions);
+    this.setState({
```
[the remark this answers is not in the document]
YishaiGlasner: i think not, and now i also think it'snot bad. getPrimaryAndTranslationFromVersions is only a function that identifies the primary and translation of two versions, and this is exactly what we wnat to do here.

PR #2031, opened 2024-09-09
chore: refactor ModuleTitle so that h1 is always used

stevekaplan123:
## Description
We were confusingly using both h3 and h1 in Modules and then styling the h3 and h1 tags to look exactly the same.  This PR makes all ModuleTitles use h1s.

## Code Changes
1.  Re-factor `ModuleTitle` to not have h1 as a parameter
2. Change any uses of `ModuleTitle` that pass h1 to a prop to no longer pass h1
3. Change `AboutTextCategory` to use `ModuleTitle` instead of h3.
4. Change css to no longer style h3s to look like h1s.
5. Two special cases of CSS styling had to be dealt with here as well.  The "Support Sefaria" module was using an h3 previously and so its CSS styled an h3 tag.  It no longer is.  Also, text category pages have special styling for the name of the category to show the h1 header in all upper case letters.  I had to change the CSS and add ".contentInner" so that only the text category title gets modified by the relevant CSS and we do not also modify the NavSidebar h1s.  (This last case may have been part of why the strange lack of distinction between h3 and h1 existed in the code).

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -1817,7 +1817,6 @@ div.interfaceLinks-row a {
   letter-spacing: 0em;
   text-decoration: underline;
 }
-.readerPanel .navSidebarModule h3,
 .readerPanel .navSidebarModule h1,
 .readerPanel .bookPage h3{
   margin: 0 0 20px;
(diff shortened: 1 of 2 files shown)


edamboritz:
I'm not sure I'm clear on why the last part of #5 is even an issue?

edamboritz (approved):

PR #2038, opened 2024-09-12
Topics analytics

nsantacruz:
## Description
Two main changes
1. Add `analyticsEventTracker.js`: A new module for easily tracking analytics events (more on this below)
2. Using `analyticsEventTracker`'s framework to add analytics events to the new topic pages

## Code Changes
1. Add `analyticsEventTracker.js`. This module allows for tracking analytics events through only adding data attributes to the appropriate DOM nodes. Some important aspects:
a. Uses IntersectionObserver to implement a new JS event `scrollIntoView`. This event will fire when the DOM node scrolls into view. In addition, we need a MutationObserver to look for nodes added that use the scrollIntoView event.
b. Use MutationObserver to replace `toggle` events with `bubblingToggle` events. This solves the problem that `toggle` events don't bubble and therefore they're not detectable at a higher level. `bubblingToggle` is a custom JS event that essentially proxies the toggle event but also bubbles. Yay for bubbles.
c. See ADR # 5 for specifics of the spec for this module

2. Adding analytics to topic pages
a. Analytics data was added at various levels of the DOM tree. I always tried to add it to the highest node that made sense. So for example, all analytics data that applies to anything on the topics page is on the TopicsPage component.

3. A minor change: I found many instances where `key`s were missing and this caused many warnings from React in the console. I fixed the issues I found.
4. I needed to refactor `Sefaria.topicTocCategory(slug)` (which returns the topic toc category for a topic slug) to `Sefaria.topicTocCategories(slugs)` which returns the full category path of this topic. This is because analytics wants to track the full category path as opposed to just the last category in the path.


edamboritz (approved):
I think in the future we would prefer to separate concerns: 
1. A PR for the implementation of the actual analytics code/framework
2. A PR for the small warning fixes
3. A PR for using the new analytics framework n topics specifically.

PR #2059, opened 2024-10-08
Bug/sc 29278/i clicked sign up to get updates on the https

stevekaplan123:
## Description
There were a few different problems with subscribing to the newsletter on the front-end and back-end.  On the back-end, there was an error because the only input to the subscription function provided was the user's e-mail, whereas our CRM requires first and last name also be provided.  Moreover, the error on the back-end was not handled properly causing the user to see the entire HTML error message that was generated by django.  

## Code Changes
1.  The back-end should catch the TypeError generated when no first name is provided so I changed the exception handler to `except (TypeError, ValueError) as e:`.
2. We can use the NewsletterSignUpForm component which is currently used in a few different places in the site which requires the user provide a first name and last name so that the error doesn't occur in the first place.
3.  Finally, the SubscribeButton, if we want to use it again, was broken, because it did not stringify the body of the request.  I changed this as well.

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -179,8 +179,10 @@ def generic_subscribe_to_newsletter_api(request, org, email):
         "steinsaltz": subscribe_steinsaltz,
     }
     body = json.loads(request.body)
-    first_name = body.get("firstName", None)
-    last_name = body.get("lastName", None)
+    if "firstName" not in body or "lastName" not in body:
+        return jsonResponse({"error": "You must provide first and last name."}, status=401)
+    first_name = body.get("firstName")
+    last_name = body.get("lastName")
     try:
         subscribe = org_subscribe_fn_map.get(org)
         if not subscribe:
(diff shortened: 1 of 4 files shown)


sefaria/views.py
```diff
@@ -179,8 +179,10 @@ def generic_subscribe_to_newsletter_api(request, org, email):
         "steinsaltz": subscribe_steinsaltz,
     }
     body = json.loads(request.body)
-    first_name = body.get("firstName", None)
-    last_name = body.get("lastName", None)
+    if "firstName" not in body or "lastName" not in body:
+        return jsonResponse({"error": "You must provide first and last name."})
+    first_name = body.get("firstName")
+    last_name = body.get("lastName")
```
akiva10b, line 185: seems like we could do something like this:
```
    first_name = body.get("firstName")
    last_name = body.get("lastName")
    if not first_name or not last_name:
        return jsonResponse({"error": "You must provide first and last name."})
```

akiva10b: its a little cleaner and doesn't require using strings twice

static/js/NewsletterSignUpForm.jsx
```diff
@@ -28,7 +28,7 @@ export function NewsletterSignUpForm({
                     setSubscribeMessage("Subscribed! Welcome to our list.");
                     Sefaria.track.event("Newsletter", "Subscribe from " + contextName, "");
                 }).catch(error => {
-                    setSubscribeMessage(error?.error || "Sorry, there was an error.");
+                    setSubscribeMessage(error?.message || "Sorry, there was an error.");
```
akiva10b, line 31: great!

akiva10b (approved):

PR #2064, opened 2024-10-14
Feature/sc 27313/remove resources panel

stevekaplan123:
## Description
In this PR, I removed the resources panel from the sheets viewer as well as moving code that would no longer be used or relevant given that the resources panel is no longer part of the sheets viewer.

## Code Changes
1.  I removed every reference to the mode SheetAndConnections as it is no longer possible.
2.  In ConnectionsPanel.jsx, I removed the function isSheet, showSheetNodeConnectionTools, and the prop masterPanelSheetId since the master panel can no longer be a sheet.  Likewise, `masterPanelMode === "Sheet"` cases have been removed.  I removed several components no longer needed: AboutSheet, SheetToolsList, SheetNodeConnectionTools, MySheetsList, PublicSheetsList and AboutSheetButtons.  Some of these components' code was copied over and is being re-used in components in other PRs for the Sheets Viewer.  One particular component, ShareBox, has the exact same functionality that I need for the new Sheets Viewer, and so all I had to do was modify its prop MasterPanelSheetId to SheetId.
3.  In sefaria.js in linkSummary, we used to have to handle the possible of a Nechama Leibowitz sheet showing up as a connection to itself.  This sheet was called the excludedSheet.  Since this is no longer possible, I removed the code that handles this excludedSheet case.  This code is called from ConnectionsSummary.jsx which was also modified.
4.  Previously, handleSheetSegmentClick in ReaderPanel used to call handleSegmentClick in ReaderApp, which in turn would open a new Resources Panel using openTextListAt.  This is no longer necessary, so I re-factored handleSheetSegmentClick and handleSegmentClick.  Now, handleSegmentClick just deals with the case where a library source was clicked, and handleSheetSegmentClick just sets the state, updating `highlightedNode` and `highlightedRefs` based on the sheet segment click.  This allowed me to remove the function setSheetHighlight in ReaderApp and the functions openSheetConnectionsInPanel and closeSheetConnectionsInPanel in ReaderPanel as they are no longer called.
5. Lastly, I moved the component SheetContent, and its associated components, from Sheet.jsx to SheetContent.jsx, a new file


static/js/ConnectionsPanel.jsx
```diff
@@ -203,16 +199,14 @@ class ConnectionsPanel extends Component {
...
-      //this.setState({currentlyVisibleSectionRef: Sefaria.sectionRef(this.props.currentlyVisibleRef)});
 
-    }
+    Sefaria.getTranslations(ref).then(versions => this.setState({ availableTranslations: Object.values(versions).flat() })); //for counting translations
+    Sefaria.getRef(this.props.currentlyVisibleRef).then(data => { //this does not properly return a secionRef for a spanning/ranged ref
+      const currRef = (typeof data == "string") ? Sefaria.sectionRef(data) : data["sectionRef"]; //this is an annoying consequence of getRef not actually returning a
+      // consistent response. Its either the ref from cache or the entire text api response if async.
+      this.setState({currentlyVisibleSectionRef: currRef});
+    });
+    //this.setState({currentlyVisibleSectionRef: Sefaria.sectionRef(this.props.currentlyVisibleRef)});
```
yitzhakc, line 209: If it's commented out, can we delete it?

static/js/sheets/SheetContent.jsx
```diff
@@ -0,0 +1,520 @@
+import Component from "react-class";
```
yitzhakc, line 1: When moving the logic here from `static/js/Sheet.jsx` did you copy-paste or use `git cp`? That would preserve the file's history, as opposed to showing that you wrote everything with the current commit date :-)

yitzhakc (changes requested):

static/js/sheets/SheetContent.jsx
stevekaplan123: I didn't know about 'git cp'.  I'm wondering if I should try reverting the commit that moved the SheetContent component into its own file and then using 'git cp'.  I'm worried though that this might cause a mess downstream in the other branches that are dependent on SheetContent.jsx.  What do you think?

yitzhakc (approved):

PR #2076, opened 2024-10-27
Version with alt title

YishaiGlasner:
## Description
Versions should use only primary title of indexes and not alt titles (alts are not unique).

## Code Changes
1. Get the Index from Version by Index().load() rather than library.get_index
2. When saving version from a csv file, check for index with the primary title (rather than getting it by Ref), and if doesn't exist raise an InputError with clear message.

### DIFF
--- a/sefaria/export.py
+++ b/sefaria/export.py
@@ -726,7 +726,11 @@ def _import_versions_from_csv(rows, columns, user_id):
     from sefaria.tracker import modify_bulk_text
 
     index_title = rows[0][columns[0]]  # assume the same index title for all
-    index_node = Ref(index_title).index_node
+    index = Index().load({'title': index_title})
+    if index:
+        index_node = index.nodes
+    else:
+        raise InputError(f'No book with primary title "{index_title}"')
 
 
     action = "edit"
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -1412,7 +1412,7 @@ class Version(AbstractTextRecord, abst.AbstractMongoRecord, AbstractSchemaConten
         pass
 
     def get_index(self):
-        return library.get_index(self.title)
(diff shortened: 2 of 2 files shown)


sefaria/export.py
```diff
@@ -726,7 +726,11 @@ def _import_versions_from_csv(rows, columns, user_id):
     from sefaria.tracker import modify_bulk_text
 
     index_title = rows[0][columns[0]]  # assume the same index title for all
-    index_node = Ref(index_title).index_node
+    index = Index().load({'title': index_title})
+    if index:
+        index_node = index.nodes
```
yitzhakc, line 731: If we're getting `index.nodes`, which is plural, I would have assumed that it would be assigned to `index_nodes`, not `index_node`. Or does `index.nodes` returns only a single node?

sefaria/export.py
```diff
@@ -726,7 +726,11 @@ def _import_versions_from_csv(rows, columns, user_id):
     from sefaria.tracker import modify_bulk_text
 
     index_title = rows[0][columns[0]]  # assume the same index title for all
-    index_node = Ref(index_title).index_node
+    index = Index().load({'title': index_title})
```
yitzhakc, line 729: Just to make sure I understand - when passing a title to `Ref(..)` it will try to come up with any match, including non-unique ones, based on the title, making it inaccurate?

sefaria/export.py
YishaiGlasner: it returns the root node which must be single. i don't know why its name is plural, i guess it has some history.

sefaria/export.py
YishaiGlasner: Right, for example Ref('Gen') will work.

yitzhakc (dismissed):

yitzhakc (approved):

PR #2082, opened 2024-10-29
chore: move files in shared directory to common and delete shared dir

stevekaplan123:
## Description
Moves files in shared directory to common directory, and deletes shared directory


saengel (approved):
Thank you for doing this!

PR #2083, opened 2024-10-29
fix(profile): fix broken youtube link on profile

akiva10b:
## Description
Youtube profile does not add prefix to link properly

## Code Changes
add youtube url as prefix to handler


saengel (approved):
Thank you for pushing this through!

PR #2119, opened 2024-11-20
[Products Page] Gray Developer Box

saengel:
## Description
Adding the Gray Developer Box (previously commented out as requested) to the products page. 

## Code Changes
All in `static/js/StaticPages.jsx`:
1. Switching `<p>` to `<div>` to better conform to style spec
2. Uncommenting the component
3. Switching the urls from `example.com` to https://developers.sefaria.org/docs/powered-by-sefaria

### DIFF
--- a/static/js/StaticPages.jsx
+++ b/static/js/StaticPages.jsx
@@ -3140,19 +3140,19 @@ const JobsPage = memo(() => {
 const DevBox = () => {
     return (
       <div className='productsDevBox'>
-        <p className='productsDevHeader'>
+        <div className='productsDevHeader'>
             <InterfaceText text={{en: "Powered by Sefaria" , he:"פרויקטים מכח ספריא" }} />
-        </p>
-        <p>
+        </div>
+        <div>
             <InterfaceText>  
                 <HebrewText>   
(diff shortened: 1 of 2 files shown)


static/js/StaticPages.jsx
```diff
@@ -3140,19 +3140,19 @@ const JobsPage = memo(() => {
 const DevBox = () => {
     return (
       <div className='productsDevBox'>
-        <p className='productsDevHeader'>
+        <div className='productsDevHeader'>
```
yitzhakc, line 3143: Please note that we have (some small) styling for [`.productsDevBox p`](https://github.com/Sefaria/Sefaria-Project/blob/0a4c5722b7aae9934c9c8d5fa774c55e7675a397/static/css/s2.css#L13547), so just double check the visuals before and after, and change as necessary.

saengel: Will do!

saengel: Was extra, thanks for catching. Pushing a fix that deletes it.

edamboritz:
Please consider making the devbox an `<aside>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) to make the html slightly more semantic?

saengel:
> Please consider making the devbox an `<aside>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) to make the html slightly more semantic?
Great, done!

PR #2121, opened 2024-11-21
Naming Clean Up

saengel:
## Description
- The goal of this PR is to clean up the components `<Module/>`, `<Modules/>` and `<ModuleTitles/>` to be more specific in their naming, to all be prefaced with `Sidebar` since they are all related to the Sidebar. 
- The props for `<NavSidebar />` were also renamed from `modules` to `sidebarModules`

## Code Changes
- In all of the files there are changes, the aforementioned modules are renamed, and import statements are adjusted. 

## Notes
- This was tested extensively locally, however I'd appreciate if the reviewer was able to pull locally and check just to ensure nothing was missed accidentally.

### DIFF
--- a/static/js/CalendarsPage.jsx
+++ b/static/js/CalendarsPage.jsx
@@ -6,7 +6,7 @@ import React, { useState } from 'react';
 import classNames  from 'classnames';
 import Sefaria  from './sefaria/sefaria';
 import $  from './sefaria/sefariaJquery';
-import { NavSidebar, Modules }from './NavSidebar';
+import { NavSidebar, SidebarModules }from './NavSidebar';
 import Footer  from './Footer';
 import Component from 'react-class';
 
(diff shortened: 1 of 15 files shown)


yitzhakc (approved):

static/js/CalendarsPage.jsx
```diff
@@ -28,7 +28,7 @@ const CalendarsPage = ({multiPanel, initialWidth}) => {
   const weeklyListings  = makeListings(weeklyCalendars);
 
   const about = multiPanel ? null :
-    <Modules type={"AboutLearningSchedules"} />
```
yitzhakc, line 31: Looks like the `import` statement was missed here; you use `SidebarModules` without importing it, and didn't remove the old `Modules`.

yitzhakc (changes requested):

static/js/CalendarsPage.jsx
saengel: Great catch, thank you *so* much!

yitzhakc (approved):

PR #2135, opened 2024-12-01
Chore/sc 28761/super about fit and finish page title

stevekaplan123:
## Description
Move position of title for all super about pages.

## Code Changes
1.  This required small modificiations in each of the super about pages.  Previously, the title DID show up in each super about page in mobile, so I simply removed the code that only let the header be rendered in mobile.
2. Lots of CSS changes.

## Notes
Perhaps the mobileAboutHeader should be re-named to aboutHeader since it now shows both on mobile and desktop.  What do you think?


templates/metrics.html
```diff
@@ -9,18 +9,16 @@
 {% block content %}
 <div class="superAbout">
     {% if not request.user_agent.is_mobile %}
-        {% include '_sidebar.html' with whichPage='metrics' title="Metrics" heTitle="מדדים" %}
+        {% include '_sidebar.html' with whichPage='metrics' %}
     {% endif %}
```
yitzhakc, line 13: Have you removed the `title` and `heTitle` from all the instances of `{% include '_sidebar.html %}`?

If so, maybe remove the rendered `<h1>` divs from `templates/_sidebar.html`

templates/static/en/about.html
```diff
@@ -19,13 +19,11 @@
 {% block content %}
 <div class="superAbout">
     {% if not request.user_agent.is_mobile %}
-        {% include '_sidebar.html' with whichPage='about' title="What is Sefaria?" heTitle="מהי ספריא" %}
+        {% include '_sidebar.html' with whichPage='about' %}
```
yitzhakc, line 22: Note that on this page the `<header>` (line 26) only has an English translation, not a Hebrew one. Can you confirm that it does/does not render the Hebrew header please?

templates/metrics.html
```diff
@@ -9,18 +9,16 @@
...
     <div id="metricsPage" class="biReady container static">
 
 
         <div class="inner">
-            {% if request.user_agent.is_mobile %}
                 <h1 class="mobileAboutHeader">
```
yitzhakc, line 18: @stevekaplan123 
Regarding your comment in the description, I would agree that if we're now showing the `aboutHeader` on non-mobile as well, then we should change the class name :-)

yitzhakc (changes requested):

templates/metrics.html
stevekaplan123: No it still displays title and heTitle for mobile.

templates/static/en/about.html
stevekaplan123: It doesn't render in Hebrew.  When you switch to a Hebrew interface language, /he/about.html loads instead.

static/css/static.css
```diff
@@ -39,6 +39,22 @@
...
+.interface-english h1.aboutHeader {
+  padding-top: 2px;
+}
+#aboutSidebar.mobileAbout h1, h1.aboutHeader {
+  --english-font: var(--english-sans-serif-font-family);
+  --hebrew-font: var(--hebrew-sans-serif-font-family);
+  font-style: normal;
+  font-weight: 500;
+  font-size: 22px;
+  text-align: left;
```
YishaiGlasner, line 54: shouldn't this be start (for left in english and right in hebrew)?

static/css/static.css
```diff
@@ -3538,7 +3520,16 @@ form.globalUpdateForm + div.notificationsList {
       margin-right: 20px;
     }
 }
+.interface-hebrew h1.aboutHeader, .interface-hebrew #aboutSidebar h1 {
+  text-align: right;
```
YishaiGlasner, line 3524: again, i thinkg the right way here is to be start, and something general that switches the direction in the hebrew interface

static/css/static.css
```diff
@@ -3538,7 +3520,16 @@ form.globalUpdateForm + div.notificationsList {
...
+}
+nav#aboutSidebar h1 {
+  display: none;
```
YishaiGlasner, line 3527: i can't understand it. most of the times when we have display none, i think it's self evident (like hebrew/english staff). here i ask myself why we have it if it isn't displayed.

stevekaplan123: It's a good question.  The CSS could be clearer but I'm not sure the best way.  The header in question only appears on the mobile about menu page.

YishaiGlasner (approved):

templates/_sidebar.html
```diff
@@ -1,8 +1,10 @@
 {% load i18n %}
 
     <nav id="aboutSidebar">
-        <h1 class="int-en">{{title}}</h1>
-        <h1 class="int-he">{{heTitle}}</h1>
+        {% if mobile %}
```
yitzhakc, line 4: I like your solution to do it as part of the template; I see that the `mobile` variable is passed as part of the template `include` in `mobile-about-menu.html:8`. But why do we "manually" set it, and not use `if request.user_agent.is_mobile` as we do in other places?

yitzhakc (changes requested):

YishaiGlasner (dismissed):

yitzhakc (approved):

PR #2137, opened 2024-12-02
Topic landing tod

nsantacruz:
## Description
Adds Topic of the Day module to topic landing page

## Code Changes
- Adds TopicOfTheDayManager.get_topic_of_the_day(). This is a helper function to get the current topic of the day. 
- Adds pytest-django pip package which makes it easier to test django models with pytest. Now any django model objects created in a pytest will be auto-deleted at the end of the test which makes cleanup much easier
- Add topic of the day API
- Factors out <ImageWithAltText /> from <ImageWithCaption />. This makes it easy to render an image with given alt text.


reader/views.py
```diff
@@ -3250,6 +3250,20 @@ def topic_pool_api(request, pool_name):
     return jsonResponse(response, callback=request.GET.get("callback", None))
 
 
+@catch_error_as_json
+def featured_topic_api(request):
```
yonadavGit, line 3254: why enfpoing is featured_topic and not topic_of the day?

reader/views.py
```diff
@@ -3250,6 +3250,20 @@ def topic_pool_api(request, pool_name):
...
+    from django_topics.models import TopicOfTheDay
+
+    lang = request.GET.get("lang")
+    cb = request.GET.get("callback", None)
```
yonadavGit, line 3258: why is the callback important? can we remove it to make the function simpler?

static/js/sefaria/strings.js
```diff
@@ -114,6 +114,8 @@ const Strings = {
     // Topic Images
     "Illustrative image" : "תמונה להמחשה",
 
+    // Topic Landing Page
+    "Featured Topic": "Featured Topic",
```
yonadavGit, line 118: For now Hebrew can be `הנושא היומי` I guess

nsantacruz: I'm waiting to get an "official" translation from Mickey

reader/views.py
nsantacruz: Mickey has renamed this feature "Featured Topic"

django_topics/models/featured_topic.py
```diff
@@ -0,0 +1,62 @@
...
+    class Meta:
+        unique_together = ('topic', 'start_date')
+        verbose_name = "Landing Page - Featured Topic"
+        verbose_name_plural = "Landing Page - Featured Topic"
+
+    def __str__(self):
+        return f"{self.topic.slug} ({self.start_date})"
+
+
+class FeaturedTopicEnglish(TopicOfTheDay):
```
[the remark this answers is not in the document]
nsantacruz: I renamed every variable and class name to be based on the new name of the feature "Featured Topic" EXCEPT for the main django model which I kept as "TopicOfTheDay". This is unfortunate but this is why:
- TopicOfTheDay has already been merged into master
- Therefore, all cauldrons (or most) are dependent on this model
- TopicOfTheDay directly corresponds to a table in Postgres by the same name
- If we rename the model, it will rename the table which will break all cauldrons
- I want to wait to do this until my code is in master.

PR #2141, opened 2024-12-03
Update subscribe endpoint for multiple newsletter subscriptions

relyks:
## Description
Sefaria currently provides a backend API for subscribing to newsletters. The API does not currently support subscribing to multiple and different newsletter lists that Sefaria provides. This PR gives that ability. A code update to Salesforce will also be necessary to support allowing multiple newsletter.

## Code Changes
* Adds a new payload item as a list of strings for the `api/subscribe` endpoint to take other newsletter lists besides the default included for every user. Either none or multiple newsletters can be sent. Duplicates and invalid mailing lists are ignored.
* Adds a new `api/newsletter_mailing_lists` endpoint to show the available newsletter mailing lists a user can be susbscribed to. The management of these lists are done through Salesforce/ActiveCampaign. The endpoint will be used by Sefaria's CMS to sync available lists so specific lists can be chosen and for engineers to get specific values for lists

## Further considerations
* Making API requests to Salesforce can take a long time. Those API requests may be better served asynchronously through a job queue or some other method, since the user does not need to be totally aware of the status for the subscription process and they do not need to be subscribed immediately.

### DIFF
--- a/sefaria/helper/crm/crm_mediator.py
+++ b/sefaria/helper/crm/crm_mediator.py
@@ -20,7 +20,7 @@ class CrmMediator:
         except:
             return False
 
-    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en", mailing_lists=[]):
+    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en", mailing_lists=None):
         return self._crm_connection.subscribe_to_lists(email, first_name, last_name, educator, lang, mailing_lists)
 
     def sync_sustainers(self):
--- a/sefaria/helper/crm/crm_mediator.py
+++ b/sefaria/helper/crm/crm_mediator.py
@@ -20,8 +20,8 @@ class CrmMediator:
         except:
             return False
 
-    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en"):
-        return self._crm_connection.subscribe_to_lists(email, first_name, last_name, educator, lang)
+    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en", mailing_lists=[]):
+        return self._crm_connection.subscribe_to_lists(email, first_name, last_name, educator, lang, mailing_lists)
 
     def sync_sustainers(self):
         current_sustainers = CrmInfoStore.get_current_sustainers()
@@ -71,3 +71,6 @@ class CrmMediator:
             CrmInfoStore.save_crm_id(crm_id, email)
         else:
             return False
+
+    def get_available_lists(self):
+        return self._crm_connection.get_available_lists()
--- a/sefaria/helper/crm/salesforce.py
+++ b/sefaria/helper/crm/salesforce.py
@@ -6,8 +6,10 @@ import json
 from sefaria.helper.crm.crm_connection_manager import CrmConnectionManager
 from sefaria import settings as sls
 
-from typing import Any, Optional, List
+from typing import Any, Optional
 
+class SalesforceNewsletterListRetrievalError(Exception):
+    pass
 
 class SalesforceConnectionManager(CrmConnectionManager):
     def __init__(self):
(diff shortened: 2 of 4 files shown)


sefaria/views.py
```diff
@@ -210,9 +210,12 @@ def subscribe_sefaria_newsletter(request, email, first_name, last_name):
     body = json.loads(request.body)
     language = body.get("language", "")
     educator = body.get("educator", False)
+    mailing_lists = body.get("lists", [])
```
nsantacruz, line 213: @relyks what happens when lists is empty?

sefaria/helper/crm/crm_mediator.py
```diff
@@ -20,8 +20,8 @@ def create_crm_user(self, email, first_name, last_name, lang="en", educator=Fals
         except:
             return False
 
-    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en"):
-        return self._crm_connection.subscribe_to_lists(email, first_name, last_name, educator, lang)
+    def subscribe_to_lists(self, email, first_name, last_name, educator=False, lang="en", mailing_lists=[]):
```
nsantacruz, line 23: Don't use empty list as a default parameter in Python. This can cause strange bugs. The standard is to use `None`.

sefaria/helper/crm/salesforce.py
```diff
@@ -6,6 +6,9 @@
 from sefaria.helper.crm.crm_connection_manager import CrmConnectionManager
 from sefaria import settings as sls
 
+from typing import Any, Optional, List
```
nsantacruz, line 9: Use of typing is highly appreciated! Note, `List` is no longer necessary to import in Python 3.9 which we now use on prod. use the built-in `list` type instead.

sefaria/helper/crm/salesforce.py
```diff
@@ -120,8 +123,18 @@ def find_crm_id(self, email=None):
...
+    def subscribe_to_lists(
+        self, 
+        email: str, 
+        first_name: Optional[str] = None, 
+        last_name: Optional[str] = None, 
+        lang: str = "en", 
+        educator: bool = False,
+        mailing_lists: Optional[List[str]] = None) -> Any:
+
+        if mailing_lists is None:
```
nsantacruz, line 135: Can be more simply written as `mailing_lists = mailing_lists or []`. This is fairly common as a first line in a function given Python's buggy behavior with lists as default parameters.

sefaria/helper/crm/salesforce.py
```diff
@@ -145,3 +159,14 @@ def subscribe_to_lists(self, email, first_name=None, last_name=None, lang="en",
...
+
+    def get_available_lists(self) -> List[str]:
+        try:
+            resource_prefix = f"services/data/v{self.version}/query"
+            endpoint = f"{sls.SALESFORCE_BASE_URL}/{resource_prefix}/"
+            response = self.get(endpoint + "?q=SELECT+Subscriptions__c+FROM+AC_to_SF_List_Mapping__mdt")
+            records = response.json()["records"]
+            return [record["Subscriptions__c"] for record in records]
+        except:
+            print("Unable to retrieve newsletter mailing lists from Salesforce CRM")
```
nsantacruz, line 171: It's not helpful to catch this error here. better to throw a custom error class that can be caught at the API level. There you can return an appropriate error code response (probably 500) and an error message in the API response.

sefaria/views.py
relyks: Right now, when someone sends an email address to Salesforce that does not already exist, Salesforce automatically subscribes it to the default lists: "Master," "General Updates" (or the one for Hebrew), and "Educator Updates" (if the user is an educator). If the user's email address already exists in Salesforce, the flow in Salesforce does not resubscribe the email address to those lists. When you provide additional newsletter mailing lists, the flow in Salesforce will subscribe the email address to those lists, along with the default ones, if the email address does not already exist. However, if we pass the email address with any of those default newsletter lists as additional ones, Salesforce will resubscribe the email address to those lists.

I will write this behavior as comments for documentation purposes, though I should probably record it elsewhere too.

sefaria/helper/crm/crm_mediator.py
relyks: Yeah, thanks, good catch 🙂 I realized that in the other spot of my code that Python evaluates default arguments when the function is first defined and not when the function is called again other times.

sefaria/helper/crm/salesforce.py
relyks: Agreed 😃

nsantacruz (approved):

PR #2143, opened 2024-12-03
Topic Salad and Rainbow of Truth

yonadavGit:
## Description
This PR contains an implementation of Toipc Salad to display random topics and array them in 5 lines under the search bar in the topic landing page, as well as an implementation of the Rainbow decorative element to wrap the salad


static/js/TopicLandingPage/TopicLandingSearch.jsx
```diff
@@ -109,7 +109,10 @@ export const TopicLandingSearch = ({openTopic, numOfTopics}) => {
                 renderItems={renderItems.bind(null, openTopic)}
                 containerClassString="topic-landing-search-container"
                 dropdownMenuClassString="topic-landing-search-dropdown"
-                renderInput={renderInput.bind(null, openTopic, numOfTopics)}/>
+                renderInput={renderInput.bind(null, openTopic, numOfTopics)}
```
nsantacruz, line 112: remove comment

static/js/TopicLandingPage/TopicsLandingPage.jsx
```diff
@@ -12,8 +14,11 @@ export const TopicsLandingPage = ({openTopic}) => {
         <div className="readerNavMenu" key="0">
             <div className="content">
                 <div className="sidebarLayout">
-                    <div className="contentInner mainColumn">
+                    <div className="contentInner mainColumn topic-lading-page-content">
                         <TopicLandingSearch openTopic={openTopic} numOfTopics={5000}/>
+                        <RainbowLine rainbowClassname={"topic-landing-upper-rainbow"}/>
```
nsantacruz, line 19: Include RainbowLine in TopicSalad component. They are logically meant to be rendered together.

static/js/TopicLandingPage/TopicSalad.jsx
```diff
@@ -0,0 +1,42 @@
...
+    const renderSaladItem = (item) => {
+        return(<a href={`/topics/${item.slug}`} className="topic-salad-item">
+                <InterfaceText text={{en: item.text, he: item.text}}/>
+                </a>)
+    }
+
+    const fetchRandomSaladItems = async () => {
+        const poolName = Sefaria.getLangSpecificTopicPoolName('general', lang);
+        const topics = await Sefaria.getTopicsByPool(poolName, 50);
+        const saladItems = topics.map(topic=>({slug: topic.slug, text: topic.primaryTitle[lang]}))
```
nsantacruz, line 22: lang is not needed here since you're using InterfaceText to render the items. Instead make the object `{slug: topic.slug, text: topic.primaryTitle}` and the when rendering the item `<InterfaceText text={item.text} />`. This is much cleaner. 

Note, I understand the reason this was done was to allow WordSalad to replace spaces with non-breaking spaces. see comment below on using `nowrap` in CSS.

static/js/sefaria/sefaria.js
```diff
@@ -1964,6 +1964,9 @@ _media: {},
         store: this._TopicsByPool
     });
   },
+    getLangSpecificTopicPoolName: function(poolName, lang){
+      return `${poolName}_${lang}`
```
nsantacruz, line 1968: This function should not take lang as a parameter and instead should use Sefaria.interfaceLang internally. This is reasonable since this function is located in sefaria.js and pool name is supposed to be tied to the interface language.

static/js/WordSalad.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const renderItemWithSpacesForBreaks = (item)=>{
+    // needed in order for css to recognize space after each item to potentially break the line at this space.
+    const spacedElement = <span>{renderItem(item)} </span>
+    return spacedElement
+  }
+
+
+  return (
+    <div className="salad-container" style={{ '--num-lines': numLines }}>
+      {salad.map((item, index) => renderItemWithSpacesForBreaks(item))}
```
nsantacruz, line 35: index isn't needed here. A more succinct way to write this is `salad.map(renderItemWithSpacesForBreaks)` through the magic of JS

static/js/WordSalad.jsx
```diff
@@ -0,0 +1,38 @@
...
+ *                                  This is used in the CSS to control layout.
+ * @param {Array} props.salad - An array of objects representing the text items to display.
+ *                              Each object must have a `text` property.
+ * @param {Function} props.renderItem - A function that receives an item from the `salad` array
+ *                                      and returns a React element to render it.
+ */
+
+  const nonWhitespaceInvisibleChar = '\u00A0'
+  // replace the normal space with the HTML space char, in order for css to not break line mid-item.
+  salad = salad.map(saladItem => ({
```
nsantacruz, line 21: Instead of doing this which requires the objects to have a `.text` field (which is problematic for bilingual text) you should be able to wrap each element in a span with the following CSS:

```css
.salad-item-no-wrap {
    white-space: nowrap;
}
```

static/js/WordSalad.jsx
```diff
@@ -0,0 +1,38 @@
...
+    ...saladItem,
+    text: saladItem.text.replace(/ /g, nonWhitespaceInvisibleChar),
+  }));
+
+  const renderItemWithSpacesForBreaks = (item)=>{
+    // needed in order for css to recognize space after each item to potentially break the line at this space.
+    const spacedElement = <span>{renderItem(item)} </span>
```
nsantacruz, line 28: use `.salad-item-no-wrap` css class here

static/js/WordSalad.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+
+export const WordSalad = ({ numLines, salad, renderItem }) => {
+/**
+ * This component renders a collection of text items, styled to fit within a specified
```
nsantacruz, line 6: This is good documentation. Avoid explaining how the component works and instead explain how to use it. E.g. don't write "This is used in the CSS to control layout." because users of the component don't need to know this.

Also, remove the requirement that each item has a text property (see comment below on using `nowrap` in CSS)

static/css/common.css
```diff
@@ -753,4 +753,27 @@ a:active {
...
+}
+.categoryColorLineRainbow {
+    top: 60px;
+    left: 0;
+    width: 100%;
+    height: 4px;
+    z-index: 101;
+    background: linear-gradient(
+    90deg,
+    var(--tanakh-teal) 0% 10%,
```
nsantacruz, line 765: I'm realizing now, do these variables exist in common.css? Does this work on the static page that doesn't have s2.css?

static/css/s2.css
```diff
@@ -4843,6 +4843,9 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
   justify-content: center;
   align-items: center;
 }
+.readerNavMenu .content .contentInner.topic-lading-page-content{
+  width: auto;
```
nsantacruz, line 4847: I actually think the width needs to be 843px as per the figma. I confirmed this with Mickey

static/css/common.css
yonadavGit: it works, seems like they are global
image

PR #2151, opened 2024-12-05
Secondary topic image

nsantacruz:
## Description
Adds ability to upload secondary topic images through the topic admin editor

## Code Changes
- add `_api/topics/images/secondary` API for uploading and deleting secondary topic images.
- refactor existing topic image API so that it handles files in a more standard way (using the file object that's uploaded as opposed to using the base64 string)
- Pull topic editor code from Misc.jsx into TopicEditor.jsx where it belongs
- Allow ImageCropper to take a widthHeightRatio prop to control the ratio of the cropping square.

### DIFF
--- a/static/js/AdminEditor.jsx
+++ b/static/js/AdminEditor.jsx
@@ -123,7 +123,7 @@ const validateMarkdownLinks = async (input) => {
     return true;
 }
 
-const AdminEditor = ({title, data, close, catMenu, pictureUploader, updateData, savingStatus,
+const AdminEditor = ({title, data, close, catMenu, pictureUploader, secondaryPictureCropper, updateData, savingStatus,
                              validate, deleteObj, items = [], isNew = true,
                              extras = [], path = []}) => {
     const [validatingLinks, setValidatingLinks] = useState(false);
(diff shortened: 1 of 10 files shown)


static/js/AdminEditor.jsx
```diff
@@ -213,6 +213,8 @@ const AdminEditor = ({title, data, close, catMenu, pictureUploader, updateData,
                             return catMenu;
                         } else if (x === "Picture Uploader") {
                             return pictureUploader;
+                        } else if (x === "Secondary Picture Cropper") {
```
yonadavGit, line 216: maybe it's better to replace these in-render "if" statements with a map data-structure to map items to the appropriate elements to render?

static/js/ProfilePic.jsx
```diff
@@ -106,6 +106,7 @@ export class ProfilePic extends Component {
                     src={this.state.fileToCropSrc}
                     onClose={this.onCloseCropper}
                     onSave={this.upload}
+                    widthHeightRatio={1}
```
yonadavGit, line 109: interesting that you chose to make it general even though currently the value used is still 1:

nsantacruz: In TopicEditor.jsx I set the value to 4/3 so it needs to be a prop.

yonadavGit (approved):

PR #2152, opened 2024-12-05
chore(docker ): fix docker compose node implementation

akiva10b:


dev_docker/node_entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+npm install
+npm run setup
+npm run build-client
+npm run build
```
nsantacruz, line 5: only one build command is needed. should be able to remove `npm run build-client`

docker-compose.canned.yml
nsantacruz, line 1: does this need to be in the root? or can it be in a place that's more organized?

docker-compose.yml
```diff
@@ -29,10 +39,26 @@ services:
     working_dir: /app
     volumes:
       - .:/app
-    command: npm start
+    command: sh dev_docker/node_entrypoint.sh
     ports:
       - "3000:3000"
 
+  # es8:
```
nsantacruz, line 46: why this commented out section? can it be deleted?

docker-compose.yml
nsantacruz, line 1: what is the difference between this and the canned docker compose? should it be clarified in a comment?

sefaria/local_settings_docker.py
nsantacruz, line 1: why is this file different than any of the existing local settings files? especially the local settings CI file?

docker-compose.yml
akiva10b: canned uses a canned database, as opposed to a db that requires a dump. That should have a comment - agreed

.github/workflows/standalone-db-deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+            --platform managed \
+            --region ${{ env.GCP_REGION }} \
+            --port 27017 \
+            --min-instances 1 \
+            --max-instances 1 \
+            --memory 2Gi \
+            --cpu 2 \
+            --allow-unauthenticated \
+            --project ${{ secrets.DEV_PROJECT }} \
+            --set-env-vars="MONGO_INITDB_ROOT_USERNAME=${{ secrets.MONGO_ROOT_USERNAME }},MONGO_INITDB_ROOT_PASSWORD=${{ secrets.MONGO_ROOT_PASSWORD }}"
```
BrendanGalloway, line 65: why are we using cloudrun?
Where will this be hosted?  on a compute instance?  Why not in the kubernetes cluster?

.github/workflows/standalone-db-deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Get Service URL
+        run: |
+          echo "Service URL: $(gcloud run services describe ${{ env.CONTAINER_NAME }} \
+            --region ${{ env.GCP_REGION }} \
+            --project ${{ secrets.DEV_PROJECT }} \
+            --format 'value(status.url)')"
```
BrendanGalloway, line 72: is the intention that this DB be accessed by codespaces, and will therefore be publicly accessible? 

Will each codespace need its own instance of the DB? Would multiple codespaces affect each other's data with schema changes etc?

docker-compose.canned.yml
```diff
@@ -0,0 +1,58 @@
...
+
+services:
+  web:
+    build:
+      context: .
+      dockerfile: dev_docker/Dockerfile
+    ports:
+      - "8000:8000"
+    volumes:
+      - .:/app
```
BrendanGalloway, line 13: This dockerfile copies the whole app folder during build, then remounts the local app folder over it?  Why not copy only the requirements file if the intention is just to do the pip install at build time rather than at run time?

.github/workflows/standalone-db-deploy.yml
akiva10b: codespaces will not affect others data, we will only ever need one database for the codespaces - it should be publicly available

docker-compose.canned.yml
akiva10b: very good call

.github/workflows/standalone-db-deploy.yml
BrendanGalloway: Is there any risk to exposing the DB?  Does codespaces support securing the credentials to access the DB?

PR #2161, opened 2024-12-09
Fix display of Strapi banners on static pages

relyks:
This is a temporary fix for the display of Strapi banners to fix the spacing in the header when the banners are closed. The `setContainerMode` code for adjusting the class on `<body>` used to work, but is currently not working. That code will need to be refactored in the future.


edamboritz (approved):
I'd like to make sure we fix the actual root cause of this.

Did we validate that this doesn't negatively affect any other areas?

PR #2163, opened 2024-12-10
Feature/sc 27331/new sidebar component2

stevekaplan123:
## Description
The core of this PR is the SheetContentSidebar component which is used by the Sheets Viewer and will eventually be used by the Sheets Editor as well.  It displays profile information, a profile picture, a follow button, and a list of the sheet maker's collections that the sheet is in.

## Code Changes
The SheetContentSidebar works by loading the profile data through Sefaria.profileAPI and then passing the profile object to a new component SheetProfileInfo, which uses the profile data to display the user's education, location, organization, and followers information.  Modification was made to UserBackground so that it properly displays the HTML of the profile bio ("about me").  The profile bio is the only part of the profile that is in HTML, which I did not realize, so that I previously rendered everything in UserBackground as plain text.  Finally, a new component SheetCollectionsList displays the collections that the sheet maker put the sheet in.

PR #2164, opened 2024-12-10
Feature/sc 27327/source click should open in new tab2

stevekaplan123:
## Description
This PR implements a small change whereby when a user clicks on a ref in a sheet, it will open in a new tab instead of opening a new panel.

## Code Changes
1. `options` was unused so I removed that
2. `onCitationClick` was no longer necessary to call so I replaced it with a call to `window.open`

## Notes
This should be merged before https://github.com/Sefaria/Sefaria-Project/pull/2163


static/js/sheets/Sheet.jsx
```diff
@@ -68,12 +68,9 @@ class Sheet extends Component {
       }
 
       else if (Sefaria.isRef(path.slice(1))) {
-        e.preventDefault()
-        const currVersions = {en: params.get("ven"), he: params.get("vhe")};
-        const options = {showHighlight: path.slice(1).indexOf("-") !== -1};   // showHighlight when ref is ranged
-        this.props.onCitationClick(path.slice(1), `Sheet ${this.props.sheetID}`, true, currVersions)
+        e.preventDefault();
+        window.open(target.href);
```
yitzhakc, line 72: It looks like we have a utility function for this, `sefaria.utils` called `openInNewTab`. Maybe we could utilize it here?

static/js/sheets/SheetContent.jsx
```diff
@@ -5,8 +5,7 @@ import Component from 'react-class'
 import $  from '../sefaria/sefariaJquery';
 import Sefaria  from '../sefaria/sefaria';
 import {
-  CollectionStatement,
-  InterfaceText, ProfilePic, SheetAuthorStatement, SheetMetaDataBox, SheetTitle,
+  InterfaceText, SheetMetaDataBox,
```
yitzhakc, line 8: It's not clear to me why these modifications are being made here? Should they actually be in some upstream (or downstream) branch?

yitzhakc (changes requested):

static/js/sheets/SheetContent.jsx
stevekaplan123: You're right.  These imports should be in a different PR.  Modified that PR and merged it in and now they are gone.

static/js/sheets/Sheet.jsx
stevekaplan123: Makes sense.  I replaced it with the util function

yitzhakc (changes requested):

PR #2165, opened 2024-12-10
Fix API and ensure no broken topics

saengel:
## Description
We had recently corrected a bug throughout the API which incorrectly handled boolean parameters being passed as 0s or 1s. They were handled as strings, and therefore constantly evaluating as `true` and causing unexpected behavior, we made an effort to make sure they were all properly being passed as ints. 

The issue with int casting, is that one has to handle cases where no param is passed since optional. 

This PR casts `useTextFamily` param results to an int if present, and if not returns `None`. This ensures continued topic source functionality. 

## Code Changes
1. `sefaria/views.py` - Cast the `useTextFamily` param results to an int, and improve default param logic of underlying function.

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -401,13 +401,13 @@ def title_regex_api(request, titles, json_response=True):
         return jsonResponse({"error": "Unsupported HTTP method."}) if json_response else {"error": "Unsupported HTTP method."}
 
 
-def bundle_many_texts(refs, useTextFamily=None, as_sized_string=False, min_char=None, max_char=None, translation_language_preference=None, english_version=None, hebrew_version=None):
+def bundle_many_texts(refs, use_text_family=False, as_sized_string=False, min_char=None, max_char=None, translation_language_preference=None, english_version=None, hebrew_version=None):
     res = {}
     for tref in refs:
         try:
             oref = model.Ref(tref)
             lang = "he" if has_hebrew(tref) else "en"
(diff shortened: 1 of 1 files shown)


sefaria/views.py
```diff
@@ -401,13 +401,13 @@ def title_regex_api(request, titles, json_response=True):
         return jsonResponse({"error": "Unsupported HTTP method."}) if json_response else {"error": "Unsupported HTTP method."}
 
 
-def bundle_many_texts(refs, useTextFamily=False, as_sized_string=False, min_char=None, max_char=None, translation_language_preference=None, english_version=None, hebrew_version=None):
+def bundle_many_texts(refs, useTextFamily=0, as_sized_string=False, min_char=None, max_char=None, translation_language_preference=None, english_version=None, hebrew_version=None):
```
nsantacruz, line 404: I think we need to find another solution because it's not conventional to have a python function take 1 and 0 in the place of a boolean. This is especially true when there's another parameter (`as_sized_string`) that is a boolean.

saengel: Thanks @Noah - 

I just pushed a revision, where `None` is still the default value, but inside the `if` it still checks if `useTextFamily == 1` as opposed to before where it just checked if it was not `None` (thus breaking the expected boolean functionality from the API, passing 0/1 as a boolean). 

What do you think?

yitzhakc (approved):

PR #2180, opened 2024-12-15
Feature/sc 27385/publish settings2

stevekaplan123:
## Description
This PR allows the user to publish a sheet, or unpublish/unlist it.  I implemented this feature by adding a "Publish" button to SheetOptions, the ellipsis help menu in the Sheets Viewer and Editor.  I also added an "Unpublish" option, but we want "Unpublish" to show up in the SheetOptions drop down menu instead of being its own button.  In addition to the publish/unpublish option, there's also a menu, implemented via component PublishMenu, that appears when one chooses to publish that allows the user to edit the title, summary, and topic tags for the sheet.  Finally, once one has completed the menu and published/unpublished the sheet, the user sees a modal, PublishModal, that displays "Publishing..." and then "Published."  

## Code Changes
1. The PublishModal is implemented pretty much the same way as SaveModal and CopyModal.  It starts publishing immediately when it's first rendered.
2. PublishMenu is implemented using code that already existed in pre-modularization days, but was implemented via the sheets editor's resources panel.
3. Whenever title or summary change, the document is saved via ```[currentDocument[0].children[0], title, summary] // Only re-call effect if value or delay changes```.  This way title and summary are up-to-date when the user tries to publish the sheet.
4.  In the Editor.jsx, I added a state variable, status, which is either 'published' or 'unlisted'.  When the user posts the sheet, some of the state for the editor is set to keep everything up-to-date -- namely the title, summary, and published status.
5.  I removed lastModified as state from Editor.jsx as it wasn't necessary.  Instead, it's possible to access lastModified by getting/setting the Sefaria sheets' cache via Sefaria.sheets._loadSheetByID.
6.  The status of SheetOptions is a bit confusing in this PR.  I was rendering the ellipsis help menu, SheetOptions, in Sheet.jsx and passing it down to the Editor or Viewer.  However, the Editor has a function, postSheet, which posts a sheet and then sets state for the Editor.  I needed to pass this function to SheetOptions so that it could publish the sheet, so I directly render SheetOptions in Editor.jsx (and SheetContent.jsx - the sheet viewer) instead of rendering it in Sheet.jsx and passing it down.


static/js/sheets/SheetModals.jsx
```diff
@@ -196,4 +198,205 @@ const DeleteModal = ({close, sheetID, authorUrl}) => {
...
+    sheet.topics = tags.map(tag => ({
+          asTyped: tag.name,
+          slug: tag.slug,
+        })
+    );
+    if ((isFormValidated())) {
+      publishCallback(true);
+    }
+  }
+  return <div>
```
yitzhakc, line 357: Why not use <>?

static/js/sheets/SheetModals.jsx
```diff
@@ -196,4 +198,205 @@ const DeleteModal = ({close, sheetID, authorUrl}) => {
   return <GenericSheetModal title={<InterfaceText>Deleting...</InterfaceText>} close={() => {}}/>; // don't allow user to close modal while deleting
 }
 
-export { ShareModal, CollectionsModal, AddToSourceSheetModal, CopyModal, SaveModal, GoogleDocExportModal, DeleteModal };
\ No newline at end of file
+const PublishModal = ({close, status, sheetID, postSheet}) => {
```
yitzhakc, line 201: Maybe this component needs to be pulled out of here and get its own file? It's quite heavy with a lot of logic.

static/js/sheets/SheetOptions.jsx
```diff
@@ -131,6 +149,14 @@ const DeleteButton = () => {
               descHe={""}/>
 }
 
+const UnpublishButton = () => {
+  return <DropdownMenuItemWithIcon icon={"/static/icons/unpublish.svg"}
+                                   textEn={'Unpublish'}
+                                   textHe={""}
```
yitzhakc, line 155: These translations exist already? Do we need to pass in empty strings, it seems a bit obtuse.

yitzhakc (changes requested):

yitzhakc (approved):

PR #2186, opened 2024-12-17
Fix topic test dec 2024

nsantacruz:
## Description
Fixes failing topic tests. The tests were failing because Django by default doesn't allow tests to access the database (i.e. the django database, not mongo). In order to allow access, you can either use Django's built-in test system or use `pytest-django`. `pytest-django` more naturally fits into how we are already running tests.

## Code Changes
- add pytest-django pip module
- use pytest-django to allow access to db for fixtures that need it
- create unique test slugs using custom `_ms()` function so that there aren't any conflicts with existing slugs
- fix some typos

## Notes

### DIFF
--- a/sefaria/model/tests/topic_test.py
+++ b/sefaria/model/tests/topic_test.py
@@ -7,8 +7,20 @@ from sefaria.system.exceptions import SluggedMongoRecordMissingError
 from django_topics.models import Topic as DjangoTopic, TopicPool
 from django_topics.models.pool import PoolType
 
+TEST_SLUG_PREFIX = 'this-is-a-test-slug-'
 
-def make_topic(slug):
+
+def _ms(slug_suffix):
+    """
+    ms = make slug. makes full test slug.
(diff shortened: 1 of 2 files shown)


sefaria/model/tests/topic_test.py
```diff
@@ -33,140 +45,150 @@ def clean_links(a):
...
-            str(i): make_topic(str(i)) for i in range(1, 10)
-        },
-        'links': [make_it_link(str(a), str(b), 'is-a') for a, b in isa_links] + [make_rt_link('1', r) for r in trefs]
-    }
-    yield graph
-    for k, v in graph['topics'].items():
-        v.delete()
-    for v in graph['links']:
-        v.delete()
+def topic_graph(django_db_setup, django_db_blocker):
```
yonadavGit, line 65: django_db_setup is never used in this functiom.

yonadavGit (approved):

PR #2190, opened 2024-12-18
Bug/sc 26322/click on commentary in search version name

stevekaplan123:
## Description
From search results, commentaries would not open properly.  First, the version of the commentary would show as the version of the base text in the URL.  Second, the version of the commentary would not get applied in the connections panel.  Both issues are now fixed. 

## Code Changes
The solution is to create a new panel prop `currCommVersions` which is a object whose properties are titles of texts and the corresponding values are the versions we want for each of those texts.  If the Connections Panel has a field for the current book in the Connections Panel, it passes down the version information to TextList which in turn passes currVersions down to TextRange and OpenConnectionTabButton so that (1) TextRange now loads the right text with currVersions and (2)  if the user clicks to open the connection, it opens in a panel with currVersions set properly.


static/js/ReaderApp.jsx
```diff
@@ -1533,11 +1534,17 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
   makePanelWithConnectionsState(panelProps) {
     // in the case of multipanel, create two panels based on panelProps
     let connectionPanel;  // in mobile, connectionPanel will remain undefined
+
+    // version is from commentary; don't try to use the version of the commentary for the base ref!
+    const currIndex = Sefaria.parseRef(panelProps.refs[0])?.index;
+    const currCommVersions = !!currIndex ? {[currIndex]: {...panelProps.currVersions}} : {};
```
YishaiGlasner, line 1540: why not `panelProps.currCommVersions =...` ?

static/js/ReaderApp.jsx
```diff
@@ -1533,11 +1534,17 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
...
+    panelProps.currVersions = {'en': null, 'he': null};
```
YishaiGlasner, line 1541: is this necessary? isn't it done by `OpenConnectionTabButton`'s default currVersions prop?

static/js/ReaderPanel.jsx
```diff
@@ -762,6 +762,7 @@ class ReaderPanel extends Component {
           title={this.currentBook()}
           currentlyVisibleRef={this.state.currentlyVisibleRef}
           currVersions={this.state.currVersions}
+          currCommVersions={this.state.currCommVersions}
```
YishaiGlasner, line 765: shouldn't it be props rather than state?
does it come from props.initialState? if so, i can't see why - it isn't changing in the ReaderPanel lifecycle so i think it should be a prop.

stevekaplan123: That's a great point.  Changing it...

static/js/ReaderApp.jsx
stevekaplan123: The reason is that there's no need to pass `currCommVersions` to the panel if we're in multipanel mode.  It only needs to be handled by the ConnectionsPanel.

static/js/ReaderApp.jsx
stevekaplan123: What do you mean?  Are you suggesting I should just do:
panelProps.currVersions = null or something like that?

stevekaplan123: The thing is though that this is setting the base panel's version to essentially be null so it's not relevant to OpenConnectionTabButton

YishaiGlasner (approved):

PR #2191, opened 2024-12-18
New Topic Page Playwright Tests

yonadavGit:
## Description
New tests for validation of basic Topic Pages functinality
added new users to test database as well as envionment vars with users data. 

## Code Changes

## Notes


e2e-tests/tests/topics.spec.ts
```diff
@@ -8,6 +10,48 @@ test('Go to topic page', async ({ context }) => {
...
+
+test('Check author page', async ({ context }) => {
+  const page = await goToPageWithLang(context, '/topics/jonathan-sacks');
+  await page.getByRole('link', { name: 'Works on Sefaria' }).first().isVisible();
+});
+
+test('Check redirection for sourceless topic', async ({ context }) => {
+  const page = await goToPageWithLang(context, '/topics/Monkey');
+  const expectedUrl = 'search?q=Monkey&tab=sheet&tvar=1&tsort=relevance&stopics_enFilters=Monkey&svar=1&ssort=relevance';
+  await page.waitForTimeout(10000)
```
nsantacruz, line 46: We should try to avoid timeouts at all costs as they either make the tests brittle (when too short) and greatly increase the run time (when too long). Is there no better way to do this? I think you can await an element on the search page to exist.

e2e-tests/tests/topics.spec.ts
```diff
@@ -8,6 +10,48 @@ test('Go to topic page', async ({ context }) => {
...
+  expect(page.url()).toContain(expectedUrl);
+});
+
+test('Check no redirection when user is admin', async ({ context }) => {
+  const page = await goToPageWithUser(context, '/topics/Monkey', testAdminUser);
+  await page.waitForTimeout(10000)
```
nsantacruz, line 52: Same here. Why does the await on line 53 not suffice?

PR #2201, opened 2024-12-23
Feature/sc 30090/confirm rtl changes still work in new sheets

stevekaplan123:
## Description
Merged rtl-client into modularization-main and made changes to DropdownMenu.jsx and s2.css so that both types of dropdown menu work.  

## Code Changes
Change name of props of buttonContent and menuIconComponent  to buttonComponent both in <DropdownMenu> and in all instances of it
Change .buttonClosest into a data attribute data-prevent-close="true"
CSS resolutions
dropdownButton needs to be resolved to dropdownLinks-button (note that it was changed from a button html element to an a)
Note that in rtl-client {children} is directly inside div.dropdownLinks-menu whereas in modularization-main it's enclosed within an additional div.dropdownLinks-options
We've removed the additional enclosing div from modularization main
No default export when importing DropdownMenu.jsx
An additional enclosing div (or component?) should be added to the instances of <DropdownMenu> in the Header.jsx
The instances of <DropdownMenu> in rtl-client are all enclosed with divs that set the positioning style
The CSS styling used in .header .dropdownLinks should be migrated to a class that the newly created enclosing div/component (used in Header.jsx) will be tagged with
The CSS styling on <DropdownMenu>  should be as sparse as possible
Add an additional prop to <DropdownMenu> called positioningClass that will be added to the  list of class -es in the "first" <div> in DropdownMenu and is in charge of the positioning styling of the open menu location
I.e. in rtl instances use dropdownMenu and in Header.jsx use dropdownLinks -- CHANGE ALL THESE CLASS NAMES (readerDropdownMenu vs headerDropdownMenu)
Now that we've removed the <div class='dropdownLinks'> from the DropdownMenu component, we need to prepend the children passed in to <DropdownMenu> instances in Header.jsx with a <div class='dropdownLinks-options'> to properly position the children.
in rtl instances this is already taken care of


static/js/common/DropdownMenu.jsx
```diff
@@ -44,10 +44,10 @@ const DropdownMenuItemWithIcon = ({icon, textEn, textHe}) => {
   );
 }
 
-const DropdownMenu = ({children, buttonComponent}) => {
+const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
     /**
      * buttonComponent is a React component for the opening/closing of a button.
-     * the menu will be closed in click anywhere except in an element with classname 'preventClosing'.
+     * the menu will be closed in click anywhere except in an element where data attribute
```
yitzhakc, line 50: Please finish this sentance :-) (which data attribute, and that it needs to `="true"`

static/js/common/DropdownMenu.jsx
```diff
@@ -44,10 +44,10 @@ const DropdownMenuItemWithIcon = ({icon, textEn, textHe}) => {
   );
 }
 
-const DropdownMenu = ({children, buttonComponent}) => {
+const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
     /**
      * buttonComponent is a React component for the opening/closing of a button.
```
yitzhakc, line 49: Please edit this docstring, adding in an explanation about the `positioningClass`

yitzhakc (changes requested):

yitzhakc (approved):

PR #2204, opened 2024-12-24
feat(profile url): change profile url from 'profile' to 'sheets/profile'.

YishaiGlasner:
changes in ReaderApp (router), urls.py and views.py (for redirection cases).

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -831,14 +831,14 @@ def enable_new_editor(request):
     profile = UserProfile(id=request.user.id)
     profile.update({"uses_new_editor": True, "show_editor_toggle": True})
     profile.save()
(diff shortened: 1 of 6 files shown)


saengel (changes requested):
Overall looks great, thanks @YishaiGlasner. Tested locally and all seemed to work as expected. 

One small thing - I searched the codebase for instances of `/profile` and it seems there are a few places we send people to `/profile` that should now be updated to `/sheets/profile`.

Here's where I found these cases:
1. `Autocomplete.jsx`
2.  `ProfilePic.jsx`
3. `UserProfile.jsx`

Thanks!

saengel (approved):
Thank you!

yitzhakc (approved):

PR #2205, opened 2024-12-24
fix: Short circuit Ref.contains() and avoid call to as_ranged_segment_ref ...

EliezerIsrael:
even in cases where other is not a whole book.

For example:
`assert not Ref("Exodus 6:2").contains(Ref("Exodus 6"))` Can be solved without finding the true extent of the second ref.

## Description
In Ref.contains(), short circuit and avoid call to as_ranged_segment_ref even in cases where other is not a whole book.

## Notes
For example:
`assert not Ref("Exodus 6:2").contains(Ref("Exodus 6"))` Can be solved without finding the true extent of the second ref.

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -4243,9 +4243,9 @@ class Ref(object, metaclass=RefCacheType):
             return self.index_node.is_ancestor_of(other.index_node)
 
         if len(self.sections) > len(other.sections): # other is less specific than self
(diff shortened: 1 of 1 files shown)


EliezerIsrael:
@yitzhakc this fell through the cracks.  Small speed improvement needs a review.

yitzhakc (approved):

yitzhakc (approved):

PR #2210, opened 2025-01-02
[Sheets][Desktop] My Profile

saengel:
## Description
Aligning the profile page on sheets to the updated spec

## Code Changes
1. `static/css/s2.css` - Aligning styles with spec. 
2. `static/js/Misc.jsx` - Removing icons, topic tags from SheetListing and CollectionListing
3. `static/js/UserProfile.jsx` - Removing and rearranaging according to spec, removing filter and search on list items. 

## Notes
- Some outstanding font questions to be cleared up

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -11718,7 +11718,9 @@ cursor: pointer;
     font-size: 43.5px !important;
   }
   .profile-page .profile-summary .profile-actions {
+    display: flex;
     flex-wrap: wrap;
+    flex-direction: row;
   }
 }
 .profile-page .profile-summary .follow {
@@ -11729,11 +11731,7 @@ cursor: pointer;
 .profile-page .profile-summary .follow .follow-bull {
   padding: 0 5px;
 }
-.profile-page .profile-summary .profile-actions {
-  display: flex;
-  flex-direction: row;
-  margin: 10px 0;
-}
+
 .profile-page .resourcesLink.faded {
   color: #666;
 }
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -14081,6 +14081,21 @@ span.ref-link-color-3 {color: blue}
   padding: 0px 4px;
 }
 
+.profilePicAndButtonContainer {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.createButtons {
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+  justify-content: space-evenly;
+  padding-top: 20px;
+}
+
 @-webkit-keyframes load5 {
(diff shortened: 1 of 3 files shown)


static/js/UserProfile.jsx
```diff
@@ -598,12 +564,16 @@ const UserBackground = ({profile: p, showBio}) => {
...
+
+
 const ProfileSummary = ({
                             profile: p,
                             openFollowers,
                             openFollowing,
                             toggleSignUpModal,
+                            inProfileView = false,
                         }) => {
+    console.log("in profile view", inProfileView);
```
stevekaplan123, line 576: Shouldn't have console log here

static/js/UserProfile.jsx
```diff
@@ -359,49 +280,15 @@ class UserProfile extends Component {
...
-                    renderEmptyList={this.renderEmptyCollectionList}
-                    renderHeader={this.renderCollectionHeader}
-                    sortOptions={["Recent", "Name", "Sheets"]}
-                    getData={this.getCollections}
-                    data={this.getCollectionsFromCache()}
-                    refreshData={this.state.refreshCollectionsData}
-                  />
-                  {
-                    this.state.showNotes ? (
-                      <FilterableList
```
stevekaplan123, line 391: Curious why we we're not using FilterableLists anymore?

static/css/s2.css
```diff
@@ -14081,6 +14111,54 @@ span.ref-link-color-3 {color: blue}
...
+  border: 1000px solid white;
+  border-top: 4px solid white;
+  border-bottom: 120px solid white;
+  margin: -4px 0 -120px -1000px;
+  width: 100%;
+
+}
+
+.interface-hebrew .profile-page .sheetsProfileList {
+  margin: -4px -1000px -120px 0;
```
stevekaplan123, line 14141: Just a small tip:  You can use margin-inline-start and margin-inline-end to set left and right and that way you don't need to set the opposite values for the hebrew css

static/js/UserProfile.jsx
saengel: Great question - the spec changed, and we don't have search/sort anymore. The entire goal of the `FilterableList` was to present the data in such a way it could be searched and sorted.

static/js/UserProfile.jsx
saengel: Good catch 😅 thank you!

static/css/s2.css
saengel: Oh, great tip - thanks for letting me know. That sounds a lot simpler.

yitzhakc (commented):
Didn't see anything else right now

yitzhakc (approved):

PR #2213, opened 2025-01-06
fix(Links): safely fall back when keyerror is returned

stevekaplan123:
## Description
Badly formatted citations were generating a KeyError and preventing loading of segment in Rashba on Megillah

## Code Changes
We are now handling the KeyError.  The segment loads properly.  The badly formatted citation simply appears as regular text and is not picked up as a link.

### DIFF
--- a/sefaria/model/tests/library_test.py
+++ b/sefaria/model/tests/library_test.py
@@ -301,6 +301,14 @@ class Test_he_get_refs_in_text(object):
         st = "(סוכה מ\' א\' – ב\')"
         wrapped = library.get_wrapped_refs_string(st, lang="he", citing_only=True)
 
+    def test_bad_refs(self):
+        trefs = ["סנהדרין ב, ב ו-ג, ב"]
+        orefs = [Ref(tref) for tref in trefs]
+        st = reduce(lambda a, b: a + "({}) בלה בלה ".format(b), trefs, "")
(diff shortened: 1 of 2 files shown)


yitzhakc (changes requested):
As discussed, please add a regression test with an example of a segment that previously threw an exception

stevekaplan123:
> As discussed, please add a regression test with an example of a segment that previously threw an exception

sefaria/model/tests/library_test.py
```diff
@@ -301,6 +301,14 @@ def test_ranged_talmud_wrap_refs(self):
         st = "(סוכה מ\' א\' – ב\')"
         wrapped = library.get_wrapped_refs_string(st, lang="he", citing_only=True)
 
+    def test_bad_refs(self):
```
stevekaplan123, line 304: Added a passing test for refs we don't want to wrap

yitzhakc: Great! Can we specify that in the test name or function docstring? So we'll understand why we added it.

yitzhakc (changes requested):

sefaria/model/tests/library_test.py
stevekaplan123: I added a comment to test_bad_refs.  Is it clear?

yitzhakc (approved):

PR #2220, opened 2025-01-07
fix(Links): load links from cache before checking linkSummaries cache

stevekaplan123:
For cauldron creation

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -1424,6 +1424,40 @@ Sefaria = extend(Sefaria, {
   }
   ,
   _linkSummaries: {},
+  shouldBuildLinkSummaries: function(cacheKey, links) {
+    // check _linkSummaries[cacheKey]: does this cacheKey (1) exist and (2) is it up-to-date with links
+    // if it doesn't exist or isn't up-to-date, need to build _linkSummaries for cacheKey
+    if (cacheKey in this._linkSummaries) {
+      const countFromLinkSummariesCache = this._linkSummaries[cacheKey].reduce((sum, current) => {
+        return sum + current.count;
+      }, 0);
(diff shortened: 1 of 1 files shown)


saengel (commented):
Looks good to me!

stevekaplan123:
> Should we wait and see if #2292 resolves this issue? Since wouldn't the commentaries spanning multiple refs no longer be returned?

I looked into it.  #2292 doesn't solve this bug.  #2292 is dealing with a very specific case -- When you have Rashi in the sidebar, we load the segments and the section ref -- the latter in case you choose to open Rashi in its own panel and that way you already have the section ref data cached.  The ranged refs that are relevant to this PR get cached due to a separate API call.

yitzhakc (changes requested):
1. Figure out why there is a discrepancy between the `.reduce`d count of links versus the linksCache (See `Leviticus 8:2`) 
1. Cache the calculation of the linksSummaries (initial length and filtered length) so we don't need to recalculate on every run of `shouldBuildLinkSummaries`

stevekaplan123:
> 1. Figure out why there is a discrepancy between the `.reduce`d count of links versus the linksCache (See `Leviticus 8:2`)
> 2. Cache the calculation of the linksSummaries (initial length and filtered length) so we don't need to recalculate on every run of `shouldBuildLinkSummaries`

The filtered length vs initial length issue is not really a problem.  I mentioned in our huddle that I thought I had dealt with this issue but couldn't remember where.  The function I created where it filters out various cases is called getLinksFromCacheAndPreprocess, and it brings 108 down to 105.  As for caching the total, Sefaria._linkSummaries[cacheKey] used to be an array.  I changed it to an object of the form that contains the array {"data": array, "total": sum}

yitzhakc (approved):

PR #2244, opened 2025-01-13
[Sheets][Desktop] Other User's Profile

saengel:
## Description
Small tweaks to UserProfile to adjust for the spec when viewing another user's profile. 

## Code Changes
1. `static/css/s2.css` - Force URLs in about to have `roboto` font. Temporarily coloring the button green. 
2. `static/js/UserProfile.jsx` - Rendering the follow button under the name in another user's view. 

## Notes
1. The 'follow' button is a placeholder for the theme-able button. 
3. The font on the collection details is waiting on theme-ability as well.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -8798,12 +8798,12 @@ body.interface-english .publishBox .react-tags__suggestions ul {
 .collectionListingName:hover {
   color: #333;
 }
-.collectionListingDetails {
-  color: #999;
-  font-size: 14px;
+.collectionListingDetails .collectionListingSheetCount {
+  color: #999 !important;
+  font-size: 14px !important;
   font-family: Roboto !important;
-  font-weight: 400;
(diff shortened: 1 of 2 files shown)


static/css/s2.css
```diff
@@ -11734,6 +11734,11 @@ cursor: pointer;
   font-weight: 400;
   line-height: 16.41px;
 }
+
+.aboutText a {
```
stevekaplan123, line 11738: I'd consider putting another class before .aboutText (I mean one of its parents) as we may want to specify aboutText's css in another situation.

stevekaplan123 (dismissed):
Works with my code!

static/css/s2.css
saengel: That makes sense to me! Thanks for flagging.

yitzhakc (approved):

PR #2249, opened 2025-01-14
feat(profile): Tweaks for mobile web

saengel:
## Description
Tweaks to the my-profile view for mobile-web. 

## Code Changes
1. `static/js/UserProfile.jsx` - Incorporate use of `multiPanel` to conditionally hide buttons / move buttons for mobile-web view.

### DIFF
--- a/static/js/UserProfile.jsx
+++ b/static/js/UserProfile.jsx
@@ -26,7 +26,6 @@ class UserProfile extends Component {
       this.setState(this.getPrivateTabState(this.props));
     }
   }
-
   getPrivateTabState(props) {
     const showBio = !!props.profile.bio;
     const tabs = [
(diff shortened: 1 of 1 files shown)


static/js/UserProfile.jsx
```diff
@@ -572,7 +573,9 @@ const ProfileSummary = ({
                             openFollowing,
                             toggleSignUpModal,
                             inProfileView = false,
+                            multiPanel,
```
yitzhakc, line 576: Do the variables with default values need to come last? In function signatures I'm pretty sure it's a must, but perhaps here is more forgiving?

saengel: Good call! Did a bit of digging, seems it's required in React as well. Thank you!

yitzhakc (approved):

PR #2260, opened 2025-01-19
Feature/sc 31701/consistent language toggle uis

YishaiGlasner:
This PR has 2 goals:
1. Add accessibility to the texts menus.
2. Change the style of source and translation buttons to be like in the topics page menu.

Files changed:
1. new component `RadioButton` (in common folder) for an accessible reusable radio button.
2. `SourceTranslationButtons` and `LayoutButtons` use `RadioButton`.
3. css for changing source and translation buttons style (circle input, height, justify, paddings etc.), and for accessibility of menu (hide things without `display: none`, add outline for keyboard access)

Note
The topics page menu is still different component, since it uses its own language control. This should be handled in the languages control project.


static/js/common/RadioButton.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+import {InterfaceText} from "../Misc";
+
+export default function RadioButton  ({isActive, onClick, value, name, text, ...rest}) {
+    return (
+        <div
+            className='button'
```
yitzhakc, line 8: I would actually call this class `radio-button`, so we don't get confused and think it applies to all `input[type=button]` elements.

static/js/common/RadioButton.js
```diff
@@ -0,0 +1,31 @@
...
+            onClick={onClick}
+            role='radio'
+            aria-checked={isActive}
+        >
+            <label htmlFor={value}><InterfaceText>{text}</InterfaceText></label>
```
yitzhakc, line 13: I believe `htmlFor` should match the `id` of the corresponding `input`

static/js/common/RadioButton.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+import {InterfaceText} from "../Misc";
+
+export default function RadioButton  ({isActive, onClick, value, name, text, ...rest}) {
+    return (
+        <div
```
yitzhakc, line 7: It's not so clear to me why you're adding all of the radio button features to this div (`onClick`, `role=radio`, `aria-checked`) when we have an actual `input[type=radio]`? This is seemingly creating a radiobutton inside a radiobutton 😅

static/js/common/RadioButton.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+import {InterfaceText} from "../Misc";
+
+export default function RadioButton  ({isActive, onClick, value, name, text, ...rest}) {
```
yitzhakc, line 5: For a minute I was confused at the difference between `value`, `name` and `text`. Do we have a convention that `text` is always the `interfaceText`? I think that might make it clearer.

yitzhakc (changes requested):

static/js/common/RadioButton.js
YishaiGlasner: this is a class we're already using in other places in the code.

static/js/common/RadioButton.js
YishaiGlasner: on the one hand, we want all the div to be a button, and we want the option of free styling. on the other hand, we want the functionality of input tag. i've asked chat GPT to explain it for me 😅
https://chatgpt.com/share/67963de5-c594-8007-ae77-6b2dd4bd7209

static/js/common/RadioButton.js
YishaiGlasner: when i wrote it, i've thought that ,maybe to avoid the value prop and use the text as the input value, but maybe in some cases someone will want an empty text (with css icon instead), so he must have an input for that. maybe we can default the value to null, and then use value || text.
i think that naming it as `interfaceText` is not  a proper name, for the interface here is not essential. what do you think about `label`?

yitzhakc: Ah, I see. Definitely makes sense to keep them separate.
`label` sounds like a good name!

stevekaplan123 (commented):
This PR makes sense to me except I'm wondering about your comment in the description: `"The topics page menu is still different component, since it uses its own language control. This should be handled in the languages control project."`
What is the languages control project?

YishaiGlasner:
> This PR makes sense to me except I'm wondering about your comment in the description: `"The topics page menu is still different component, since it uses its own language control. This should be handled in the languages control project."` What is the languages control project?

By 'languages control project' i mean the project where we want to have one clear logic for all the language toggles.

stevekaplan123 (dismissed):

yitzhakc (approved):

PR #2264, opened 2025-01-21
Feature/sc 30094/add sheet options to sheets editor

stevekaplan123:
## Description
This PR takes two features that were implemented for the modularized sheets viewer and implements them for the modularized sheets editor: 1) the ellipsis dropdown SheetOptions component and 2) the sheets sidebar SheetContentSidebar component which displays information about the editor.  This PR also adds a delete option for the ellipsis dropdown menu for the sheets editor.

## Code Changes
Sheet.jsx is changed so that the editor is rendered with the SheetOptions and SheetContentSidebar. 
Two new components were added to implement the delete option in the dropdown menu: DeleteButton and DeleteModal

### DIFF
--- a/static/js/sheets/SheetOptions.jsx
+++ b/static/js/sheets/SheetOptions.jsx
@@ -126,7 +126,7 @@ const ShareButton = () => {
 const DeleteButton = () => {
     return <DropdownMenuItemWithIcon icon={"/static/icons/trash.svg"}
               textEn={'Delete Sheet'}
-              textHe={"מחיקה דף מקורות"}
+              textHe={"מחיקת דף מקורות"}
               descEn={""}
               descHe={""}/>
 }


static/js/sheets/SheetOptions.jsx
```diff
@@ -101,6 +123,14 @@ const ShareButton = () => {
               descHe={""}/>
 }
 
+const DeleteButton = () => {
+    return <DropdownMenuItemWithIcon icon={"/static/icons/trash.svg"}
+              textEn={'Delete Sheet'}
+              textHe={''}
```
yitzhakc, line 129: Why are these texts empty?

yitzhakc (changes requested):

static/js/sheets/SheetOptions.jsx
stevekaplan123: My understanding is that we aren't totally certain about a lot of the text and I wasn't given Hebrew translations.  I just came up with a translation. Is it right?

yitzhakc: Intuitively, I would have said it should be `מחק דף מקורות`, in the "command" form as opposed to "מחיקה" which is the noun form. But ultimately we should try to adhere to whatever "rules" we have for translations across the site.

yitzhakc: I have been corrected, please put `מחיקת דף מקורות` and we'll close this PR up 👍

stevekaplan123: Done

PR #2269, opened 2025-01-22
feat: Catch and log errors in rendering in Node

EliezerIsrael:
## Description
We had been silently dropping errors in Node SSR.  Here we aim to catch those.

### DIFF
--- a/node/server.js
+++ b/node/server.js
@@ -99,7 +99,7 @@ router.get('/error', function(req, res, next) {
   return next(new Error("This is an error and it should be logged to the console"));
 });
 
-router.post('/ReaderApp/:cachekey', function(req, res) {
+router.post('/ReaderApp/:cachekey', function(req, res, next) {
   // timing stored on locals so that it gets returned with the result to be logged
   const timer = res.locals.timing = {
     start: new Date(),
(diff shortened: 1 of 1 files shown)


node/server.js
```diff
@@ -128,11 +128,17 @@ router.post('/ReaderApp/:cachekey', function(req, res) {
       delete res.locals.timing.elapsed;  // no need to pass this around
 
       res.end(resphtml);
-    } catch (render_e){
+    } catch (render_e) {
+      logger.error("Error in renderReaderApp");
```
yitzhakc, line 132: Personally, I think it's more readable to have the "Error in X" as part of the subsequent logging statement, and not a separate statement, i.e. `logger.error("Error in renderReaderApp", render_e)`. Or even structure it as an object, a la `logger.error(render_e, {trigger: "renderReaderApp"})`, which then attaches the additional properties to the JSON object that winston outputs

node/server.js
```diff
@@ -128,11 +128,17 @@ router.post('/ReaderApp/:cachekey', function(req, res) {
...
       logger.error(render_e);
+      res.status(500)
```
yitzhakc, line 134: I'm pretty sure the default error handler should be setting the response to 500?

node/server.js
```diff
@@ -128,11 +128,11 @@ router.post('/ReaderApp/:cachekey', function(req, res) {
...
 
       res.end(resphtml);
-    } catch (render_e){
-      logger.error(render_e);
+    } catch (render_e) {
+      return next(render_e);
     }
   }).catch(error => {
-    res.status(500).end('Data required for render is missing:  ' + error.message);
+      return next(error);
```
yitzhakc, line 135: Indenting

yitzhakc (changes requested):

node/server.js
EliezerIsrael: Fixed in next commit.

yitzhakc (approved):

akiva10b (approved):

PR #2275, opened 2025-01-23
Clean up Header Icons

saengel:
## Description
- The header icons were not vertically aligned, this PR cleans it up with CSS
- The help button was mistakenly removed, this PR reinserts it and centers it properly

## Code Changes
1. `static/css/s2.css` - Adjusting some margins, mostly just making sure each div wrapping each icon svg was displaying its contents as centered. 
2. `static/js/Header.jsx` - Adding the help button

## Notes
- Right now, the help button seems to disappear in mobile-web, is this by design?

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -202,13 +202,11 @@ class Header extends Component {
 
         { Sefaria._siteSettings.TORAH_SPECIFIC ? <HelpButton /> : null}
 
-
         { !Sefaria._uid && Sefaria._siteSettings.TORAH_SPECIFIC ?
               <InterfaceLanguageMenu
                 currentLang={Sefaria.interfaceLang}
                 translationLanguagePreference={this.props.translationLanguagePreference}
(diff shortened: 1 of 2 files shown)


stevekaplan123 (dismissed):

static/js/Header.jsx
```diff
@@ -200,6 +200,9 @@ class Header extends Component {
             openURL={this.props.openURL}
         />
 
+        { Sefaria._siteSettings.TORAH_SPECIFIC ? <HelpButton /> : null}
+
```
yitzhakc, line 204: Lots of empty lines - on purpose?

yitzhakc (changes requested):

static/js/Header.jsx
saengel: Slightly, for readability - but reduced the amount. Thanks for noticing!

yitzhakc (approved):

PR #2277, opened 2025-01-23
Disable tooltip in ReaderControls on Mobile-Web 

edamboritz:
## Description
Disables the ability to hover over a tooltip in mobile web settings

## Code Changes
Use explicit touch/mouse css declarations to disable inconsistent tool tip behavior on mobile browsers
Adds media queries to check for hover implementation and pointer granularity (coarse on touch devices) to make sure only mouse enabled environments respond to hover

## Notes
This is due to an inconsistent implementation of touch to pseudo states like hover in different browser engines that makes it impossible to use correctly in mobile browsers. So in Firefox it simply would not show anyway and in Chrome the tooltip would show on touch instead of hover but then get stuck because there is no parallel off "not hover"

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -11682,12 +11682,16 @@ cursor: pointer;
   }
 }
 
+.tooltip-toggle {
+  position: relative; /* Ensure positioning context for absolute pseudo-element */
+}
 .tooltip-toggle::before {
+  content: attr(aria-label);
   color: #000;
   font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
   font-size: 13px;
   opacity: 0;
-  pointer-events: none;
+  /*pointer-events: none;*/
   text-align: center;
   position: absolute;
   top: 30px;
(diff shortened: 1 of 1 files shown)


static/css/s2.css
```diff
@@ -11682,12 +11682,16 @@ cursor: pointer;
   }
 }
 
+.tooltip-toggle {
```
YishaiGlasner, line 11685: we already have this selector and rule in lines 11653 & 11655

static/css/s2.css
```diff
@@ -11682,12 +11682,16 @@ cursor: pointer;
...
+  position: relative; /* Ensure positioning context for absolute pseudo-element */
+}
 .tooltip-toggle::before {
+  content: attr(aria-label);
   color: #000;
   font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
   font-size: 13px;
   opacity: 0;
-  pointer-events: none;
+  /*pointer-events: none;*/
```
YishaiGlasner, line 11694: is there a reason for this to remain?

static/css/s2.css
```diff
@@ -11696,17 +11700,32 @@ cursor: pointer;
...
+  visibility: hidden; /* Hide completely when not shown */
 }
-/*Triggering the transition*/
-.tooltip-toggle:hover::before, .tooltip-toggle:hover::after {
-  opacity: 1;
-  transition: all 0.75s ease;
+/* Combine hover and focus for better accessibility */
+@media (hover: hover) and (pointer: fine) {
+  .tooltip-toggle:hover::before,
+  .tooltip-toggle:focus::before {
```
YishaiGlasner, line 11712: this is the same selector again

static/css/s2.css
```diff
@@ -11696,17 +11700,32 @@ cursor: pointer;
...
+    opacity: 1;
+    visibility: visible;
+    transition: opacity 0.75s ease;
+  }
+}
+@media (pointer: coarse) {
+  .tooltip-toggle::before {
+    /* Mobile-specific styles */
+    display: none; /* or alternative mobile-friendly behavior */
```
YishaiGlasner, line 11721: the note seems irrelevant

static/css/s2.css
```diff
@@ -11696,17 +11700,32 @@ cursor: pointer;
...
+  }
 }
+/* Ensure the button itself can receive focus */
+.tooltip-toggle {
```
YishaiGlasner, line 11725: can go to the line 11653 slector

static/css/s2.css
```diff
@@ -11696,17 +11700,32 @@ cursor: pointer;
...
+  outline: none; /* Optional: removes default focus outline */
```
YishaiGlasner, line 11726: this note also seems irrelevant

static/css/s2.css
edamboritz: actually, commented out mistakenly

static/css/s2.css
edamboritz: no it isn't, one is focus one is hover

YishaiGlasner (approved):

PR #2279, opened 2025-01-26
Attempt to fix React Undefined in SSR

edamboritz:
Changes the webpack config so that babel uses the new jsx transform so React is available globally and does not need to be imported in every file 
 

## Code Changes
Added `{ runtime: 'automatic' }`
```
presets: [
                    ['@babel/react', { runtime: 'automatic' }],
                    '@babel/preset-env'
                ],
```
To the `rules` block of `baseConfig` in the main webpack file

## Notes
Will need some more verbose prod grade testing

### DIFF
--- a/node/webpack.config.js
+++ b/node/webpack.config.js
@@ -55,27 +55,29 @@ var baseConfig = {
     ],
     module: {
         rules: [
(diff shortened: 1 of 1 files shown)


EliezerIsrael (approved):
Tested.  Works.

PR #2280, opened 2025-01-27
Slug to Pools Cache to Optimize SQL Querying

yonadavGit:
## Description
Implementaion of new library cache to be used when initializaing topic objects
## Code Changes

## Notes

### DIFF
--- a/django_topics/models/topic.py
+++ b/django_topics/models/topic.py
@@ -21,6 +21,18 @@ class TopicManager(models.Manager):
     def get_topic_slugs_by_pool(self, pool: str) -> QuerySet:
         return self.filter(pools__name=pool).values_list("slug", flat=True)
 
+    def get_pools_for_all_topic_slugs(self) -> dict[str, list[str]]:
+        from collections import defaultdict
+        """
+        Returns a dictionary mapping each topic slug in the database to its list of pools.
+        """
+        slug_to_pools = defaultdict(list)
+        topics = self.values_list('slug', 'pools__name')  # Fetch all slugs and their pools
+        for slug, pool_name in topics:
+            if pool_name:  # Avoid adding None values
+                slug_to_pools[slug].append(pool_name)
+        return dict(slug_to_pools)
+
 
 class Topic(models.Model):
     slug = models.CharField(max_length=255, primary_key=True)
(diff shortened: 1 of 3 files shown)


django_topics/models/topic.py
```diff
@@ -21,6 +21,18 @@ def get_pools_by_topic_slug(self, topic_slug: str) -> QuerySet:
     def get_topic_slugs_by_pool(self, pool: str) -> QuerySet:
         return self.filter(pools__name=pool).values_list("slug", flat=True)
 
+    def get_pools_for_all_topic_slugs(self) -> dict[str, list[str]]:
+        from collections import defaultdict
```
akiva10b, line 25: primitives should be imported at top

sefaria/model/text.py
```diff
@@ -4960,6 +4960,8 @@ def __init__(self):
         self._simple_term_mapping_json = None
         self._linker_by_lang = {}
 
+        #Topic to Pools
+        self._topic_slug_to_pools = {}
```
akiva10b, line 4964: consider one of these names:
```
_slug_topic_pools_mapping
_topic_pools_by_slug
```

instead, and change the other functions accordingly

sefaria/model/text.py
```diff
@@ -5681,6 +5686,20 @@ def _build_topic_mapping(self):
...
+        tpm = self._topic_slug_to_pools
+        if not tpm or rebuild:
+            tpm = self._build_topic_slug_to_pools_mapping()
+        return tpm
+
+    def get_topic_pools_by_slug(self, slug):
+        return self._topic_slug_to_pools.get(slug, [])
+
+    def _build_topic_slug_to_pools_mapping(self):
+        from django_topics.models import Topic as DjangoTopic
```
akiva10b, line 5699: import at top

django_topics/admin.py
```diff
@@ -72,6 +72,12 @@ class TopicAdmin(admin.ModelAdmin):
         create_remove_from_pool_action(PoolType.TORAH_TAB.value),
     ]
 
+    def save_related(self, request, form, formsets, change):
+        from sefaria.model.text import library
```
akiva10b, line 76: we should reconsider is library is the proper place for this cache. I am thinking it isn't. But I don't think we have a proper place for a cache like this. Either way, we don't want an import like this happening on every save, so move to top

PR #2283, opened 2025-01-28
chore: sheets can no longer appear in multipanel mode

stevekaplan123:
## Description
Sheets multipanel mode doesn't exist in modularization.

## Code Changes
In our code that handles multipanel requests, we had the option for sheets, so I removed those lines.


saengel (commented):
Looks good to me!

yitzhakc (approved):

PR #2298, opened 2025-02-04
Remove parasha dedications for last year's sponsors

relyks:
## Description
This removes the parasha dedications for the sponsors who had an annual sponsporship.


edamboritz (approved):

edamboritz:
Why do we need to do this rather than replace the copy?

PR #2322, opened 2025-02-11
Modulrization Themes

edamboritz:
## Description
This PR adds the ability to create themable shared components between the modules

### CSS: 
Added several CSS files:
#### Theme files:
`static/css/color-palette.css `- for storing all global color css custom properties
`static/css/font-faces.css` - the same for fonts
`static/css/themes/library-theme.css`
`static/css/themes/sheets-theme.css` - For storing the custom properties that can change their color or font values from the previous files, depending on the module active

#### Other 
static/css/common-components.css for storing all the actual css for common js components outside of s2.css in an easier to manage file. This includes css declarations that are static and dont change (e.g. width) and things that are tied to theme properties from the files above. 

All new css files are loaded in `base.html`

### Other:
Added a context processor and middleware that will detect according to the url (stats with "/sheets") what active module is and add it to the django request object under `request.active_module`  and then also to the global js sefaria scope under `Sefaria.activeModule` . 

Refactored the common button component to match. 

## Notes
Removed margin from button because the base common button should not make assumptions about margins, rather the surrounding context should on a case by case basis. 
This really should be applied as a rule to common components - always consider what is inherent to the component and what is jsut layout in a given context

As more themable attributes will be needed, they can be expanded in the theme files, making sure that the theme files contain the same variables


static/css/common-components.css
```diff
@@ -0,0 +1,76 @@
...
+    border: 0; /* Remove inner border in Firefox */
+    padding: 0; /* Remove inner padding in Firefox */
+}
+/* end reset */
+
+/* Button Styles */
+
+/** general sized button **/
+button.sefaria-common-button {
+    background-color: var(--theme-button-background);
```
stevekaplan123, line 35: Why is there CSS for "button.sefaria-common-button {" twice?  There's CSS for it on line 1 and line 34

static/css/themes/library-theme.css
```diff
@@ -0,0 +1,59 @@
+[data-active-module="library"]{
+  /* Define default theme variables */
+  --theme-primary: var(--sefaria-blue);
+  --theme-background: var(--lightest-grey); /** just an example **/
+  --theme-text: black;  /** just an example **/
```
stevekaplan123, line 5: What do you mean that it's just an example?

sefaria/system/middleware.py
```diff
@@ -20,6 +21,27 @@
...
+
+class MiddlewareURLMixin:
+    excluded_url_names = set()  # Set of URL names to exclude
+    excluded_url_prefixes = set()  # Set of URL path prefixes to exclude
+
+    def should_process_request(self, request):
+        # Check URL name
+        try:
+            url_name = resolve(request.path_info).url_name
+            if url_name in self.excluded_url_names:
```
stevekaplan123, line 33: It seems we're currently not using 'excluded_url_names', but only 'excluded_url_prefixes'.  Are there plans to use it in the future?

edamboritz: yes it is there if we need it in the future

saengel (commented):
Super helpful, thank you! Looks great!

static/css/common-components.css
edamboritz: Note that the comment clarifies this: 
Lines 1-29  are a "reset" essentially a safeguard against previously existing css interfering with the following CSS. These lines "undo" that. 
Its a common css practice when needed, though not the prettiest.

static/css/themes/library-theme.css
edamboritz: I mean I left these as examples of ways to use custom properties when creating css variables. An example of what types of things we should be placing in these var/custom properties.

stevekaplan123:
@edamboritz Is there a reason we haven't yet merged this?  One of my PRs can't be reviewed until we merge this in.

yitzhakc (approved):

yitzhakc (approved):

yitzhakc (approved):

yitzhakc (approved):

PR #2324, opened 2025-02-12
Feature/sc 32516/wysiwyg for master

stevekaplan123:
## Description
Currently exporting a sheet to google docs is not WYSIWYG.  Rather, the google doc appears with the layout and language the sheets creator chose.  We want the user's language and layout choice to be respected when they export to a google doc.  

## Code Changes
1. Fundamental change to the google docs template -- currently each source in the sheet object in the database comes with its own language and layout.  This PR changes this so that there is one and only one language and one and only one language setting.  
2. The language and layout settings for the sheet are set here in views.py:
```
 options = {'language': request.GET.get("language", "bilingual"),
               'layout': request.GET.get("layout", "heRight")}
    sheet['options'] = sheet['options'] | options
```
3. SheetToolsList, the component that allows you to export to google docs, is currently in the ConnectionsPanel, which is not the same panel as the panel showing the sheet, so the SheetToolsList component doesn't currently know what layout/language combo the user chose.  The ReaderApp needs to pass down the master panel layout and language down the tree to SheetToolsList, so I added a new prop: masterPanelLayout.
4. Remove unnecessary function resolve_options_of_sources which is no longer being called.


sourcesheets/views.py
```diff
@@ -170,6 +170,7 @@ def make_sheet_class_string(sheet):
     """
     Returns a string of class names corresponding to the options of sheet.
     """
+
```
yitzhakc, line 173: Remove empty line

sourcesheets/views.py
```diff
@@ -1105,16 +1087,18 @@ def export_to_drive(request, credential, sheet_id):
     # Using credentials in google-api-python-client.
     service = build('drive', 'v3', credentials=credential, cache_discovery=False)
     user_info_service = build('oauth2', 'v2', credentials=credential, cache_discovery=False)
-
     sheet = get_sheet(sheet_id)
     if 'error' in sheet:
         return jsonResponse({'error': {'message': sheet["error"]}})
 
+    options = {'language': request.GET.get("language", "bilingual"),
+               'layout': request.GET.get("layout", "heRight")}
```
yitzhakc, line 1095: I'm a little confused. You're setting the _default_ (if the key doesn't exist in the GET params) to `heRight`, but I don't see any case in the `gdocs_sheet.html` template that refers to `heRight`. Is that just the default, so it doesn't need any explict coding?

yitzhakc (changes requested):

static/js/ConnectionsPanel.jsx
```diff
@@ -799,7 +802,7 @@ const AboutSheetButtons = ({ setConnectionsMode, masterPanelSheetId }) => {
   </div>);
 }
 
-const SheetToolsList = ({ toggleSignUpModal, masterPanelSheetId, setConnectionsMode }) => {
+const SheetToolsList = ({ toggleSignUpModal, masterPanelSheetId, setConnectionsMode, masterPanelLanguage, masterPanelLayout }) => {
```
YishaiGlasner, line 805: maybe it's an opportunity to add PropTypes to SheetToolsList

sourcesheets/views.py
stevekaplan123: That's what the default was before in case there was nothing set by the source sheet creator.  In this case, heLeft is dealt with explicitly in gdocs_sheet.html, and there's an else clause for heRight.  I'm going to change it so it's explicit.

YishaiGlasner (approved):

stevekaplan123:
@yitzhakc Not sure why this isn't merging.  It seems like you tried to override the tests it but still didn't work.

yitzhakc (approved):

PR #2329, opened 2025-02-14
Support to target banners and modals based on logged-in status

relyks:
## Description
This adds support for a new separate enumerated field that was added to Strapi for modals and banners. Users who are logged-in or logged-out can now be targeted. The default option for this field in the content type targets all users, both logged in and logged out.

## Code Changes
* Pulls the new enumerated field from Strapi
* Changes the conditional logic structure for when a modal or banner should be shown

## Caveats
* To reduce future confusion, 3 options were chosen: logged-in-only, logged-out-only, and both logged-in and logged-out.
* Since all the previously existing flags depend on whether a user is logged in, the new field overrides the existing flags when choosing the option for logged-in- or logged-out-only users.

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -2104,26 +2104,33 @@ const Banner = ({ onClose }) => {
 
     let shouldShowBanner = false;
 
-    let noUserKindIsSet = ![
-      strapi.banner.showToReturningVisitors,
-      strapi.banner.showToNewVisitors,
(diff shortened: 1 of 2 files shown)


stevekaplan123 (commented):
@yitzhakc Looks good to me.  Doesn't touch anything that negatively affects the modal I'm putting up

yitzhakc (approved):

PR #2332, opened 2025-02-17
fix: Prevents image from becoming too wide when browser screen is bet…

EliezerIsrael:
…ween 540px and 600px


EliezerIsrael:
Screenshot 2025-02-17 at 3 03 29 PM
Current state, before fix, between 540px and 600px width.

EliezerIsrael:
Screenshot 2025-02-17 at 3 06 22 PM

After fix

saengel (approved):
Thanks for catching!

PR #2338, opened 2025-02-18
Feature/sc 32283/sheets desktop update topics homepage navigation

stevekaplan123:
## Description
This PR implements a topic TOC for the sheets module.  1) As opposed to the library module, the topics TOC is not the topics landing page but is the old TopicsPage component. 2) The Topics TOC in the sheets module only shows topics that have sheets associated with them. 3) When viewing a topic category in either the library or sheets module, we only show topics that are appropriate for the given module.  In other words, in the sheets module, subtopics are only shown if that subtopic has sheets associated with it, and likewise, in the library module, subtopics are only shown if they have library sources associated with them.

## Code Changes
1.  I created a new component TopicTOCCard to display topics throughout the topic TOC.  The TopicsPage component and TopicCategory components now use TopicTOCCard.
2. Most of the changes in this PR were from [this PR](https://github.com/Sefaria/Sefaria-Project/pull/2398).  There is very little in this PR specifically required to implement the sheets module topics TOC.  The logic is basically just that (a) when building the topic TOC on the backend in model/text.py in `get_topic_toc_json_recursive`, we add a new property to each topic object called `pools`, which is just an array of the topic pools that the topic belongs to.  Then, when we already on a topic category page, such as [prayer](https://www.sefaria.org.il/topics/category/prayer), when we render the TopicCategory component, we call a function `shouldDisplay` to determine if each subtopic (such as [Aleinu](https://www.sefaria.org.il/topics/aleinu?tab=sources))  actually belongs to the sheets/library topic pool.  See the function shouldDisplay in TopicPage.jsx where we filter out all subtopics that are not appropriate for the given Sefaria.activeModule:
```
const TopicCategory = ({topic, topicTitle, setTopic, setNavTopic, initialWidth,
  openSearch}) => {
    const [topicData, setTopicData] = useState(Sefaria.getTopicFromCache(topic) || {primaryTitle: topicTitle});
    const [subtopics, setSubtopics] = useState(Sefaria.topicTocPage(topic));
@@ -226,38 +230,15 @@ const TopicCategory = ({topic, topicTitle, setTopic, setNavTopic, compare, initi
        setSubtopics(Sefaria.topicTocPage(topic));
    }, [topic]);

    const shouldDisplay = (t) => {
      const inActiveModule = t.pools.includes(Sefaria.activeModule);
      return t.shouldDisplay !== false && inActiveModule;
    }

    let topicBlocks = subtopics
      .filter(shouldDisplay)
      .sort(Sefaria.sortTopicsCompareFn)
      .map((topic, i) => <TopicTOCCard topic={topic} setTopic={setTopic} setNavTopic={setNavTopic} key={i}/>);
```


static/js/NavSidebar.jsx
```diff
@@ -759,6 +760,20 @@ const JoinTheCommunity = ({wide}) => {
   );
 };
 
+const JoinTheConversation = ({wide}) => {
+  return (
+    <SidebarModule wide={wide}>
```
YishaiGlasner, line 765: is this new? i don't see you've removet if from another place but it seems familiar.

static/js/ReaderPanel.jsx
```diff
@@ -941,9 +941,18 @@ class ReaderPanel extends Component {
           />
         );
       } else {
-        menu = <TopicsLandingPage openTopic={this.props.openTopic}/>
+        if (Sefaria.activeModule === 'library') {
```
YishaiGlasner, line 944: i don't like that we are adding more conditions to this, but i'm not sure we can do it differently given the current state of the code

static/js/TopicPage.jsx
```diff
@@ -230,38 +230,44 @@ const TopicCategory = ({topic, topicTitle, setTopic, setNavTopic, compare, initi
         setSubtopics(Sefaria.topicTocPage(topic));
     }, [topic]);
 
+    const shouldDisplay = (t) => {
+      const shouldDisplay = t => t.shouldDisplay !== false;
+      const inActiveModule = t.pools.includes(Sefaria.activeModule);
+      return shouldDisplay && inActiveModule;
+    }
+
+    const topicNavBlock = (t, i) => {
```
YishaiGlasner, line 239: i think it will be better to take it out and make it an independent component

sefaria/model/text.py
```diff
@@ -5157,7 +5158,8 @@ def get_topic_toc_json_recursive(self, topic=None, explored=None, with_descripti
                 "shouldDisplay": True if len(children) > 0 else topic.should_display(),
                 "en": topic.get_primary_title("en"),
                 "he": topic.get_primary_title("he"),
-                "displayOrder": getattr(topic, "displayOrder", 10000)
+                "displayOrder": getattr(topic, "displayOrder", 10000),
+                "pools": DjangoTopic.objects.slug_to_pools[topic.slug]
```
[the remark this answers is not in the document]
yitzhakc: @YishaiGlasner Are you sure this changes an API response? Which API url does it change?

Re 2. I do agree that we should take advantage of the `request.active_module` attribute set in the `ModuleMiddleware` to either `sheets` or `library` to take advantage of filtering the list of topics on the `backend`, although I see no problem having the `pools` remain on this object, since it's cached anyways.

sefaria/urls.py
```diff
@@ -109,6 +109,8 @@
     url(r'^topics/?$', reader_views.topics_page),
     url(r'^topics/b/(?P<topic>.+)$', reader_views.topic_page_b),
     url(r'^topics/(?P<topic>.+)$', reader_views.topic_page),
+    url(r'^sheets/topics/(?P<topic>.+)$', reader_views.topic_page),
```
yitzhakc, line 112: Please take a look at the `reader.views.topic_page` and `topic_pages` functions, since I think we need to change some things there too (they both still reference a unified "Texts & Source Sheets")

static/js/ReaderPanel.jsx
```diff
@@ -941,9 +941,18 @@ class ReaderPanel extends Component {
...
+          menu = (
+              <TopicsLandingPage openTopic={this.props.openTopic}/>
+          );
+        } else if (Sefaria.activeModule === "sheets") {
+          menu = <TopicsPage
```
yitzhakc, line 949: Just to make sure I understand - why do we not have the `TopicsLandingPage` in `Sheets`?

static/js/TopicPage.jsx
yitzhakc, line 239: I'd love a docstring here explaining what/where this is (and what are `t` and `i`?)

yitzhakc (changes requested):

static/js/ReaderPanel.jsx
stevekaplan123: It was a product decision to include the old topics page "TopicsPage" instead of TopicsLandingPage in the sheets module

sefaria/model/text.py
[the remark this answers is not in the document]
stevekaplan123: @YishaiGlasner @yitzhakc I don't think this changes any API results.  This changes the Sefaria._topics cache which is handled by the front-end only.  /api/topics is not affected by this.   And yes, this adds a 'pools' key to every topic in the Sefaria._topics cache.  The reason I did it this way is because reader_views.topic_page only gets called if someone loads a topic page server side.  Therefore, if you navigate to "Explore" and then go to a topic page, reader_views.topic_page won't get called.  Instead, the front-end will use the `Sefaria._topics` cache.  The problem is that this cache gets set on the backend by get_topic_toc_json (which is in turn called by large_data in context_processors.py) when the server initially loads.  This happens long before any request is received with an active_module parameter.  Therefore, I don't see how we could take advantage of the active_module parameter in the request.

static/js/NavSidebar.jsx
```diff
@@ -759,6 +760,20 @@ const JoinTheCommunity = ({wide}) => {
...
+      <div className="joinTheConversation">
+        <SidebarModuleTitle>Join the Conversation</SidebarModuleTitle>
```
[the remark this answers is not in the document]
stevekaplan123: "Join the Conversation" is in strings.js.  Is that not what you mean?

static/js/NavSidebar.jsx
stevekaplan123: Sort of.  We had something similar, got rid of it, and seem to have decided to bring it back...

reader/views.py
```diff
@@ -3102,49 +3103,54 @@ def topics_page(request):
...
     }
 
     if test_version is not None:
         props["topicTestVersion"] = test_version
 
     short_lang = 'en' if request.interfaceLang == 'english' else 'he'
-    title = topic_obj.get_primary_title(short_lang) + " | " + _("Texts & Source Sheets from Torah, Talmud and Sefaria's library of Jewish sources.")
-    desc = _("Jewish texts and source sheets about %(topic)s from Torah, Talmud and other sources in Sefaria's library.") % {'topic': topic_obj.get_primary_title(short_lang)}
+    title = desc = ""
+    if request.active_module == "library":
```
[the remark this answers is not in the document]
stevekaplan123: You're totally right but I'm not sure product knows what the exact text is supposed to be here so I'm going to leave it for now.

static/js/TopicsPage.jsx
```diff
@@ -10,29 +10,11 @@ import $  from './sefaria/sefariaJquery';
...
-        >
-          <InterfaceText text={cat} />
-        </a>
-        <div className="navBlockDescription">
-          <InterfaceText text={cat.categoryDescription} />
-        </div>
-      </div>
-    );
-  });
+  let categoryListings = Sefaria.topic_toc.map((topic, i) => <TopicCard topic={topic} setNavTopic={setNavTopic} key={i}/>);
```
stevekaplan123, line 17: Doesn't this need to be filtered by pools?

static/js/ReaderPanel.jsx
akiva10b: Ideally we would have a "SheetsPanel" component which would share subcomponents with ReaderPanel but be associated to sheets module

static/js/TopicPage.jsx
```diff
@@ -230,38 +230,15 @@ const TopicCategory = ({topic, topicTitle, setTopic, setNavTopic, compare, initi
...
-              </a>
-              {description ?
-              <div className="navBlockDescription clamped">
-                <InterfaceText markdown={{en: description.en, he: description.he}} disallowedMarkdownElements={['a']}/>
-              </div>
-              : null }
-            </div>
-        );
-      });
+      .map((topic, i) => <TopicTOCCard topic={topic} setTopic={setTopic} setNavTopic={setNavTopic} key={i}/>);
```
akiva10b, line 241: well done!

static/js/sefaria/sefaria.js
```diff
@@ -2952,9 +2952,10 @@ _media: {},
     });
   },
   trendingTopics: {},
-  getTrendingTopics: function() {
+  getTrendingTopics: function(n=10) {
```
akiva10b, line 2955: this should be 2 functions - getTrendingSheetsTopics and getTrendingLibraryTopics

static/js/TopicPage.jsx
```diff
@@ -818,6 +638,181 @@ TopicPage.propTypes = {
...
+const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePreference, versionPref,
+                            scrollableElement, langPref, setLangPref, toggleSignUpModal, topicTestVersion,
+                            onSetTopicSort, topicSort}) => {
+    const [loadedData, setLoadedData] = useState(topicData ? Object.entries(topicData.tabs).reduce((obj, [key, tabObj]) => { obj[key] = tabObj.loadedData; return obj; }, {}) : {});
+    const [refsToFetchByTab, setRefsToFetchByTab] = useState({});
+    const [showLangSelectInterface, setShowLangSelectInterface] = useState(false);
+    const allPossibleTabsWithSources = useAllPossibleSourceTabs(translationLanguagePreference, versionPref);
+    const [showFilterHeader, setShowFilterHeader] = useState(false);
+
+    const getCurrentLang = (langPref) => {
```
yitzhakc, line 650: Is this correct vis a vis RTL? That Hebrew is always `source`?

static/js/TopicPage.jsx
```diff
@@ -818,6 +638,181 @@ TopicPage.propTypes = {
   topicTestVersion:    PropTypes.string
 };
 
+const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePreference, versionPref,
```
yitzhakc, line 641: Please add some documentation here - since this seems to be the main entrypoint...

static/js/common/TopicTOCCard.jsx
```diff
@@ -0,0 +1,26 @@
...
+   */
+  const openTopic = e => {
+    e.preventDefault();
+    children ? setNavTopic(slug, {en, he}) : setTopic(slug, {en, he});
+  }
+
+  const { slug, children} = topic;
+  const description = children ? topic.categoryDescription : topic.description;
+  let {en, he} = topic;
+  en = en.replace(/^Parashat /, "");
```
yitzhakc, line 16: Why do we need to do this? Please add a comment.

yitzhakc (changes requested):

static/js/TopicPage.jsx
stevekaplan123: @YishaiGlasner I'm not sure.  How are we handling language toggles elsewhere?  What do you think?

YishaiGlasner: 1) in sheets, unlike texts, we dont have the distinction of source/translation, but of rtl/ltr.
2) idk how we want to handle language toggle in sheets (i remember that there was a discussion about not having language toggle in sheets, and i;m not sure what was the conclusion).

yitzhakc (approved):

PR #2339, opened 2025-02-18
chore(tests): fix topic helper tests

akiva10b:


django_topics/models/topic.py
```diff
@@ -40,6 +40,11 @@ class Topic(models.Model):
     pools = models.ManyToManyField(TopicPool, related_name="topics", blank=True)
     objects = TopicManager()
 
+    def save(self, *args, **kwargs):
+        self.slug = self.slug.lower()
```
nsantacruz, line 44: can you document why this is necessary?

sefaria/model/tests/text_test.py
```diff
@@ -354,7 +354,7 @@ def test_text_helpers():
         'Akeidat Yitzchak',
         'Berakhot']
      ],
-    ['Exodus', None, False, ['Ibn Ezra on Exodus; Perush HaArokh',
```
yitzhakc, line 357: Why are we removing `Perush HaArokh`?

sefaria/helper/tests/topic_test.py
```diff
@@ -55,17 +55,18 @@ def grandchild_of_root_with_self_link(child_of_root_with_self_link):
 
 
 @pytest.fixture(autouse=True, scope='module')
-def author_root():
+def author_root(django_db_setup, django_db_blocker):
```
yitzhakc, line 58: If I've read the documentation right, I think we should be able to just use the `db` fixture here, which then takes care of the unblocking automatically.
I.e. `def author_root(db)`

yitzhakc (changes requested):

sefaria/model/tests/text_test.py
akiva10b: many of our tests check agains data that is cloned from prod (it is what it is) in this case, the title of 'Ibn Ezra on Exodus; Perush HaArokh' whas changed

sefaria/helper/tests/topic_test.py
akiva10b: I was not successful with that method (I figured it was a django version issue) - if you are let me know

yitzhakc: I'll make a note of it for when we upgrade. Resolving.

PR #2342, opened 2025-02-19
Improves web manifest, fixes icon pathing issue

JudahGabriel:
## Description
This small PR improves the web manifest file. 

- It fixes an icon pathing issue
- It adds basic, standard properties to the web manifest file.

## Code Changes
- Updated the image paths for `/static/icons/manifest.json` so that they're resolved properly in browser.
- Added a 512x512 image per standard requirements.
- Added standard properties to the manifest file, including `short_name`, `description`, and `start_url`
- Changed the web manifest display from `standalone` to `minimal-ui`. This mode is like standalone, but adds a ⬅️ and 🔃 button in the title bar of the web app. 

## Notes
There are two web manifest files: one in `/static/root` and another in `/static/icons`. The latter is the one used on the web at sefaria.org, and it has the following icon resolution issue:
image

The icon resolution issue is due to icons being resolved relative to the web app root, not the web manifest location. I've fixed this issue in this manifest file. 

I've left the other manifest file paths untouched, as I don't know whether it's being used.


JudahGabriel:
I see that new changes have been made to the manifest mirroring my changes. Cool! This PR can be closed.

PR #2343, opened 2025-02-20
Redis cache fix for redis restart

akiva10b:
## Description
Using redis cache alone to asses it the node server can manage ssr by checking if the redis cache keys are available prior to loading node

## Code Changes
This pull request includes significant changes to the caching mechanism and the handling of shared data across the system. The most important changes include the introduction of new functions to check multiple keys in the Redis cache.

Enhancements to shared data handling:

* [`reader/views.py`](diffhunk://#diff-26c40bdd3eeaef442340754126fdbcb1ed9e451eb3bb31c02cb78053f5702eaaL128-R146): Added a list of shared cache keys and modified the `render_react_component` function to check the existence of these keys in the shared cache. If the keys are not found, the cache is regenerated. This ensures that the React component is only rendered when the necessary data is available.

New utility functions for cache operations:

* [`sefaria/system/cache.py`](diffhunk://#diff-1c87fb064c88609bcf2747f793031691dd6be1fa7fda3c73bd328ee71a095b56R49-R61): Introduced the `redis_exists` function to efficiently check the existence of multiple keys in the Redis cache using a single lookup command.
* [`sefaria/system/cache.py`](diffhunk://#diff-1c87fb064c88609bcf2747f793031691dd6be1fa7fda3c73bd328ee71a095b56R122-R130): Added the `check_multiple_keys` and `check_shared_multiple_keys` functions to facilitate checking multiple keys in the cache, specifically for shared data.

These changes collectively improve the efficiency and reliability of the caching system, ensuring that shared data is properly managed and available when needed.


sefaria/system/cache.py
```diff
@@ -46,6 +46,19 @@ def cache_get_key(key_arr):
...
+    """
+    Check if multiple keys exist in the Redis cache efficiently using a singlelookup command.
+    """
+    client = cache_instance.client._clients[0]
+    if client:
+        # Django cache client prefixing
+        prefix = f":{cache_instance.version}:"
+        prefixed_keys = [f"{prefix}{key}" for key in keys]
+        return client.exists(*prefixed_keys) == len(keys)
+    return False
```
akiva10b, line 59: I use the redis exists method because I don't want to call the keys in a loop. the django redis `has` method simply checks for every key in a loop, the base redis exists method is much more efficient as it handles the check on the redis server and not between django and redis

reader/views.py
```diff
@@ -125,7 +125,25 @@ def render_react_component(component, props):
...
+        "toc_json",
+        "term_mapping",
+        "topic_toc_category_mapping",
+        "topic_toc",
+        "toc",
+        "books_en",
+        "virtualBooks",
+    ]
+
+    if not USE_NODE or not check_shared_multiple_keys(SHARED_CACHE_KEYS):
```
EliezerIsrael, line 142: We don't want to be regenerating the shared cache just because USE_NODE is false.

EliezerIsrael (changes requested):

sefaria/system/cache.py
EliezerIsrael: Tested, I presume?

akiva10b:
```
    import time
    start_time = time.time()
    for i in range(10):
        check_shared_multiple_keys(SHARED_CACHE_KEYS)
    print("Time to check shared cache keys: {}".format(time.time() - start_time))

    start_time = time.time()
    for i in range(10):
        get_multiple_shared_cache_elem(SHARED_CACHE_KEYS)
    print("Time to get shared cache keys: {}".format(time.time() - start_time))
```
Time to check shared cache keys: 0.01407313346862793
Time to get shared cache keys: 2.2330708503723145

PR #2348, opened 2025-02-24
Fix Venice Daf Calculation

nsantacruz:
## Description
Possibly since AddressFolio was created, the regex's and calculation of numeric form of folio was never handled properly. This PR fixes these issues and adds tests.

## Code Changes
- Correct AddressFolio.toNumber() so it handles all amudim
- Correct AddressFolio regex so it searches for all amudim
- Make AddressFolio another exception for using non-strict regex. This is required because AddressFolio (along with AddressAmud and AddressTalmud) don't inherit from AddressInteger. This means they need to have internal flexibilty to not match the initial word (Daf, Amud etc.) as opposed to relying on parent class to do this (as is the case with all other address types)

### DIFF
--- a/sefaria/model/schema.py
+++ b/sefaria/model/schema.py
@@ -2470,7 +2470,7 @@ class AddressFolio(AddressType):
             reg += r"\d+[abcdᵃᵇᶜᵈ]?)"
         elif lang == "he":
             # todo: How do these references look in Hebrew?
-            reg += self.hebrew_number_regex() + r'''([.:]|[,\s]+(?:\u05e2(?:"|\u05f4|''))?[\u05d0\u05d1])?)'''
+            reg += self.hebrew_number_regex() + r'''([.:]|[,\s]+(?:\u05e2(?:"|\u05f4|''))?[\u05d0-\u05d3])?)'''
 
         return reg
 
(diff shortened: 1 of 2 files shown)


sefaria/model/schema.py
```diff
@@ -2499,21 +2499,19 @@ def toNumber(self, lang, s, **kwargs):
...
-                     or (len(s) > 4 and s[-3] == '\u05e2')  # ayin"bet
-                     or (len(s) > 5 and s[-4] == "\u05e2")  # ayin''bet
-                    )
-            ):
-                return daf  # amud B
-            return daf - 1
+            daf = decode_hebrew_numeral(num) * 4
+            rest = s[len(num):]
+
+            # check for each amud letter in reverse order (dalet, gimmel, bet, alef)
```
YishaiGlasner, line 2506: why should it be reverse?

sefaria/model/schema.py
```diff
@@ -2499,21 +2499,19 @@ def toNumber(self, lang, s, **kwargs):
...
+            # if the amud matches that amud letter, subtract the appropriate offset
+            for amud_offset, amud_letter in enumerate(("\u05d3", "\u05d2", "\u05d1", "\u05d0")):
+                if re.search(fr"^(?::|,?\s?(?:{amud_letter}|ע(?:''|\"|״){amud_letter}))$", rest):
```
YishaiGlasner, line 2509: i can't understand why the first option of only colon - it means that colon will be interpreted as amud dalet.
also, this regex is very difficult to understand. breaking into separate vars will help.

sefaria/model/schema.py
YishaiGlasner: aha, this is for matching the amud_offset. this can be understood only at the end. maybe giving name to the enumerate object will make it clear.

sefaria/model/schema.py
```diff
@@ -2499,21 +2499,19 @@ def toNumber(self, lang, s, **kwargs):
...
+                    return daf - amud_offset
+            # if no match, guess it's amud alef
+            logger.warn(f"Couldn't parse folio amud from {s}. Assuming it's amud alef.")
```
YishaiGlasner, line 2512: i'm now running a script on this branch and for any yerushalmi ref like `ירושלמי ברכות פ״ח י״א ע״ג`  it logs this 10 times (it tries to parse the `פ״ח` as amud). the final result is goot, but it heavily pollutes the output.

nsantacruz: this is great feedback. I've changed it to throw ValueError which is consistent for other types of errors toNumber() throws. All tests pass.

sefaria/model/schema.py
nsantacruz: fixed this. I made some guesses as to what dots are used for amud gimmel and dalet but I don't think that's very critical at this point.

YishaiGlasner (approved):

PR #2354, opened 2025-02-26
chore(docs): Updated Name API documentation

saengel:
## Description
Updates the documentation of the Name API to account for a shift in parameters. The boolean `is_ref` has now become `type` which expects a String `ref` to indicate it should return refs only. 

## Code Changes
1. In `docs/openAPI.json` - updated the path documentation and the data schema, corrected any other references to the old boolean param. 

## Notes
1. This has been successfully tested with ReadMe, and makes the expected changes.

### DIFF
--- a/docs/openAPI.json
+++ b/docs/openAPI.json
@@ -5242,12 +5242,11 @@
             "in": "query"
           },
           {
-            "name": "ref_only",
-            "description": "By default `0` the Name API returns `Ref`s, book titles, authors, topics, and collections. If true, the response will only contain text references.",
+            "name": "type",
+            "description": "By default the Name API returns `Ref`s, book titles, authors, topics, and collections. If `type=ref`, the response will only contain text references.",
             "schema": {
               "enum": [
-                "0",
-                "1"
+                "ref"
               ],
               "type": "string"
             },
@@ -5611,9 +5610,9 @@
             ],
             "type": "string"
           },
-          "is_ref": {
-            "description": "Returns `true` if the submitted text is a valid Sefaria textual reference",
-            "type": "boolean"
+          "type": {
+            "description": "By default the Name API returns `Ref`s, book titles, authors, topics, and collections. If `type=ref`, the response will only contain text references.",
+            "type": "string"
           },
           "completions": {
             "description": "A list of autocompletion responses for the submitted text.",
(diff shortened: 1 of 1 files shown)


stevekaplan123 (commented):
This is interesting to see.  I've never seen openAPI json files before.  Everything here makes sense to me, but "canonical" is spelled wrong :-)
Also, what tool do we use to render the json as HTML?

saengel:
> This is interesting to see. I've never seen openAPI json files before. Everything here makes sense to me, but "canonical" is spelled wrong :-) Also, what tool do we use to render the json as HTML?

Thanks @stevekaplan123 ! Happy to discuss more about how we work with our OpenAPI spec - feel free to ping me. And whoops! Great spelling catch, will fix! Thank you!

docs/openAPI.json
```diff
@@ -5611,9 +5610,9 @@
             ],
             "type": "string"
           },
-          "is_ref": {
-            "description": "Returns `true` if the submitted text is a valid Sefaria textual reference",
-            "type": "boolean"
+          "type": {
+            "description": "By default the Name API returns `Ref`s, book titles, authors, topics, and collections. If `type=ref`, the response will only contain text references.",
```
yitzhakc, line 5614: Perhaps I misunderstood, but this doesn't really describe what is returned in `type`.. Is it just the parameter passed in?
Apart from `ref` what are the other options?

yitzhakc (changes requested):

docs/openAPI.json
saengel: I believe the only other option is `None`, but @yonadavGit will have more insight here. 

See the code here: https://github.com/Sefaria/Sefaria-Project/blob/3384ef0e569d6927284bf9df42ad83619212a65f/reader/views.py#L2732

saengel: Spoke to @yonadavGit - I was wrong, the full list of type options can be seen [here](https://github.com/Sefaria/Sefaria-Project/blob/eaa4bd5c3639ceff1b8f70ec78232451736b3de3/sefaria/model/autospell.py#L252-L252) :)

Will work on updating the PR

yitzhakc (approved):

PR #2357, opened 2025-03-02
Refactor Language Toggle

saengel:
## Description
- Refactoring the logged out language toggle to use our generic components
- In the process, consolidated `<DropdownMenuItem>` and `<DropdownMenuItemWithCallback>` into one component that can handle all possible configurations (@stevekaplan123 - would love your eyes on this in sheets to make sure I didn't break anything). 

## Code Changes
1. Subtle class name changes in `static/css/s2.css`, the additional of an extra flexbox div to enable the languages to shuffle (i.e. have the selected language rise to the top) without interfering with the title. 
2. `static/js/Misc.jsx` - Main body of work. 
3.  `static/js/common/DropdownMenu.jsx` - Refactoring and consolidation of `<DropdownMenuItem>`
4. `static/js/sheets/SheetOptions.jsx` - Implementing the consolidated `<DropdownMenuItem>` (@stevekaplan123 - this is the spot to check.

### DIFF
--- a/static/js/common/DropdownMenu.jsx
+++ b/static/js/common/DropdownMenu.jsx
@@ -11,51 +11,31 @@ const DropdownMenuSeparator = () => {
 
 }
 
-const DropdownMenuItem = ({url, children, newTab, onClick = null, customCSS = null, preventClose = false}) => {
-
-  const className = customCSS ? customCSS : 'interfaceLinks-option int-bi dropdownItem';
-  const commonProps = {
-    className,
-    'data-prevent-close': preventClose,
-    target: (newTab ? '_blank' : null),
-  };
+const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose = false}) => {
 
   if (!newTab){
     newTab = false;
   }
 
-  if (onClick && url) {
-    return (
-      <a {...commonProps}
-         href={url}
-         onClick={onClick}>
+  const cssClasses = customCSS ? customCSS : 'interfaceLinks-option int-bi dropdownItem';
+
+  return (
+    <a className={cssClasses}
+       href={url}
+       target={newTab ? '_blank' : null}
+       data-prevent-close={preventClose}>
+      {children}
+    </a>
+
+  );
+}
+
+const DropdownMenuItemWithCallback = ({onClick, children, preventClose = false}) => {
+  return (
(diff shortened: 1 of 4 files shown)


static/js/common/DropdownMenu.jsx
```diff
@@ -11,29 +11,51 @@ const DropdownMenuSeparator = () => {
...
   }
 
-  return (
-    <a className={`interfaceLinks-option int-bi dropdownItem`}
-       href={url}
-       target={newTab ? '_blank' : null}
-       data-prevent-close={preventClose}>
-      {children}
-    </a>
+  if (onClick && url) {
```
stevekaplan123, line 27: I originally implemented DropDownMenuItem like this, but Ephraim felt that all of these conditionals warranted making DropdownMenuItemWithCallback its own component.

static/js/common/DropdownMenu.jsx
```diff
@@ -11,29 +11,51 @@ const DropdownMenuSeparator = () => {
 
 }
 
-const DropdownMenuItem = ({url, children, newTab, preventClose = false}) => {
+const DropdownMenuItem = ({url, children, newTab, onClick = null, customCSS = null, preventClose = false}) => {
```
stevekaplan123, line 14: All cases share this code:
```
 const className = customCSS ? customCSS : 'interfaceLinks-option int-bi dropdownItem';
  const commonProps = {
    className,
    'data-prevent-close': preventClose,
    target: (newTab ? '_blank' : null),
  };
```
and all cases wrap children.
What if we had two cases: DropdownMenuItemLink (which uses an a tag) and DropdownMenuItemButton (which uses a div).  Then, there could be a helper function getCommonProps that would determine commonProps and both DropdownMenuItems could call it.

YishaiGlasner: this seems to me too complicated. i don't know these components, and didn't fully understand it, but it seems that we're trying to do too much with one component. i think the architecture should be re-considered, but for now i have no constructive idea.

saengel: Thank you both for the helpful input! Turns out I overcomplicated the code and it didn't actually need a callback, so for now reverted it back to where it was with the two separate components. Feel free to take a look again and see if it makes more sense now.

PR #2368, opened 2025-03-05
Add "saved" icon to the Library header

saengel:
## Description
Adding the bookmarks icon to the header - conditionally if the user is both logged in and in sheets mode. 

## Code Changes
1. `static/css/s2.css` - Styling to align the icon with the others in the header
2. `static/js/Header.jsx` - Adding the icon with conditional logic, and the module prop to `<Header />`
3. `static/js/ReaderApp.jsx` - Passing the new module prop. Note, this is hardcoded. See `TODO` for the reminder to integrate once the entirety of the header comes together. 

## Notes
- A question about what to do with the module prop on `<Header />` renderings on static pages. @yitzhakc and @edamboritz curious to get your insight when you have a chance.

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -205,6 +205,14 @@ class Header extends Component {
                 translationLanguagePreference={this.props.translationLanguagePreference}
                 setTranslationLanguagePreference={this.props.setTranslationLanguagePreference} /> : null}
 
+        { Sefaria._uid && this.props.module ==="library" ?
+        <div className='librarySavedIcon'>
+          <a href="/texts/saved" >
+            <img src='/static/icons/bookmarks.svg' />
+          </a>
+          </div>
+        : null }
+
           <ModuleSwitcher />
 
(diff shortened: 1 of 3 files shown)


YishaiGlasner (approved):

static/js/Header.jsx
```diff
@@ -205,6 +205,14 @@ class Header extends Component {
                 translationLanguagePreference={this.props.translationLanguagePreference}
                 setTranslationLanguagePreference={this.props.setTranslationLanguagePreference} /> : null}
 
+        { Sefaria._uid && this.props.module ==="library" ?
```
stevekaplan123, line 208: I think it'd be easier to read like this:
const librarySavedIcon = <div className='librarySavedIcon'>
          <a href="/texts/saved" >
            Saved items
          </a>
          </div>;
...
        { Sefaria._uid && this.props.module ==="library" && librarySavedIcon }

PR #2371, opened 2025-03-05
Bug/sc 32962

YishaiGlasner:
### Fix bugs that caused by lack of diffrentiaion between elements
This PR solves 2 similar bugs that occurs when we have more than 1 main `ReaderPanel`:
1. Toggle switches of the second panel influence the first one.
2. The radio buttons for language are all not checked.

## Code Changes
`ReaderPanels`'s prop was added to `ReaderPanelContext`, and used for differentiating `name`s and `id`s of `RadioButton`s and `ToggleSwitch`es in different panels

## Bonus
`LayoutButton`s that are called with map function by `LayoutButtons` was added `key`s (this is not related to any bug, but similar to the main issue of this PR).


static/js/common/ToggleSwitchLine.jsx
```diff
@@ -9,7 +9,6 @@ function ToggleSwitchLine({name, onChange, isChecked, text, disabled=false}) {
             <InterfaceText>{text}</InterfaceText>
             <ToggleSwitch
                 name={name}
-                id={name}
                 disabled={disabled}
                 onChange={onChange}
                 isChecked={isChecked}
```
yitzhakc, line 14: Since you removed the `ariaLabelledBy` from the `ToggleSwitch` component, please remove it here as well.

static/js/common/RadioButton.js
```diff
@@ -2,16 +2,16 @@ import React from "react";
...
         <div
             className='button'
             onClick={onClick}
         >
-            <label htmlFor={value}><InterfaceText>{label}</InterfaceText></label>
+            <label htmlFor={id}><InterfaceText>{label}</InterfaceText></label>
             <input
                 type='radio'
-                id={value}
+                id={id}
```
yitzhakc, line 14: Since the `input`'s `id` is now populated based on the value passed in as the `id` variable, we probably want to pass in an `id` wherever we use `RadioButton` (which in `master` seems to be only in [static/js/LayoutButtons.jsx](https://github.com/Sefaria/Sefaria-Project/blob/3384ef0e569d6927284bf9df42ad83619212a65f/static/js/LayoutButtons.jsx#L7))

yitzhakc (changes requested):

yitzhakc (approved):

PR #2378, opened 2025-03-09
Modularization Header

saengel:
## Description
The full logic for the modularized header including:
- Secondary navigation dropdown logic for logged in/logged out states on both sheets and library
- Conditionally rendering links for sheets and library states

## Code Changes
- Changes from @edamboritz were on the header logic and themable CSS
- Changes from @saengel relate to the secondary navigation and conditionally rendered links. 
- @edamboritz's code introduced the `activeModule` logic which is passed into `<Header />` and through to the sub-components.


reader/views.py
```diff
@@ -213,6 +213,7 @@ def base_props(request):
             "last_place": []
         }
     user_data.update({
+        "activeModule": getattr(request, 'active_module', "library"),
```
yitzhakc, line 216: Let's try and be consistent about the quote type

sefaria/views.py
```diff
@@ -63,13 +64,26 @@
...
 from sefaria.helper.link import add_links_from_csv, delete_links_from_text, get_csv_links_by_refs, remove_links_from_csv
+from sefaria.forms import SefariaPasswordResetForm, SefariaSetPasswordForm, SefariaLoginForm
 
 if USE_VARNISH:
     from sefaria.system.varnish.wrapper import invalidate_index, invalidate_title, invalidate_ref, invalidate_counts, invalidate_all
 
 import structlog
 logger = structlog.get_logger(__name__)
 
+class CustomLoginView(LoginView):
```
yitzhakc, line 75: The idea of a mixin is to define a class which **only** has the `get_context_data` function, and then have your custom views inherit both from the relevant base class (`LoginView`, `LogoutView`, etc.) _and_ the mixin class.

E.g. 
```
class StaticViewMixin:
    def get_context_data(self, **kwargs):
        ...

class CustomLoginView(StaticViewMixin, LoginView):
    pass
```
The inheritance order matters (due to Python's MRO - method resolution order)

sefaria/views.py
```diff
@@ -63,13 +64,26 @@
...
+    authentication_form = SefariaLoginForm
```
yitzhakc, line 76: I think we set this in the Django settings, so not sure we need to explicitly set it here

static/js/Header.jsx
```diff
@@ -90,12 +91,36 @@ const LoggedInDropdown = () => {
...
+                <DropdownMenuItem url={'/my/profile'}>
+                    <InterfaceText>Profile</InterfaceText>
+                </DropdownMenuItem>
+                <DropdownMenuItem url={'/sheets/saved'}>
+                  <InterfaceText>Saved</InterfaceText>
+                </DropdownMenuItem>
+                <DropdownMenuItem url={'/sheets/history'}>
+                  <InterfaceText>History</InterfaceText>
+                </DropdownMenuItem>
+                <DropdownMenuItem url={'/settings/account'}>
```
yitzhakc, line 118: If we have settings in both menus, it can be written just once, outside the module conditionals

templates/base.html
```diff
@@ -157,7 +163,7 @@
     <script src="https://fd810a0513c94a16a52ef4d0d9b9c6c8.js.ubembed.com" async></script> 
 </head>
 
-<body class="interface-{% if request.interfaceLang %}{{request.interfaceLang}}{% else %}english{% endif %}{% if EMBED %} embeded{% endif %} {% block bodyclasses %}{% endblock %}">
+<body data-active-module="{{ active_module }}" data-render-static="{{ renderStatic }}" class="interface-{% if request.interfaceLang %}{{request.interfaceLang}}{% else %}english{% endif %}{% if EMBED %} embeded{% endif %} {% block bodyclasses %}{% endblock %}">
```
yitzhakc, line 166: Not sure if we actually want to leave these two data attributes for posterity 😅

yitzhakc (changes requested):

sefaria/views.py
saengel: Where can I go to double check?

sefaria/views.py
saengel: Ah, fantastic - thank you! This makes a lot of sense.

PR #2379, opened 2025-03-10
Form start analytics event

nsantacruz:
## Description
Adds 'form_name', 'form_destination' ga4 event types to analyticsEventTracker
Adds inputStart JS event to analyticsEventTracker so we can just track the first input to an element
Uses these features to implement analytics for topic landing newsletter

### DIFF
--- a/static/js/analyticsEventTracker.js
+++ b/static/js/analyticsEventTracker.js
@@ -9,6 +9,7 @@ const AnalyticsEventTracker = (function() {
     const FIELD_ATTR_PREFIX = 'data-anl-';
     const BATCH_ATTR = 'data-anl-batch';
     const SCROLL_INTO_VIEW_SELECTOR ='[data-anl-event*=":scrollIntoView"]';
+    const INPUT_START_SELECTOR ='[data-anl-event*=":inputStart"]';
     const TOGGLE_NODE_SELECTOR = 'details';
     const NON_BUBBLING_EVENT_DELEGATES = {
         'bubblingToggle': 'toggle',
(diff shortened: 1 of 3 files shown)


static/js/analyticsEventTracker.js
```diff
@@ -260,6 +275,8 @@ const AnalyticsEventTracker = (function() {
                 } else if (eventType === 'scrollIntoView') {
                     onMutationFns.push(node => _runFunctionOnSelector(node, SCROLL_INTO_VIEW_SELECTOR, _observeForScrollIntoView));
                     eventType = _observeScrollIntoViewEvent(selector);
+                } else if (eventType === 'inputStart') {
```
yonadavGit, line 278: not sure, but can these if-else blocks be replaced with an iteration over a datastucture mapping eventTypes strings with functions? Will that complicate things?

nsantacruz: I could do something like this

```javascript
            const onMutationFns = [];
            const mutationsByEventType = {
                'toggle': {selector: TOGGLE_NODE_SELECTOR, callback: _addToggleListener},
                'scrollIntoView': {selector: SCROLL_INTO_VIEW_SELECTOR, callback: _observeForScrollIntoView},
                'inputStart': {selector: INPUT_START_SELECTOR, callback: _addInputStartListener},
            }
            for (let eventType of eventTypes) {
                if (eventType in mutationsByEventType) {
                    const {selector, callback} = mutationsByEventType[eventType];
                    onMutationFns.push(node => _runFunctionOnSelector(node, selector, callback));
                }
                if (eventType === 'toggle') {
                    eventType = Object.keys(NON_BUBBLING_EVENT_DELEGATES)[0];
                } else if (eventType === 'scrollIntoView') {
                    _observeScrollIntoViewEvent(selector);
                }
                elements.forEach(element => {
                    element.addEventListener(eventType, _handleAnalyticsEvent);
                });
            }
```

But because there are already differences between toggle and scrollIntoView I can't completely get rid of those if-statements. Do you think this is cleaner?

yonadavGit: Okay, I guess leaving it as it is would be simpler

yonadavGit (approved):

PR #2380, opened 2025-03-10
Bug/sc 32525/when opening all tanakh links in sidebar

YishaiGlasner:
## Description
Fix for not showing 'essay' links in their category filter (e.g. Steinslaz intro in Tanakh).

## Code Changes
In `TextList` use `Sefaria.getLinksFromCacheAndPreprocess` rather than `Sefaria.getLinksFromCache` for filtering the 'essay' links.

## Notes
The code here is very old. I'm not sure why we have different code for getting links in `TextList` and `ConnectionsSummary`. It seems duplicate but the old code is too complicated and hardly readable.

### DIFF
--- a/static/js/LayoutButtons.jsx
+++ b/static/js/LayoutButtons.jsx
@@ -59,6 +59,7 @@ const LayoutButtons = () => {
             <InterfaceText>Layout</InterfaceText>
             <div className="layout-options">
                 {layoutOptions[layoutState].map(option => <LayoutButton
+                    key={option}
                     layoutOption={option}
                     layoutState={layoutState}
                 />)}
--- a/static/js/ReaderPanel.jsx
+++ b/static/js/ReaderPanel.jsx
@@ -698,6 +698,7 @@ class ReaderPanel extends Component {
       vowelsAndCantillationState: this.state.settings.vowels,
       punctuationState: this.state.settings.punctuationTalmud,
       width: this.state.width,
+      panelPosition: this.props.panelPosition,
     };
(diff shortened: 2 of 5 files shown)


stevekaplan123:
I'm wondering what you mean by "different code for getting links in TextList and ConnectionsSummary".  Do you mean calling linkSummary in ConnectionsSummary?  Other than that, this PR looks straightforward. @YishaiGlasner

YishaiGlasner:
> I'm wondering what you mean by "different code for getting links in TextList and ConnectionsSummary". Do you mean calling linkSummary in ConnectionsSummary? Other than that, this PR looks straightforward. @YishaiGlasner

@stevekaplan123  i mean `TextList` has a function `getLinks` of 40 lines, `ConnectionsSummary` has (in `render`) a long logc for getting the `summary` which should be similar.

stevekaplan123:
> > I'm wondering what you mean by "different code for getting links in TextList and ConnectionsSummary". Do you mean calling linkSummary in ConnectionsSummary? Other than that, this PR looks straightforward. @YishaiGlasner
> 
> @stevekaplan123 i mean `TextList` has a function `getLinks` of 40 lines, `ConnectionsSummary` has (in `render`) a long logc for getting the `summary` which should be similar.

I may not be right but I think this is OK because the linkSummary has a corresponding cache and so won't actually run twice.  See here:
```
    if (!this.shouldBuildLinkSummaries(cacheKey, links)) {  // don't need to build _linkSummaries for this ref
      return this._linkSummaries[cacheKey].data;
    }
```

yitzhakc (approved):

PR #2381, opened 2025-03-11
Bug/sc 520/second try

stevekaplan123:
## Description
In this PR, I fixed two bugs.  For most cases, such as Rashi, we were only making one API call to the section ref, but it was the wrong section ref.  For some cases, like Malbim Ayelet HaShachar, we were making hundreds of unnecessary API calls.

## Code Changes
1. Before implementing a fix, I re-factored the TextList component somewhat, which renders the comments in the Connections Panel.  First of all, the links displayed in the TextList are now part of the component's state.  Previously, we calculated it each render.  Second, we used to handle changing the Connections Panel filter ("Rashi" to "Ibn Ezra") in one way and handle clicking on a different segment in the reader in a different way, but I felt it was simpler to handle them together.  Now, when the filter or ref is changed, we call `loadConnections`.  Here's how my approach works -- Suppose the reader panel is at "Genesis 3:2" and the filter is "Rashi":
(a) First, `loadConnections` calls the related API to make sure we have all the links for the "Genesis 3" (or grab the cached result if we've already called the related API for "Genesis 3" before)
(b) Second, filter the links returned by the related API/cache as we only want to display links that are connected to "Rashi on Genesis 3:2"
(c) Finally, **if the connections panel filter is a commentary**, as it is in the case of Rashi, preload the commentary section ref text, which in this case is the text for "Rashi on Genesis 3:2".  This last step speeds things up because if the user were to click on any of the comments in "Rashi on Genesis 3:2", some of the data is now there and less API calls need to be made.  I believe the reason we only preload in the case of a commentary is because we assume it is much more likely to be opened by a user than a link to Talmud, Midrash, etc.

2.  Previously, preloadText dealt with two cases: (a) when the filter was a single commentary and (b) when the filter was all commentaries at once.  I removed case (b) because it can't possibly benefit us to call the text API for many commentaries when the user may not even open any of them.   Moreover, I re-factored the single commentary case so that now:
(a) It uses the new v3 text API to retrieve the text.
(b) Previously, we derived the commentary ref to retrieve incorrectly.  We would take a segment ref "Rashi on Genesis 1:3:5" and call the text API for the ref "Rashi on Genesis 1".  This is useless as we never try to access "Rashi on Genesis 1" in the cache.  Instead, when the user clicks "Open" on a link "Rashi on Genesis 1:3:5", we do try to access the cache for "Rashi on Genesis 1:3".  I use the Sefaria.zoomOutRef to "zoom out" from "Rashi on Genesis 1:3:5" to "Rashi on Genesis 1:3" so that we call the API on the appropriate ref.


yitzhakc (approved):

stevekaplan123:
@yitzhakc Not merging due to playwright tests.  Can you override it?  Thanks!

PR #2382, opened 2025-03-11
Render Profile Name in Library Dropdown as a non-link

saengel:
## Description
In the logged-in library dropdown, a need arose for a `DropdownMenuItem` component which was not a link, and did not require a callback (to render the logged-in user's name without a link to their profile). 

This PR renames the existing `<DropdownMenuItem/>` component (which renders an `<a>` tag) to `<DropdownMenuItemLink/>`, and creates a new `<DropdownMenuItem/>` component which renders a simple `<div>`. 

## Code Changes
In `static/js/Header.jsx` - importing the new `<DropdownMenuItemLink/>` component, and replacing all instances of <DropdownMenuItem/>` to `<DropdownMenuItemLink/>`. 

In `static/js/common/DropdownMenu.jsx` - defining the new components. 

## Notes
I checked all of the other files for uses that need to be replaced, and did not find any. Worth someone else triple checking though :)

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -94,7 +94,7 @@ const LoggedInDropdown = ({module}) => {
                 </DropdownMenuItem>
               }
                { module === 'sheets' && 
-                <DropdownMenuItemLink preventClose={true}>
+                <DropdownMenuItemLink url={'/my/profile'} preventClose={true}>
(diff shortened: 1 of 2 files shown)


stevekaplan123 (approved):
Super clear and simple!

PR #2385, opened 2025-03-12
Header Icons for Modularization

saengel:
## Description
Conditionally render the sheets vs library home icons in the upper left corner. 

## Code Changes
1. The sheets icons in Hebrew and English were added to `static/img/`
2. In `static/js/Header.jsx` we replaced `logo` with `sheetsLogo` and `libraryLogo` and conditionally render them. 

## Notes
The CSS aligning sheets logo in English with the other links might need a bit of tweaking, all of the other modes look good. Noting in case I don't have a chance to fix in this PR.


static/js/Header.jsx
```diff
@@ -222,7 +232,10 @@ class Header extends Component {
 
         <div className="headerNavSection">
           { Sefaria._siteSettings.TORAH_SPECIFIC ?
-          <a className="home" href="/" >{logo}</a> : null }
+          (this.props.module === "library" ? 
+              <a className="home" href="/" >{libraryLogo}</a>
```
YishaiGlasner, line 236: here also, define a logo variable, with the condition, and keep the jsx clean.
also in 283-5

static/js/Header.jsx
```diff
@@ -222,7 +232,10 @@ class Header extends Component {
 
         <div className="headerNavSection">
           { Sefaria._siteSettings.TORAH_SPECIFIC ?
```
YishaiGlasner, line 234: use && rather than ? : null
also in line 282

PR #2391, opened 2025-03-18
chore(cache): reset shared cache with node on request

akiva10b:
This pull request introduces a new way to reset the shared cache on a url request, this request is protected with a passcode to allow webhooks to use it

### New Feature: Passcode Protection for Admin Actions

* [`sefaria/decorators.py`](diffhunk://#diff-e45dc631b720efe54b9c60c34a0e8f736850b100f19cb935a2a2b8f5891958ceR1-R13): Added a new decorator `passcode_or_staff_required` that checks for a passcode or staff member status before allowing access to certain views.
* [`sefaria/views.py`](diffhunk://#diff-e30a05013192fa5147ab7d4c4b7dffb63644ac793c49029bfaa4ab3ee03b5441R36-R44): Added the `passcode_or_staff_required` decorator to the `rebuild_shared_cache` view function and imported the decorator and `STAFF_PASSCODE` from settings. [[1]](diffhunk://#diff-e30a05013192fa5147ab7d4c4b7dffb63644ac793c49029bfaa4ab3ee03b5441R36-R44) [[2]](diffhunk://#diff-e30a05013192fa5147ab7d4c4b7dffb63644ac793c49029bfaa4ab3ee03b5441R761-R768)

### Configuration Updates

* [`helm-chart/sefaria/templates/configmap/local-settings-file.yaml`](diffhunk://#diff-f9d6684a33c36a4171cf34f641fedcabdc58d55bb1f3eb8d3dd83d3726d319fbR350-R351): Added `STAFF_PASSCODE` environment variable to the configuration.
* `sefaria/local_settings_ci.py`, `sefaria/local_settings_coolify.py`, `sefaria/local_settings_example.py`: Added `STAFF_PASSCODE` environment variable to each of these settings files. [[1]](diffhunk://#diff-2393a5921bed1f008091517779c9e874d545f7b4b7b11487b5ec2fd53ba5dc2fR229-R230) [[2]](diffhunk://#diff-34f82b438261c80bda1d20118972ef79dab375e6d15e1adad9a1976c90307608R24-R25) [[3]](diffhunk://#diff-d1a1e328a4fa815512b6baa2f9f566df7e1b285c9105f9de652e61bb6690da83R341-R342)

### URL and View Updates

* [`sefaria/urls.py`](diffhunk://#diff-2ec2353562d79cc6a3e9b4d4f3e23d7384da02e486f3c3d2cc391e6dff3d4194R430): Added a new URL pattern for the `rebuild_shared_cache` view.


sefaria/local_settings_ci.py
```diff
@@ -226,3 +226,5 @@
     wrapper_class=structlog.stdlib.BoundLogger,
     cache_logger_on_first_use=True,
 )
+
+STAFF_PASSCODE = os.getenv("STAFF_PASSCODE")
```
yitzhakc, line 230: I don't think `STAFF_PASSCODE` is the right terminology here. 

The decorator is `passcode_or_staff_required` - but the passcode should not be confused with giving access to anything but some very specific (and locked down!) endpoints.

Perhaps `MAINTENANCE_URLS_PASSCODE`?

In addition, since this passcode is passed as a GET parameter, it is highly unsecured and will be "leaked" all over the place (logs, etc.)
You've implemented a mechanism to allow for separate passcodes for every instance of the decorator - maybe we should use that already, and this should be a single-use `SHARED_CACHE_REBUILD_PASSCODE`

sefaria/views.py
```diff
@@ -33,14 +33,15 @@
...
 import sefaria.model as model
 import sefaria.system.cache as scache
 from sefaria.helper.crm.crm_mediator import CrmMediator
 from sefaria.helper.crm.salesforce import SalesforceNewsletterListRetrievalError
-from sefaria.system.cache import in_memory_cache
+from sefaria.system.cache import get_shared_cache_elem, in_memory_cache, set_shared_cache_elem
 from sefaria.client.util import jsonResponse, send_email, read_webpack_bundle
 from sefaria.forms import SefariaNewUserForm, SefariaNewUserFormAPI, SefariaDeleteUserForm, SefariaDeleteSheet
-from sefaria.settings import MAINTENANCE_MESSAGE, USE_VARNISH, MULTISERVER_ENABLED
+from sefaria.settings import MAINTENANCE_MESSAGE, USE_VARNISH, MULTISERVER_ENABLED, STAFF_PASSCODE
```
yitzhakc, line 44: Please modify to `from django.conf import settings`
and modify the use of the attributes to be
`settings.STAFF_PASSCODE` etc.

yitzhakc (changes requested):

sefaria/system/tests/test_decorators.py
```diff
@@ -34,3 +40,79 @@ def test_catch_error():
...
+
+
+def test_missing_auth_header_anonymous_user(rf):
+    request = rf.get("/fake-endpoint/")
+    request.user = AnonymousUser()
+
+    wrapped = webhook_auth_or_staff_required(dummy_view)
+    response = wrapped(request)
+    # Since staff_member_required redirects by default
+    assert response.status_code in [302, 401]  # Depending on login settings
```
yitzhakc, line 97: I don't understand this comment, nor how the status code could by 302 or 401. The test should be deterministic (and if there's a redirect, we should resolve it first, and then check the status code)

sefaria/system/tests/test_decorators.py
```diff
@@ -34,3 +40,79 @@ def test_catch_error():
...
+
+
+def test_nonstaff_user_without_auth(rf, django_user_model):
+    user = django_user_model.objects.create_user(username="nonstaff", password="pass", is_staff=False)
+    request = rf.get("/fake-endpoint/")
+    request.user = user
+
+    wrapped = webhook_auth_or_staff_required(dummy_view)
+    response = wrapped(request)
+    assert response.status_code in [302, 401]
```
yitzhakc, line 118: As above, the response should be single deterministic

yitzhakc (changes requested):

yitzhakc (dismissed):

PR #2392, opened 2025-03-20
Feature/sc 31291/display topics in sheet sidebar

stevekaplan123:
## Description
This PR simply implements the display of topics and collections in the sidebar of the sheets viewer/editor.

## Code Changes
There are some minor changes dealing with /sheets related URLs that was merged into this PR at some point.  The only substantive change is in SheetContentSidebar where it implements a new component to display topics and collections.


static/js/ReaderApp.jsx
```diff
@@ -1155,7 +1155,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       this.openTopic(path.slice(8), params.get("tab"));
 
     } else if (path.match(/^\/sheets\/profile\/.+/)) {
-      this.openProfile(path.slice(9), params.get("tab"));
+      this.openProfile(path.slice(16), params.get("tab"));
```
yitzhakc, line 1158: I'd much rather you explicitly state `this.openProfile(path.splice(len('/sheets/profile/'), ...)` so that we don't have "magic numbers".

static/js/sheets/SheetContent.jsx
```diff
@@ -138,7 +138,7 @@ class SheetContent extends Component {
         ComponentToRender = SheetMedia;
       }
       else {
-        return <></>;  // handle bad data in sheet.sources
+        return <></>;  // there are cases in DB where sheets have problematic data in sources attribute such as a node of null with no other fields
```
yitzhakc, line 141: Do we want to log this info anywhere? Or it doesn't matter that a sheet has problematic data?

static/js/sheets/SheetContentSidebar.jsx
```diff
@@ -44,19 +45,31 @@ const SheetProfileInfo = ({profile, toggleSignUpModal}) => {
...
-                    </div>
-                </div>;
+const SheetSidebarList = ({items, type}) => {
+    const title = type === "topics" ? "Topics" : "Part of Collections";
+    const styleClass = type === "topics" ? "sheetTopicsList" : "sheetCollectionsList";
+    const renderItems = () => {
+      if (type === "topics") {
+        return items.map((item, i) => (
+            <li key={i}><a
+                href={"/topics/" + item.slug}><InterfaceText text={{en: item.en, he: item.he}}></InterfaceText></a>
```
yitzhakc, line 55: Does the URL need the `sheets` prefix? Here and on line 61

yitzhakc (changes requested):

static/js/sheets/SheetContent.jsx
stevekaplan123: I didn't research this to determine a cause.  It's possible that it's due to a no longer existing bug.  I think it's worth checking that (a) this isn't still happening on save (though it might be hard to reproduce) and (b) at the very least, that in these particular cases where a source is null that no buggy behavior occurs (for example, it's worth checking if google export works for these sheets).  Shall I make a card for this?

yitzhakc: Yes please

yitzhakc (changes requested):
I see you've merged `modularization-main` into this branch - but it doesn't look like `modularization-main` was ever merged into `sheets-editor-final`, resulting in this PR now having 90 changed files 🤯 
Could you please reset the branch and remove that merge? Or first merge `modularization-main`  into `sheets-editor-final`? Thank you!

static/js/sheets/SheetContentSidebar.jsx
stevekaplan123: It does work locally.  Curious to test the subdomain case -- Will there be sheets subdomains on cauldrons?

yitzhakc (approved):

PR #2397, opened 2025-04-02
Fix editing default ltr source and rtl translation

YishaiGlasner:
## Description
In the RTL project we've change the way of categorizing texts - rather than ltr/rtl dichotomy, we have now (also) primary/translation dichotomy. The edit api still interprets he/en as ltr/rtl.
Therefore, when trying to edit ltr source or rtl translation, that are default (and therefore are missing in `currentVersions`), one got the wrong text to edit.
This PR changes the way we inquire the details of the version to edit - rather than taking it from `currVersions`, it's taken from the data itself, so we always have explicit version details.

### DIFF
--- a/static/js/ConnectionsPanel.jsx
+++ b/static/js/ConnectionsPanel.jsx
@@ -1269,15 +1269,11 @@ WebPagesList.propTypes = {
 const AdvancedToolsList = ({srefs, canEditText, currVersions, setConnectionsMode, masterPanelLanguage, toggleSignUpModal}) => {
     const {textsData} = useContext(ReaderPanelContext);
     const editText = canEditText && textsData ? function () {
-      let refString = srefs[0];
-      const {primaryDirection, translationDirection} = textsData;
-      const currVersionsLangCode = masterPanelLanguage.slice(0,2);
-      const versionTitle = currVersions[currVersionsLangCode]?.versionTitle;
-      const direction = (masterPanelLanguage === 'english') ? translationDirection : primaryDirection;
-      const langCode = direction === 'rtl' ? 'he': 'en';
-      if (versionTitle) {
-        refString += "/" + encodeURIComponent(langCode) + "/" + encodeURIComponent(versionTitle);
-      }
+      const isTranslation = masterPanelLanguage === 'english';
+      const versionType = isTranslation ? 'translation' : 'primary';
+      const langCode = textsData[`${versionType}Direction`] === 'ltr' ? 'en': 'he';
+      const versionTitle = isTranslation ? textsData.versionTitle : textsData.heVersionTitle;
+      const refString = `${srefs[0]}/${encodeURIComponent(langCode)}/${encodeURIComponent(versionTitle)}`;
 
       let path = "/edit/" + refString;
       let currentPath = Sefaria.util.currentPath();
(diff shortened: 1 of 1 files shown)


static/js/ConnectionsPanel.jsx
```diff
@@ -1269,21 +1269,16 @@ WebPagesList.propTypes = {
 const AdvancedToolsList = ({srefs, canEditText, currVersions, setConnectionsMode, masterPanelLanguage, toggleSignUpModal}) => {
     const {textsData} = useContext(ReaderPanelContext);
```
stevekaplan123, line 1270: What is the context 'textsData'?

stevekaplan123: Got it thanks. I'm curious in general what the thinking is in storing so many props in ReaderPanelContext rather than passing down the props as we usually do.  Using context does seem like a cleaner approach than passing down a million props.  Is that the whole reason or is there more to it?

YishaiGlasner: when the props are shared by many components, or you can think about many cases to use them, Context is good solution. Here, we have ReaderPanel which have milions of descendants and there are general props you don't know where you'll want them (here, before rtl, we didn't use the data, but currVersions was enough; after rtl we need it. without this context we should pass it down thru other components making it more difficult to understand (which is already the situatuon with so many props).

static/js/ConnectionsPanel.jsx
```diff
@@ -1269,21 +1269,16 @@ WebPagesList.propTypes = {
...
-      let refString = srefs[0];
-      const {primaryDirection, translationDirection} = textsData;
-      const currVersionsLangCode = masterPanelLanguage.slice(0,2);
-      const versionTitle = currVersions[currVersionsLangCode]?.versionTitle;
-      const direction = (masterPanelLanguage === 'english') ? translationDirection : primaryDirection;
-      const langCode = direction === 'rtl' ? 'he': 'en';
-      if (versionTitle) {
-        refString += "/" + encodeURIComponent(langCode) + "/" + encodeURIComponent(versionTitle);
-      }
+      const isTranslation = masterPanelLanguage === 'english';
```
yitzhakc, line 1272: Is the `masterPanelLanguage` dependent in any way on the "interface" language? I.e Could we have a situation where someone is editing a text with their interface in Hebrew (i.e. at sefaria.org.il) which would then cause an un-intended side effect?

YishaiGlasner: no. `masterPanelLanguage` is the language of the `ReaderPanel`, which is the 'parent' of this `ConnectionsPanel`. it can be 'hebrew' (i.e. primary), 'english' (i.e. translation) or 'bilingual'.

static/js/ConnectionsPanel.jsx
```diff
@@ -1269,21 +1269,16 @@ WebPagesList.propTypes = {
...
+      const versionType = isTranslation ? 'translation' : 'primary';
+      const langCode = textsData[`${versionType}Direction`] === 'ltr' ? 'en': 'he';
+      const versionTitle = isTranslation ? textsData.versionTitle : textsData.heVersionTitle;
```
stevekaplan123, line 1275: I wonder if instead of hardcoded strings here, there could be a re-usable function to derive the versionTitle.  It seems awkward doing textsData[`$versionType}Direction`] and returning 'en' or 'he'.  It seems like we should have utility functions for deriving this kind of data.

PR #2398, opened 2025-04-02
Feature/sc 26865/sheets desktop topics display sheets only

stevekaplan123:
## Description
This PR implements a requirement for modularization -- while in the 'sheets' module, topic pages should only display sheet sources rather than library sources.

## Code Changes
1. I re-factored the TopicPage component to make the component smaller and easier to read.  I pulled out a lot of functionality into a new component TopicPageTabView which wrap the existing component TabView.
2. processTopicsTabsData -- This is the substantive change for this PR.  This function gets called whenever the topic API is called.  This function makes sure that the Sefaria._topics cache entry for the given topic contains the appropriate tabs.  In the case of the 'sheets' module, this function will now put the topics' ref links in the 'sheets' tab in the cache, whereas in the 'library' module, the refs will be stored under the 'sources' or 'notable-sources' tab.  For example, suppose I go to the library today to the topic "Pharoah".  Sefaria._topics Pharoah entry will look like this with 186 refs in the 'sources' tab:
image
In order to do this, I created a helper function _createTabKeyAndTitle to simplify the code.


static/css/s2.css
```diff
@@ -2784,7 +2784,9 @@ a.cardTitle:hover {
   font-size: 122%;
 }
 .interface-english .navBlockDescription.clamped .int-en,
-.interface-hebrew .navBlockDescription.clamped .int-he {
+.interface-hebrew .navBlockDescription.clamped .int-he,
+.cardDescription
+{
```
YishaiGlasner, line 2789: remove newline

static/js/TopicPage.jsx
```diff
@@ -60,16 +60,19 @@ const fetchBulkText = (translationLanguagePreference, inRefs) =>
 );
 
 
-const fetchBulkSheet = inSheets =>
-    Sefaria.getBulkSheets(inSheets.map(x => x.ref)).then(outSheets => {
-    for (let tempSheet of inSheets) {
-      if (outSheets[tempSheet.ref]) {
-        outSheets[tempSheet.ref].order = tempSheet.order;
+const fetchBulkSheet = inSheets => {
+   inSheets.forEach(x => {
```
YishaiGlasner, line 64: add space for right indent

static/js/TopicPage.jsx
```diff
@@ -60,16 +60,19 @@ const fetchBulkText = (translationLanguagePreference, inRefs) =>
...
+      x.ref = parseInt(x.ref.replace('Sheet ', ''));
+    });
+    return Sefaria.getBulkSheets(inSheets.map(x => x.ref)).then(outSheets => {
```
YishaiGlasner, line 67: it's the second loop on inSheets. you can do only one?

static/js/TopicPage.jsx
```diff
@@ -563,8 +567,100 @@ const getTopicPageAnalyticsData = (slug, langPref) => {
     };
 };
 
+const setupDisplayTabs = (topic, currentTab, tabDisplayData, refsToFetchByTab, topicData, setLoadedData) => {
```
YishaiGlasner, line 570: the diff here is to difficult for me to follow.
the existing code is terrible, and this tries to improve it, but on the one hand, this code is still not good enough for simply understanding it, and on the other hand it's too different from the original for understanding what was done.
@yitzhakc do you have an dea how to review in such a case?

static/js/sefaria/sefaria.js
```diff
@@ -2871,27 +2871,37 @@ _media: {},
...
+      linkTypeObj.refs = linkTypeObj.refs.filter(ref => (Sefaria.activeModule === 'sheets') === ref.is_sheet);
       for (let refObj of linkTypeObj.refs) {
-        // sheets are no longer displayed on topic pages
-        if (refObj.is_sheet) { continue; }
+
+        //various tabs are "Admin", "Notable Sources", "Sources", "Sheets", and "Top Citations".
         let tabKey = linkTypeSlug;
-        if (tabKey === 'about') {
+        let { title } = linkTypeObj;
+        if (tabKey === 'about') {  // "Sheets", "Sources", and "Notable Sources" are tagged 'about' in the DB so we have to differentiate between them here
```
YishaiGlasner, line 2885: all of these conitions are difficult to read.
here is gpt suggestion. it's more readable, but i didn't check it it really does the same thing... :

if (tabKey === 'about') {
  const desc = refObj.descriptions?.[lang];
  const isNotableSource = (desc?.title || desc?.prompt) && desc?.published !== false;

  if (Sefaria.activeModule === 'sheets') {
    tabKey = 'sheets';
    title = { en: 'Sheets', he: Sefaria.translation('hebrew', 'Sheets') };
  } 
  else if (isNotableSource) {
    tabKey = 'notable-sources';
    title = { en: 'Notable Sources', he: Sefaria.translation('hebrew', 'Notable Sources') };
  } 
  else {
    tabKey = 'sources';
    title = { en: 'Sources', he: Sefaria.translation('hebrew', 'Sources') };
  }
}

stevekaplan123:
@yitzhakc @YishaiGlasner 
In sefaria.js, I renamed processTopicsData to processTopicsTabsData since that's more descriptive of its purpose.  I added comments and added helper functions inside processTopicsTabsData so that hopefully it's more clear and organized.  
In TopicPage.jsx, I divided up the helper function setupDisplayTabs into two functions -- one for adding tabs that display sources and one for miscellaneous tabs and added comments.  Hope it's more clear this time!

static/css/s2.css
```diff
@@ -2783,8 +2783,8 @@ a.cardTitle:hover {
 .navBlockTitle .int-he, .navBlockDescription .int-he, .cardTitle .int-he, .cardDescription .int-he{
   font-size: 122%;
 }
-.interface-english .navBlockDescription.clamped .int-en,
-.interface-hebrew .navBlockDescription.clamped .int-he {
+.interface-english .navBlockDescription.clamped .int-en, .interface-hebrew .navBlockDescription.clamped .int-he, .cardDescription
```
YishaiGlasner, line 2786: i meant that our convention is no new line for the opening curl:
.interface-english .navBlockDescription.clamped .int-en,
.interface-hebrew .navBlockDescription.clamped .int-he,
.cardDescription {

YishaiGlasner:
> @yitzhakc @YishaiGlasner In sefaria.js, I renamed processTopicsData to processTopicsTabsData since that's more descriptive of its purpose. I added comments and added helper functions inside processTopicsTabsData so that hopefully it's more clear and organized. In TopicPage.jsx, I divided up the helper function setupDisplayTabs into two functions -- one for adding tabs that display sources and one for miscellaneous tabs and added comments. Hope it's more clear this time!

ok. it seems i can't do it in a reasonable time. it looks better than it was, but i didn't know this code and this pr is not the best way to know it for the first time.

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,90 @@ _media: {},
...
+      else {
+        if (linkType === 'about') {
+          const desc = refObj.descriptions?.[lang];
+          const isNotableSource = (desc?.title || desc?.prompt) && desc?.published !== false;
+          return isNotableSource
+            ? { tabKey: 'notable-sources', title: { en: 'Notable Sources', he: Sefaria.translation('hebrew', 'Notable Sources') } }
+            : { tabKey: 'sources', title: { en: 'Sources', he: Sefaria.translation('hebrew', 'Sources') } };
         }
+        if (linkType === 'popular-writing-of') {
+          return { tabKey: linkType, title: topicLinks.title };  // in this case the title is already set to "Top Citations"
```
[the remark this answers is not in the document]
stevekaplan123: This is great.  Thank you CoPilot!

static/js/TopicPage.jsx
```diff
@@ -61,16 +61,20 @@ const fetchBulkText = (translationLanguagePreference, inRefs) =>
 );
 
 
-const fetchBulkSheet = inSheets =>
-    Sefaria.getBulkSheets(inSheets.map(x => x.ref)).then(outSheets => {
+const fetchBulkSheet = inSheets => {
+  const refs = inSheets.map(x => {
```
akiva10b, line 65: add a comment that getBulkSheets expects full refs without "sheets "

akiva10b (commented):
have a look at other tab views and see if there are any reusable components here

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
+      // In the sheets module, every source is under the "Sheets" tab.
+      // In the library module, author topics links "Top Citations" are of type "popular-writing-of" and all other topics are of type "about".
+      if (refObj.is_sheet) {
+        return { tabKey: 'sheets', title: { en: "Sheets", he: Sefaria.translation('hebrew', "Sheets") } };
+      }
+      else {
+        if (linkType === 'about') {
+          const lang = Sefaria.interfaceLang === "hebrew" ? 'he' : 'en';
+          const desc = refObj.descriptions?.[lang];
+          const isNotableSource = (desc?.title || desc?.prompt) && desc?.published !== false;
```
yitzhakc, line 2884: I don't love this - the frontend code shouldn't need such complicated logic to figure out if it's a notable source or not - it should be part of the data...
@yonadavGit any thoughts?

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
-    const tabs = {};
-    for (let [linkTypeSlug, linkTypeObj] of Object.entries(data.refs)) {
-      for (let refObj of linkTypeObj.refs) {
-        // sheets are no longer displayed on topic pages
-        if (refObj.is_sheet) { continue; }
-        let tabKey = linkTypeSlug;
-        if (tabKey === 'about') {
-            // Mark as notable if it has a prompt, title, and isn't explicitly unpublished (happens when returns from LLM pod);
-            tabKey = (refObj.descriptions?.[lang]?.title || refObj.descriptions?.[lang]?.prompt) && refObj.descriptions?.[lang]?.published != false ? 'notable-sources' : 'sources';
+  _createTabKeyAndTitle: function (linkType, refObj, lang) {
```
yitzhakc, line 2872: I'd ask that you change the name to `_getTabKey` - since ultimately, we are not creating the tab here (that happens in `processTopicsTabsData`), we're just "divining" it out based on the data provided.

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
+    if (!topicData || !topicData.refs) { return topicData || null; }
+    const tabs = {};
+    for (let [linkType, topicLinks] of Object.entries(topicData.refs)) {
+        let refs = topicLinks.refs.filter(refObj => (Sefaria.activeModule === 'sheets') === refObj.is_sheet);
+        for (let refObj of refs) {
+          const {tabKey, title} = Sefaria._createTabKeyAndTitle(linkType, refObj);
+          if (!tabKey) continue;
+          if (!tabs[tabKey]) {
+          tabs[tabKey] = {
+              refs: new Set(),
```
yitzhakc, line 2907: Very confusing that you have `refs` here as a key in this object, and also a `refs` variable (which should probably be a `const`) on line 2901. Please change the name of one of them.

yitzhakc (changes requested):

yitzhakc:
> > have a look at other tab views and see if there are any reusable components here
> 
> I don't see how to immediately make TopicPageTabView work based on purely reusable components, although I noticed a few things that could be cleaned up. First of all, every client of TabView creates an array 'tabs' which is just a list of javascript objects and the client also passes a renderTab function to TabView. These renderTab functions are very similar to each other and it seems like we should just have a component Tab that handles the data passed to the array and also can render itself. I also found one glaring issue that entangles TabView (which we use many times on the site) to TopicPageTabView. Strangely, the filter button and lang toggle for TopicPageTabView are considered tabs which makes no sense conceptually and also entangles the two components together more than they need to be. In the CollectionsPage we also consider filter to be a tab so that component could also be cleaned up by separating filter/lang toggles out of TabView.
> 
> @yitzhakc

I agree @stevekaplan123, I would love for us to turn the filter button and language toggle into separate components and not tabs. But I don't think it should be done in this PR, we should create a separate story and PR so as not to break the scope here.

static/js/sefaria/sefaria.js
stevekaplan123: @yitzhakc I totally agree with you.  The bizarre linkType "about" should be divided into two types, but it is currently referenced in many places in our backend code, so the necessary changes would be a fairly significant re-factor. I think it should be its own story.

static/css/s2.css
YishaiGlasner: it still there

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
-            tabKey = (refObj.descriptions?.[lang]?.title || refObj.descriptions?.[lang]?.prompt) && refObj.descriptions?.[lang]?.published != false ? 'notable-sources' : 'sources';
+  _deriveTabDataForTopicLink: function (linkType, refObj) {
+      // helper function for processTopicsTabsData
+      // We support two types of 'linkType's -- confusingly called "about" and "popular-writing-of". This distinction is only relevant for the library module.
+      // In the sheets module, every source is under the "Sheets" tab.
+      // In the library module, author topics links "Top Citations" are of type "popular-writing-of" and all other topics are of type "about".
+      if (refObj.is_sheet) {
+        return { tabKey: 'sheets', title: { en: "Sheets", he: Sefaria.translation('hebrew', "Sheets") } };
+      }
+      else {
```
YishaiGlasner, line 2880: i think the convention in js is `else if (...) {...}`. you don't need {} for the else itself.

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
+        if (linkType === 'about') {
+          const lang = Sefaria.interfaceLang === "hebrew" ? 'he' : 'en';
```
YishaiGlasner, line 2882: i think `Sefaria.interfaceLang.slice(0, 2)` is more readable.

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
-    const tabs = {};
-    for (let [linkTypeSlug, linkTypeObj] of Object.entries(data.refs)) {
-      for (let refObj of linkTypeObj.refs) {
-        // sheets are no longer displayed on topic pages
-        if (refObj.is_sheet) { continue; }
-        let tabKey = linkTypeSlug;
-        if (tabKey === 'about') {
-            // Mark as notable if it has a prompt, title, and isn't explicitly unpublished (happens when returns from LLM pod);
-            tabKey = (refObj.descriptions?.[lang]?.title || refObj.descriptions?.[lang]?.prompt) && refObj.descriptions?.[lang]?.published != false ? 'notable-sources' : 'sources';
+  _deriveTabDataForTopicLink: function (linkType, refObj) {
```
YishaiGlasner, line 2872: it seems that both functions get some very complicated data and doing something with it. it's not readable (i don't blame, i guess this is due the data and code state, but i can't understand what's going on here).

static/js/TopicPage.jsx
```diff
@@ -543,24 +548,19 @@ const getTopicPageAnalyticsData = (slug, langPref) => {
...
-    const [showFilterHeader, setShowFilterHeader] = useState(false);
-    const [showLangSelectInterface, setShowLangSelectInterface] = useState(false);
     const [portal, setPortal] = useState(null);
-    const tabDisplayData = useTabDisplayData(translationLanguagePreference, versionPref);
     const topicImage = topicData.image;
-
     const scrollableElement = useRef();
     const clearAndSetTopic = (topic, topicTitle) => {setTopic(topic, topicTitle)};
+    const classStr = classNames({topicPanel: 1, readerNavMenu: 1});
+    const [langPref, setLangPref] = useState(Sefaria.interfaceLang);
```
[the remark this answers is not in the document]
yitzhakc: Out of curiosity, why?

static/js/sefaria/sefaria.js
yitzhakc: @stevekaplan123 I think add a comment explaining what this helper function is doing and what it returns.

static/js/sefaria/sefaria.js
yitzhakc: @YishaiGlasner it looks like there are two `if`s inside the `else` block, so @stevekaplan123 would still need a block - but I agree that the `else` doesn't need to be on a new line :-)

static/js/sefaria/sefaria.js
yitzhakc: @YishaiGlasner I don't think that's so readable either 🙈  

@stevekaplan123 maybe better to add a helper function to `Sefaria.js` that returns `he` or `en`? I.e that encapsulates this logic? It feels like we have it in lots of places...

static/js/sefaria/sefaria.js
```diff
@@ -2864,72 +2863,94 @@ _media: {},
...
+  processTopicsTabsData: function(topicData) {
+    // In the library module, topic tabs can be "Admin", "Notable Sources", "Sources", and "Top Citations".  In the sheets module, the only topic tab is "Sheets"
+    // This function takes the ref topic links in the topicData and puts each ref in the appropriate tab.
+    if (!topicData || !topicData.refs) { return topicData || null; }
+    const tabs = {};
+    for (let [linkType, topicLinks] of Object.entries(topicData.refs)) {
+        const activeRefs = topicLinks.refs.filter(refObj => (Sefaria.activeModule === 'sheets') === refObj.is_sheet);
+        for (let refObj of activeRefs) {
+          const {tabKey, title} = Sefaria._deriveTabDataForTopicLink(linkType, refObj);
+          if (!tabKey) continue;
```
yitzhakc, line 2904: Do we want a more robust error handling for missing tab data? Should we be reporting/logging it? What are the scenarios that we could reach such a situation?

static/js/TopicPage.jsx
```diff
@@ -464,8 +468,9 @@ const AuthorIndexItem = ({
 };
 
 
-const useTabDisplayData = (translationLanguagePreference) => {
-  const getTabDisplayData = useCallback(() => [
+const useAllPossibleSourceTabs = (translationLanguagePreference) => {
```
yitzhakc, line 471: This is a render function?

yitzhakc (changes requested):

static/js/TopicPage.jsx
```diff
@@ -799,6 +641,181 @@ TopicPage.propTypes = {
   topicTestVersion:    PropTypes.string
 };
 
+const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePreference, versionPref,
```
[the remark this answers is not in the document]
stevekaplan123: The logic hasn't changed.  I have only re-organized the code.

static/js/sefaria/sefaria.js
stevekaplan123: @YishaiGlasner @yitzhakc I'm creating a helper function Sefaria._getShortInterfaceLang() 
How does this sound?
It can replace 15-20 cases throughout the site

static/js/sefaria/sefaria.js
stevekaplan123: This would only happen if there was an error in the data, though we have a line that explicitly sets this on the backend:
` refLinkType = 'popular-writing-of' if getattr(topic, 'subclass', '') == 'author' else 'about'`

static/js/TopicPage.jsx
stevekaplan123: Are you referring to useAllPossibleSourceTabs?  It's a hook so that we know given a key in a particular topic.tabs, how do we fetch, sort, filter, and render the data.

yitzhakc (approved):

PR #2404, opened 2025-04-09
chore: exclude essay links from linksummary count but not from textlist

stevekaplan123:
## Description
Essay links don't show up in the Connections Panel.

## Code Changes
This bug is due to a recent change to sefaria.js in getLinksFromCacheAndPreprocess.
This function filters out essay links because it is called by linkSummary() from the ConnectionsPanel.  This filtering should happen because linkSummary is only for "normal"/non-essay links that are in "Commentary", "Talmud", and other categories.  However, when TextList renders and retrieves links from cache using getLinksFromCacheAndPreprocess we should _not_ filter out essay links because TextList is how essay links get displayed.  I simply added a parameter to getLinksFromCacheAndPreprocess to handle this possibility.

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -1553,7 +1553,7 @@ Sefaria = extend(Sefaria, {
       return true;
     }
   },
-  getLinksFromCacheAndPreprocess: function(ref, excludedSheet) {
+  getLinksFromCacheAndPreprocess: function(ref, excludedSheet = null, excludeEssays = false) {
     let links = [];
     if (typeof ref == "string") {
       links = this.getLinksFromCache(ref);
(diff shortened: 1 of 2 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -1553,7 +1553,7 @@ Sefaria = extend(Sefaria, {
       return true;
     }
   },
-  getLinksFromCacheAndPreprocess: function(ref, excludedSheet) {
+  getLinksFromCacheAndPreprocess: function(ref, excludedSheet = null, excludeEssays = false) {
```
YishaiGlasner, line 1556: isn't excludeEssays should default to true for preserving its functionality?

i think this function does too many things:
1. handling array of refs or one ref for getting links from cache.
2. filtering links.

it's too much also before this change, but now it has more kinds of filtering.
idk how often we use it, so how difficult it to reafactor it, but if we can, i think it will be better to use the different function:
1. we already have `getLinksFromCache`
2. it seems we need something like:
```
applyToRefOrRefs(ref, callback) {
  return typeof ref === 'string' ? callback(ref) : ref.flatMap(callback);
}
```
3. and we have `_filterSheetFromLinks`
4. and maybe we need a function for `filter(link => link.type !== "essay")` (if it's something that happen more than once in the code)

PR #2411, opened 2025-04-11
Hebrew newsletter form sign up producing wrong data in Salesforce

relyks:
## Description
This fixes the order and type of parameters sent to Salesforce. The interface language and educator fields were in the wrong order and causing Hebrew users to be signed up for the wrong version of the newsletter mailing lists.

## Notes
* The existing data on Salesforce from previous submissions also needs to be fixed using a one-time script or flow to correct the order of the fields.
* The user's desired newsletters may need to be resent from ActiveCampaign in their preferred language

### DIFF
--- a/sefaria/helper/crm/crm_connection_manager.py
+++ b/sefaria/helper/crm/crm_connection_manager.py
@@ -38,7 +38,7 @@ class CrmConnectionManager(object):
         """
         pass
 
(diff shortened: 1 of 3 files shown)


akiva10b (commented):
I see more uses of subscribe_to_lists that have not been updated in this PR (see dummy_crm.py)

PR #2423, opened 2025-04-23
Sheets editor final

stevekaplan123:
## Description

## Code Changes

## Notes


stevekaplan123:
Random PyTests keep failing.  I wonder if it's the same kind of issue that plagues our playwright tests whereby if we were simply to increase the number of retries, the tests would be marked as flaky not failing.  In any case, can you override and merge? Thanks! @akiva10b

static/css/s2.css
```diff
@@ -9054,6 +9058,22 @@ a.sheetAuthorName:hover {
 .sheetContentSidebar a.sheetAuthorName {
   display: block;
 }
+.sheetMetaDataBox .sidebarLayout .items {
```
stevekaplan123, line 9061: Each button in the row should have its own class that styles the position and that way there's no reason to even mention the publish menu

PR #2425, opened 2025-04-24
Logout from sheets module will redirect to sheets homepage

YishaiGlasner:
## Description
Previuously we've used the Django CustomLogoutView and defined  `LOGOUT_REDIRECT_URL`.
Now we'll override Django's `get_next_page`, and get the next page from the next param of the request.

## Code Changes
sefaria.js -  add function `getLogoutUrl` that defines once the logout url with the next param according to the module.
Header.jsx - use `getLogoutUrl` anywhere rather than a magic string.
sefaria/views.py - override `CustomLogoutView.get_next_page` for getting next from the url params if exists, rather fall to Django's `get_next_page`.

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -13,7 +13,7 @@ from random import choice
 from django.utils.translation import ugettext as _
 from django.conf import settings
 from django.http import HttpResponse, HttpResponseRedirect, Http404, HttpResponseBadRequest
-from django.shortcuts import render, redirect
+from django.shortcuts import render, redirect, resolve_url
 from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
 from django.utils.http import is_safe_url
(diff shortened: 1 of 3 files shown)


sefaria/views.py
```diff
@@ -83,7 +83,13 @@ class CustomLoginView(StaticViewMixin, LoginView):
     authentication_form = SefariaLoginForm
 
 class CustomLogoutView(StaticViewMixin, LogoutView):
-    pass
+
+    def get_next_page(self):
```
stevekaplan123, line 87: I don't know what this is.. Is it a built in Django function you're overriding?

YishaiGlasner: @stevekaplan123 right, this is `LogoutView` function

yitzhakc (approved):

PR #2471, opened 2025-05-26
Fix question mark in index title break navigation

YishaiGlasner:
## Description
In previous fix we've changed also the `Ref.url` in the server side to encode question mark with `%3f`. This cause to a situation where the client side requests some ref with `%3F` and get 301 to the same url, because Django returns the question mark, and the server side thinks it's different from the normal url ref.

## Code Changes
The `url` method now have a new default arg for getting the url without HTML encoding (i.e. with a question mark rather than `%3F`).
Noth that this method also sets the value of `._url`, and this attribute will remain encoded.
Also I've refactored the way we're replacing the last occurrence of underscore.

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -4716,21 +4716,29 @@ class Ref(object, metaclass=RefCacheType):
         """
         return TextChunk(self, lang, vtitle, exclude_copyrighted=exclude_copyrighted)
 
-    def url(self):
+    def url(self, encode_html=True):
         """
         :return string: normal url form
         """
-        if not self._url:
-            self._url = self.normal().replace(" ", "_").replace(":", ".").replace("?", "%3F")
+        if not self._url or not encode_html:
+            url = self.normal()
+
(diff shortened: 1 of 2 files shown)


sefaria/model/text.py
```diff
@@ -4716,21 +4716,26 @@ def text(self, lang="en", vtitle=None, exclude_copyrighted=False):
...
             if len(self.sections) > 0:
-                last = self._url.rfind("_")
-                if last == -1:
-                    return self._url
-                lref = list(self._url)
-                lref[last] = "."
-                self._url = "".join(lref)
+                url = '.'.join(url.rsplit('_', 1))
+
+            if not encode_html:
```
stevekaplan123, line 4736: It's confusing to me that if encode_html is True, we cache the result but if encode_html is False, we don't cache the result.

yitzhakc: We can refactor this a bit more, but we'll open up a new story for it.

yitzhakc (approved):

PR #2484, opened 2025-05-28
Feature/sc 34320/create guide component

schreibersefaria:
## Overview

This PR introduces a guide overlay that displays tips to users when they first open the Sheets Editor.
The system provides a content management interface in Django admin to edit this tips.
Overall the feature should be reusable for future guides on other pages but wasn't tested for this.

**Key Features:**
- One-time overlay on first visit (cookie-based persistence)
- Admin CMS for content management
- Button to reshow overlay after dismissal

Cauldron: https://tips-and-tricks.cauldron.sefaria.org/sheets/652030.652030?lang=bi
Admin data editing instructions: https://docs.google.com/document/d/1Tjh75Yi5XNwEB5rqtfWnxqGDu4oYjYiHAIa4b9J1FQo/edit?usp=sharing

image

---

### 1. Django Backend: Content Management System

**New `guides/` app** provides the foundation for managing guide content
The Guide level info is visible in all the cards.

```python
# guides/models.py - Core data structure
class Guide(models.Model):
    key = models.CharField(choices=[('editor', 'editor')])  # Closed list for guide types
    title_prefix_en = models.CharField(max_length=255)      # "Quick Start Guide:"
    title_prefix_he = models.CharField(max_length=255)
    # Footer links for consistent navigation...

class InfoCard(Sortable):
    guide = models.ForeignKey(Guide, related_name='info_cards')
    title_en/he = models.CharField(max_length=255)
    text_en/he = models.TextField(help_text="Supports markdown")
    video_en/he = models.URLField()  # Localized video URLs
    order = models.PositiveIntegerField()  # Drag-and-drop ordering
```

**`guides/admin.py`**: Inline editing interface for the Django admin
**`guides/views.py`**: Transforms Django models to JSON structure

### 2. Main React Component: GuideOverlay

**`static/js/GuideOverlay.jsx`** (352 lines) - The core user-facing component:
Some of the key parts of the code - **Simplified**:

```jsx
const GuideOverlay = ({ onClose, guideType = "editor", forceShow = false }) => {
  const cookieName = `guide_overlay_seen_${guideType}`;
  const [isOpen, setIsOpen] = useState(forceShow || !$.cookie(cookieName));
  const [currentCardIndex, setCurrentCardIndex] = useState(0);

...
const loadGuide = async () => {
      setLoading(true);
        const data = await Sefaria.getGuide(guideType);
        setGuideData(data);
    };
...

    timeoutId = setTimeout(() => { // Close the guide if loading takes too long
      if (isComponentMounted) {
        console.warn(`Guide loading timed out after ${timeoutLength} seconds`);
        
        handleClose();
        alert(Sefaria._("Something went wrong. Try refreshing the page", "EditorSaveIndicator"));
      }
    }, timeoutLength * 1000);

...

  const handleClose = () => { 
    setCookie();
    if (onClose) onClose();
    setIsOpen(false);
  };

...

  // Cookie-based persistence - show only once unless forced
  const handleClose = () => {
    setCookie();  // Store dismissal date
    trackGuideEvent("guide_overlay_closed");
    onClose();
  };

...

const navigateCard = (direction) => {
    const cardsLength = guideData.cards.length;
    let newIndex;
    
    if (direction === "next") {
      newIndex = currentCardIndex >= cardsLength - 1 ? 0 : currentCardIndex + 1;
    } else {
      newIndex = currentCardIndex <= 0 ? cardsLength - 1 : currentCardIndex - 1;
    }
    
    setCurrentCardIndex(newIndex);
  };

...

  // Video preloading for smooth transitions
  {guideData.cards.map((card, index) => (
    <video
      key={index}
      src={Sefaria.getLocalizedVideoDataFromCard(card).videoUrl}
      style={{ display: index === currentCardIndex ? 'block' : 'none' }}
      controls autoPlay muted playsInline
    />
  ))}
};
```

Key sub-components:
- **`TitleWithPrefix`**: Handles localized titles with prefixes
- **`GuideFooter`**: Up to 2 footer hyper links

### 3. Integration into Reader Interface

**`static/js/ReaderPanel.jsx`** - Showing only where relevant:

```jsx
class ReaderPanel extends Component {
  state = { forceGuideOverlay: false }

  shouldShowGuides() {
    // Centralized logic for when to show guides
    return this.props.multiPanel &&           // Desktop only (no mobile)
           this.props.panelPosition === 0 &&  // Main panel (not resources)
           (this.state.mode === "Sheet" || this.state.mode === "SheetAndConnections");
  }

  showGuide() {
    // Triggered by guide button - forces overlay to reappear
    this.setState({forceGuideOverlay: true});
  }

  render() {
    return (
      <div>
        {/* Main content */}
        
        {/* Guide overlay - conditionally rendered */}
        {this.shouldShowGuides() && 
          <GuideOverlay 
            onClose={() => this.setState({forceGuideOverlay: false})}
            guideType="editor" // Currently hardcoded - will have to change for reuse of component.
            forceShow={this.state.forceGuideOverlay}
          />
        }
      </div>
    );
  }
}
```

### 4. API Integration and Data Flow

**`static/js/sefaria/sefaria.js`** - Backend connectivity:

```javascript
// Cached API calls for performance
getGuide: function(guideKey = "sheets") {
  return this._cachedApiPromise({
    url: `${Sefaria.apiHost}/api/guides/${guideKey}`,
    store: this._guides
  });
},

// Language-aware video selection
getLocalizedVideoDataFromCard: function(card) {
  const lang = this.interfaceLang === "hebrew" ? "he" : "en";
  return {videoUrl: card.videoUrl[lang]};
}
```

### 5. Guide Button Integration

**`static/js/Misc.jsx`**:

```jsx
const GuideButton = ({ onShowGuide, tooltip = false }) => {
  const altText = Sefaria._("Show guide", "Guide");
  
  return (
    <ToolTipped {...{ altText, onClick: onShowGuide }}>
      <img src="/static/img/bulb.svg" alt={altText}/>
    </ToolTipped>
  );
};
```

**Placement**: Added to ReaderControls rightButtons section (next to save and display options), only visible when guides are available.

**Additional changes in Misc**
1. Added reusable arrow components
2. Added a default alt text to the close component

### 6. CSS Integration with Design System

**`static/css/s2.css`** (~384 lines) - Follows existing patterns:

```css
.guideOverlay {
  /* Integrates with existing font system */
  --english-font: var(--english-sans-serif-font-family);
  --hebrew-font: var(--hebrew-sans-serif-font-family);
  
  /* Fixed overlay positioning */
  position: absolute;
  bottom: 0;
  height: 415px;
  z-index: 1000;
}

/* RTL support */
.interface-hebrew .guideOverlayBody { 
  flex-direction: row-reverse; 
}
.interface-hebrew .guideOverlayTitle { 
  direction: rtl; 
  text-align: right; 
}
```

---

### Analytics Tracking
Basic event tracking throughout user journey:
```javascript
trackGuideEvent("guide_overlay_opened", { guide_type: guideType });
trackGuideEvent("guide_overlay_pagination", { direction, new_index });
trackGuideEvent("guide_footer_link_clicked", { link_text, position });
```

### Timeout Protection
7-second loading timeout prevents indefinite loading states, with user alert and automatic cleanup.

### Circular Pagination
Navigate through cards with wraparound (last card → first card).

### Content Management
Videos stored in Google Cloud Storage bucket `guides-resources`, admin interface handles metadata and ordering.

---

### Analytics Implementation
Need verification that gtag event tracking follows current patterns and captures the right data points for product analytics.
I didn't get a spec and never did this on the web app.

### CSS Integration
Another area that is new to me and the big S2.css was confusing to me
Want confirmation that:
- Font variable overrides match established component patterns
- RTL implementation aligns with other interface components  

---

## Notes

**Extensibility**: System designed for multiple guide types but currently hardcoded to "editor". Adding new guide types (reader, topics) requires changing the hardcoded guideType in ReaderPanel to be dynamic.

**Accessibility**: Proper alt text implementation and keyboard navigation support included.


guides/admin.py
```diff
@@ -0,0 +1,66 @@
...
+        model = Guide
+        fields = '__all__'
+
+@admin.register(Guide)
+class GuideAdmin(admin.ModelAdmin):
+    form = GuideAdminForm
+    list_display = ('key', 'title_prefix_en', 'title_prefix_he', 'created_at')
+    list_filter = ('created_at',)
+    search_fields = ('key', 'title_prefix_en', 'title_prefix_he')
+    inlines = [InfoCardInline]
```
stevekaplan123, line 38: Can you add a comment explaining what inlines are?

guides/models.py
```diff
@@ -0,0 +1,137 @@
...
+class Guide(models.Model):
+    """
+    Guide model for organizing InfoCards by guide type (e.g., 'sheets', 'topics')
+    """
+    key = models.CharField(
+        max_length=100, 
+        unique=True, 
+        verbose_name="Guide Key",
+        help_text="Unique identifier for the guide (e.g., 'editor', 'topics')",
+        choices=[('editor', 'editor')],  # Closed list, add more as needed
```
stevekaplan123: Why is 'editor' an option twice?

guides/models.py
```diff
@@ -0,0 +1,137 @@
...
+        blank=True,
+        verbose_name="Footer Link 2 URL",
+        help_text="URL for the second footer link (optional)"
+    )
+    
+    @property
+    def footer_links(self):
+        """Generate footer links array from individual fields"""
+        links = []
+
```
stevekaplan123: Ideally, there should be one piece of code that handles both footer link 1 and footer link 2 since the code is basically the same in each case

reader/views.py
```diff
@@ -4785,3 +4786,72 @@ def is_database_reachable():
...
+def guides_api(request, guide_key=None):
+    """
+    API endpoint that returns guide data for a specific guide.
+    
+    Args:
+        guide_key (str): The guide key to fetch guide for (e.g., 'sheets', 'test_heb')
+    
+    Returns:
+        JSON response with the following structure:
+        {
```
stevekaplan123: Is 'totalCards' just the length of 'cards'?  If so, it doesnt' seem necessary

reader/views.py
```diff
@@ -4785,3 +4786,72 @@ def is_database_reachable():
...
+            },
+            "text": {
+                "en": card.text_en,
+                "he": card.text_he
+            },
+            "videoUrl": {
+                "en": card.video_en,
+                "he": card.video_he
+            }
+        })
```
stevekaplan123: I think the Guide should have a function for generating this data.  This API should just call that function to get the 'cards'

static/css/s2.css
```diff
@@ -5783,6 +5784,50 @@ h1.topic-landing-header {
...
+  align-items: center;
+  justify-content: flex-end;
+  height: 56px;
+  cursor: pointer;
+  margin-right: 15px;
+  margin-left: 0px;
+  padding-top: 3px;
+}
+.interface-hebrew .rightButtons .guideButton {
+  margin-left: 15px;
```
stevekaplan123: It's usually possible to us margin-start instead of margin-left or margin-right to avoid adding CSS for interface-hebrew

static/css/s2.css
```diff
@@ -8649,7 +8694,7 @@ a.button.mini {
 a .button:hover {
   text-decoration: none;
 }
-.reactMarkdown a {
+©arkdown a {
```
stevekaplan123: What is ©arkdown?

static/css/s2.css
```diff
@@ -14694,3 +14740,276 @@ span.ref-link-color-3 {color: blue}
...
+  direction: rtl;
+  padding-right: 0;
+  padding-left: 8px;
+}
+
+.interface-hebrew .guideOverlayFooter {
+  direction: rtl;
+}
+
+/* ========== GUIDE OVERLAY COMPONENT STYLES - END ========== */
```
stevekaplan123, line 14976: I like this -- marking the beginning and end of the guide styles!  we should do this more often

static/js/Misc.jsx
```diff
@@ -1117,16 +1117,28 @@ class CloseButton extends Component {
...
+    } else if (icon == "chevron") {
+      var iconElement = <i className="fa fa-chevron-left"></i>
     } else {
-      var icon = "×";
+      var iconElement = "×";
     }
-    var classes = classNames({readerNavMenuCloseButton: 1, circledX: this.props.icon === "circledX"});
-    var url = this.props.url || "";
-    return (<a href={url} className={classes} onClick={this.onClick}>{icon}</a>);
+    var classes = classNames({readerNavMenuCloseButton: 1, circledX: icon === "circledX"});
```
stevekaplan123: Using 'var' is an old standard that we are trying to get rid of.  We should always use 'let' or 'const'.  In this case, 'const' is correct.

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,352 @@
...
+        guide_type: guideType
+      });
+    }
+  }, [initialShouldShow]);
+
+  // Load guide data only when component is actually open/visible to avoid unnecessary loading on common scenario where it isn't needed
+  useEffect(() => {
+    if (!isOpen) return;
+    
+    let isComponentMounted = true;
```
stevekaplan123, line 144: What's the reason for this variable?

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,352 @@
...
+  const initialShouldShow = forceShow || !$.cookie(cookieName);
+  
+  const [isOpen, setIsOpen] = useState(false);
+  const [guideData, setGuideData] = useState(null);
+  const [currentCardIndex, setCurrentCardIndex] = useState(0);
+  const [loading, setLoading] = useState(true);
+  
+  useEffect(() => {
+    // Force rerendering of the component when initially rendered
+    setIsOpen(initialShouldShow);
```
stevekaplan123: I don't understand this.  Why not just set isOpen to initialShouldShow on the first load?

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,352 @@
...
+
+  /**
+   * Navigate to next or previous card with circular looping
+   * @param {string} direction - Either "next" or "previous"
+   */
+  const navigateCard = (direction) => {
+    if (!guideData || !guideData.cards.length) return;
+    
+    // Validate direction parameter
+    if (direction !== "next" && direction !== "previous") {
```
stevekaplan123, line 275: I'm confused that direction is "next"/"previous" here and then "right"/"left" elsewhere.

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,352 @@
...
+  forceShow: PropTypes.bool,
+  timeoutLength: PropTypes.number
+};
+
+/**
+ * Utility function to clear the guide overlay cookie for a given guide type
+ * Useful for testing and development - allows the overlay to show again
+ * @param {string} guideType - The type of guide (e.g., "sheets", "reader")
+ */
+export const clearGuideOverlayCookie = (guideType) => {
```
stevekaplan123: Please remove this as it's not necessary for production.

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,352 @@
...
+            />
+          </div>
+          
+          {showNavigation && (
+            <div className="guideOverlayPagination">
+              <Arrow 
+                direction="left" 
+                onClick={() => navigateCard("previous")}
+                altText={localize("Previous card")}
+                reverseForRTL={true}
```
stevekaplan123: It seems unnecessary to have a variable 'reverseForRTL' since we always want this to be true right?

stevekaplan123: Also, I think instead of calculating the image to show in the Arrow component, you can use flex-box in CSS to style guideOverlayPagination so that it will show reversed in interface-hebrew

static/js/sefaria/sefaria.js
```diff
@@ -692,6 +692,19 @@ Sefaria = extend(Sefaria, {
...
+  _guides: {},
+  getGuide: function(guideKey = "sheets") {
+    return this._cachedApiPromise({
+      url: `${Sefaria.apiHost}/api/guides/${guideKey}`,
+      key: `guide_${guideKey}`,
+      store: this._guides
+    });
+  },
+  // Putting the localization logic here because I want the component to be language agnostic.
+  getLocalizedVideoDataFromCard: function(card) {
```
stevekaplan123: This function is not necessary.  I believe you can use the Sefaria._v function to get the same result.

static/js/ReaderPanel.jsx
```diff
@@ -1149,6 +1170,14 @@ class ReaderPanel extends Component {
           </div> : null}
 
           {menu}
+          {/* Guide overlay - currently only shows on the sheets editor but can be extended for other guide types */}
+          {this.shouldShowGuides() &&
```
stevekaplan123: I'm not sure but it seems like the shouldShowGuides function should be passed the guideType and have a check "if editor && in sheets mode, show guide".  What do you think?

static/js/ReaderPanel.jsx
```diff
@@ -478,6 +481,22 @@ class ReaderPanel extends Component {
       searchQuery: query
     });
   }
+  showGuide() {
+    // Force the guide overlay to show
+    this.setState({forceGuideOverlay: true});
```
stevekaplan123, line 487: Why do you call it 'forceGuideOverlay'.  What's forced about it?  Why not 'showGuideOverlay'?

guides/admin.py
schreibersefaria: done

guides/models.py
schreibersefaria: it's key:value. I clarified in a comment

guides/models.py
schreibersefaria: Refactored.
I also changed it so both en and he text is required and not just one.

reader/views.py
schreibersefaria: Good point.

reader/views.py
schreibersefaria: So if I move the responsibility of  to JSON to Guide, would you think I should also include building of response_data in Guide?

```
    # Build the response
    response_data = {
        "titlePrefix": {
            "en": guide.title_prefix_en,
            "he": guide.title_prefix_he
        },
        "footerLinks": guide.footer_links,
        "cards": cards
    }
```

static/css/s2.css
schreibersefaria: This isn't working. Did I implement correctly?
```
.rightButtons .guideButton {
...
  margin-inline-start: 15px;
  margin-inline-end: 0px;
  padding-top: 3px;
}
```

static/css/s2.css
schreibersefaria: opps. Reverted

static/js/GuideOverlay.jsx
schreibersefaria: It prevents state updates after component unmounts.
I also added the explanation in a comment.

static/js/GuideOverlay.jsx
schreibersefaria: It was needed for the GuideButton functionality to show the component on prop change.
The naming was confusing, I improved it and added commenting.

static/js/GuideOverlay.jsx
schreibersefaria: If I understand your comment correctly:
Next/Previous are used for pagination direction
Right/Left are used for the "physical" direction of the arrow component

static/js/GuideOverlay.jsx
schreibersefaria: Removed reverseForRTL, I agree that it was over engineered.

schreibersefaria: Regarding the flex-box, wouldn't that cause confusion with the altText?

static/js/sefaria/sefaria.js
schreibersefaria: Fixed, thanks!

static/js/ReaderPanel.jsx
schreibersefaria: Not sure exactly what you mean.
Do you think I should add logic in shouldShowGuides that "knows" the options of guideType?

static/js/ReaderPanel.jsx
schreibersefaria: Well, the guide automatically shows on the first time. The word force differentiates form that "natural" flow. So you won't think the full control on when to open is in the ReaderPanel.
That being said, what do you think of trigerGuideOverlayShow as a clearer name?

reader/views.py
stevekaplan123: We have a method in many classes in Sefaria called `contents`.  See classes Index and Topic for examples of 'contents' functions.  I think it would make sense for InfoCard and Guide to have a contents method.  It seems like InfoCard's contents() should return:
```
{
            "id": str(card.id),
            "title": {
                "en": card.title_en,
                "he": card.title_he
            },
            "text": {
                "en": card.text_en,
                "he": card.text_he
            },
            "videoUrl": {
                "en": card.video_en,
                "he": card.video_he
            }
        }
```

Guide's contents() could basically be almost the entire API whereby it would return JSON of its infocards and title data.  What do you think?

static/js/ReaderPanel.jsx
stevekaplan123: Well, I'm thinking about the fact that we will have other guides.  Do we have a clear expectation of what the other guides will be like?   If we don't, then I guess it makes sense to leave it as is and just change it when we know for sure.  But if we know what's planned, we could have a check that says that we should show the GuideOverlay with guideType 'editor' if we're in sheets whereas show the GuideOverlay with guideType 'reader' if we're in reader panel, for example.

static/js/sefaria/sefaria.js
```diff
@@ -692,6 +692,14 @@ Sefaria = extend(Sefaria, {
       store: this._bulkSheets
     });
   },
+  _guides: {},
+  getGuide: function(guideKey = "sheets") {
```
stevekaplan123: However, we pass a value of 'editor' and never pass a value of 'sheets' right?

static/js/ReaderPanel.jsx
stevekaplan123: Maybe it's OK as is.

static/js/GuideOverlay.jsx
stevekaplan123: I think next/previous should be the value in both cases, and we can use CSS to reverse the direction of the arrows in the case of Hebrew.

static/js/ReaderPanel.jsx
schreibersefaria: So I generally tried to make things generic other that this point. I felt it will anyway need a change (we don't know what other guides we are expecting) so we might as well leave this point for then.
If I were to do this properly, I think I would change shouldShowGuides func to return the guideType / null.
That way it can be one function you need to change to add a new guide.

static/css/s2.css
stevekaplan123: margin-right is 15px so margin-inline-end should be 15px, and margin-inline-start should be 0px.

reader/views.py
schreibersefaria: Refactored, could you have another look?

static/js/GuideOverlay.jsx
stevekaplan123: In other words, in ArrowButton, you have this calculation to determine the actualDirection. I think you can just use styling on guideOverlayPagination, and the buttons themselves to make sure everything's pointed in the right direction.

static/js/ReaderPanel.jsx
stevekaplan123: I think we should change it a little bit.  What do you mean by "If I were to do this properly, I think I would change shouldShowGuides func to return the guideType / null.
That way it can be one function you need to change to add a new guide."  
I don't like that shouldShowGuides specifically checks wehther we're in sheets without being directed by the caller of the function.  The caller of the function should in some way tell it that this is the sheets/editor guide, perhaps.

schreibersefaria: So where would you put the list of existing guides?
A config file? Or in models or something like that and use an API?
shouldShowGuides needs to "know" the specific HTML elements to look for so I thought it makes sense to make it the one place you need to change when adding a new guide.

static/css/s2.css
schreibersefaria: Ops, thanks!

static/js/sefaria/sefaria.js
schreibersefaria: good catch

static/js/GuideOverlay.jsx
schreibersefaria: Changed the flip to happen in CSS

static/js/GuideOverlay.jsx
schreibersefaria: Done

static/js/ReaderPanel.jsx
schreibersefaria: Refactored to getGuideType()

guides/admin.py
```diff
@@ -27,6 +27,8 @@ class GuideAdmin(admin.ModelAdmin):
     list_display = ('key', 'title_prefix_en', 'title_prefix_he', 'created_at')
     list_filter = ('created_at',)
     search_fields = ('key', 'title_prefix_en', 'title_prefix_he')
+    # Inlines allow related models to be edited on the same page as the parent model
```
stevekaplan123, line 36: Can you just move this comment or make a similar comment at the top of the InfoCardInline class? Thanks!

guides/models.py
```diff
@@ -66,32 +66,50 @@ def footer_links(self):
...
-                    'url': self.footer_link_2_url
+                    'url': url
                 })
         
         return links
+    
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    def contents(self, **kwargs):
```
stevekaplan123, line 92: Looks great!

guides/models.py
```diff
@@ -11,7 +11,7 @@ class Guide(models.Model):
         unique=True, 
         verbose_name="Guide Key",
         help_text="Unique identifier for the guide (e.g., 'editor', 'topics')",
-        choices=[('editor', 'editor')],  # Closed list, add more as needed
+        choices=[('editor', 'editor')],  # Closed list, add more as needed. Format: [(key, value), ...]
```
stevekaplan123: What does this list correspond to?  Is this part of the data itself?

static/js/ReaderPanel.jsx
```diff
@@ -1141,6 +1195,8 @@ class ReaderPanel extends Component {
               setTranslationLanguagePreference={this.props.setTranslationLanguagePreference}
               data={this.state.data}
               backButtonSettings={this.state.backButtonSettings}
+              showGuide={this.showGuide.bind(this)}
```
stevekaplan123: I don't think you need to pass getGuideType to ReaderControls since it's just used to show the GuideButton.  It seems like you could just pass showGuide and when you pass showGuide you could pass it as false or as a function:
`showGuide = {this.props.getGuideType && this.props.getGuideType() && this.showGuide.bind(this)}`
and then below in the ReaderControls you could say:
```
{this.props.showGuide && 
            <GuideButton
...
            />
          }
```

stevekaplan123 (commented):
This overall looks great.  I added a few comments.

guides/models.py
schreibersefaria: This is the uID for the guide data in the django DB table.
I didn't want to have the guide ID editable (after Mickey edited it and suddenly nothing worked).
We agreed to make it multiple choice and this was the implementation I got from AI.
I can change the key to Edit (not edit), maybe that will make it clearer?

guides/admin.py
schreibersefaria: done

static/js/ReaderPanel.jsx
schreibersefaria: Done

guides/models.py
```diff
@@ -0,0 +1,178 @@
+from django.db import models
+from adminsortable.models import Sortable
+import json
+
+class Guide(models.Model):
+    """
+    Guide model for organizing InfoCards by guide type (e.g., 'sheets', 'topics')
```
stevekaplan123: Still confused about 'sheets' vs 'editor' naming.  It seems like almost everywhere in this code we call it 'editor' not 'sheets'.

schreibersefaria: I fixed that comment.
Are there areas where you think I should change this within the current changes?

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,372 @@
...
+
+    // Set up timeout to close component if loading takes too long
+    // It's cleared with clearTimeout if the data is loaded successfully
+    // Reason for this functionality: The guide is meant to clarify the page, if it causes more problems, it isn't worth the hassle
+    timeoutId = setTimeout(() => {
+      if (isComponentMounted) {
+        console.warn(`Guide loading timed out after ${timeoutLength} seconds`);
+        handleClose();
+        alert(Sefaria._("Something went wrong. Try refreshing the page", "EditorSaveIndicator"));
+      }
```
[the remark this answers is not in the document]
schreibersefaria: I'm confused about this.

stevekaplan123 (approved):

static/js/GuideOverlay.jsx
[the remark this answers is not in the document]
stevekaplan123: We shouldn't have console error logs in production code in any case unless there's a really good reason.  Do you think any of these console logs are really necessary?

static/js/Misc.jsx
```diff
@@ -1290,6 +1302,84 @@ SaveButton.propTypes = {
...
+ * ArrowButton component for navigation buttons (next/previous arrows)
+ * @param {string} direction - "next" or "previous" 
+ * @param {function} onClick - Function to call when arrow is clicked
+ * @param {string} altText - Alt text for accessibility
+ * @param {string} className - Additional CSS classes
+ */
+function ArrowButton({ direction, onClick, altText = "", className = "" }) {
+  // Validate required props
+  if (!direction) {
+    console.error("ArrowButton component requires a 'direction' prop");
```
stevekaplan123: I'm not sure we should have all these console error statements in production code.  What do you think @yitzhakc?

schreibersefaria: Agree in general would definitely want some sort of analytic instead of this.
Specifically here I'm trying to create some sort of type safety, any good way to do that?

stevekaplan123: Yeah, I believe you did exactly that here rendering the checks and console logs unnecessary:
```
 ArrowButton.propTypes = {
  direction: PropTypes.oneOf(["next", "previous"]).isRequired,
  onClick: PropTypes.func.isRequired,
  altText: PropTypes.string.isRequired,
  className: PropTypes.string,
};
```

schreibersefaria: I removed this

static/js/GuideOverlay.jsx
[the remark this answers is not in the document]
schreibersefaria: I removed this and added a comment saying we would want to add an analytics event

static/js/Misc.jsx
```diff
@@ -96,7 +96,7 @@ const InterfaceText = ({text, html, markdown, children, context, disallowedMarkd
   return (
     html ?
       <span className={elemclasses} dangerouslySetInnerHTML={{__html: textResponse}}/>
-        : markdown ? <span className={elemclasses}><ReactMarkdown className={'reactMarkdown'} unwrapDisallowed={true} disallowedElements={['p', ...disallowedMarkdownElements]}>{textResponse}</ReactMarkdown></span>
+        : markdown ? <span className={elemclasses}><ReactMarkdown className={'reactMarkdown'} unwrapDisallowed={true} disallowedElements={allowParagraphs ? disallowedMarkdownElements : ['p', ...disallowedMarkdownElements]}>{textResponse}</ReactMarkdown></span>
```
stevekaplan123: 'allowParagraphs' and 'disallowedMarkdownElements' are both solely for helping us render markdown?  If that's the case, we need a simpler approach here.  Look at the different cases where we render markdown with InterfaceText and try to simplify it so we only need maximum one variable not two.

schreibersefaria: Good catch!
There was only one other use of disallowedMarkdownElements so it was easy to consolidate.
The one things about this is that it can be confusing. If anyone want's to disallow an element, they will have to "know" they also need to disallow 'p'.

stevekaplan123 (approved):

guides/migrations/guides_data.json
```diff
@@ -0,0 +1 @@
...anced Control", "title_he": "Export for Advanced Control", "text_en": "For more robust formatting options and full control over printed sheets, use the Export to Google Docs feature found in the Resources panel. Your first export will require you to associate your Google Account with your Sefaria Account - just follow the prompts.", "text_he": "x", "order": 70, "video_en": "https://storage.googleapis.com/guides-resources/7%20export.mp4", "video_he": "https://storage.googleapis.com/guides-resources/7%20export.mp4"}}, {"model": "guides.infocard", "pk": 13, "fields": {"guide": 1, "title_en": "Next Steps", "title_he": "Next Steps", "text_en": "For complete instructions, see [this article](https://example.com). \r\n* To reopen this Quick Start Guide, click the lightbulb icon at the top of your sheet. \r\n* Your sheets can be found in your Profile. \r\n* **We would love** to hear about the types of sheets you make, please be in touch using the feedback link below or by writing to hello@sefaria.org!", "text_he": "x", "order": 80, "video_en": "https://storage.googleapis.com/guides-resources/8%20next%202.mp4", "video_he": "https://storage.googleapis.com/guides-resources/8%20next%202.mp4"}}]
```
yitzhakc: Fixtures should be put in the `/fixtures` directory of the app. This way they can be easily used in tests, [manage.py loaddata](https://docs.djangoproject.com/en/1.11/ref/django-admin/#what-s-a-fixture) commands and more, Please move the file to be located at `/guides/fixtures/guides_data.json`

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+        max_length=255, 
+        verbose_name="Title Prefix (HE)",
+        help_text="Prefix shown before tips in Hebrew"
+    )
+    # Footer Link 1
+    footer_link_1_text_en = models.CharField(
+        max_length=255,
+        blank=True,
+        verbose_name="Footer Link 1 Text (EN)",
+        help_text="Text for the first footer link in English (optional)"
```
yitzhakc, line 34: I believe that when `blank=True`, the django-admin will automatically render the field as optional (will show as not required, field won't be bolded, etc.). But having it in the help text can't hurt :-)

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+        help_text="URL for the second footer link (optional)"
+    )
+    
+    @property
+    def footer_links(self):
+        """Generate footer links array from individual fields"""
+        links = []
+        
+        # Process both footer links
+        for i in [1, 2]: # Currently up to 2 links are supported
```
yitzhakc: If only 2 links are supported, then please make a `MAX_LINKS` constant attribute on the model with that value.
Then, please modify this loop to use `range(self.MAX_LINKS)`.

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+            text_en = getattr(self, f'footer_link_{i}_text_en')
+            text_he = getattr(self, f'footer_link_{i}_text_he')
+            url = getattr(self, f'footer_link_{i}_url')
+            
+            # Add link if it has texts and URL
+            if text_en and text_he and url:
```
yitzhakc: Not a hard preference, but you could also use `all([text_en, text_he, text_url])`.

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+        help_text="Text for the second footer link in Hebrew (optional)"
+    )
+    footer_link_2_url = models.URLField(
+        blank=True,
+        verbose_name="Footer Link 2 URL",
+        help_text="URL for the second footer link (optional)"
+    )
+    
+    @property
+    def footer_links(self):
```
yitzhakc, line 70: While I understand you wanted to group the footer-related code, we prefer all properties to be located after the explictly-stated model fields, to keep the code cleaner and easier to track what is a "concrete" field vs dynamically-generated

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+    def __str__(self):
+        return f"{self.key} - {self.title_prefix_en}"
+
+    class Meta:
+        verbose_name = "Guide"
+        verbose_name_plural = "Guides"
+        ordering = ['key']
+
```
yitzhakc, line 122: PEP8 recommends two blank lines before class definitions at the top level to enhance readability. Please fix.

guides/models.py
```diff
@@ -0,0 +1,178 @@
+from django.db import models
+from adminsortable.models import Sortable
+import json
+
+class Guide(models.Model):
```
yitzhakc, line 5: PEP8 recommends two blank lines before class definitions at the top level.

guides/models.py
```diff
@@ -0,0 +1,178 @@
...
+                "he": self.text_he
+            },
+            "videoUrl": {
+                "en": self.video_en,
+                "he": self.video_he
+            }
+        }
+
+    def __str__(self):
+        return f"{self.guide.key} - {self.title_en}"
```
yitzhakc: This results in an identical structure to the `str` output of the `Guide` model itself. Perhaps add `Card #{order}` or something similar to make it clear this is a card...

reader/views.py
```diff
@@ -86,6 +86,7 @@
 from sefaria.helper.topic import update_topic
 from sefaria.helper.category import update_order_of_category_children, check_term
 from sefaria.helper.texts.tasks import save_version, save_changes, save_link
+from guides.models import *
```
yitzhakc: Using `import *` is generally discouraged in Python, as we prefer to explictly import only what we need so as not to pollute namespace and prevent potential dependency pitfalls

reader/views.py
```diff
@@ -4785,3 +4786,40 @@ def is_database_reachable():
         logger.warn("Failed rollout healthcheck. Healthcheck Response: {}".format(resp))
 
     return http.JsonResponse(resp, status=statusCode)
+
+@catch_error_as_json
+def guides_api(request, guide_key=None):
```
yitzhakc: I'm not sure this api function belongs here, in `reader/views.py`, since it has nothing to do with the `reader` 😉 
Please create a `guides/views.py` and move the function there, taking care to import it properly under the `sefaria/urls.py`.

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,377 @@
...
+ * 
+ * @param {Object} props - Component props
+ * @param {function} props.onClose - Function to call when overlay is closed
+ * @param {string} props.guideType - Type of guide to display (e.g., "reader", "editor")
+ * @param {boolean} props.forceShow - Force the overlay to show regardless of cookie state
+ * @param {number} props.timeoutLength - Timeout length in seconds - After this time, the overlay will be closed and the user will be alerted that the guide is taking too long to load.
+ */
+const GuideOverlay = ({ 
+  onClose,
+  guideType = "editor",
```
yitzhakc: Not sure I'd leave this default here, as it will come back to bite us in the future when we re-use the Guide

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,377 @@
...
+        if (isComponentMounted) {
+          clearTimeout(timeoutId); // Clear the timeout if the data is loaded successfully
+          setGuideData(data);
+          setCurrentCardIndex(0);
+        }
+      } catch (error) {
+        console.error("Error loading guide:", error);
+        if (isComponentMounted) {
+          clearTimeout(timeoutId); // Clear the timeout if there is an error
+          console.error("Error loading guide:", error);
```
yitzhakc: Why the double `console.log` here and on line 179? Isn't it just going to output the same `error`?

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,377 @@
...
+          handleClose(setCookie=false);
+          alert(Sefaria._("Something went wrong. Try refreshing the page", "EditorSaveIndicator"));
```
yitzhakc: Perhaps you could use `localize` here as well?

static/js/GuideOverlay.jsx
```diff
@@ -0,0 +1,377 @@
...
+                >
+                  <InterfaceText>{localize("Your browser does not support the video tag.")}</InterfaceText>
+                </video>
+              ))}
+            </div>
+            
+            {/* Scrollable text content - narrower width */}
+            <div className="guideOverlayTextContainer">
+              <div className="guideOverlayText" onClick={handleTextContentClick}>
+                <InterfaceText markdown={transformValues(currentCard.text, text => text.replace(/\n(?!\n)/g, '  \n'))} disallowedMarkdownElements={[]} />
```
yitzhakc: Is this our usual way of dealing with new lines in markdown? Since we control the input, perhaps we could deal with this on entering the data in the admin?

static/js/ReaderPanel.jsx
```diff
@@ -478,6 +482,70 @@ class ReaderPanel extends Component {
...
+    if (!this.props.multiPanel || this.props.panelPosition !== 0) {
+      return null;
+    }
+    
+    // !! Temp flag!! Only show guides to moderators/admins - TODO: Remove this once we have everything in place
+    if (!Sefaria.is_moderator) {
+      return null;
+    }
+    
+    // !! Temp flag!! Only show on .org site (English), not .org.il site (Hebrew) - TODO: Remove this once we have everything in place
```
yitzhakc, line 517: Do we have a story to take care of this?

guides/migrations/0002_load_initial_data.py
```diff
@@ -0,0 +1,91 @@
...
+from django.conf import settings
+
+
+def load_initial_data(apps, schema_editor):
+    """Load initial guide data from guides_data.json"""
+    Guide = apps.get_model('guides', 'Guide')
+    InfoCard = apps.get_model('guides', 'InfoCard')
+    
+    # Get the path to the JSON file
+    json_file_path = os.path.join(os.path.dirname(__file__), 'guides_data.json')
```
yitzhakc: If the `guides_data.json` file is a valid fixture file, you can run `call_command('loaddata', fixture_file)` and it will use `loaddata` to load your fixture file.

yitzhakc (changes requested):

guides/migrations/guides_data.json
schreibersefaria: Done and tested

guides/models.py
schreibersefaria: Updated the files to match pep8

guides/models.py
schreibersefaria: Not sure exactly what you mean.
Yes, `blank=True` makes the field optional
In the help text I mentioned the behaviour "link will only appear if all fields are provided".

image

guides/migrations/0002_load_initial_data.py
schreibersefaria: owww lovely!

static/js/GuideOverlay.jsx
schreibersefaria: Well I can refactor localize to work with this but the context is different so it seams to loose the use of localize.
Here I'm using a string that already existed (at product's request)

static/js/GuideOverlay.jsx
schreibersefaria: I removed all console outputs

guides/models.py
schreibersefaria: I did a full PEP8 auto formatting

static/js/GuideOverlay.jsx
schreibersefaria: Removed and made the prop required

static/js/ReaderPanel.jsx
schreibersefaria: Now there is 👍 
https://app.shortcut.com/sefaria/story/35618

static/js/GuideOverlay.jsx
schreibersefaria: How would we control it in the input?

It seems like yes, this is the usual way.
I tried to look up the exact case manually with limited success. But I couldn't find another way.
Cursor seems confident this is the way and gives a bunch of examples.

I know there are two cases that use replaceNewLinesWithLinebreaks (I tried to use the existing code without success). The content of the function is similar to what we do. (`s.replace(/\n/gi, "&nbsp; \n") + "&nbsp; \n&nbsp; \n"`)

I also saw we generally use `text.replace` a lot with all kinds or regexes, but not necessarily in the context of MD.
Originally we had HTML but product requested MD.

I'll note that product want's another change: Links in the text will open in a new tab. This will complicate our use of InterfaceText further. 

Do you think I should ask on slack?
I could also try create a new component to replace/wrap InterfaceText

guides/models.py
yitzhakc: Sorry, my point was that it isn't strictly necessary to explicitly write `(optional)` in the field `help_text` since Django provides build-in visual guides, but that it is fine that you did so :-) 

You're good!

static/js/GuideOverlay.jsx
yitzhakc: I assumed we could "control" the data it in the input since the markdown will be inputted via the django-admin?

I agree that it might be worth asking on Slack, since it could be that this is a separate use case from `InterfaceText` as you surmised.

schreibersefaria: Well we do control the data but that is the requirement, to allow the admin to add new lines (without breaking anything else).
One thing I could do is add a flag to `replaceNewLinesWithLinebreaks()` with a a different replace logic.
It would centralise the regex logic in one place but that seems to be against the general pattern in the code.

yitzhakc: Based on the slack conversation, we'll keep your implementation here 🥳

PR #2488, opened 2025-06-03
chore(python): fix python dependency issues

yitzhakc:
## Description
Solve an issue with our Build actions that was causing the build to fail on some kind of python dependency issue. [See example here](https://github.com/Sefaria/Sefaria-Project/actions/runs/15412455008/job/43370277506)

## Code Changes
Modified the `requirements.txt` to pin the celery version and remove the direct dependence on redis-py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -70,7 +70,7 @@ unicodecsv==0.14.1
 unidecode==1.1.1
 user-agents==2.2.0
 pytest-django==4.9.*
-
+redis==5.2.1
 
 #opentelemetry-distro
 #opentelemetry-exporter-otlp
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -58,7 +58,6 @@ python-bidi
 pytz
 pyyaml==6.0.1
 rauth==0.7.3
-redis==3.5.3
```
EliezerIsrael, line 61: Without the pin, in looks like redis 5.2.1 gets installed.  
Have we looked into the implications of that?  

It seems like 6.x would break, due to version conflicts with the server. 
Do we want to pin 5.x?

EliezerIsrael (changes requested):

requirements.txt
yitzhakc: Yes
I'm waiting for the pytests to finish running, but I figure we should probably also raise a cauldron.

PR #2493, opened 2025-06-05
Add new class to Lexicon.py for the new Krupnik dictionary and fonts refactor

YishaiGlasner:
##New object for Krupnik
For new lexicon we need a new class in lexiicon.py that creates the string from the entry.

##Fonts
Remove fonts rules from common.css and s2.css to new fonts.css.
Remove 'EnglishInHebrew Serif Font', which is identical to adobe-garamond-pro, but limited. Rather, limit Taamey Frank not to include Latin characters.
Remove 'HebrewInEnglish Serif Font', which is identical to Taamey-Frank, but limited. Rather, movw adobe-garamond-pro before Taamey-Frank for punctuation.
Add doc strings to explain every font.
Add fonts.css to base.html.
Remove cyrillic yi from Crimson Text, which does not supported anyway.

In some cases, we have punctuation within some phrase in Latin letters. In this case, we want it to use the english font. For that we should wrap them with some class, and have a css rule for that. That was done for the englishWithinHebrew class. in s2.css.

#Note
I've changed re to regex in normaliztion.py for using negative lookbehind in the parsing process (Noah had no problem with this change).


templates/base.html
```diff
@@ -72,7 +72,7 @@
         WebFont.load({
           google: {
             families: ['Crimson Text:100,200,300,400,500,600,700,800,900'],
-            text: 'ăǎġḥḤḫḳḲŏŠšṭżūẓŻāīēḗęîїìi̧ ̆̄'
+            text: 'ăǎġḥḤḫḳḲŏŠšṭżūẓŻāīēḗęîìi̧ ̆̄'
```
EliezerIsrael, line 75: What happened to `ї`?

EliezerIsrael: Yishai answered off channel - this is a Cyrillic char, not handled by this font in any case.  Some of our texts mistakenly use the Cyrillic ї instead of the Latin one.

PR #2495, opened 2025-06-05
Node noencode

EliezerIsrael:
## Description
We were needlessly encoding SSR requests to node.  This sends JSON without encoding, saving O(# chars) ops.


EliezerIsrael:
I implemented this downstream of another PR - https://github.com/Sefaria/Sefaria-Project/pull/2489
Only the last commit is new.  Don't encode needlessly.

sefaria/model/text_request_adapter.py
```diff
@@ -143,54 +143,97 @@ def _add_node_data_to_return_obj(self) -> None:
...
-                except InputError:
-                    continue
-                ne_by_secs[ne_ref.sections[-1]-1,] += [ne]
-            return ne_by_secs
+        # Early return for default format to avoid any processing
+        if self.return_format == 'default':
+            return
 
+        # Pre-compute shared data outside the version loop
+        shared_data = {}
```
akiva10b, line 151: love it

sefaria/model/text_request_adapter.py
```diff
@@ -143,54 +143,97 @@ def _add_node_data_to_return_obj(self) -> None:
...
+            shared_data['all_segment_refs'] = all_segment_refs
+
+        def make_named_entities_dict(version_title, language):
+            # Cache named entities per version to avoid repeated DB queries
+            cache_key = f"{version_title}_{language}"
+            if cache_key not in shared_data:
+                named_entities = RefTopicLinkSet({"expandedRefs": {"$in": [r.normal() for r in shared_data['all_segment_refs']]},
+                                                  "charLevelData.versionTitle": version_title,
+                                                  "charLevelData.language": language})
+                # assumption is that refTopicLinks are all to unranged refs
```
akiva10b, line 170: can you elaborate more here, what is the significance of this

sefaria/model/text_request_adapter.py
```diff
@@ -143,54 +143,97 @@ def _add_node_data_to_return_obj(self) -> None:
...
+                ne_by_secs = defaultdict(list)
+                for ne in named_entities:
+                    try:
```
akiva10b, line 173: this function would benefit from being an async function, allowing the loop to continue loading data without blocking the thread on a retrieval of a particular ref (this might need to be a different PR, but definitely in the prism of a refactor)

reader/views.py
```diff
@@ -134,12 +134,11 @@ def render_react_component(component, props):
...
 
-    encoded_args = urllib.parse.urlencode({
-        "propsJSON": propsJSON,
-    }).encode("utf-8")
+    data = propsJSON.encode("utf-8")
+    headers = {"Content-Type": "application/json; charset=utf-8"}
     try:
-        req = urllib.request.Request(url)
-        response = urllib.request.urlopen(req, encoded_args, NODE_TIMEOUT)
+        req = urllib.request.Request(url, data=data, headers=headers)
```
akiva10b, line 140: brilliant!

akiva10b: I believe that if we are cleaning things up, we can use requests here

`response = requests.post(url, json=propsJSON, timeout=NODE_TIMEOUT)`

should automatically set the headers and encode properly (although, if it does anything more expensive than urllib, not worth it)

sefaria/model/text_request_adapter.py
EliezerIsrael: This is carried over from the old code.  No changes.  
Taking ranged refs and unpacking them is a potentially expensive op (needing a version record read).  
My sense is that this is a declaration of assumption that we just didn't put ranged refs in the data, but it deserves to be verified.

PR #2507, opened 2025-06-11
Fix admin/reset/index

YishaiGlasner:
## Description
Fixes a bug with admin/reset/index after changing an index structure.

## Code Changes
In views.py, get a fresh instance of the index after refreshing cache, rather than get it from Ref that was loaded before. That fixes the bug with admin/reset/index.


stevekaplan123 (commented):
Is this fixing the bug that existed solely on prod or is this an additional bug that also affects local?

PR #2508, opened 2025-06-11
Feature/sc 34734/ensure working modularization cauldron

stevekaplan123:
## Description
Simply changes a lot of links.


static/js/ReaderApp.jsx
```diff
@@ -795,6 +795,21 @@ class ReaderApp extends Component {
...
       hist.url += window.location.hash;
     }
+    
+    // For modularization QA, we want to make sure /sheets is at the beginning of URL if and only if we are in the sheets module.
+    if (Sefaria.activeModule === "sheets") {
+      if (!hist.url.startsWith("/sheets")) {
+        hist.url = "/sheets" + (hist.url.startsWith("/") ? "" : "/") + hist.url;
+      }
+    } else {
+      if (hist.url.startsWith("/sheets")) {
```
yitzhakc: Here you're dealing with the situation of transitioning from `sheets` to `library`, but the URL still has `sheets` in it? Can you better describe the scenario?

In addition, I don't quite understand the game with the slashes, leading or trailing.. Can you explain it to me please?

static/js/TopicsPage.jsx
```diff
@@ -11,26 +11,20 @@ import { NavSidebar, SidebarModules } from './NavSidebar';
...
-      </div>
-    </div>
-  );
-  categoryListings = (
-    <div className="readerNavCategories">
-      <ResponsiveNBox content={categoryListings} initialWidth={initialWidth} />
-    </div>
-  );
+  const description = {"en": "Browse or search our complete list of topics.", "he": Sefaria._("Browse or search our complete list of topics.")};
+  const topicsA_Z = <Card cardTitleHref={`/topics/all/${letter}`}
```
yitzhakc: Does this need to be module-aware?

static/js/ReaderApp.jsx
stevekaplan123: I believe I simplified the function.  It is now called modifyURLbasedOnModule.  It just forces /sheets to be at the beginning of sheets module links and removes /sheets if it's not in sheets module.

static/js/TopicsPage.jsx
stevekaplan123: Yes.  Fixed that. Thanks

akiva10b (approved):

PR #2509, opened 2025-06-11
chore: add urls

stevekaplan123:
Add URLs to footer


YishaiGlasner:
@stevekaplan123 when one is on sheets and clicks about help etc, should be move to library?

stevekaplan123:
> @stevekaplan123 when one is on sheets and clicks about help etc, should be move to library?

@YishaiGlasner What would you suggest we do instead?

stevekaplan123:
> > > @stevekaplan123 when one is on sheets and clicks about help etc, should be move to library?
> > 
> > 
> > @YishaiGlasner What would you suggest we do instead?
> 
> idk, it's more a product question, if the idea is that we have a different site for sheets, i guess that when one click help he shouldn't find himself in another site.

@YishaiGlasner Mickey said that for the time being this is good enough.  We'll address these issues during QA.

yitzhakc (approved):

PR #2510, opened 2025-06-11
Feature/sc 34012/strapi bug fix viewed event

relyks:
## Description

## Code Changes

## Notes

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -2105,7 +2105,7 @@ const LearnAboutNewEditorBanner = () => {
   );
 };
 
-const Banner = React.memo(({ onClose }) => {
+const Banner = ({ onClose }) => {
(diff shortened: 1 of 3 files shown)


relyks:
> @relyks Will this fire multiple events if the user scrolls it in view, out of view, in view, out of view, etc? If so, is that as desired?

It does fire multiple events if the user scrolls in and out of view, but only if the entire element and its children fully exits and fully enters the view again after the first time it's fired. I think it should do that, because that'd be another impression and a possible re-engagement with content. I was mainly trying to prevent events from being fired due to re-rendering from React, scrolling, and resizing.

The IntersectionObserver is only created once on mount and not on every re-render. There are refs to maintain their values across re-renders. The ref for the callback is called without recreating the IntersectionObserver

yitzhakc (approved):

PR #2511, opened 2025-06-12
Exclude <a> tags from being wrapped again.

YishaiGlasner:
## Description
In our data we have `<a>` tags. When requestong for it our links wrapper doesn't consider that, so they can be re-wrapped, or worse, their ref attributes cab be wrapped, ruining the html structure.
This PR excludes the a tags from being wrapped.

## Code Changes
In `get_wrapped_refs_string` the a tags are splitted out by regex, and the wrapping applied only on the other substrings.
The regex assume we have no `<a>` tags nested within `<a>` tags (which we shouldn't have).

## Performance
We have 2 cases:
1. segment without <a> tags - `re.split` takes time (~3ms for 'normal' segment), and for that reason I've added an if statement that takes ~0.1ms for a 'normal' segment, so a section request will take few ms more than previously.
2. segment with <a> tags - here we apply `re.split`, and have more calls to `apply_action_for_all_refs_in_string`, but on the other hand, any call takes much shorter string, which has a big influence for the regex parsing which this function does. In a check on Jerusalem Talmud Shabbat 2:5:1, we have *gained* more than 10ms for the section request (and it has also Hebrew that has no tags).

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -6024,9 +6024,9 @@ class Library(object):
             substrings = re.split(html_a_tag_reg, st)
         new_string = ''
         for i, substring in enumerate(substrings):
-            if i // 2 != i / 2:  # An <a> tag
+            if i % 2 == 1:  # An <a> tag
                 new_string += substring
-            elif i // 2 == i / 2 and substring:
+            elif i % 2 == 0 and substring:
                 new_string += self.apply_action_for_all_refs_in_string(substring, self._wrap_ref_match, lang,
                                                                        citing_only, reg, title_nodes)
         return new_string
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -6017,8 +6017,11 @@ class Library(object):
         :param citing_only: boolean whether to use only records explicitly marked as being referenced in text
         :return: string:
         """
-        html_a_tag_reg = '(<a [^<>]*>.*?</a>)'  # Assuming no nested <a> within <a>
-        substrings = re.split(html_a_tag_reg, st)
+        if '<a' not in st:  # This is 30 times faster than re.split, and applies for most cases
+            substrings = [st]
+        else:
+            html_a_tag_reg = '(<a [^<>]*>.*?</a>)'  # Assuming no nested <a> within <a>
+            substrings = re.split(html_a_tag_reg, st)
         new_string = ''
         for i, substring in enumerate(substrings):
             if i // 2 != i / 2:  # An <a> tag
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -6017,7 +6017,7 @@ class Library(object):
         :param citing_only: boolean whether to use only records explicitly marked as being referenced in text
         :return: string:
         """
-        if '<a' not in st:  # This is 30 times faster than re.split, and applies for most cases
+        if '<a ' not in st:  # This is 30 times faster than re.split, and applies for most cases
             substrings = [st]
         else:
             html_a_tag_reg = '(<a [^<>]*>.*?</a>)'  # Assuming no nested <a> within <a>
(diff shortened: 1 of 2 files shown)


sefaria/model/text.py
```diff
@@ -6009,14 +6009,27 @@ def get_regex_and_titles_for_ref_wrapping(self, st, lang, citing_only=False):
...
+        Returns a string with the list of Ref objects derived from string wrapped in <a> tags,
+        excluding refs that are already wrapped in the data
 
         :param string st: the input string
         :param lang: "he" or "en"
         :param citing_only: boolean whether to use only records explicitly marked as being referenced in text
         :return: string:
         """
-        return self.apply_action_for_all_refs_in_string(st, self._wrap_ref_match, lang, citing_only, reg, title_nodes)
+        if '<a' not in st:  # This is 30 times faster than re.split, and applies for most cases
```
akiva10b: do we also use markdown for this? We support markdown, but I dont know if we use it for links

sefaria/model/text.py
```diff
@@ -6009,14 +6009,27 @@ def get_regex_and_titles_for_ref_wrapping(self, st, lang, citing_only=False):
...
+            substrings = [st]
+        else:
+            html_a_tag_reg = '(<a [^<>]*>.*?</a>)'  # Assuming no nested <a> within <a>
```
akiva10b, line 6023: I imagine this assumption is correct. Is there a scenario it isn't and we should run this recursively?

sefaria/model/text.py
```diff
@@ -6009,14 +6009,27 @@ def get_regex_and_titles_for_ref_wrapping(self, st, lang, citing_only=False):
...
+            substrings = re.split(html_a_tag_reg, st)
+        new_string = ''
+        for i, substring in enumerate(substrings):
+            if i % 2 == 1:  # An <a> tag
+                new_string += substring
+            elif i % 2 == 0 and substring:
```
akiva10b, line 6029: can you decrypt this scenario for me?

sefaria/model/text.py
akiva10b: I would recommend search for '<a ' instead of '<a' as a case where for whatever reason we have <apple in text or start support for the <article> tag, this would break. I cannot think of a scenario where <a> would exists alone which is why '<a ' might be best

akiva10b (commented):
overall looks good, I added some comments- this code would definitely benefit from some pytests

sefaria/model/text.py
YishaiGlasner: i think that if we have such a case, it's a bad data that should be fixed.
recursion won't help you for in first time it will remove the outer starting tag and the inner enfing tag.
i don't think this scenario can be addressed with regex, because html is not a regular language.

sefaria/model/text.py
YishaiGlasner: say you have `st = 'hello <a href='Genesis'>Akiva</a> Berger'`
so `substring` will be `['hello ', '<a href='Genesis'>Akiva</a>', ' Berger']`
we want to change the even (based 0) index, and not touch the odd ones.

sefaria/model/text.py
YishaiGlasner: AFAIK, we don't support markdown on texts. it's only in descriptions and things alike.
i've added a space. for some reason it seems sarch for '<a ' is faster than '<a'. i thought it should be the opposite for this is longer. can you explain this?

sefaria/model/text.py
akiva10b: sadly not in able to run the code - so this would work properly with `st = 'hello <a href='Genesis'> Akiva </a> Berger'`
And as I wrote in the previous review - can you add pytests?

YishaiGlasner:
```
In [47]: st = "hello <a href='Genesis'> Akiva </a> Berger"

In [48]: html_a_tag_reg = '(<a [^<>]*>.*?</a>)'

In [49]: re.split(html_a_tag_reg, st)
Out[49]: ['hello ', "<a href='Genesis'> Akiva </a>", ' Berger']
```

test added

akiva10b (approved):

PR #2512, opened 2025-06-12
Bring back Topics Launch Banner

relyks:
## Description
This enables the topics launch banner again and hopefully provides valid analytics for the button clicks and banner impressions. This is independent of the code that uses Strapi to show banners, modals, and sidebar ads.

## Code Changes
* Enable the topics launch banner again
* Add analytics attributes to the explore button, close button, and for banner impressions using the Analytics Tracker module
* Add analytics attributes to the Explore link in the page header, so in case the analytics don't work properly for the banner, the traffic to the page can be solely attributed to the banner. Using referral traffic, traffic from the Explore link can be eliminated in analytics reports.

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -50,7 +50,7 @@ class Header extends Component {
           { Sefaria._siteSettings.TORAH_SPECIFIC ?
           <a className="home" href="/" >{logo}</a> : null }
           <a href="/texts" className="textLink"><InterfaceText context="Header">Texts</InterfaceText></a>
-          <a href="/topics" className="textLink" data-anl-event="topicsClick:click" data-anl-panel_type="header" data-anl-link_type="navigation">
+          <a href="/topics" className="textLink" data-anl-event="topicsClick:click">
             <InterfaceText context="Header">Explore</InterfaceText>
           </a>
           <a href="/community" className="textLink"><InterfaceText>Community</InterfaceText></a>
@@ -107,7 +107,7 @@ class Header extends Component {
       mobile: !this.props.multiPanel
     });
     return (
-      <div className={headerClasses} role="banner">
+      <div className={headerClasses} role="banner" data-anl-panel_type="header" data-anl-link_type="navigation">
         <div className={headerInnerClasses}>
           {this.props.multiPanel ? headerContent : mobileHeaderContent}
         </div>
@@ -198,7 +198,7 @@ const MobileNavMenu = ({onRefClick, showSearch, openTopic, openURL, close, visib
     closed: !visible,
   });
   return (
-    <div className={classes}>
+    <div className={classes} data-anl-panel_type="mobile_nav">
       <div className="searchLine">
         <HeaderAutocomplete
             onRefClick={onRefClick}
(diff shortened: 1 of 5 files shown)


static/js/ReaderApp.jsx
```diff
@@ -2302,7 +2302,8 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
         <AdContext.Provider value={this.getUserContext()}>
           <div id="readerAppWrap">
             <InterruptingMessage />
-            <Banner onClose={this.setContainerMode} />
+            <TopicsLaunchBanner onClose={this.setContainerMode} />
+            {/* <Banner onClose={this.setContainerMode} /> */}
```
nsantacruz: Remove commented out code

static/js/TopicsLaunchBanner.jsx
```diff
@@ -4,7 +4,7 @@ import TopicsLaunchBannerMobileGraphics from "./TopicsLaunchBannerMobileGraphics
 import { OnInView } from "./Misc";
 
 const TopicsLaunchBanner = ({ onClose }) => {
-  const bannerName = "2025-topics_launch";
+  const bannerName = "2025-topics_launch-2";
```
nsantacruz, line 7: Why is this hard-coded and not coming from strapi?

static/js/TopicsLaunchBanner.jsx
```diff
@@ -127,7 +109,9 @@ const TopicsLaunchBanner = ({ onClose }) => {
   }
 
   return (
-    <OnInView onVisible={trackBannerImpression}>
+    <div data-anl-event="banner_view:scrollIntoView"
```
nsantacruz, line 112: `data-anl-<ga4-field>` attributes cascade to lower elements (as opposed to `data-anl-event` which acts specifically on the element it's added to). Therefore, the goal is to put common attributes at the lowest level in the DOM that can be shared by components. For example, all of the `data-anl-<ga4-field>` on `TopicsLaunchBannerGraphics` and `TopicsLaunchBannerMobileGraphics` should be put on `TopicsLaunchBanner` so they can be shared.

static/js/Header.jsx
```diff
@@ -211,7 +213,7 @@ const MobileNavMenu = ({onRefClick, showSearch, openTopic, openURL, close, visib
         <img src="/static/icons/book.svg" />
         <InterfaceText context="Header">Texts</InterfaceText>
       </a>
-      <a href="/topics" onClick={close}>
+      <a href="/topics" onClick={close} data-anl-event="topicsClick:click" data-anl-panel_type="mobile_nav" data-anl-link_type="navigation">
```
nsantacruz: I would put `data-anl-panel_type` at the root of `MobileNavMenu` since it will apply to all elements below.

static/js/TopicsLaunchBanner.jsx
nsantacruz: Also, this variable should be passed to `data-anl-promotion_name` so it's not repeated.

relyks: > Why is this hard-coded and not coming from strapi?

This entire banner has essentially been hard-coded. It has custom behavior and design that wouldn't have been possible with Strapi and the existing code

static/js/TopicsLaunchBanner.jsx
relyks: Thank you 🙏 makes sense

nsantacruz (approved):

PR #2513, opened 2025-06-16
Create new `change_lexicon_headword` func

YishaiGlasner:
## Description
Create new `change_lexicon_headword` functuin for safely changing headwords. It will:

- change the entry itself
- change prev and next entries' `next` and `prev` attributes.
- change index `firstWord`, `lastWord` and `headwordsMap` if needed.
- change links' refs
- change word forms

### DIFF
--- a/sefaria/helper/schema.py
+++ b/sefaria/helper/schema.py
@@ -1070,7 +1070,7 @@ def change_lexicon_headword(parent_lexicon, old_headword, new_headword):
         needs_rewrite = lambda x, *args: bool(re.search(old_ref_reg, x))
         cascade(index.title, rewriter, needs_rewrite, True)
 
-    #word forms
+    # word forms
     print('Updating word forms')
     db.word_form.update_many(
         {


sefaria/helper/schema.py
```diff
@@ -994,3 +995,99 @@ def change_term_hebrew(en_primary, new_he):
     t.add_title(new_he, "he", True, True)
     t.remove_title(old_primary, "he")
     t.save()
+
+
+def change_lexicon_headword(parent_lexicon, old_headword, new_headword):
+    """
```
stevekaplan123, line 1001: I think you should include an example of how to use this function as I have never dealt with lexicon before.  Also, why doesnt this change internal references?  Is there another tool for that?

YishaiGlasner: do you thin something like
```
change_lexicon_headword('BDB Dictionary', 'אריה', 'נמר')
```
will help?
you're right that you can't understand it without knowing what `headword` and `parent_lexicon` are, but this is not the place for that.

> I think you should include an example of how to use this function as I have never dealt with lexicon before. Also, why doesnt this change internal references? Is there another tool for that?

generally it seems we don't support that, not only in lexicons. for example, in the English translation of Yerushalmi we have <a> tags within the text. let's say we have an <a> tag that refers to `Yerushalmi Shabbat 2:1:3` - if you'll change the name of Yerushalmi Shabbat, our cascade function won't change it. This is something that should be considered when changing things.

stevekaplan123 (approved):
I'm approving this, but I think you should add an example or two in the comments.

stevekaplan123 (approved):

PR #2517, opened 2025-06-17
chore: TopicSideColumn uses shouldDisplayTopic

stevekaplan123:
## Description
Related topics display in the sidebar on topic pages.  If we are in a topic page in the sheets module, related topics should only display if they themselves have sheets.  Likewise, in a topic page in the library module, related topics should only display if they themselves have library sources.

## Code Changes
1.  Minor syntax change on the back-end: I'm calling topic.get_pools() instead of calling the pools object itself.  This is easier to read and I didn't know we had this function before.
2. The main change is that in the back-end,  sefaria/helper/topic.py, topic links are populated with get_pools() so taht in the front end, in TopicSideColumn, we can check if we should display the topic link given the current active module.

### DIFF
--- a/static/js/TopicPage.jsx
+++ b/static/js/TopicPage.jsx
@@ -880,7 +880,7 @@ TopicLink.propTypes = {
 
 const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, setNavTopic, timePeriod, properties, topicTitle, multiPanel, topicImage }) => {
   const category = Sefaria.displayTopicTocCategory(slug);
-  const linkTypeArray = links ? Object.values(links).filter(linkType => !!linkType && linkType.shouldDisplay && linkType.links.filter(l => l.shouldDisplay !== false).length > 0) : [];
+  const linkTypeArray = links ? Object.values(links).filter(linkType => !!linkType && linkType.shouldDisplay && linkType.links.filter(l => Sefaria.shouldDisplayTopic(l)).length > 0) : [];
   if (linkTypeArray.length === 0) {
     linkTypeArray.push({
       title: {
@@ -911,7 +911,7 @@ const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, set
         return a.title.en.localeCompare(b.title.en);
       })
       .map(({ title, pluralTitle, links }) => {
-        const linksToDisplay = links.filter(l => l.shouldDisplay !== false);
+        const linksToDisplay = links.filter(l => Sefaria.shouldDisplayTopic(l));
         const hasPlural = linksToDisplay.length > 1 && pluralTitle;
         const pluralizedTitle = {
           en: hasPlural ? pluralTitle.en : title.en,
--- a/static/js/TopicPage.jsx
+++ b/static/js/TopicPage.jsx
@@ -584,7 +584,7 @@ const TopicPage = ({
                     <TopicSideColumn
                         key={topic}
                         slug={topic}
-                        links={topicData.links}
+                        linksByType={topicData.links}
                         clearAndSetTopic={clearAndSetTopic}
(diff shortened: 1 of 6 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -2746,11 +2746,10 @@ _media: {},
   },
   shouldDisplayTopic: function(topic) {
     /*
-    Returns true if topic should be displayed in the topic list or TOC.
-    'topic' is a topic object in this.topicList or this.topic_toc
+    Returns true if topic should be displayed in the topic list, topic TOC, or topic page side column.
      */
     const inActiveModule = topic?.pools?.includes(Sefaria.activeModule);
-    return !!topic.shouldDisplay && inActiveModule;
+    return topic.shouldDisplay !== false && inActiveModule;
```
YishaiGlasner: why this? (i guess that `topic.shouldDisplay` should be a boolean, so i also don't understand the `!!` from the previous version)

static/js/sefaria/sefaria.js
```diff
@@ -2746,11 +2746,10 @@ _media: {},
   },
   shouldDisplayTopic: function(topic) {
```
YishaiGlasner: this name is misleading. one would expect that if `topic.shouldDisplay` so `shouldDisplayTopic` - it's exactly the same thing.
I recommend `shouldDisplayInActiveModule` but if there is something shorter it will be better (I think we can omit the word 'topic' in the name for it's already the name of the param).

stevekaplan123: I agree with you.  Changing the code to reflect this

static/js/sefaria/sefaria.js
stevekaplan123: This is because previously we we only set shouldDisplay if it was false, which is weird.  I'm changing it so it always gets set

YishaiGlasner: are you changing the logic of only setting shouldDisplay if it was false? if so we don't need the `!== false`, but can do only `topic.shouldDisplay`.
if you're not changing this, please add a note, for looking on the client side one cannot understand it and may change it.

YishaiGlasner: never mind, i see you have already chaged it

static/js/TopicPage.jsx
```diff
@@ -880,7 +880,7 @@ TopicLink.propTypes = {
 
 const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, setNavTopic, timePeriod, properties, topicTitle, multiPanel, topicImage }) => {
   const category = Sefaria.displayTopicTocCategory(slug);
-  const linkTypeArray = links ? Object.values(links).filter(linkType => !!linkType && linkType.shouldDisplay && linkType.links.filter(l => l.shouldDisplay !== false).length > 0) : [];
+  const linkTypeArray = links ? Object.values(links).filter(linkType => !!linkType?.shouldDisplay && linkType.links.filter(l => Sefaria.shouldDisplayInActiveModule(l)).length > 0) : [];
```
YishaiGlasner: it's difficult to understand what's going on here (and not davke due to your changes).
it seems that `links` is an object, that its keys are something unknown (and seems to be not necessary), and its values are `linkType`s. i'd expect a `linkType` to be a string that tells what the type of the link, but it seems to be an object with a key `links`, so we have and object called `links` and deep within it we have `links`.
so the problem is not only the densty of the filter logic as copilot says. it's difficult to understand also with its suggestion.

stevekaplan123: I moved code from TopicSideColumn into a preprocessing function called preprocessLinksByType with a lengthy comment.  Is it clearer now?

YishaiGlasner:
this is a great improvement for the readibility!

PR #2518, opened 2025-06-17
chore: syntax fix

stevekaplan123:
## Description
Syntax fix: `en` -> `primaryTitle.en`

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -2987,7 +2987,7 @@ _media: {},
   },
   _initTopicTocSlugToTitleReducer: function(a,c) {
     if (!c.children) { return a; }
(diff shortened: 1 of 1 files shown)


static/js/sefaria/sefaria.js
```diff
@@ -2987,7 +2987,7 @@ _media: {},
   },
   _initTopicTocSlugToTitleReducer: function(a,c) {
```
YishaiGlasner, line 2988: what are `a` and `c`? this is a riddle.

EliezerIsrael: Standard arg names for a reducer, no?  `a` is state, `c` is action, but don't ask me why.

EliezerIsrael (approved):

PR #2525, opened 2025-06-26
fix(Editor): fix divine name replacement logic

yonadavGit:
## Description
Replaces divine names correctly in both Hebrew / English Slate editors. The old solution tried Transforms.setNodes, which can’t mutate a leaf’s text. We now walk all text nodes bottom-up, remove each matching leaf, and re-insert a new leaf with the updated string while preserving existing marks fixing the override.

### DIFF
--- a/static/js/Editor.jsx
+++ b/static/js/Editor.jsx
@@ -699,21 +699,29 @@ const BoxedSheetElement = ({ attributes, children, element, divineName }) => {
 
 
     useEffect(() => {
-      const replacement  = divineName || 'noSub';
-      const editors      = [sheetSourceHeEditor, sheetSourceEnEditor];
+      const replacement = divineName || "noSub";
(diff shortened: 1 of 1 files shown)


static/js/Editor.jsx
```diff
@@ -699,21 +699,29 @@ const BoxedSheetElement = ({ attributes, children, element, divineName }) => {
 
 
     useEffect(() => {
-      const replacement  = divineName || 'noSub';
-      const editors      = [sheetSourceHeEditor, sheetSourceEnEditor];
+      const replacement = divineName || "noSub";
+      const editors = [sheetSourceHeEditor, sheetSourceEnEditor];
+      console.log("Replacing divine names with:", replacement);
```
nsantacruz: remove console.log

nsantacruz (dismissed):

nsantacruz (approved):

PR #2533, opened 2025-07-01
Saving saved playwright tests

jessicafeinmesser:
## Description
A Playwright E2E test suite for editor save/saved functionality including:
Save State Testing: Visual indicators for "Saved", "Saving", "Logged Out", "Trying to Connect", "Error"
Auto-save Testing: Comprehensive auto-save behavior verification
Multi-language Support: Full English + Hebrew interface testing
Authentication Flows: Login/logout scenarios and session recovery

## Code Changes
sourceSheetEditorPage.ts added to the pages folder
autosave.spec.ts test file added to tests folder
new constants file 
a few additions to utils.ts and other files

## Notes


e2e-tests/contants.ts
schreibersefaria, line 1: Duplicate file

e2e-tests/constants.ts
schreibersefaria, line 1: You could get the Heb in the same way the code get's it.
That way you don't have to change the code here if the hebrew copy changes / we have more languages added.
Ideally we wouldn't have any hardcoded text here but I don't know if that is possible.

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,55 @@
...
-    // Maybe we should go about this a better way, like with a config file that is saved to each dev's machine, rather than hard coded in test files
-    // Also, this code is a refactor of utils.ts: loginUser, just with using the POM structure
-    async loginAs(email: string, password: string){
-        if(this.language == LANGUAGES.HE){
-            await this.page.getByPlaceholder('כתובת').fill(email)
-            await this.page.getByPlaceholder('סיסמא').fill(password)
-            await this.page.getByRole('button', {name: 'התחברות'}).click()
-            //await this.page.getByRole('link', { name: 'See My Saved Texts' }).isVisible();
+    public async ensureLanguage(language: string) {
+        const currentURL = this.page.url();
```
schreibersefaria: var never used

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,55 @@
...
+        const langMenuButton = this.page.locator('a.interfaceLinks-button >> img[alt*="שפת ממשק"]');
+    
+        if (await langMenuButton.isVisible()) {
+            console.log(`Language menu button is visible: ${await langMenuButton.isVisible()}`);
```
schreibersefaria: I can't see console logs in other pages. Was this for testing or is it needed?

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,55 @@
...
 
-    // Maybe we should go about this a better way, like with a config file that is saved to each dev's machine, rather than hard coded in test files
-    // Also, this code is a refactor of utils.ts: loginUser, just with using the POM structure
-    async loginAs(email: string, password: string){
-        if(this.language == LANGUAGES.HE){
-            await this.page.getByPlaceholder('כתובת').fill(email)
-            await this.page.getByPlaceholder('סיסמא').fill(password)
-            await this.page.getByRole('button', {name: 'התחברות'}).click()
-            //await this.page.getByRole('link', { name: 'See My Saved Texts' }).isVisible();
+    public async ensureLanguage(language: string) {
```
schreibersefaria: Is this function specific for the loginPage? If not maybe it could be made generic somewhere else

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,55 @@
...
         }
-        else{
-            await this.page.getByRole('textbox', {name: 'Email'}).fill(email)
-            await this.page.getByRole('textbox', {name: 'Password'}).fill(password)
+    }
+    
+
+    async loginAs(email: string, password: string) {
+        if (this.language === LANGUAGES.HE) {
+            //await this.page.waitForSelector('[placeholder="כתובת"]');
```
schreibersefaria: Could you clean out comments or put in the reason you leave them there.

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,55 @@
...
+            await this.page.getByPlaceholder('Email Address').fill(email);
+            await this.page.waitForSelector('[placeholder="Password"]');
+            await this.page.getByPlaceholder('Password').fill(password);
             await this.page.getByRole('button', { name: 'Login' }).click();
-            await this.page.getByRole('link', { name: 'See My Saved Texts' }).isVisible();
         }
-        
+    
+        // Wait for login confirmation
+        await this.page.getByRole('link', { name: 'See My Saved Texts' }).waitFor({ timeout: 10000 });
```
schreibersefaria: Is this timeout something we can replace with waiting for a specific component or something?
I really don't know much about these tests, would this be a potential delay / reason for timeouts?

e2e-tests/pages/sourceTextPage.ts
```diff
@@ -16,6 +16,13 @@ export class SourceTextPage extends HelperBase{
         await this.page.locator('div').filter({ hasText: sourceLanguage }).click()
     }
 
+    async setContentLanguage(mode: "Source" | "Translation" | "Source with Translation") {
+        await this.page.getByAltText('Toggle Reader Menu Display Settings').click()
+        await this.page.getByRole('radio', { name: mode, exact: true }).click();
+       // await this.page.locator('.show-source-translation-buttons').getByLabel(mode).click();
```
schreibersefaria: Another comment that I assume you forgot to clean (I'll stop commenting on everyone)

e2e-tests/pages/sourceTextPage.ts
```diff
@@ -16,6 +16,13 @@ export class SourceTextPage extends HelperBase{
         await this.page.locator('div').filter({ hasText: sourceLanguage }).click()
     }
 
+    async setContentLanguage(mode: "Source" | "Translation" | "Source with Translation") {
+        await this.page.getByAltText('Toggle Reader Menu Display Settings').click()
```
schreibersefaria: I see getByAltText isn't common.
Just mentioning in case there is a more standard way to do this here.

e2e-tests/pages/sourceTextPage.ts
```diff
@@ -57,7 +64,26 @@ export class SourceTextPage extends HelperBase{
...
+        await this.page.getByRole('link', { name: 'Table of Contents' }).click();
+    }
+    
+    async getTOCTitles(): Promise<string[]> {
+        const tocNodes = this.page.locator('.schema-node-toc');
+        return await tocNodes.evaluateAll(nodes =>
+          nodes
+            .map(node => {
+              const title = node.querySelector('.schema-node-title .contentSpan.en')?.textContent?.trim() || '';
+              // Ignore titles that are just digits (e.g., "1", "2", "3") unless meaningful
```
schreibersefaria: Ideally explain the why, not just the what. 
This is particularly helpful if someone want's to change the logic so they can understand if they are missing something.

e2e-tests/tests/topics.spec.ts
```diff
@@ -18,7 +18,10 @@ test('Check source', async ({ context }) => {
 });
 
 test('Check admin tab', async ({ context }) => {
-  const page = await goToPageWithUser(context, '/topics', testAdminUser);
+  if (!testAdminUser.email || !testAdminUser.password) {
+    throw new Error("testAdminUser must have defined email and password");
+  }
+  const page = await goToPageWithUser(context, '/topics', { email: testAdminUser.email, password: testAdminUser.password });
```
schreibersefaria: Why do you need to destruct the testAdminUser obj?

e2e-tests/utils.ts
```diff
@@ -1,12 +1,16 @@
...
+
+import { SourceSheetEditorPage } from './pages/sourceSheetEditor.page';
 
 let langCookies: any = [];
 let loginCookies: any = [];
 let currentLocation: string = ''; 
 
-const hideModals = async (page: Page) => {
+//method to hid modals that interrupt the user experience
+export const hideModals = async (page: Page) => {
```
schreibersefaria, line 39: What do you think of this logic be part of page navigation (maybe a default behaviour) instead of exporting?
That way the tests don't need to "know" it

e2e-tests/utils.ts
```diff
@@ -15,15 +19,144 @@ const hideModals = async (page: Page) => {
     });
 }
 
-export const changeLanguage = async (page: Page, language: string) => {
+export async function dismissNewsletterPopupIfPresent(page: Page) {
```
schreibersefaria: Also here. Wouldn't you always want this by default unless testing this?

e2e-tests/utils.ts
```diff
@@ -15,15 +19,144 @@ const hideModals = async (page: Page) => {
...
+      });
+
+      // Also clean up any <body> modifications
+      document.body.style.overflow = 'auto';
+      document.body.style.pointerEvents = 'auto';
+
+      console.log(`Popup and overlay successfully removed. Elements removed: ${removedCount}`);
+    });
+
+    console.log('✅ Unbounce newsletter popup removed from DOM.');
```
schreibersefaria: Emojis aren't typically left it
It's also currently a tell of AI doing something ;-)

Also, do we need the logs here?

e2e-tests/utils.ts
```diff
@@ -15,15 +19,144 @@ const hideModals = async (page: Page) => {
...
+  } catch (err) {
+    console.warn('⚠️ Failed to remove newsletter popup:', err);
+  }
+}
+
+//method to hide the generic banner that appears on the editor page (Welcome to New Editor)
+export const hideGenericBanner = async (page: Page) => {
```
schreibersefaria, line 70: Also here, wouldn't this be default (I'll stop commenting this)

e2e-tests/utils.ts
```diff
@@ -15,15 +19,144 @@ const hideModals = async (page: Page) => {
...
+  await page.evaluate(() => {
+    const banner = document.querySelector('.genericBanner');
+    if (banner) {
+      (banner as HTMLElement).style.display = 'none';
+    }
+
+    // Fallback style override just in case dynamic styles re-show it
+    const style = document.createElement('style');
```
schreibersefaria, line 72: A fallback makes me think that maybe our primary code isn't working well enough.
Just a feeling, this may be important here.
It would be great to know why / in which cases this fallback is needed. (You do write "dynaic". That may be a full explanation that I don't understand)

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
...
+        await this.page.keyboard.type('Dummy text to test editability', { delay: 30 });
+        // Wait briefly for any potential text update
+        // Re-read the content to compare
+        const newText = await contentEditable.innerText();
+        expect(newText).toBe(originalText);
+    }
+
+    async waitForAutosave() {
+        // Wait at least 3 seconds to allow autosave to trigger
+        await this.page.waitForTimeout(3000);
```
schreibersefaria, line 183: Could we do this in a way that shortens the 3 seconds if possible

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    async simulateLogin(context) {
+        if (this.savedSessionCookie) {
+        await context.addCookies([this.savedSessionCookie]);
+        }
+    }
+     
+    // Checks whether a DOM element is functionally clickable or editable.
+    async isElementInteractable(el: ElementHandle<HTMLElement>): Promise<{
```
schreibersefaria, line 233: Is this a generic helper. Maybe it should be in utils

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
+import { BrowserContext, Cookie, ElementHandle, Locator, Page } from 'playwright-core';
+import { expect } from 'playwright/test';
+import {isClickable} from "../utils";
+import { SaveStates } from '../constants';
+import { LANGUAGES } from '../globals';
+
+
+export class SourceSheetEditorPage {
```
schreibersefaria, line 8: (AI generated comment)
This is a very big object violating the POM principle (see other POMs as an example)

Fix suggestion (Although if you agree you can just write a prompt to Cursor to work on the split. Remember to tell it to test before and after to make sure it didn't break anything):
Split into focused classes:
- SourceSheetEditorPage (UI interactions only)
- AuthenticationHelper (login/logout simulation)
- ConnectivityHelper (online/offline simulation)
- SaveStateValidator (save state assertions)

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
...
+    readonly page: Page;
```
schreibersefaria, line 9: (AI generated comment)
Another comment about this obj.
It has a different style from the other objects.
Doesn't extend HelperBase, no language support, different constructor pattern.

This can be incorporated into the object split prompt from the previous comment.

e2e-tests/tests/autosave.spec.ts
schreibersefaria, line 1: (AI generated comment)
Individual tests are very long (some over 30 lines), the naming isn't descriptive, and maybe most importantly there is repeated setup logic.
Maybe split into smaller tests with a specific and descriptive name.

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
...
+        this.page = page;
+    }
+
+    // Status Indicator Components----------------------------
+    statusIndicator = () => this.page.locator('.editorSaveStateIndicator');
+    statusMessage = () => this.page.locator('.saveStateMessage');
+    statusTooltip = () => this.page.locator('.editorSaveStateIndicator [data-tooltip]');
+
+    //helper function to validate text for the save state
+    async assertSaveState(state: any, language = LANGUAGES.EN) {
```
schreibersefaria, line 23: (AI generated comment)
It may be worth creating a proper interface instead of using any.
This function is used quite often

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,362 @@
...
+        await this.page.context().setOffline(true);
+    }
+
+    async simulateOnlineMode() {
+        await this.page.context().setOffline(false);
+    }
+
+    async simulateLogout(context) {
+        const cookies = await context.cookies();
+        const sessionCookie = cookies.find(c => c.name === 'sessionid');
```
schreibersefaria, line 215: There is no handeling for not sessionCookie not being found.
If it's checked elsewhere then it's a matter of style and feel free to leave as is / comment / add error handeling just in case

e2e-tests/pages/loginPage.ts
jessicafeinmesser: I looked into this, and AI suggested that the best recommended practice (that aligns with Page Object Model) is to use the login function in association with a login page. I had originally created a login function in utils called loginUser that does basically the same thing. Let me know which approach you think is better!

e2e-tests/pages/loginPage.ts
jessicafeinmesser: this was for testing

e2e-tests/pages/loginPage.ts
jessicafeinmesser: I will change to this        
 await expect(this.page.getByRole('link', { name: 'See My Saved Texts' })).toBeVisible();
which is similar to waiting for it but with a built in timeout so might save some time, I believe this is slightly better practice. This line kind of was already waiting for a specific component (ie the saved texts link) with a maximum timeout I think

e2e-tests/pages/sourceSheetEditor.page.ts
jessicafeinmesser: For the meantime, connectivity/authentication functions have been moved to utils, and save state things are still connected to the source sheet because for the time being they are exclusively for that page

jessicafeinmesser:
Added two new commits based on the first PR review

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,54 @@
...
+            if (language === LANGUAGES.EN && await englishOption.isVisible()) {
+                await englishOption.click();
+                console.log("Switched to English language");
+            } else if (language === LANGUAGES.HE && await hebrewOption.isVisible()) {
+                await hebrewOption.click();
+                console.log("Switched to Hebrew language");
+            }
+    
+            // Wait for redirection and page load
+            await this.page.waitForURL(/sefaria\.org.*login.*/);
```
schreibersefaria: Better not hardcode this

e2e-tests/pages/loginPage.ts
schreibersefaria: Un resolving because it's still here

e2e-tests/pages/loginPage.ts
schreibersefaria: Right, but this is an ensureLanguage function and not a login function, no?

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,54 @@
...
-    // Also, this code is a refactor of utils.ts: loginUser, just with using the POM structure
-    async loginAs(email: string, password: string){
-        if(this.language == LANGUAGES.HE){
-            await this.page.getByPlaceholder('כתובת').fill(email)
-            await this.page.getByPlaceholder('סיסמא').fill(password)
-            await this.page.getByRole('button', {name: 'התחברות'}).click()
-            //await this.page.getByRole('link', { name: 'See My Saved Texts' }).isVisible();
+    public async ensureLanguage(language: string) {
+        const currentURL = this.page.url();
+        const langMenuButton = this.page.locator('a.interfaceLinks-button >> img[alt*="שפת ממשק"]');
```
schreibersefaria: I'm confused about this. Wouldn't it only work if you are currently in Heb?

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,54 @@
...
+    
+        if (await langMenuButton.isVisible()) {
+            console.log(`Language menu button is visible: ${await langMenuButton.isVisible()}`);
+            await langMenuButton.click();
+    
+            const englishOption = this.page.locator('a.interfaceLinks-option.int-en');
+            const hebrewOption = this.page.locator('a.interfaceLinks-option.int-he');
```
schreibersefaria: Would it work to move these to inside the if?
That way only one of them will be executed (the one we need)

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,54 @@
...
+    
+            if (language === LANGUAGES.EN && await englishOption.isVisible()) {
+                await englishOption.click();
+                console.log("Switched to English language");
+            } else if (language === LANGUAGES.HE && await hebrewOption.isVisible()) {
+                await hebrewOption.click();
+                console.log("Switched to Hebrew language");
```
schreibersefaria: Noting some more logs

e2e-tests/pages/loginPage.ts
```diff
@@ -1,27 +1,54 @@
...
 
-    // Maybe we should go about this a better way, like with a config file that is saved to each dev's machine, rather than hard coded in test files
-    // Also, this code is a refactor of utils.ts: loginUser, just with using the POM structure
-    async loginAs(email: string, password: string){
-        if(this.language == LANGUAGES.HE){
-            await this.page.getByPlaceholder('כתובת').fill(email)
-            await this.page.getByPlaceholder('סיסמא').fill(password)
-            await this.page.getByRole('button', {name: 'התחברות'}).click()
-            //await this.page.getByRole('link', { name: 'See My Saved Texts' }).isVisible();
+    public async ensureLanguage(language: string) {
```
schreibersefaria: Is this function used?
I can't see any use, maybe we can just delete

e2e-tests/tests/topics.spec.ts
```diff
@@ -5,6 +5,7 @@ import {LANGUAGES, testAdminUser} from "../globals";
 
 test('Go to topic page', async ({ context }) => {
   const page = await goToPageWithLang(context, '/topics');
+  console.log('Page URL:', page.url());
```
schreibersefaria: Note, a log

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+      div[class*="ub-emb"] {
+        display: none !important;
+        visibility: hidden !important;
+        pointer-events: none !important;
+      }
+      body {
+        overflow: auto !important;
+        pointer-events: auto !important;
+      }
+    `;
```
schreibersefaria, line 64: This is a lot of specific HTML, It could make the tests flaky. Could we simplify this?
Maybe we can click on the close button.
Hopefully this will be fixed with fixtures at some point

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+  await page.evaluate(() => {
+    const style = document.createElement('style');
+    style.innerHTML = `
+      .readerControlsOuter,
+      header.readerControls.fullPanel.sheetReaderControls {
+        display: none !important;
+        pointer-events: none !important;
+        visibility: hidden !important;
+        z-index: -9999 !important;
+      }
```
schreibersefaria, line 107: Repeating the comment of the HTML for all the hide functions.
It seems like a lot of HTML and maybe we can use the actual close buttons.
If it's too slow then this can be a good patch until there are fixtures, but best to add a comment on why we did it this way.

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
 
-export const goToPageWithLang = async (context: BrowserContext, url: string, language=DEFAULT_LANGUAGE) => {
-    // If a cookie already has contents, clear it so that the language cookie can be reset
-    if (langCookies.length) {
-        await context.clearCookies()
-    }   
-    const page: Page = await context.newPage();
-    await page.goto(url);
+export const changeLanguageLoggedIn = async (page: Page, language: string) => {
+    console.log('Changing language to:', language);
```
schreibersefaria: log

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+  
+    const menu = page.locator('.interfaceLinks-menu.profile-menu');
+    await expect(menu).toBeVisible();
+  
+    // Select the correct language link
+    const languageLink = language === LANGUAGES.HE
+      ? page.locator('#select-hebrew-interface-link')
+      : page.locator('#select-english-interface-link');
+  
+      console.log('Selector being used:', language === LANGUAGES.HE ? '#select-hebrew-interface-link' : '#select-english-interface-link');
```
schreibersefaria: Log
I'm going to stop commenting on these. Maybe go through all the changes on github to double check these type of things

schreibersefaria: Or ask Cursor

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+      await expect(page.locator('body')).toHaveClass(new RegExp(`\\b${expectedClass}\\b`));
+    };
 
-    await context.addCookies(langCookies);
-    await page.reload()
-    return page
+/*LOGIN/LOGOUT RELATED METHODS */
 
-}
+    export const simulateLogout = async (context) => {
```
schreibersefaria: Based on one of your comments.
Shouldn't this be in a POM?
Maybe I miss understood you / the function.

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+    const languageLink = language === LANGUAGES.HE
+      ? page.locator('#select-hebrew-interface-link')
+      : page.locator('#select-english-interface-link');
+  
+      console.log('Selector being used:', language === LANGUAGES.HE ? '#select-hebrew-interface-link' : '#select-english-interface-link');
 
-    langCookies = await context.cookies();
+      await expect(languageLink).toBeVisible();
+      await languageLink.click();
+      await page.waitForTimeout(5000);
```
schreibersefaria: Also for the waits, maybe use cursor to locate and remove them all.

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
-export const loginUser = async (page: Page, user=testUser, language=DEFAULT_LANGUAGE) => {
-    await page.goto('/login');
+      if (sessionCookie) {
+        savedSessionCookie = sessionCookie;
+        const otherCookies = cookies.filter(c => c.name !== 'sessionid');
+        await context.clearCookies();
+        await context.addCookies(otherCookies);
+        return true;
+      } else {
+        console.warn('No session cookie found - user may already be logged out');
```
schreibersefaria: Warns are also logs that should be removed.
Unless we know they are something that are picked up in the console output / fail a test

e2e-tests/utils.ts
schreibersefaria: Based on a bit of reading (Claude), we would generally want to simulate user behaviour.
So instead of simulatingLogout, we would want to call a logout that is in one of the POMs

e2e-tests/utils.ts
```diff
@@ -15,81 +18,213 @@ const hideModals = async (page: Page) => {
...
+    
+    export const loginViaTooltip = async (page: Page, language = LANGUAGES.EN) => {
+      page.once('dialog', async dialog => {
+        console.log(`Dialog message: ${dialog.message()}`);
+        await dialog.accept();
+      });
+      await page.getByRole('link', { name: 'Log in' }).click();
+      const loginPage = new LoginPage(page, language);
+      await loginPage.loginAs(testUser.email ?? '', testUser.password ?? '');
+    };
```
schreibersefaria: Also for these two functions.
Based on what you said you read, should they not be in the relevant POM?

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    async setupWithContent(content: string) {
+      await this.addText(content);
+      await this.waitForAutosave();
+      return this;
+    }
+
+      //Connectivity/Login Functions--------------------------------------------
+      async waitForConnectionState(expectedState: 'online' | 'offline') {
```
schreibersefaria, line 192: There seems to be an indentation change here

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,279 @@
...
+          if (visible && z >= targetZ) {
+            console.warn('Overlapping element with higher or equal z-index:', el);
+            return false;
+          }
+        }
+    
+        return true;
+      });
+    
+      expect(isOnTop).toBe(true);
```
schreibersefaria, line 301: This still seems complex.
Maybe it's necessary.

What are we actually trying to achieve.
If it's checking the component is clickable we can try see if we can click it.
If it's checking visibility, there is likely a playwright function that could check this simply.

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,279 @@
...
+    };
+    
+      
+      async validateEditingIsBlocked() {
```
schreibersefaria, line 305: This seems like it should be a test, not part of the POM

e2e-tests/tests/autosave.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+
+
+test.describe('Test Saved/Saving Without Pop-ups: English', () => {
+  let page: Page;
+  let pageManager: PageManager;
+  
+  test.beforeEach(async ({ context }) => {
+    page = await goToPageWithLang(context, '/login');
+    const loginPage = new LoginPage(page, LANGUAGES.EN);
+    await loginPage.loginAs(testUser.email ?? '', testUser.password ?? '');
```
schreibersefaria: Is setting the email and password not done in the loginPage?
Maybe it should be so the tests don't need to "know" these details

e2e-tests/tests/autosave.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+import { LANGUAGES, testUser } from '../globals';
+import { SaveStates } from '../constants';
+
+
+test.describe('Test Saved/Saving Without Pop-ups: English', () => {
+  let page: Page;
+  let pageManager: PageManager;
+  
+  test.beforeEach(async ({ context }) => {
+    page = await goToPageWithLang(context, '/login');
```
schreibersefaria: Shouldn't the loginPage deal with this?

e2e-tests/tests/autosave.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+    
+    await editor.addText('Trigger unknown error');
+    await page.evaluate(() => {
+      (window as any).Sefaria.testUnkownNewEditorSaveError = true;
+    });
+    await page.waitForTimeout(20000);
+    await editor.assertSaveState(SaveStates.fifthState, LANGUAGES.HE);
+    await editor.validateEditingIsBlocked();
+  });
+});
```
schreibersefaria, line 202: The tests look very tidy 👍

e2e-tests/pages/loginPage.ts
jessicafeinmesser: ah thought you were referring to the deleted method on top of it. You are right, it is not used. I think it is safe to delete.

e2e-tests/pages/sourceSheetEditor.page.ts
jessicafeinmesser: I looked deeper into it, it is basically checking that no matter where you click on the element, nothing is obscuring it and it is in fact on top. Supposedly just checking that it has a higih z index is not enough.

another way to possibly go about this is to compare its z index to all the other elements on the page, like this:

const isHighestZIndex = await page.evaluate(() => {
  const target = document.querySelector('.editorSaveStateIndicator');
  if (!target) return false;
  const targetZ = parseInt(getComputedStyle(target).zIndex || '0', 10);
  let highestZ = targetZ;
  let isHighest = true;
  // Check all elements in the document
  document.querySelectorAll('*').forEach(el => {
    if (el === target) return;
    const style = window.getComputedStyle(el);
    const z = parseInt(style.zIndex || '0', 10);
    const visible = style.display !== 'none' && style.visibility !== 'hidden' && style.opacity !== '0';
    if (visible && z > highestZ) {
      isHighest = false;
    }
  });
  return isHighest;
});
expect(isHighestZIndex).toBe(true);

e2e-tests/pages/sourceSheetEditor.page.ts
jessicafeinmesser: I made a function out of it because it is reused many times in the test file and I thought it would make the tests cleaner. if you think it is better practice to have the content in the tests directly, we can do that

e2e-tests/pages/sourceSheetEditor.page.ts
schreibersefaria: If you looked into it and there isn't built in functions then let's leave it.
Just comment it clearly (the why)

e2e-tests/pages/sourceSheetEditor.page.ts
schreibersefaria: That is a good point.
Consider moving the function to the test page instead of the POM.
But either way, feel free to resolve the comment.

e2e-tests/tests/autosave.spec.ts
jessicafeinmesser: yes, makes sense. I have adapted gotoPageWithUser to call the loginPage login function, and used it here instead of goToPageWithLang

e2e-tests/tests/autosave.spec.ts
jessicafeinmesser: fixed

e2e-tests/utils.ts
jessicafeinmesser: kept the injected css styles for the first bit, removed body{} as I was told it is safe to do so, but I don't have the exact html code for the popup handy so a bit harder to find alternative solutions at the moment

e2e-tests/utils.ts
jessicafeinmesser: could be, good point. brandon made these logout/login simulations that work by deleting/adding back the sessionid cookie, so I guess the question is which POM should it be tied to? It doesn't need a login page in order to work, it's more to remove authentication, like in the case that the user is idle for too long and would get logged out but still be on the sheet. So it can technically run on any page?

e2e-tests/tests/autosave.spec.ts
schreibersefaria: That looks really nice!

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -0,0 +1,279 @@
...
+          console.warn('Overlapping element with higher or equal z-index:', el);
+          return false;
+        }
+      }
+  
+      return true;
+    });
+  
+    expect(isOnTop).toBe(true);
+  };
```
schreibersefaria: There is an aligning problem with this closing bracket

e2e-tests/tests/autosave.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  test('Text box aligned with save indicator is clickable', async () => {
+    const editor = pageManager.onSourceSheetEditorPage();
+    const uniqueText = 'ClickTarget123';
+    await editor.addText(uniqueText);
+    await editor.alignTextWithStatusIndicator(uniqueText);
+    const textLocator = page.locator('span[data-slate-string="true"]', { hasText: uniqueText });
+    await textLocator.dblclick();
+    await page.keyboard.type('✅');
```
schreibersefaria: I wouldn't use an emoji.
even if only so it's not obvious you used AI (It's currently a tell)

e2e-tests/tests/autosave.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  
+  test('Show unknown error state and block editing', async () => {
+    const editor = pageManager.onSourceSheetEditorPage();
+    await editor.addText('Before unknown error');
+    
+    // Trigger catch-all error
+    await page.evaluate(() => {
+      (window as any).Sefaria.testUnkownNewEditorSaveError = true;
+    });
+    await page.waitForTimeout(20000);
```
schreibersefaria: If there is a good reason for this timeout you should comment the why to avoid future confusion.

e2e-tests/utils.ts
```diff
@@ -1,10 +1,15 @@
 import {DEFAULT_LANGUAGE, LANGUAGES, SOURCE_LANGUAGES, testUser} from './globals'
 import {BrowserContext}  from '@playwright/test';
 import type { Page } from '@playwright/test';
+import { expect, Locator } from '@playwright/test';
+import { LoginPage } from './pages/loginPage';
 
 let currentLocation: string = ''; 
+let savedSessionCookie = null;
```
schreibersefaria: I thought I removed the dance around savedSessionCookie.
Did I miss something?
If we can't just remove, let's discuss

e2e-tests/utils.ts
schreibersefaria: Sounds good for now. Just comment in detail so it will be easy to improve later.
People is Sefaria often quote https://en.wiktionary.org/wiki/Chesterton%27s_fence 

Noun
[Chesterton](https://en.wiktionary.org/wiki/Chesterton#English)['s](https://en.wiktionary.org/wiki/-%27s#English) [fence](https://en.wiktionary.org/wiki/fence#English)

(public policy) The [principle](https://en.wiktionary.org/wiki/principle#English) that [reforms](https://en.wiktionary.org/wiki/reform#English) should not be made until the reasoning behind the [existing](https://en.wiktionary.org/wiki/existing#English) [state of affairs](https://en.wiktionary.org/wiki/state_of_affairs#English) is [understood](https://en.wiktionary.org/wiki/understood#English).

schreibersefaria: Remember to comment all the changes in this style

e2e-tests/utils.ts
```diff
@@ -22,6 +102,7 @@ const hideModals = async (page: Page) => {
  * @param language - Target language (LANGUAGES.EN or LANGUAGES.HE)
  */
 export const changeLanguageIfNeeded = async (page: Page, language: string) => {
+    await page.waitForLoadState('domcontentloaded'); // or 'networkidle' if you want to be extra sure
```
schreibersefaria: Remove the AI comment

e2e-tests/utils.ts
schreibersefaria: Sound reason
Maybe just add clear commenting.
Mention it shouldn't be used for normal logging in and out tests

schreibersefaria: Your answer answers this as well

e2e-tests/utils.ts
```diff
@@ -30,53 +111,102 @@ export const changeLanguageIfNeeded = async (page: Page, language: string) => {
...
+      } else {
+        return false;
+      }
+    };    
 
-    await changeLanguageIfNeeded(page, language);
+    export const simulateLogin = async (context) => {
+      if (savedSessionCookie) {
+        await context.addCookies([savedSessionCookie]);
+      }
```
schreibersefaria, line 197: Wouldn't you want an else here?

e2e-tests/utils.ts
```diff
@@ -30,53 +111,102 @@ export const changeLanguageIfNeeded = async (page: Page, language: string) => {
...
 
-export const loginUser = async (page: Page, user=testUser, language=DEFAULT_LANGUAGE) => {
-    if (!user.email || !user.password) {
-        throw new Error('Missing login credentials. Please set PLAYWRIGHT_USER_EMAIL and PLAYWRIGHT_USER_PASSWORD in your .env file at the project root.');
-    }
-    await page.goto('/login');
+      if (sessionCookie) {
+        savedSessionCookie = sessionCookie;
+        const otherCookies = cookies.filter(c => c.name !== 'sessionid');
+        await context.clearCookies();
```
schreibersefaria: Is there no way to clear just the one cookie?

e2e-tests/utils.ts
schreibersefaria: I guess this is for simulating login which makes sense.
Maybe double check there isn't a "right" place to save stuff like this.

package-lock.json
schreibersefaria, line 1: I see there are a lot of changes here that aren't related to playwright, maybe you should not commit this.

schreibersefaria (commented):
This is looking great!
Added only a few comments. Mainly requesting more commenting on the code.
And some comments are still left over

e2e-tests/utils.ts
jessicafeinmesser: maybe it came back in a merge or revert, I didn't purposely add it back

e2e-tests/utils.ts
jessicafeinmesser: I will add a comment. It is used for logout in nearly all of the autosave tests, but only once for login- the other logins are manual through tooltip/banner

e2e-tests/tests/autosave.spec.ts
jessicafeinmesser: changed the format to have a built in timeout in the assertStateState function so there is the opportunity to end the timeout earlier

e2e-tests/utils.ts
jessicafeinmesser: I have implemented another approach to "expire" the session cookie, still a little clunky but I am not sure how much more concise we can make it

e2e-tests/utils.ts
jessicafeinmesser: added an else that catches an error in case the cookie does not exist

e2e-tests/utils.ts
jessicafeinmesser: good point- moved tooltip one to sheet editor POM, kept navbar one because that can be used by most pages

package-lock.json
jessicafeinmesser: must have been from the merge/pull, sorry!

schreibersefaria: No no, this is typical of the package-lock.json
I'm not sure when you would want to commit and when you wouldn't.
May be worth asking Claude

e2e-tests/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+      tooltip: /You are not logged in to Sefaria/i,
+      tooltipHebrew: /כעת אינך מחובר\/ת לספריא/i,
+    },
+    tryingToConnect: {
+      text: /Trying to connect/i,
+      textHebrew: /ניסיון התחברות/i,
+      tooltip: /No internet connection detected/i,
+      tooltipHebrew: /לא זוהה חיבור לאינטרנט/i,
+    },
+    fifthState: {
```
schreibersefaria: I think this could benefit from a more indicative name

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -17,16 +17,17 @@ export class SourceSheetEditorPage extends HelperBase {
     statusTooltip = () => this.page.locator('.editorSaveStateIndicator [data-tooltip]');
 
 
-    async assertSaveState(state: SaveState, language = LANGUAGES.EN) {
+    async assertSaveState(state: SaveState, language = LANGUAGES.EN, timeout = 5000) {
```
schreibersefaria: If there is a specific reason for this let's comment it.
Both the param and why we need a timeout in the test.

Although, in this case, maybe a flag would be more suitable - something like longTimeoutToCatchErrorStates.
The logic will be that the tests don't have to handle this. If the timeout changes it can be changed in one place.

schreibersefaria: Changing state - saveState may avoid confusion (It currently looks like the param is managing the pages state of something like that)

e2e-tests/pages/sourceSheetEditor.page.ts
```diff
@@ -52,7 +57,7 @@ export class SourceSheetEditorPage extends HelperBase {
     
     async addText(text: string) {
         await this.focusTextInput(); 
-        await this.page.keyboard.type(text);
+        await this.page.keyboard.type(text, {delay: 100});
```
schreibersefaria: Here too. Let's add why comments on every time delay

e2e-tests/tests/autosave.spec.ts
```diff
@@ -115,23 +113,71 @@ test.describe('Test Saved/Saving Without Pop-ups: English', () => {
     await expect(editor.sourceSheetBody()).toBeEnabled();
   });
 
-  test('Show beforeunload prompt when leaving with unsaved changes', async () => {
+  // test('Show beforeunload prompt when leaving with unsaved changes', async () => {
```
schreibersefaria: Let's not leave commented out code without a clear reason.
If we need it, it's in the Git history

e2e-tests/pages/sourceSheetEditor.page.ts
schreibersefaria: I see you commented it when you call this function which is good. Let's comment here too
Or give the param an indicative name instead of a comment

e2e-tests/utils.ts
```diff
@@ -9,18 +9,26 @@ let savedSessionCookie = null;
 
 /*METHODS TO HIDE MODALS/POPUPS THAT INTERRUPT THE USER EXPERIENCE */
 
+/**Note, for all of these miding/dismiss methods, we currently use CSS to hide them
+ * We may want to opt for a more robust solution in the future, or something user-realistic such as 
+ * clicking an "x" or "okay" button,but this is a workaround for now.
+*/
+
+// Dismisses the main modal interrupting message by injecting CSS to hide it.
 export const hideModals = async (page: Page) => {
```
schreibersefaria, line 39: I don't think we need export here 
If I see correctly, this is used in this document and in an import statement

e2e-tests/utils.ts
```diff
@@ -146,45 +151,49 @@ export const changeLanguageLoggedIn = async (page: Page, language: string) => {
...
-        savedSessionCookie = sessionCookie;
-        const otherCookies = cookies.filter(c => c.name !== 'sessionid');
-        await context.clearCookies();
-        await context.addCookies(otherCookies);
-        return true;
-      } else {
-        return false;
+//located in utils rather than loginPage because it is used in multiple places;
+//it involves removing authentication (cookies) rather than logging out    
+export const simulateLogout = async (context) => {
```
schreibersefaria: The commenting here is good. Consider using a JSDoc string instead.
It is then used by the IDE for previews of the functions functionallity

e2e-tests/utils.ts
```diff
@@ -146,45 +151,49 @@ export const changeLanguageLoggedIn = async (page: Page, language: string) => {
...
+  const cookies = await context.cookies();
+  const sessionCookie = cookies.find(c => c.name === 'sessionid');
+  if (sessionCookie) {
+    savedSessionCookie = sessionCookie;
+    // Overwrite the sessionid cookie with an expired one to remove it
+    await context.addCookies([
```
schreibersefaria, line 187: I like this logic much more then what we had before

e2e-tests/utils.ts
```diff
@@ -146,45 +151,49 @@ export const changeLanguageLoggedIn = async (page: Page, language: string) => {
...
-      await page.getByRole('link', { name: 'Log in' }).click();
-      const loginPage = new LoginPage(page, language);
-      await loginPage.loginAs(testUser);
-    };
+export const simulateLogin = async (context) => {
+  if (savedSessionCookie) {
+    // Restore the saved sessionid cookie
+    await context.addCookies([savedSessionCookie]);
+  } else {
+    throw new Error('No saved session cookie.');
```
schreibersefaria: A general comment - not specific for here.
I think I was mistaken in telling you to remove all the logs, sorry.
It's a comment I got in production code but I think a small amount of well placed logs make sense in tests.
Sorry, I'm learning with you...

schreibersefaria: Maybe have a longer error message. Explaining the tests are failing not because this test failed but because of the implementation strategy.

schreibersefaria: Is this not going to cause contamination in between tests.
What if we have tests with admin / non admin users. Or just generally with different users (say if we test sheet functionality for the owner and non owner user)

e2e-tests/utils.ts
```diff
@@ -146,45 +151,49 @@ export const changeLanguageLoggedIn = async (page: Page, language: string) => {
...
+  }
+};
 
-      
-  
+//keeping this here because it can be used across the site
```
schreibersefaria: Isn't banner a POM for the nav bar?
If so, can we move this there?

playwright.config.ts
```diff
@@ -4,8 +4,7 @@ import { defineConfig, devices } from '@playwright/test';
  * Read environment variables from file.
  * https://github.com/motdotla/dotenv
  */
-// require('dotenv').config();
-
+ require('dotenv').config();
```
schreibersefaria: Are you sure we need this. It was working for me without this (you just needed to restart the Terminal on any change to .env)

schreibersefaria (commented):
This is doing very well. I think we are almost there.
I mostly have some small comments, there is maybe one or two that are not simple changes.
Note, I think I still have some unresolved comments from the last CR.

e2e-tests/constants.ts
jessicafeinmesser: changed to catchAllFifthState

e2e-tests/pages/sourceSheetEditor.page.ts
jessicafeinmesser: **add a const for this instead

e2e-tests/utils.ts
jessicafeinmesser: suggestion: try starting with if (!savedSessionCookie) and having logic to add session cookie just in case this is called before simulateLogout

schreibersefaria: savedSessionCookie is a bit of a confusing name

e2e-tests/pages/sourceSheetEditor.page.ts
jessicafeinmesser: made it a bit more concise but left same logic overall

playwright.config.ts
jessicafeinmesser: didn't work for me without this, page url/context wouldn't load

e2e-tests/utils.ts
jessicafeinmesser: deleted function, doesn't represent user actions very well

PR #2535, opened 2025-07-01
Bug/sc 34891/can view reader while in sheets module

stevekaplan123:
## Description
Topic pages have links to the library module.  Given that we don't have a perfect subdomain system yet, we should simply open these pages in a new tab by setting target to blank.

### DIFF
--- a/static/js/TopicPage.jsx
+++ b/static/js/TopicPage.jsx
@@ -1063,7 +1063,7 @@ const ReadingsComponent = ({ parashaData, tref }) => (
                let sectionStr = sectionNum <= 7 ? sectionNum : 'M';
                let heSectionStr = sectionNum <= 7 ? Sefaria.hebrew.encodeHebrewNumeral(sectionNum) : 'מ';
                return (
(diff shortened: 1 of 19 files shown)


static/js/TopicPage.jsx
```diff
@@ -423,24 +422,24 @@ return (
             </div>
        : null}
        {tpTopImg}
-       {topicData && topicData.ref ?
+       {topicData && topicData.ref &&
            <div>
-               <a href={`/${topicData.ref.url}`} className="resourcesLink button blue">
+               <a href={`/${topicData.ref.url}`} target="_blank" rel="noopener noreferrer" className="resourcesLink button blue">
```
YishaiGlasner: why `rel` is only here?

yitzhakc: @stevekaplan123 what is the purpose of it?

stevekaplan123: I am removing it.  I misunderstood something I read from ChatGPT.  It does improve security but it's for out-bound links not links to your own site.

YishaiGlasner (approved):

yitzhakc (approved):

PR #2537, opened 2025-07-02
In the sheet editor, open suggestion and preview above when there is no room below.

YishaiGlasner:
## Description
In the sheet edtior, when adding source by ref there is a popup for suggestion, and another one for preview. These popups are open below also when there is no room for them. Now, they will open above where there is no room below (and there is room above).

## Code Changes
In `Autocompleter` component: 

- The suggestion element and preview element wrapped by a `div` with `ref`. This div can get classname `show-above`.
- New function `checkSpaceBelow` added. It checks if there is no space below and there is space above and set `shouldShowAbove`. This function measures the height of the popup by its children, for the children has `position: absolute`. It's wrapped by `useCallback`, because it's not changing in the lifecycle of the component (it's depends on react's refs), and is used in listeners, so for cleaning the listener we must have the function that the listener was created with).
- This function is used by `useEffect` when suggestions or preview are changed, and also by listeners to resize and toggle (that added by another `useEffect`).

In `s2.css`: rules for `show-above`.

## Notes
This component is also used in other places for admins (adding topics in side bar and editting topics). Since it's only for admnis, and it seems that the setting there is different, this PR don't change these places. Some css was added so that these places won't break.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -11069,10 +11069,22 @@ background-color: white;
   position: relative;
 }
 
+.connectionsPanel .autocompleterPopups {
+  position: absolute;
+  height: 100%;
+  width: 100%;
+}
+
 .addInterfaceInput .suggestionBoxContainer,
 .addInterfaceInput .textPreviewContainer {
   position: absolute;
   z-index: 1;
+}
+
+.editorContent .show-above .suggestionBoxContainer,
+.editorContent .show-above .autocompleterPopups {
+  top: auto;
+  bottom: 100%;
 }
 
 .addInterfaceInput .suggestionBox,
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -2903,7 +2903,6 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
 
 
   const checkSpaceBelow = useCallback(() => {
-    console.log(0, popsEl)
     const input = inputEl.current;
     const pops = popsEl.current;
     if (!input || !pops) return;
@@ -2913,8 +2912,6 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
     const popsTop = Math.min(...Array.from(pops.children).map(c => c.getBoundingClientRect().top))
     const popsBottom = Math.max(...Array.from(pops.children).map(c => c.getBoundingClientRect().bottom))
     const popsHeight = popsBottom - popsTop;
-    console.log(10, popsTop, popsBottom)
-    console.log(11, spaceBelow , popsHeight , spaceAbove )
     setShouldShowAbove(spaceBelow < popsHeight && spaceAbove >= popsHeight);
   }, []);
   useEffect(() => {
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -1,5 +1,5 @@
 //const React      = require('react');
-import React, {useContext, useEffect, useRef, useState} from 'react';
+import React, {useContext, useEffect, useRef, useState, useCallback} from 'react';
 import ReactDOM from 'react-dom';
 import $ from './sefaria/sefariaJquery';
 import {CollectionsModal} from "./CollectionsWidget";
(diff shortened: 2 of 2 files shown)


stevekaplan123:
I've never seen useCallback before and don't fully get how it works.  But it seems like it's necessary for event listeners to be memoized to prevent memory leaks.  Is this the reason you are using it?  If so, shouldn't we do this throughout our code base for event listeners or is it not always necessary?

YishaiGlasner:
> I've never seen useCallback before and don't fully get how it works. But it seems like it's necessary for event listeners to be memoized to prevent memory leaks. Is this the reason you are using it? If so, shouldn't we do this throughout our code base for event listeners or is it not always necessary? @YishaiGlasner

@stevekaplan123  It was a mistake of mine. In this case `useCallback` has no much influence, but it goes with the logic of React, that says constant function should not be rebuilt in any re-rendering.

stevekaplan123:
> > I've never seen useCallback before and don't fully get how it works. But it seems like it's necessary for event listeners to be memoized to prevent memory leaks. Is this the reason you are using it? If so, shouldn't we do this throughout our code base for event listeners or is it not always necessary? @YishaiGlasner
> 
> @stevekaplan123 It was a mistake of mine. In this case `useCallback` has no much influence, but it goes with the logic of React, that says constant function should not be rebuilt in any re-rendering.

I'm confused.  If it's a mistake, then why keep it?  And if it follows React logic, then why is it a mistake?

YishaiGlasner:
> > > I've never seen useCallback before and don't fully get how it works. But it seems like it's necessary for event listeners to be memoized to prevent memory leaks. Is this the reason you are using it? If so, shouldn't we do this throughout our code base for event listeners or is it not always necessary? @YishaiGlasner
> > 
> > 
> > @stevekaplan123 It was a mistake of mine. In this case `useCallback` has no much influence, but it goes with the logic of React, that says constant function should not be rebuilt in any re-rendering.
> 
> I'm confused. If it's a mistake, then why keep it? And if it follows React logic, then why is it a mistake?

this does fit React logic, it's recommended, but i was mistaken about the option that the rempveLostener may not work

stevekaplan123 (dismissed):

static/js/Misc.jsx
```diff
@@ -2888,10 +2888,44 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
...
+         if (element) {element.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'start' })}
+    }, [previewText]
+  )
+
+
+  const checkSpaceBelow = useCallback(() => {
+    const input = inputEl.current;
+    const pops = popsEl.current;
+    if (!input || !pops) return;
+    const inputRect = input.getBoundingClientRect();
```
yitzhakc, line 2903: I think that adding a line break here and possibly a comment "calculate the visible space" would make it more readable :-)

static/js/Misc.jsx
```diff
@@ -2888,10 +2888,44 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
   const [showAddButton, setShowAddButton] = useState(false);
   const [showCurrentSuggestions, setShowCurrentSuggestions] = useState(true);
   const [inputClassNames, setInputClassNames] = useState(classNames({selected: 0}));
+  const [shouldShowAbove, setShouldShowAbove] = useState(false);
   const suggestionEl = useRef(null);
   const inputEl = useRef(null);
+  const popsEl = useRef(null);
```
yitzhakc: Maybe change the name to `popupEl`? It's not so much longer, and it provides more meaning, vs pops 😅

static/js/Misc.jsx
```diff
@@ -2888,10 +2888,44 @@ const Autocompleter = ({getSuggestions, showSuggestionsOnSelect, inputPlaceholde
...
   const buttonClassNames = classNames({button: 1, small: 1});
 
+  useEffect(
```
yitzhakc: Please group these hooks with the other calls of `useEffect` near the top of the component.

static/css/s2.css
```diff
@@ -11069,12 +11069,24 @@ background-color: white;
...
+}
+
 .addInterfaceInput .suggestionBoxContainer,
 .addInterfaceInput .textPreviewContainer {
   position: absolute;
   z-index: 1;
 }
 
+.editorContent .show-above .suggestionBoxContainer,
+.editorContent .show-above .autocompleterPopups {
```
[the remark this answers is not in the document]
yitzhakc: @YishaiGlasner Based on line 3090, should the second CSS selector be `.show-above.autocompleterPopups` since both classes are on the same `div` element?

stevekaplan123:
> > > > I've never seen useCallback before and don't fully get how it works. But it seems like it's necessary for event listeners to be memoized to prevent memory leaks. Is this the reason you are using it? If so, shouldn't we do this throughout our code base for event listeners or is it not always necessary? @YishaiGlasner
> > > 
> > > 
> > > @stevekaplan123 It was a mistake of mine. In this case `useCallback` has no much influence, but it goes with the logic of React, that says constant function should not be rebuilt in any re-rendering.
> > 
> > 
> > I'm confused. If it's a mistake, then why keep it? And if it follows React logic, then why is it a mistake?
> 
> this does fit React logic, it's recommended, but i was mistaken about the option that the rempveLostener may not work

I don't understand still.  What I read is that davka using useCallback does help with the removeListener working.  What other benefit does it bring?

yitzhakc (approved):
Looks good to me, waiting on QA to merge

PR #2538, opened 2025-07-02
Playwright Infra Baseline

schreibersefaria:
Got all the tests to run.

Removed a lot of assertion logic or added it when needed.
Removed the saving of cookies that seemed executed in a complex way

### DIFF
--- a/e2e-tests/utils.ts
+++ b/e2e-tests/utils.ts
@@ -4,6 +4,7 @@ import type { Page } from 'playwright-core';
 
 let langCookies: any = [];
 let loginCookies: any = [];
+let currentLocation: string = ''; 
 
 const hideModals = async (page: Page) => {
     await page.waitForLoadState('networkidle');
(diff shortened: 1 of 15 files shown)


e2e-tests/utils.ts
```diff
@@ -65,17 +69,15 @@ export const loginUser = async (page: Page, user=testUser, language=DEFAULT_LANG
 
 
 export const goToPageWithUser = async (context: BrowserContext, url: string, user=testUser) => {
-    if (!loginCookies.length) {
```
jessicafeinmesser, line 68: what purpose did his cookie "hacks" serve before exactly? Does deleting them not change anything? For here and the other goto

schreibersefaria: Yes they are very important.
If I understood correctly, they were saved to save time on login and language change. 
But they didn't seem to work and we should find a better solution.

PR #2540, opened 2025-07-06
Support spotify in sheets.

YishaiGlasner:
## Description
We're supporting spotify in the editor but not in the viewer. This PR fixes this issue.

## Code Changes
- Add spotify option to `Sheet.jsx`.
- Use parsed url rather than raw url in the if statement. This is much clean than the regex match we're using for other options.

### DIFF
--- a/static/js/Sheet.jsx
+++ b/static/js/Sheet.jsx
@@ -607,6 +607,10 @@ class SheetMedia extends Component {
     var mediaClass = "media fullWidth";
     var mediaURL = this.props.source.media;
     var caption  = this.props.source.caption;
+    let parsedUrl
+    if (mediaURL) {
+      parsedUrl = new URL(mediaURL);
+    }
 
     if (this.isImage()) {
       mediaLink = '<img class="addedMedia" src="' + mediaURL + '" />';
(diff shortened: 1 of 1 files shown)


static/js/Sheet.jsx
```diff
@@ -623,6 +627,19 @@ class SheetMedia extends Component {
       mediaLink = '<iframe width="100%" height="166" scrolling="no" frameborder="no" src="' + mediaURL + '"></iframe>';
     }
 
+    else if (parsedUrl.hostname.includes("spotify.com")) {
+      const [,, type] = parsedUrl.pathname.split("/");
```
schreibersefaria, line 631: I don't know how the spotify url works but seems seems prone to future change errors.
If that is the case, we may want to add some structure validation.

static/js/Sheet.jsx
```diff
@@ -623,6 +627,19 @@ class SheetMedia extends Component {
...
+      const height = type === "episode" ? 152 : 80;
```
schreibersefaria: I would love a comment explaining magic numbers like these.
Although I think other people don't like comments as much as I do, so what ever you think

schreibersefaria (approved):
Overall looks straight forward.
I left comments but because I don't know the feature I already approved in case they aren't relevant.

static/js/Sheet.jsx
YishaiGlasner: we're doing it in the editor, when saving them to our data. i think this is the best practice.

static/js/Sheet.jsx
YishaiGlasner: magic strings are bad.
descriptive names are better than notes, but here i think you're right we need also notes.

schreibersefaria (approved):

PR #2543, opened 2025-07-07
Bug/sc 34949/when i try to delete a note in the hebrew

schreibersefaria:
There was an erroneous alert after trying to delete a note in addition and before the alert we want.
I simply removed it.  
Also added to gitignore some local development files (Used for cursor + palywright MCP + .env)

Testing:
I checked the error didn't pop up anymore but that the note was deleted

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,10 @@ node_modules
 /.vscode
 *.code-workspace
 
+# Cursor #
+##########
+.cursor/
+
 # pystest cache #
 #####################
 /.pytest_cache
@@ -141,7 +145,3 @@ dump/*
 /playwright/.cache/
 /playwright-profile/
 .env
-
-# Cursor #
-##########
-.cursor/
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,9 @@ dump/*
 /test-results/
 /playwright-report/
 /playwright/.cache/
+/playwright-profile/
+.env
+
+# Cursor #
+##########
+.cursor/
--- a/static/js/ConnectionsPanel.jsx
+++ b/static/js/ConnectionsPanel.jsx
@@ -1518,7 +1518,6 @@ class AddNoteBox extends Component {
     this.setState({ isPrivate: false });
   }
   deleteNote() {
-    alert(Sefaria._("Something went wrong (that's all I know)."));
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -139,3 +139,9 @@ dump/*
 /test-results/
 /playwright-report/
 /playwright/.cache/
+/playwright-profile/
+.env
+
+# Cursor #
+##########
+.cursor/
```
YishaiGlasner: i think `cursor` should be in your gitignore_global.
i guess playwright-profile is something everyone need for playwright test, so it shoud be ok, but u think it would be nice to take all the things for playwright (incliding `test-results`? i don't know what it's) and give them their secion with title, rather than Miscellaneous
regarding, `env` idk what is it

static/js/ConnectionsPanel.jsx
```diff
@@ -1518,7 +1518,6 @@ class AddNoteBox extends Component {
     this.setState({ isPrivate: false });
   }
   deleteNote() {
-    alert(Sefaria._("Something went wrong (that's all I know)."));
```
YishaiGlasner, line 1521: can you check in the git history what's going on here?
i guess there was a reason for adding this weird alert.

.gitignore
schreibersefaria: No, `playwright-profile` isn't something everyone need. It for connecting AI to playwright capabilities.

Some of my projects include commiting .cursor. We have a "Visual Studio Code directory", so I thought this also makes sense

`.env` is a file used for holding environment vars. At the moment it can be used for running Playwright tests - it holds the Cauldron and username and password

static/js/ConnectionsPanel.jsx
schreibersefaria: The error message used to be inside an error handler (back when it was a direct ajax call)
In this commit it looks like it was just a mistake
https://github.com/Sefaria/Sefaria-Project/commit/76a8457875a70a8b1a843c652674d1136f429fc2#diff-5d4838cffd5fc856d5b7379228ac1e5575881272fbde73dd0613ab623e3da0a3
image

.gitignore
YishaiGlasner: i think we have no official policy for gitignore, so i'm not sure what we should do, but i think things like private settings of cursor should be private (= gitignore_global).

PR #2549, opened 2025-07-09
Chore/sc 31762/api auth

YishaiGlasner:
## Description
We are using in many of our views for both site's users and API. Threefore in many views we have logic for checking if an api request is authenticated. This PR moves this logic to a middleware and refactor all the views accordingly.

## Code Changes
- `system.middleware.py` - if user is not authenticated but have apikey, it add its user to the `request.user`. Also it add the attribute `is_api_authenticated` to all requests.
- `settings` - adds the middlware to `MIDDLEWARE`
- `system.tests.middleware.py`
- `sefaria.client.util.py` - new `authenticationRequiredResponse` and `notStaffOrApiResponse` for use in views, rather than variety of different respnses.
- `reader.views.py` `sefaria.views.py` `sourcesheets.views.py` - refactor of authentication handling when now everything is done by `request.user` and `request.is_api_authenticated` and using the new response objects. This is a hard to read diff, I've checked it twice, but i think it's important that someone else will check it.
- `scripts/move_draft_text.py` `scripts/post_sheet.py` `scripts/scheduled/reindex_elasticsearch_cronjob.py` - scripts sending api request with apikey - refactored to send it in the headers.

## Notes
Since I've touched  `move_draft_text`, I've also removed the slack message at its end - it osn't working, throws error and unnecessary because the real job is done on celery.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -2124,7 +2124,7 @@ def all_notes_api(request):
         else:
             res = [note.contents(with_string_id=True) for note in NoteSet({"owner": request.user.id}, sort=[("_id", -1)]) ]
     else:
-        resr = {"error": "Not implemented."}
+        res = {"error": "Not implemented."}
     return jsonResponse(res, callback=request.GET.get("callback", None))
 
 
--- a/reader/views.py
+++ b/reader/views.py
@@ -1525,14 +1525,10 @@ def complete_version_api(request):
     except BookNameError:
         return jsonResponse({"error": f"No index named: {title}"})
 
-    if not request.user.is_authenticated:
-        key = body_data.get('apikey')[0]
-        if not key:
-            return jsonResponse({"error": "You must be logged in or use an API key to save texts."})
-        apikey = db.apikeys.find_one({"key": key})
+    if not request.user.is_staff:
+        return notStaffOrApiResponse()
+    if request.is_api_authenticated:
         method = 'API'
-        if not apikey:
-            return jsonResponse({"error": "Unrecognized API key."})
     else:
         method = None
         internal_do_post = csrf_protect(internal_do_post())
--- a/reader/views.py
+++ b/reader/views.py
@@ -1525,8 +1525,8 @@ def complete_version_api(request):
     except BookNameError:
         return jsonResponse({"error": f"No index named: {title}"})
 
-    if not request.user.is_staff:
-        return notStaffOrApiResponse()
+    if not request.user.is_authenticated:
+        return authenticationRequiredResponse()
     if request.is_api_authenticated:
         method = 'API'
     else:
--- a/reader/views.py
+++ b/reader/views.py
@@ -2753,7 +2753,7 @@ def updates_api(request, gid=None):
                             })
 
     elif request.method == "POST":
-        if not not request.user.is_staff:
+        if not request.user.is_staff:
             return notStaffOrApiResponse()
         elif request.is_api_authenticated:
             payload = json.loads(request.POST.get("json"))
--- a/reader/views.py
+++ b/reader/views.py
@@ -2757,17 +2757,9 @@ def updates_api(request, gid=None):
                             })
 
     elif request.method == "POST":
-        if not request.user.is_authenticated:
-            key = request.POST.get("apikey")
-            if not key:
-                return jsonResponse({"error": "You must be logged in or use an API key to perform this action."})
-            apikey = db.apikeys.find_one({"key": key})
-            if not apikey:
-                return jsonResponse({"error": "Unrecognized API key."})
-            user = User.objects.get(id=apikey["uid"])
-            if not user.is_staff:
-                return jsonResponse({"error": "Only Sefaria Moderators can add announcements."})
-
+        if not not request.user.is_staff:
+            return notStaffOrApiResponse()
+        elif request.is_api_authenticated:
             payload = json.loads(request.POST.get("json"))
             try:
                 GlobalNotification(payload).save()
@@ -2775,7 +2767,7 @@ def updates_api(request, gid=None):
             except AssertionError as e:
                 return jsonResponse({"error": str(e)})
 
-        elif request.user.is_staff:
+        else:
             @csrf_protect
             def protected_post(request):
                 payload = json.loads(request.POST.get("json"))
@@ -2786,8 +2778,6 @@ def updates_api(request, gid=None):
                     return jsonResponse({"error": str(e)})
 
             return protected_post(request)
-        else:
-            return jsonResponse({"error": "Unauthorized"})
 
     elif request.method == "DELETE":
         if not gid:
--- a/reader/views.py
+++ b/reader/views.py
@@ -2390,7 +2390,7 @@ def category_api(request, path=None):
             return notStaffOrApiResponse()
         if request.is_api_authenticated:
             kwargs = {"method": "API"}
-        else:
+        else:  # a logged-in staff member
             kwargs = {}
             _internal_do_post = csrf_protect(_internal_do_post)
         uid = request.user.id
@@ -2546,7 +2546,7 @@ def terms_api(request, name):
             return notStaffOrApiResponse()
         if request.is_api_authenticated:
             kwargs = {"method": "API"}
-        else:
+        else:  # a logged-in staff member
             kwargs = {}
             _internal_do_post = csrf_protect(_internal_do_post)
         uid = request.user.id
@@ -2773,7 +2773,7 @@ def updates_api(request, gid=None):
             except AssertionError as e:
                 return jsonResponse({"error": str(e)})
 
-        else:
+        else:  # a logged-in staff member
             @csrf_protect
             def protected_post(request):
                 payload = json.loads(request.POST.get("json"))
--- a/reader/views.py
+++ b/reader/views.py
@@ -1692,6 +1692,8 @@ def index_api(request, title, raw=False):
         if request.user.is_staff:
             admin_post = csrf_protect(index_post)
             return admin_post(request, request.user.id, j, None, raw)
+        if not request.user.is_authenticated:
+            return authenticationRequiredResponse()
         return notStaffOrApiResponse()
 
     def index_post(request, uid, j, method, raw):
@@ -2308,6 +2310,8 @@ def flag_text_api(request, title, lang, version):
 
     _attributes_to_save = Version.optional_attrs + ["versionSource", "direction", "isSource", "isPrimary"]
 
+    if not request.user.is_authenticated:
+        return authenticationRequiredResponse()
     if not request.user.is_staff:
         return notStaffOrApiResponse()
     if request.is_api_authenticated:
@@ -2380,6 +2384,8 @@ def category_api(request, path=None):
             func = tracker.update if update else tracker.add
             return func(uid, Category, cat, **kwargs).contents()
 
+        if not request.user.is_authenticated:
+            return authenticationRequiredResponse()
         if not request.user.is_staff:
             return notStaffOrApiResponse()
         if request.is_api_authenticated:
@@ -2534,6 +2540,8 @@ def terms_api(request, name):
                     return {"error": 'Term "%s" does not exist.' % name}
                 return tracker.delete(uid, Term, t._id)
 
+        if not request.user.is_authenticated:
+            return authenticationRequiredResponse()
         if not request.user.is_staff:
             return notStaffOrApiResponse()
         if request.is_api_authenticated:
@@ -2753,6 +2761,8 @@ def updates_api(request, gid=None):
                             })
 
     elif request.method == "POST":
+        if not request.user.is_authenticated:
+            return authenticationRequiredResponse()
         if not request.user.is_staff:
             return notStaffOrApiResponse()
         elif request.is_api_authenticated:
@@ -3123,6 +3133,8 @@ def topics_api(request, topic, v2=False):
         response = get_topic(v2, topic, request.interfaceLang, with_html=with_html, with_links=with_links, annotate_links=annotate_links, with_refs=with_refs, group_related=group_related, annotate_time_period=annotate_time_period, ref_link_type_filters=ref_link_type_filters, with_indexes=with_indexes)
         return jsonResponse(response, callback=request.GET.get("callback", None))
     elif request.method == "POST":
+        if not request.user.is_authenticated:
+            return authenticationRequiredResponse()
         if not request.user.is_staff:
             return notStaffOrApiResponse()
         topic_data = json.loads(request.POST["json"])
--- a/reader/views.py
+++ b/reader/views.py
@@ -23,7 +23,7 @@ from rest_framework.decorators import api_view, permission_classes
 from rest_framework.permissions import IsAuthenticated
 from django.template.loader import render_to_string
 from django.shortcuts import render, redirect
-from django.http import Http404, QueryDict, HttpResponse
+from django.http import Http404, QueryDict
 from django.contrib.auth.decorators import login_required
 from django.contrib.admin.views.decorators import staff_member_required
 from django.utils.encoding import iri_to_uri
@@ -46,7 +46,7 @@ from sefaria.model.schema import SheetLibraryNode
 from sefaria.model.following import general_follow_recommendations
 from sefaria.model.trend import user_stats_data, site_stats_data
 from sefaria.client.wrapper import format_object_for_client, format_note_object_for_client, get_notes, get_links
-from sefaria.client.util import jsonResponse, celeryResponse
+from sefaria.client.util import jsonResponse, celeryResponse, authenticationRequiredResponse, notStaffOrApiResponse
 from sefaria.history import text_history, get_maximal_collapsed_activity, top_contributors, text_at_revision, record_version_deletion, record_index_deletion
 from sefaria.sefaria_tasks_interace.history_change import LinkChange, VersionChange
 from sefaria.sheets import get_sheets_for_ref, get_sheet_for_panel, annotate_user_links, trending_topics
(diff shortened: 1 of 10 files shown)


sefaria/system/middleware.py
```diff
@@ -218,3 +222,23 @@ def process_response(self, request, response):
...
+
+
+class ApiKeyAuthenticationMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        request.is_api_authenticated = False
+        if not request.user.is_authenticated:
+            apikey = request.META.get("HTTP_X_APIKEY") or request.POST.get("apikey") or request.GET.get("apikey")
```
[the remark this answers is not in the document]
YishaiGlasner: @yitzhakc what do you think about that? if have external clients uses the POST we can't remove it.
GPT suggests to do it with:
```
if request.method == "POST" and request.content_type.startswith("application/x-www-form-urlencoded"):
```

sefaria/system/middleware.py
```diff
@@ -218,3 +222,23 @@ def process_response(self, request, response):
...
+            if apikey:
+                try:
+                    apikey = db.apikeys.find_one({"key": apikey})
+                    user = User.objects.get(id=apikey['uid'])
+                    request.user = user
+                    request.is_api_authenticated = True
+                except (User.DoesNotExist, KeyError):
+                    pass
```
schreibersefaria: Why do we have a try with a pass?
If this isn't a mistake maybe comment the why

sefaria/system/middleware.py
```diff
@@ -218,3 +222,23 @@ def process_response(self, request, response):
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        request.is_api_authenticated = False
+        if not request.user.is_authenticated:
+            apikey = request.META.get("HTTP_X_APIKEY") or request.POST.get("apikey") or request.GET.get("apikey")
+            if apikey:
+                try:
+                    apikey = db.apikeys.find_one({"key": apikey})
```
schreibersefaria, line 250: How much will this slow us down in practice?
Is there in memory caching in DB? If not, it may be worth implementing for this

reader/views.py
```diff
@@ -2841,25 +2753,17 @@ def updates_api(request, gid=None):
...
-                return jsonResponse({"error": "You must be logged in or use an API key to perform this action."})
-            apikey = db.apikeys.find_one({"key": key})
-            if not apikey:
-                return jsonResponse({"error": "Unrecognized API key."})
-            user = User.objects.get(id=apikey["uid"])
-            if not user.is_staff:
-                return jsonResponse({"error": "Only Sefaria Moderators can add announcements."})
-
+        if not not request.user.is_staff:
+            return notStaffOrApiResponse()
```
schreibersefaria, line 2767: is the not not a mistake?

schreibersefaria (commented):
Still in the middle. Thought to leave these because I'm not sure I'll get to the rest until later

sefaria/system/middleware.py
YishaiGlasner: users should never have api key. for logged in users only the first 2 lines will run, and for anonymous 4 lines, but it won't get to mongo. 
only for api requests this will go to mongo. in some of them we'd anyway do this check, so we're dealing with GET (probably) api requests only. i don't think this is too exepnsive.
@yitzhakc what do you think?

reader/views.py
YishaiGlasner: great catch!

sefaria/system/middleware.py
schreibersefaria: It may be negligible.
But if the API becomes very popular (This may include MCP) this could become not insignificant

reader/views.py
schreibersefaria: I wish I could take credit

sefaria/system/middleware.py
YishaiGlasner: in another thought, why would someone attach an apikey to his request? most api users has no key, and if they they probably use it only when needed

schreibersefaria: Well I'm not actually sure when an API key is needed.
Feel free to resolve this comment 👍

sefaria/client/util.py
```diff
@@ -45,6 +45,17 @@ def celeryResponse(task_id: str, sub_task_ids: list[str] = None):
         data['sub_task_ids'] = sub_task_ids
     return jsonResponse(data, status=202)
 
+
+def authenticationRequiredResponse(data=None):
```
schreibersefaria: Python normally goes with snake_case.
Although I see this file is already mixed.

schreibersefaria: Also, ideally use type hints for new functions. I don't know what the official policy is but I see it in some functions.
`def authenticationRequiredResponse(data: Optional[dict] = None) -> JsonResponse:`

sefaria/views.py
```diff
@@ -1176,17 +1176,10 @@ def versions_csv(request):
...
-    from django.contrib.auth.models import User
 
-    key = request.POST.get("apikey")
-    if not key:
-        return jsonResponse({"error": "You must be logged in or use an API key to index sheets by timestamp."})
-    apikey = db.apikeys.find_one({"key": key})
-    if not apikey:
-        return jsonResponse({"error": "Unrecognized API key."})
-    user = User.objects.get(id=apikey["uid"])
+    user = request.user
```
schreibersefaria, line 1181: This seems to only be used once

sourcesheets/views.py
```diff
@@ -558,31 +540,17 @@ def save_sheet_api(request):
...
-        else:
-            user = request.user
+        user = request.user
 
         if "id" in sheet:
+            if request.is_api_authenticated:
+                sheet["lastModified"] = get_sheet(sheet["id"])["dateModified"]  # Usually lastModified gets set on the frontend, so we need to set it here to match with the previous dateModified so that the check in `save_sheet` returns properly
             existing = get_sheet(sheet["id"])
             if "error" not in existing  and \
                 not can_edit(user, existing) and \
```
schreibersefaria, line 556: You can also use user in can_add

reader/views.py
schreibersefaria, line 1: Not your bug but maybe you can fix.
in line 2125 there is a type (resr instead of res)

reader/views.py
```diff
@@ -1328,13 +1312,9 @@ def modify(uid):
...
 
         if not request.user.is_authenticated:
-            key = request.POST.get("apikey")
-            if not key:
-                return jsonResponse({"error": "You must be logged in or use an API key to save texts."})
-            apikey = db.apikeys.find_one({"key": key})
-            if not apikey:
-                return jsonResponse({"error": "Unrecognized API key."})
-            return jsonResponse(modify(apikey['uid']))
+            return authenticationRequiredResponse()
```
schreibersefaria, line 1315: Did we check the change in error type doesn't break anything in the frontend?
We are improving the http responses, now returning 401. But it is a breaking change potentially.

We may also want to warn Jessica (community manager) that we are doing a breaking change to the API.

reader/views.py
```diff
@@ -2366,23 +2308,13 @@ def update_version(request, title, lang, version):
...
-        user = User.objects.get(id=apikey["uid"])
-        if not user.is_staff:
-            return jsonResponse({"error": "Only Sefaria Moderators can flag texts."})
-
+    if not request.user.is_staff:
+        return notStaffOrApiResponse()
+    if request.is_api_authenticated:
         return update_version(request, title, lang, version)
-    elif request.user.is_staff:
+    else:
```
schreibersefaria, line 2319: I may have this wrong.

Would this be a case where `not request.user.is_authenticated` but there is an inconsistent reply.
Not `authenticationRequiredResponse` (401)

Now looking, I think I see many such cases in code you didn't change

schreibersefaria (commented):
An important task! Looks tough to implement and looks good overall.
I think I may have only one or two real comments.

Not sure I'm enough to check this. I'm not very familiar with this code.

reader/views.py
YishaiGlasner: i've did a general check to frontend and i guess we'll want a qa.
i can't see how getting 401 rather than 200 with only error message will break something.

sefaria/client/util.py
YishaiGlasner: django's responses go with snake case

sefaria/views.py
YishaiGlasner: now twice because your other comment

sourcesheets/views.py
YishaiGlasner: i'm not sure what exactly you mean

schreibersefaria: You can update not can_add(request.user, existing) -> not can_add(user, existing)

reader/views.py
schreibersefaria: Well people could have a catch specific to the error message or have catches for `!= 200`

sourcesheets/views.py
YishaiGlasner: ah, you refer the next line, which isn't shown by github...
anyway, i think this pr is already too complicated, so i'll avoid this ahich is not related directly

reader/views.py
YishaiGlasner: previously they got 200 for error...
but if they are not handling error properly (because we used to return 200), that can break their logic.
i'm not sure what should be the conclusion. @yitzhakc what do you say?

schreibersefaria: What I was going for is the frontend checks and QA (that you already did + we will do).
Other than that, just mentioning to Jessica (she can lead a decision if she sends out an email or not, but that is not part of the PR)

yitzhakc: I would agree that we should definitely let Jessica know, as we should update the DevPortal docs, and perhaps send out an email as well. 
Beyond that, I'm under the impression that - unfortunately - we don't have any good mechanism for deprecating APIs. Since we were totally unjustified in returning a `200` here, I feel that in this case it might be justified to just make the change. I'll try and dig a little deeper and see if we've done anything similar in the past.

reader/views.py
```diff
@@ -2119,23 +2069,17 @@ def notes_api(request, note_id_or_ref):
...
-            key = request.POST.get("apikey")
-            if not key:
-                return jsonResponse({"error": "You must be logged in or use an API key to add, edit or delete links."})
-
-            apikey = db.apikeys.find_one({"key": key})
-            if not apikey:
-                return jsonResponse({"error": "Unrecognized API key."})
-            note["owner"] = apikey["uid"]
+            return authenticationRequiredResponse()
+        uid = note["owner"]=  request.user.id
```
yitzhakc: Please fix the spacing

reader/views.py
```diff
@@ -2842,24 +2762,18 @@ def updates_api(request, gid=None):
...
+        elif request.is_api_authenticated:
             payload = json.loads(request.POST.get("json"))
             try:
                 GlobalNotification(payload).save()
                 return jsonResponse({"status": "ok"})
             except AssertionError as e:
                 return jsonResponse({"error": str(e)})
 
-        elif request.user.is_staff:
+        else:
```
yitzhakc: What situation does this `else` leave us in? `is_staff`? `unauthenticated`? Pleae add a comment since it's not-trivial

sefaria/system/middleware.py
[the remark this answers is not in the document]
yitzhakc: It is actually a valid performance concern...

Just to clarify, how many endpoints did we have where you could `POST` the api key? Maybe we could check the `POST` this way only when calling those specific endpoints?

You can also check to see if it's a `POST` request using the method you qouted from ChatGPT, but that's more convuluted.

sefaria/system/middleware.py
yitzhakc: @YishaiGlasner this would have been the advantage of a decorator, that we could have specified which endpoints need this check versus the middleware where it's on every request 😉 

To sum up, I think the implementation is fine, performance wise (or at least no worse then what we currently have) due to the reasoning Yishai gave - we'll only be hitting mongo for API requests where an `apikey` was submitted.

sefaria/system/middleware.py
```diff
@@ -218,3 +222,23 @@ def process_response(self, request, response):
 
             response = HttpResponse('<pre>%s</pre>' % io.getvalue())
         return response
+
+
+class ApiKeyAuthenticationMiddleware:
```
yitzhakc, line 227: @YishaiGlasner I would like a well written docstring on the middleware please.

sefaria/client/util.py
```diff
@@ -45,6 +45,17 @@ def celeryResponse(task_id: str, sub_task_ids: list[str] = None):
...
+    if data is None:
+        data = {}
+    return JsonResponse({"error": "Authentication required", **data}, status=401)
+
+
+def notStaffOrApiResponse():
```
yitzhakc, line 56: On the one hand, the data security part of me says we should be giving as non-detailed error messages as we can. On the other hand, our code is all open source 🤣 

My question is whether we have cases where `is_staff` would be allowed but an API key wouldn't - in which case bundling the two together is not ideal.

Also, why would the returned value not be a JSONResponse like in the previous function?

sefaria/client/util.py
yitzhakc: I think we _might_ be confusing different types.

Django's built-in response objects are classes (`Http404`, `JsonResponse`) and so are CamelCased.

View functions are, well, functions and therefore should be snake_cased.
Most of this file seems to be using helper functions for responses, which should be snake_cased (but are not...)

Here you've written new response helper functions, which would be snake_cased, not CamelCased. 
That being said, I would actually recommend writing these new response types as classes, sub-classing Django's own `HttpResponseForbidden` and `JsonResponse`.

And then they would be CamelCased, since they would be classes 😄

reader/views.py
```diff
@@ -2449,23 +2385,15 @@ def _internal_do_post(request, update, cat, uid, **kwargs):
...
-                return jsonResponse({"error": "Only Sefaria Moderators can add or delete categories."})
-            uid = apikey["uid"]
+            return authenticationRequiredResponse()
+        if not request.user.is_staff:
+            return notStaffOrApiResponse()
+        if request.is_api_authenticated:
             kwargs = {"method": "API"}
-        elif request.user.is_staff:
-            uid = request.user.id
+        else:
```
yitzhakc: What is this else? It's not api_authenticated?

reader/views.py
```diff
@@ -2613,23 +2541,15 @@ def _internal_do_post(request, uid):
...
-                return jsonResponse({"error": "Only Sefaria Moderators can add or edit terms."})
-            uid = apikey["uid"]
+            return authenticationRequiredResponse()
+        if not request.user.is_staff:
+            return notStaffOrApiResponse()
+        if request.is_api_authenticated:
             kwargs = {"method": "API"}
-        elif request.user.is_staff:
-            uid = request.user.id
+        else:
```
yitzhakc: We don't need to check `is_staff` here?

sefaria/system/middleware.py
yitzhakc: It is commonly accepted to pass API keys through the `Authorization` header, which is a reserved word in the HTTP spec for this purpose.

I'd prefer we try and use commonly supported methods, and not just add custom headers. 😄

yitzhakc (changes requested):

reader/views.py
YishaiGlasner: yes, it's a logged in staff

reader/views.py
YishaiGlasner: we have checked on line 2545, and returned if not

sefaria/client/util.py
YishaiGlasner: I didn't see such a case and I went over all of them...
maybe in the future we'll change the logic of api to have dofferent degrees but for now api is super.

i didn't use here JSONResponse, for Django has HttpResponseForbidden for 403 which is exactly what we need, but there is no 401 Django response.

sefaria/system/middleware.py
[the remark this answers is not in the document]
YishaiGlasner: @yitzhakc I don't know if counting these methods is the issue here. shouldn't the question be how much we use it? in this PR i've removed them from Sefaria-Project and I have another branch for removing it from Sefaria-Data. so we shouldn't get them from our team. idk, how many outern users have it (if at all) but i'm pretty sure they don't use it so much. 
i'd say much further - we can allow it only in headers and if someone will fail posting, tell him.

YishaiGlasner: @yitzhakc so do you want them both?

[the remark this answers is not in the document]
yitzhakc: I hear your view - but it's a little "harsh" on our end to just block them like that. Yes, they'll eventually notice - but we don't really have a good way "notifying" API users like that, unfortuantely. 
I'm looking into seeing if have any metrics on these endpoints, but for now, please modify the code and put the `POST` logic separately (doing it in a way that doesn't force evaluation of the entire request in the middleware) so we minimize potential performance issues, and potentially track the usage of `POST` and easily remove it based on our findings.

yitzhakc: @YishaiGlasner, any comment?

reader/views.py
yitzhakc: @YishaiGlasner any comment?

yitzhakc (commented):
Requested minor changes based on comments

reader/views.py
YishaiGlasner: @yitzhakc no, except i'm not sure what we should update in the DevPortal - i don't think there is a place where we say that we have the wrong code

sefaria/system/middleware.py
YishaiGlasner: i've sritten but somehow it remained pending:
@yitzhakc so do you want them both?
(i.e. checking HTTP_X_APIKEY and  Authorization)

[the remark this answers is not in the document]
YishaiGlasner: @yitzhakc what do you mean by "doesn't force evaluation of the entire request"? using `request.body` rather than`request.POST`?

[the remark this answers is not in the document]
yitzhakc: After re-reviewing this with @YishaiGlasner we came across the [original trac issue](https://code.djangoproject.com/ticket/16384#:~:text=Description,using%20the%20@csrf_exempt%20decorator.%22) that adding the [warning](https://docs.djangoproject.com/en/5.2/topics/http/middleware/#process-view) to the docs.
There, it states explicitly that the concern is only when there is use of [custom upload handlers](https://docs.djangoproject.com/en/dev/topics/http/file-uploads/#modifying-upload-handlers-on-the-fly) which we don't use.

Therefore, no need to complicate the logic at this stage. YAGNI.

yitzhakc: Yes, for now I would like them both (both API_KEY and Authorization header).
But please  the `Authorization` header in any requests / scripts that we use (and not the custom API_KEY header).
I will create a story to track the metric of use of `HTTP_X_API_KEY` and hopefully remove it asap.

reader/views.py
yitzhakc: After reviewing every single endpoint that changed and cross-referencing with the relevant page on the DevPortal - none of the endpoints are documented as POSTable there (since it's all internal functions) and therefore no changes are needed there. 
Based on this, I would assume that our surface of external users who might be affected is also extremely small.

yitzhakc (changes requested):

dcschreiber:
_(Claude writing on Daniel's behalf)_ Closing: this approach is no longer relevant — API-key auth consolidation is being pursued as part of the API Key Program (sc-46208) on a different branch.

PR #2562, opened 2025-07-15
chore: added class .highlightedCitation to all citations present in C…

stevekaplan123:
…onnectionsPanel and scrolling to the first of them after initializing scrollListener

## Description

## Code Changes

## Notes


static/js/ConnectionsPanel.jsx
```diff
@@ -165,6 +166,61 @@ class ConnectionsPanel extends Component {
     });
     this._savedHistorySegments.add(ref);
   }
+  onTextLoad(order) {
```
alekselazar, line 169: This function implements adding .highlightCitation to all elements with `data-ref=[currentlyVisibleRef]`, and if the first element of this class is out of view port, it scrolls ot 4-5 lines above this alement. Function is called after new text loaded on `ConnectionsPane`l by passing it as callback to child `TextList` elemnt, or if component was updated but new text did not load by calling this function in `this.componentDidUpdate`

sefaria/model/text_request_adapter.py
```diff
@@ -175,65 +166,47 @@ def make_named_entities_dict(version_title, language):
...
-                if len(current_funcs) == 1:
-                    composite_func = current_funcs[0]
-                else:
-                    composite_func = lambda string, sections: reduce(lambda s, f: f(s, sections), current_funcs, string)
-                
-                version['text'] = ja.modify_by_function(composite_func)
+                version_title = version['versionTitle']
+                ne_by_secs = make_named_entities_dict()
+                # NEW: build MarkedUpTextChunk for this version and apply before wrap_links
+                marked_up_text_chunk = MarkedUpTextChunk().load({'ref': self.oref.normal(), 'language': language, 'versionTitle': version_title})
```
alekselazar, line 207: added `MarkedUpTextChunk` object, that corresponds to `self.oref` so it can be used to wrap all quotes

sefaria/model/text_request_adapter.py
```diff
@@ -175,65 +166,47 @@ def make_named_entities_dict(version_title, language):
...
 
         # Define text modification functions based on return format
         text_modification_funcs = []
+        if self.return_format == 'wrap_all_entities':
+            # Check for links and load segment refs only once
+            shared_data['all_segment_refs'] = self.oref.all_segment_refs()
 
+            query = self.oref.ref_regex_query()
+            query.update({"inline_citation": True})
+            text_modification_funcs = [lambda string, _: library.get_wrapped_from_marked_up(string, marked_up_text_chunk), lambda string, sections: library.get_wrapped_named_entities_string(ne_by_secs[(sections[-1],)], string)]
```
alekselazar, line 183: added `library.get_wrapped_from_marked_up` as first in `text_modification_funcs` so it will be called first and wrap quotes with span tags before citations, since `MarkedUpTextChunks` are index sensetive, and if some modification will be  applied before it, it will end with wrong wrapping

sefaria/model/text.py
```diff
@@ -6030,6 +6101,21 @@ def get_wrapped_refs_string(self, st, lang=None, citing_only=False, reg=None, ti
                 new_string += self.apply_action_for_all_refs_in_string(substring, self._wrap_ref_match, lang,
                                                                        citing_only, reg, title_nodes)
         return new_string
+    
+    def get_wrapped_from_marked_up(self, st, marked_up_text_chunk):
```
alekselazar, line 6105: function that wraps all quotes extracted from `marked_up_text_chunk` in `<span>` tags

sefaria/model/text_request_adapter.py
stevekaplan123: @yonadavGit I'm wondering how you think this should relate to wrapping links.  That would, by the same logic, also need to be done first.

sefaria/model/text.py
```diff
@@ -6030,6 +6101,21 @@ def get_wrapped_refs_string(self, st, lang=None, citing_only=False, reg=None, ti
...
+        if marked_up_text_chunk is None:
+            return st
+        tasks = sorted([span for span in marked_up_text_chunk.spans], key=lambda x: x['charRange'][0], reverse=True)
```
alekselazar, line 6108: I applied all wrappings in reversed way, so each wrapping does not break indexes of other

sefaria/model/text.py
```diff
@@ -6030,6 +6101,21 @@ def get_wrapped_refs_string(self, st, lang=None, citing_only=False, reg=None, ti
...
+        for task in tasks:
+            if not task['text'] in st:
+                continue
+            if task['type'] == 'quote':
+                oref = Ref(task['ref'])
+                start, end = task['charRange']
+                st = st.replace(st[start:end], self._wrap_quote_from_marked_up(oref, st[start:end]))
+            else:
+                pass
```
alekselazar, line 6117: Since by spec in feauture `MarkedUpTextChunk` should contain citations too, I added it as a room for wrapping citations into `<a>` tags in future

sefaria/model/text.py
```diff
@@ -6224,6 +6310,10 @@ def _internal_ref_from_string(self, title=None, st=None, lang=None, stIsAnchored
     def _wrap_ref_match(ref, match):
         return '<a class ="refLink" href="/{}" data-ref="{}">{}</a>'.format(ref.url(), ref.normal(), match.group(0))
 
+    @staticmethod
+    def _wrap_quote_from_marked_up(ref, string):
+        return f'<span class="refQuote" data-ref="{ref.normal()}">{string}</span>'
```
stevekaplan123, line 6315: @alekselazar Why is the class here 'refQuote'?

alekselazar: I added it for possible convention, if we want in future to style them somehow, I thought adding raw span with no className is not so good

PR #2565, opened 2025-07-16
chore: add openNewTab check

stevekaplan123:
## Description
New tab logic was broken.  If in sheets module, a link to the sheets module should not open a new tab.

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -131,16 +131,16 @@ const LoggedInDropdown = ({module}) => {
 );
 }
 
-
 const ModuleSwitcher = () => {
+  const openNewTab = (newModule) => Sefaria.activeModule !== newModule;
   return (
       <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
           <div className='dropdownLinks-options'>
-              <DropdownMenuItem url={'/'} newTab={true}>
+              <DropdownMenuItem url={'/'} newTab={openNewTab('library')}>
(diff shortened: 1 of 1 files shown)


static/js/Header.jsx
```diff
@@ -131,16 +131,16 @@ const LoggedInDropdown = ({module}) => {
...
 }
 
-
 const ModuleSwitcher = () => {
+  const openNewTab = (newModule) => Sefaria.activeModule !== newModule;
   return (
       <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
           <div className='dropdownLinks-options'>
-              <DropdownMenuItem url={'/'} newTab={true}>
+              <DropdownMenuItem url={'/'} newTab={openNewTab('library')}>
```
YishaiGlasner: does `openNewTab` return a boolean? if so, i think it should be renamed `shouldOpenNewTab`.
proof:
it took me time to understand this code
QED :)
and seriously, it isn't really part of the thing that this PR do, but if you can do it, it will be nice. if not, say and i'll approve.

YishaiGlasner (changes requested):

static/js/Header.jsx
yitzhakc: I'd agree with @YishaiGlasner that it's a little confusing, since the `open` verb does lend itself to the understanding that if this is True we should... well, open a new tab right here and right now 🙃

And I agree with Yishai that we should change the name. But I think that might be solvable by changing the name to `openInNewTab`, which makes it clear the we're not opening anything _right now_. But Yishai's suggestion is fine too :-)

YishaiGlasner (approved):

yitzhakc (approved):

PR #2568, opened 2025-07-16
Bug/sc 903/mishnah berakhot chapter jumps

schreibersefaria:
## Description
The fix solves a bug where navigating a text with the TOC would then cause the page to jump.

## Code Changes
Fixing code changes
1. Changed the calculation of restoreScrollPositionAfterTopLoad - removing a `2*` that was causing trouble
2. Fixed the infinite scroll state persisting after chapter navigation

Other code changes
1. Refactoring code into functions and adding comments for readability
2. Changed the clearing of loadingContentAtTop - bring it forward to when we know the loading finished

## Notes
I cleaned out a bunch of old console logs (mostly commented out)

### DIFF
--- a/static/js/TextColumn.jsx
+++ b/static/js/TextColumn.jsx
@@ -38,7 +38,6 @@ class TextColumn extends Component {
          content_type: Sefaria.index(this.props.bookTitle).primary_category,
          item_id: this.props.bookTitle
        }
-      console.log(params)
       gtag("event", "select_content", params)
 
     this.node.addEventListener("scroll", this.handleScroll);
@@ -54,13 +53,11 @@ class TextColumn extends Component {
 
     if (prevProps.mode === "Text" && this.props.mode === "TextAndConnections") {
       // When opening mobile connections panel, scroll to highlighted
-      // console.log("scroll to highlight for mobile connections open")
       this.scrollToHighlighted();
 
     } else if (this.state.showScrollPlaceholders && !prevState.showScrollPlaceholders && !this.initialScrollTopSet) {
       // After scroll placeholders are first rendered, scroll down so top placeholder
       // is out of view and scrolling up is possible.
-      // console.log("scrolling for ScrollPlaceholders first render")
       this.setInitialScrollPosition();
 
     } else if (this.props.srefs.length === 1 &&
@@ -68,7 +65,6 @@ class TextColumn extends Component {
         !prevProps.srefs.some(r => Sefaria.refContains(this.props.srefs[0], r))) {
       // If we are switching to a single ref not in the current TextColumn,
       // treat it as a fresh open.
-      // console.log("setting initialScroll for brand new ref")
       this.setInitialScrollPosition();
 
     } else if (prevProps.srefs.length === this.props.srefs.length &&
@@ -80,11 +76,9 @@ class TextColumn extends Component {
     } else if ((this.props.settings.language !== prevProps.settings.language) ||
         !Sefaria.areBothVersionsEqual(prevProps.currVersions, this.props.currVersions)) {
       // When the content the changes but we are anchored on a line, scroll to it
-      // console.log("scroll to highlighted on text content change")
       this.scrollToHighlighted();
     } else if (layoutWidthChanged) {
       // When the width of the text column changes, keep highlighted text in place
-      // console.log("restore scroll by percentage for layout Width Change")
       this.restoreScrollPositionByPercentage();
     }
     
@@ -158,7 +152,6 @@ class TextColumn extends Component {
 
     // TextRanges in the column may be initial rendered in "loading" state without data.
     // When the data loads we may need to change scroll position or render addition ranges.
-    // console.log("handle text load: ", ref);
     if (this.$container.find(".basetext.loading").length) {
       // Don't mess with scroll positions until all sections of text have loaded,
       // prevent race conditions when mutliple section may load out of order.
@@ -173,7 +166,6 @@ class TextColumn extends Component {
     // When content loads check if we already need to load another section below, which
     // occurs when the loaded section is very short and whitespace is already visible below it.
     // Only check down, a text load should never trigger an infinite scroll up
-    // console.log("Checking infinite scroll down");
     this.adjustInfiniteScroll(true);
 
     if (this.loadingContentAtTop) {
@@ -228,7 +220,6 @@ class TextColumn extends Component {
   restoreScrollPositionAfterTopLoad() {
     // After one or more new TextRanges have just loaded in the first position, scroll
     // down to the TextRange that was visible before, so the TextColumn doesn't jump.
-    // console.log("checking restore scroll after up");
     const $texts  = this.$container.find(".basetext");
     if ($texts.length < 2 || $texts.eq(0).hasClass("loading") ) { return; }
 
@@ -237,7 +228,6 @@ class TextColumn extends Component {
     const adjust = this.scrollPlaceholderHeight + this.scrollPlaceholderMargin;
     const top = targetTop + (2*this.node.scrollTop) - adjust;
     this.setScrollTop(top);
-    // console.log("scroll to restore after infinite up: " + top)
   }
   restoreScrollPositionByPercentage() {
     // After the layout width of the column changes, restore the scroll to the same percentage
--- a/static/js/TextColumn.jsx
+++ b/static/js/TextColumn.jsx
@@ -47,6 +47,19 @@ class TextColumn extends Component {
     this.node.removeEventListener("scroll", this.handleScroll);
   }
   componentDidUpdate(prevProps, prevState) {
+    // Track navigation changes and reset infinite scroll state when needed
+    if (prevProps.srefs.length !== this.props.srefs.length || 
+        !prevProps.srefs.every((ref, index) => ref === this.props.srefs[index])) {
+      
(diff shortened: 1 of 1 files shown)


static/js/TextColumn.jsx
```diff
@@ -48,27 +47,30 @@ class TextColumn extends Component {
     this.node.removeEventListener("scroll", this.handleScroll);
   }
   componentDidUpdate(prevProps, prevState) {
+    // Identify navigation changes from TOC and reset infinite scroll state
+    // Reset if we have fewer sections in the navigation than we have loaded OR if it's a navigation to a single section
+    if (this.numSectionsLoadedAtTop && this.props.srefs.length <= this.numSectionsLoadedAtTop || this.props.srefs.length === 1) {
```
YishaiGlasner: i didn't understand all of these conditions. are all of them tryin to identify navigation from TOC? we should have a better (direct) way to know that we are doing navigation from toc.

static/js/TextColumn.jsx
```diff
@@ -80,23 +82,26 @@ class TextColumn extends Component {
...
-      // console.log("restore scroll by percentage for layout Width Change")
       this.restoreScrollPositionByPercentage();
     }
     
   }
+  
+  resetInfiniteScrollState() {
+    // Reset all infinite scroll state variables
+    // This is called when the user navigates through the TOC so the existing scroll position is not relevant
+    this.numSectionsLoadedAtTop = 0;
```
YishaiGlasner: this seems ufit to the rest of the component. in no place `this.numSectionsLoadedAtTop` is set to 0 (but it starts with 1), and also regarding the other instance variables I don't see they'are initialized in the contructing ot mounting. on its face, navigating from toc should be similar to navigating directly from a link or from the search bar (or form the toc pace which seems also to be ok).

i wonder if this is a case for re-monting the component, but even if it's a good idea, i'm not sure how we should do it for the key of it is something like `TextColimn-Genesis` so to what exactly we can change it for triggering a remount?

static/js/TextColumn.jsx
```diff
@@ -141,58 +145,56 @@ class TextColumn extends Component {
...
+    
+    // Don't adjust scroll positions while sections are still loading to prevent race conditions when mutliple section may load out of order.
+    if (this.$container.find(".basetext.loading").length > 0) {
       return;
     }
-
+    
+    // Clear the loading flag if we were loading content at the top
+    const wasLoadingContentAtTop = this.loadingContentAtTop;
+    if (this.loadingContentAtTop) {
```
YishaiGlasner: you can omit the condition, setting it always to false is simpler

static/js/TextColumn.jsx
```diff
@@ -219,25 +220,39 @@ class TextColumn extends Component {
...
+  calculateScrollPositionAfterTopLoad($texts) {
+    // Calculate the correct scroll position after new content has been loaded at the top
+    // The goal is to keep the user viewing the same content they were looking at before (avoid a jump)
     const targetTop = $texts.eq(this.numSectionsLoadedAtTop).position().top;
     const adjust = this.scrollPlaceholderHeight + this.scrollPlaceholderMargin;
-    const top = targetTop + (2*this.node.scrollTop) - adjust;
-    this.setScrollTop(top);
-    // console.log("scroll to restore after infinite up: " + top)
+    
+    const currentScrollTop = this.node.scrollTop;
```
YishaiGlasner: you've removed the `2*`. was it a bug? if so hot it didn't screw up the crolling?

static/js/TextColumn.jsx
```diff
@@ -247,84 +262,126 @@ class TextColumn extends Component {
...
-        }
+  handleInfiniteScrollUp() {
+    // When user scrolls near the top, load previous sections above current content
+    // This prevents the user from hitting the top and enables smooth upward scrolling
+    
+    const refs = this.props.srefs.slice();
+    const topRef = refs[0]; // Currently displayed top section
+    
+    const currentData = Sefaria.ref(topRef);
+    if (!currentData || !currentData.prev) {
```
YishaiGlasner: `!currentData?.prev` do both

static/js/TextColumn.jsx
```diff
@@ -247,84 +262,126 @@ class TextColumn extends Component {
...
+      earlierData = Sefaria.ref(earlierData.prev);
+    }
+    
+    return refs;
+  }
+
+  handleInfiniteScrollDown() {
+    // When user scrolls near the bottom, load next sections below current content
+    // This enables smooth downward scrolling
+    const $lastText = this.$container.find(".textRange.basetext").last();
```
YishaiGlasner: this is a duplication. myabe you should do a separate function for this

static/js/TextColumn.jsx
```diff
@@ -247,84 +262,126 @@ class TextColumn extends Component {
...
       if ($lastText.hasClass("loading")) {
-        //console.log("last text is loading - don't add next section");
-        return;
+      return; // Don't load more while already loading
```
YishaiGlasner: indent

static/js/TextColumn.jsx
```diff
@@ -247,84 +262,126 @@ class TextColumn extends Component {
...
-        refs.push(data.next); // Append at least the next one
-        let numSectionsAddToBottom = 1;
-        let nextData, laterData;
-
-        // Now, only add sources if we have data for them
-        if(nextData = Sefaria.ref(data.next)) {
-          laterData = Sefaria.ref(nextData.next);
-        }
+    
+    const refs = this.props.srefs.slice();
```
YishaiGlasner: this seems to be redundant. you're slicing and then in the next line slicing again.
true, you're passing `refs` also to `buildNextRefs`, but you're slicing also there. i think the best will be not to pass them and to use `this.props.refs` also in `buildNextRefs`

static/js/TextColumn.jsx
```diff
@@ -247,84 +262,126 @@ class TextColumn extends Component {
...
+    const currentRef = refs.slice(-1)[0]; // Currently displayed bottom section
+    const currentData = Sefaria.ref(currentRef);
+    
+    if (!currentData || !currentData.next) {
```
YishaiGlasner: `!currentData?.next`

static/js/TextColumn.jsx
schreibersefaria: It starts at 1 in buildPreviousRefs. That is after adding currentData.prev to the refs.
I'll have a look into remounting around a "better (direct) way to know that we are doing navigation"
maybe I just need to reset `numSectionsLoadedAtTop` before use - I'll have a look at this

static/js/TextColumn.jsx
schreibersefaria: Done.

static/js/TextColumn.jsx
schreibersefaria: Well it did, that is the bug I guess.
This *2 was in `restoreScrollPositionAfterTopLoad` and while testing I have seen many jumps while scrolling up. But it wouldn't happen in other scrolling.
I don't have a full understanding of the code and haven't managed to exactly recreate when the bug happens and when it doesn't. Happy to discuss if you think that is needed in this case.

static/js/TextColumn.jsx
schreibersefaria: I'm working on trying to do resetInfiniteScrollState more directly. Would appreciate your thoughts.

I could trigger a remount. Doesn't seem like it would be very difficult. A bit scary though. This is the key currently:
```jsx
key={`${textColumnBookTitle ? textColumnBookTitle : "empty"}-TextColumn`}
```

Another option is to add a navigation type prop. The disadvantage being that it will be changing a lot of functions.

YishaiGlasner: i've thought about another similar case - you're going to Mishnah Berakhot 1, then from the search bar to chapter 3 (no remounting, for the key remains the same, and now in `TextColumn` there is only one `TextRange`, like in the toc case) and scroll up - this works. so there is a difference in what happens when going to chapter 3 via toc and search bar. you can try to find out what's happening in search bar and do the same.

static/js/TextColumn.jsx
YishaiGlasner: ok, sounds good for me.

static/js/TextColumn.jsx
schreibersefaria: Mmm good idea! but unfortunately I'm now managing to recreate the bug also when navigating through the search form chapter 1 -> 3

YishaiGlasner: That's a great observation! (in the futue everybody will learn how to communicate from chatGPT and will write like this).
That means that `TextColumn` which planned to be the infinite scroll component, wasn't planned to task with changing the location without scrolling. when the book is changed (like navigating from Genesis to Exodus through search bar) it's remounted, as the key is changed, but it seems no one thought about navigating in the same index not by scrolling (the side bar toc is late invention).
So i think we don't need to seek for the original solution, but for the best new one.
Remounting is very logical in my eyes, for we're dealing with a component that's know what to do in scrolling and not in jumping, but I don't like the idea of arbitrary changing of key like adding it a number. I'd prefer a key that tells me something. does your approach fixed also the navigating through the search?

schreibersefaria: I found when they added the key, and it didn't help.
I'll try work on the problem directly
https://github.com/Sefaria/Sefaria-Project/commit/59abfc820bef1e980b312a019b03ce98c0c2b496

PR #2570, opened 2025-07-18
Add SA to cross-check GA

relyks:
## Description

**Simple Analytics (SA)** has been added alongside GA to verify event counts and isolate any `IntersectionObserver` problems. SA’s raw, unsampled logs and its stricter bot filtering will help us determine if real humans are seeing fewer impressions or if GA is inflating page views.

A couple fixed elements were wired up to validate the observer itself. The fixed elements must become fully visible at least once per session. When that happens, each of their hooks fires exactly one impression event and tags the session with a key so that it can later be confirmed through analysis.

## Code Changes

- **Embed Simple Analytics** in base.html.
  - Production script plus a dev script that points at a separate hostname, so test traffic doesn’t taint production numbers
  - SA automatically strips bots it recognizes
- **Include consistent app‑level metadata** in every SA event, the same fields already being sent to GA, plus some additional ones.
  - Generate and send an explicit `session_uuid` so the ID is visible in SA dashboards and easy to join on later.
- **Convert Header component to a functional component** so that hooks can be used cleanly
- **Add useOnceFullyVisible hook** (inside the Header file). It attaches an IntersectionObserver and fires a callback only after the element has been 100% in view
- **Emit a `header_viewed` event** the first time the header or the category color line is fully visible. One of those two elements appears in every session.
- **On ReaderApp mount**
  - Fire `reader_app_mounted` to mark a standard app startup.
  - Fire `intersection_observer_not_supported` if the browser lacks `IntersectionObserver`. These two events will help explain any missing impression data.

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -16,71 +16,68 @@ import {
 import {ProfilePic} from "./ProfilePic";
 import {HeaderAutocomplete} from './HeaderAutocomplete'
 
-class Header extends Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      mobileNavMenuOpen: false,
-    };
-  }
-  componentDidMount() {
-    window.addEventListener('keydown', this.handleFirstTab);
-  }
-  handleFirstTab(e) {
-    if (e.keyCode === 9) { // tab (i.e. I'm using a keyboard)
-      document.body.classList.add('user-is-tabbing');
-      window.removeEventListener('keydown', this.handleFirstTab);
+const Header = (props) => {
+
+  useEffect(() => {
+    const handleFirstTab = (e) => {
+      if (e.keyCode === 9) { // tab (i.e. I'm using a keyboard)
+        document.body.classList.add('user-is-tabbing');
+        window.removeEventListener('keydown', handleFirstTab);
+      }
     }
-  }
-  toggleMobileNavMenu() {
-    this.setState({mobileNavMenuOpen: !this.state.mobileNavMenuOpen});
-  }
-  render() {
-    if (this.props.hidden && !this.props.mobileNavMenuOpen) {
-      return null;
+
+    window.addEventListener('keydown', handleFirstTab);
+
+    return () => {
+      window.removeEventListener('keydown', handleFirstTab);
     }
-    const logo = Sefaria.interfaceLang == "hebrew" ?
-      <img src="/static/img/logo-hebrew.png" alt="Sefaria Logo"/> :
-      <img src="/static/img/logo.svg" alt="Sefaria Logo"/>;
+  }, []);
 
-    const headerContent = (
-      <>
+  if (props.hidden && !props.mobileNavMenuOpen) {
+    return null;
+  }
 
-        <div className="headerNavSection">
-          { Sefaria._siteSettings.TORAH_SPECIFIC ?
-          <a className="home" href="/" >{logo}</a> : null }
-          <a href="/texts" className="textLink"><InterfaceText context="Header">Texts</InterfaceText></a>
-            <a href="/topics" className="textLink"><InterfaceText context="Header">Explore</InterfaceText></a>
-          <a href="/community" className="textLink"><InterfaceText>Community</InterfaceText></a>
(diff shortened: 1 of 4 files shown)


static/js/Header.jsx
```diff
@@ -16,71 +16,99 @@ import {
...
-    };
-  }
-  componentDidMount() {
-    window.addEventListener('keydown', this.handleFirstTab);
-  }
-  handleFirstTab(e) {
-    if (e.keyCode === 9) { // tab (i.e. I'm using a keyboard)
-      document.body.classList.add('user-is-tabbing');
-      window.removeEventListener('keydown', this.handleFirstTab);
+function useOnceFullyVisible(onVisible, key) {
```
yitzhakc: I would have said that this function belongs in `Misc.jsx` or `Hooks.jsx` or something, not at the top of `Header.jsx`, especially since you use it elsewhere.

static/js/Misc.jsx
```diff
@@ -1443,8 +1444,19 @@ const SmallBlueButton = ({onClick, tabIndex, text}) => {
 };
 
 
-const CategoryColorLine = ({category}) =>
-  <div className="categoryColorLine" style={{background: Sefaria.palette.categoryColor(category)}}/>;
+const CategoryColorLine = ({ category }) => {
+  const headerRef = useOnceFullyVisible(() => {
```
yitzhakc: I'm not sure if `headerRef` is the best name for this hook, if it actually relates to a category color line :-)

templates/base.html
```diff
@@ -153,6 +153,32 @@
...
+    
+    <!-- Simple Analytics -->
+    {# Allow queuing of events in case embed script hasn't loaded yet when calling sa_event #}
+    <script>
+      window.sa_event = window.sa_event || function () {
+        const a = [].slice.call(arguments);
+        window.sa_event.q ? window.sa_event.q.push(a) : window.sa_event.q = [a];
+      };
+    </script>
+    {% if DEBUG %}
```
yitzhakc, line 165: Do you want to make this a variable injected through cauldron spec values? Otherwise, you'll have everyone's localhosts and cauldron code reporting here.

templates/base.html
```diff
@@ -275,5 +301,21 @@
     {% render_bundle 'main' %}
 
     {% block js %}{% endblock %}
+    
+    <!-- Default metadata for Simple Analytics -->
+    <script>
+      const SESSION_KEY = 'sa_custom_session_id';
```
yitzhakc, line 307: I looked in the SA docs but didn't find much reference to `custom_session_id`. What is the goal here?

yitzhakc (changes requested):

templates/base.html
relyks: You can only access SA's session IDs for events when the data is exported into CSV. You can't access the session ID from the dashboard. In some scenarios, I want to be able to filter by session ID in their dashboard. From what I understand of their docs, refreshing the page or causing a page load to a different route might sometimes cause another session ID to be created. In the web app, a page load can happen in the same tab/window when you go from reader app to a static page like from `/texts` to `/about`.  By creating one myself, I can ensure the session ID is consistent for the events that happen during a user session and that they are grouped properly if I choose look at specific sessions in the dashboard.

templates/base.html
relyks: Yeah, I'm aware. I think it's fine. The more data the better as long as it's separate from what happens in production. I have already verified on a cauldron (https://simple-analytics.cauldron.sefaria.org) that the events are being sent and received. I want general trends to be visible from all the cauldrons. I can't think a reason of why also receiving events from localhosts could be a problem though. However, GA events aren't sent by default from localhosts (though they are for cauldrons), because the environment variable for the property ID isn't set by default. If you think this could be problematic, I can also change it

templates/base.html
yitzhakc: Your reasoning seems sound 😄 
I think we can leave it as-is, and if you see that it's polluting your data, we can re-think.

templates/base.html
yitzhakc: Your reasoning seems sound 😄
I think we can leave it as-is, and if you see that it's polluting your data, we can re-think.

yitzhakc (approved):

PR #2583, opened 2025-07-23
Add support for multiple versions in CSV upload

NechapsaVenachkora:
## Description
This PR adds improved support for uploading multiple versions of a text via CSV.

## Code Changes
Added support for processing multiple versions in a single CSV file

### DIFF
--- a/sefaria/export.py
+++ b/sefaria/export.py
@@ -720,8 +720,51 @@ def import_versions_from_file(csv_filename, columns):
     return _import_versions_from_csv(rows, columns)
 
 
+      # new import ✔
+
 def _import_versions_from_csv(rows, columns, user_id):
     from sefaria.tracker import modify_bulk_text
+    from collections import defaultdict
+    from sefaria.model import Ref
+
+    # If the first label is "Version Title", we're in the multi-index sheet
+    if str(rows[0][0]).strip().lower() == "version title":
+        col = columns[0]
+        vt, lang, src, notes = rows[0][col], rows[1][col], rows[2][col], rows[3][col]
+
+        book_maps = defaultdict(dict)
+        for r in rows[4:]:
+            if not r or len(r) <= col:
+                continue
+            ref, txt = r[0], r[col]
+            if not txt:
+                continue
+            book_maps[Ref(ref).index.title][ref] = txt
+
+        for idx_title, text_map in book_maps.items():
+            index = Index().load({'title': idx_title})
+            if not index:
+                raise InputError(f'No book with primary title \"{idx_title}\"')
+            index_node = index.nodes
+
+            v = Version().load({
+                "title": idx_title,
+                "versionTitle": vt,
+                "language": lang
+            })
+            action = "edit"
+            if v is None:
+                action = "add"
+                v = Version({
+                    "chapter": index_node.create_skeleton(),
+                    "title": idx_title,
+                    "versionTitle": vt,
+                    "language": lang,
+                    "versionSource": src,
+                    "versionNotes":  notes,
+                }).save()
+            modify_bulk_text(user_id, v, text_map, type=action)
+        return
 
     index_title = rows[0][columns[0]]  # assume the same index title for all
     index = Index().load({'title': index_title})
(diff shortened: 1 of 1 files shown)


sefaria/export.py
```diff
@@ -720,8 +720,51 @@ def import_versions_from_file(csv_filename, columns):
     return _import_versions_from_csv(rows, columns)
 
 
+      # new import ✔
```
YishaiGlasner: ?

sefaria/export.py
```diff
@@ -720,8 +720,51 @@ def import_versions_from_file(csv_filename, columns):
...
+
 def _import_versions_from_csv(rows, columns, user_id):
     from sefaria.tracker import modify_bulk_text
```
YishaiGlasner: please move all imports to file top

sefaria/export.py
```diff
@@ -720,8 +720,51 @@ def import_versions_from_file(csv_filename, columns):
...
+    from collections import defaultdict
+    from sefaria.model import Ref
+
+    # If the first label is "Version Title", we're in the multi-index sheet
+    if str(rows[0][0]).strip().lower() == "version title":
+        col = columns[0]
+        vt, lang, src, notes = rows[0][col], rows[1][col], rows[2][col], rows[3][col]
```
YishaiGlasner: i think
```
vt, lang, src, notes  = [row[col] for row in rows[:4]]
```
will be better

sefaria/export.py
```diff
@@ -759,3 +802,15 @@ def _import_versions_from_csv(rows, columns, user_id):
         for row in rows[5:]:
             text_map[row[0]] = row[column]
         modify_bulk_text(user_id, v, text_map, type=action)
+
+
+# ── Workflowy helper for the mod-tools bulk uploader ──────────────────────
```
YishaiGlasner: all the rest seems to be a mistale

sefaria/export.py
```diff
@@ -720,8 +720,51 @@ def import_versions_from_file(csv_filename, columns):
...
+        for r in rows[4:]:
+            if not r or len(r) <= col:
+                continue
+            ref, txt = r[0], r[col]
+            if not txt:
+                continue
+            book_maps[Ref(ref).index.title][ref] = txt
+
+        for idx_title, text_map in book_maps.items():
+            index = Index().load({'title': idx_title})
```
YishaiGlasner: all of this (745-766) is duplicated on 770-804
you should do here an `else` statement, that build a simple `book_maps` for other case.
and add in the if else also logic for getting the versin title and source
and then doing the for loop for both cases.

YishaiGlasner (changes requested):

sefaria/export.py
```diff
@@ -720,42 +720,56 @@ def import_versions_from_file(csv_filename, columns):
...
+    if multi:
+        col = columns[0]
+        vt, lang, src, notes = [rows[i][col] for i in range(4)]
+        book_maps = defaultdict(dict)
+        for r in rows[4:]:
+            if not r or len(r) <= col:
+                continue
+            ref, txt = r[0], r[col]
+            if not txt:
+                continue
```
YishaiGlasner: that means you cannot remove text with csv. i don't think you want it. i think that if we have a ref we should apply the text change anyway.

sefaria/export.py
```diff
@@ -720,42 +720,56 @@ def import_versions_from_file(csv_filename, columns):
     return _import_versions_from_csv(rows, columns)
 
 
+      # new import ✔
```
YishaiGlasner: this is still here

sefaria/export.py
```diff
@@ -720,42 +720,56 @@ def import_versions_from_file(csv_filename, columns):
...
+                continue
+            ref, txt = r[0], r[col]
+            if not txt:
+                continue
+            book_maps[Ref(ref).index.title][ref] = txt
+        for idx_title, text_map in book_maps.items():
+            jobs.append((idx_title, vt, lang, src, notes, text_map))
+    else:
+        idx_title = rows[0][columns[0]]
+        for col in columns:
```
YishaiGlasner: i know this loop was before, but i can't see why it's needed. also now it adds a job for any column which seems like a mistake.
i think we can do
```
else:
  col = columns[-1]
  idx_title, vt, lang, src, notes = [rows[i][col] for i in range(5)]
  text_map = {r[0]: r[col] for r in rows[5:] if len(r) > col and r[col]}
  jobs.append((idx_title, vt, lang, src, notes, text_map))
```

sefaria/export.py
```diff
@@ -720,42 +720,56 @@ def import_versions_from_file(csv_filename, columns):
...
+            vt, lang, src, notes = [rows[i][col] for i in range(1, 5)]
+            text_map = {r[0]: r[col] for r in rows[5:] if len(r) > col and r[col]}
```
YishaiGlasner: why this if?
what's wrong with `text_map = {r[0]: r[col] for r in rows[5:]}`?
this again will prevent removing text.
by the way, all this code seems to assume that the csv can have more than 2 columns and just the first and last are intersting. i think that no one uses it with more than 2.

YishaiGlasner (approved):

PR #2585, opened 2025-07-24
First version of caching method for Strapi API calls

relyks:
## Description

## Code Changes

## Notes

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -1196,6 +1196,120 @@ def index_sheets_by_timestamp(request):
     response_str = search_index_sheets_by_timestamp(timestamp)
     return jsonResponse({"success": response_str})
 
+@csrf_exempt
+def strapi_graphql_cache(request):
+    """
+    Cache mediator for Strapi GraphQL queries with date parameters.
+    POST request with start_date, end_date params and GraphQL query in body.
+    """
+    if request.method != 'POST':
+        return jsonResponse({"error": "Only POST method supported"}, status=405)
+    
+    try:
+        import json
+        from datetime import datetime
+        from sefaria.system.cache import get_cache_elem, set_cache_elem
+        
+        # Parse request parameters
+        start_date = request.POST.get('start_date')
+        end_date = request.POST.get('end_date')
+        
+        if not start_date or not end_date:
+            return jsonResponse({"error": "start_date and end_date parameters required"}, status=400)
+        
+        # Validate date format (YYYY-MM-DD)
+        try:
+            datetime.strptime(start_date, '%Y-%m-%d')
+            datetime.strptime(end_date, '%Y-%m-%d')
+        except ValueError:
+            return jsonResponse({"error": "Dates must be in YYYY-MM-DD format"}, status=400)
+        
+        # Get GraphQL query from request body
+        query = request.body.decode('utf-8')
+        if not query:
+            return jsonResponse({"error": "GraphQL query required in request body"}, status=400)
+        
+        # Create cache key from dates only (query structure is static)
+        cache_key = f"strapi_graphql_{start_date}_{end_date}"
+        
+        # Try to get from cache first
+        cached_result = get_cache_elem(cache_key, cache_type='default')
+        if cached_result:
+            return HttpResponse(cached_result, content_type='application/json')
+        
+        # If not in cache, fetch from Strapi
+        import requests
+        from django.conf import settings
+        
+        strapi_url = getattr(settings, 'STRAPI_INSTANCE', None)
+        if not strapi_url:
+            return jsonResponse({"error": "STRAPI_INSTANCE not configured"}, status=500)
+        
+        # Make request to Strapi GraphQL endpoint
+        response = requests.post(
+            f"{strapi_url}/graphql",
+            json={"query": query},
+            headers={"Content-Type": "application/json"},
+            timeout=30
+        )
+        
+        if response.status_code != 200:
+            return jsonResponse({"error": f"Strapi request failed with status {response.status_code}"}, status=500)
+        
+        result_json = response.text
+        
+        # Cache the result for 7 days (604800 seconds) - will be invalidated by webhook
+        set_cache_elem(cache_key, result_json, timeout=604800, cache_type='default')
+        
+        return HttpResponse(result_json, content_type='application/json')
+        
+    except Exception as e:
+        logger.error(f"Error in strapi_graphql_cache: {str(e)}")
+        return jsonResponse({"error": "Internal server error"}, status=500)
+
+
+@csrf_exempt
+@webhook_auth_or_staff_required
+def strapi_cache_invalidate(request):
+    """
+    Webhook endpoint to invalidate Strapi GraphQL cache when content changes.
+    Can be called by Strapi webhook or staff members.
+    """
+    if request.method != 'POST':
+        return jsonResponse({"error": "Only POST method supported"}, status=405)
+    
+    try:
+        from sefaria.system.cache import delete_cache_elem
+        
+        # Get cache instance to search for keys
+        from sefaria.system.cache import get_cache_factory
+        cache_instance = get_cache_factory('default')
+        
+        # For MongoDB cache backend, invalidate all strapi_graphql cache entries
+        try:
+            if hasattr(cache_instance, '_get_collection'):
+                coll = cache_instance._get_collection()
+                # Find and delete all cache entries that start with 'strapi_graphql'
+                result = coll.delete_many({"key": {"$regex": "^.*strapi_graphql.*"}})
+                deleted_count = result.deleted_count
+                
+                logger.info(f"Invalidated {deleted_count} Strapi GraphQL cache entries")
+                return jsonResponse({"success": f"Invalidated {deleted_count} cache entries"})
+            else:
+                # Fallback for other cache backends - attempt pattern-based deletion
+                logger.info("Attempting cache invalidation for non-MongoDB backend")
+                return jsonResponse({"success": "Cache invalidation completed"})
+                
+        except Exception as cache_error:
+            logger.error(f"Error during cache invalidation: {str(cache_error)}")
+            # Don't fail the webhook if cache invalidation has issues
+            return jsonResponse({"success": "Cache invalidation attempted", "warning": str(cache_error)})
+            
+    except Exception as e:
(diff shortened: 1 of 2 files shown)


sefaria/system/tests/strapi_cache_test.py
```diff
@@ -0,0 +1,383 @@
...
+def pytest_configure(config):
+    validate_redis_cache_backend()
+
+def validate_redis_cache_backend():
+    # Validate that the Django cache backend is using Redis
+    # Fail all tests if Redis backend is not configured
+    from django.core.cache import caches
+    
+    default_cache = caches['default']
+    cache_backend = default_cache.__class__.__module__ + '.' + default_cache.__class__.__name__
```
akiva10b: interesting method, clever for testing. An easier and more reliable way to do this I think would be to import the `CASHES` object from settings and look at the `default` `backend` keys

```
CACHES = {
    "shared": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://127.0.0.1:6379/1", #The URI used to look like this "127.0.0.1:6379:0"
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient",
            #"SERIALIZER": "django_redis.serializers.json.JSONSerializer", #this is the default, we override it to ensure_ascii=False
            "SERIALIZER": "sefaria.system.serializers.JSONSerializer",
        },
        "TIMEOUT": None,
    },
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://127.0.0.1:6379/0", #The URI used to look like this "127.0.0.1:6379:0"
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient",
            #"PASSWORD": "secretpassword", # Optional
        },
        "TIMEOUT": 60 * 60 * 24 * 30,
    },
}
```

sefaria/views.py
```diff
@@ -1196,6 +1197,182 @@ def index_sheets_by_timestamp(request):
...
+        request: Django HttpRequest object
+
+    Returns:
+        HttpResponse: JSON response with GraphQL data or error message
+    """
+    if request.method != "POST":
+        return jsonResponse({"error": "Only POST method supported"}, status=405)
+
+    try:
+        from datetime import datetime
```
akiva10b: datetime is already imported at the top - also move the other imports to the top

sefaria/views.py
```diff
@@ -1196,6 +1197,182 @@ def index_sheets_by_timestamp(request):
...
+                {"error": "end_date parameter must be a single value"}, status=400
+            )
+
+        # They are strings!
+        start_date: str = start_date_raw
+        end_date: str = end_date_raw
+
+        # Validate date format (YYYY-MM-DD)
+        try:
+            datetime.strptime(start_date, "%Y-%m-%d")
```
akiva10b: great! another validation is often "start date before end date"

sefaria/views.py
```diff
@@ -1196,6 +1197,182 @@ def index_sheets_by_timestamp(request):
...
+
+    Returns:
+        HttpResponse: JSON response with success/error message and cache invalidation count
+    """
+    if request.method != "POST":
+        return jsonResponse({"error": "Only POST method supported"}, status=405)
+
+    try:
+        from sefaria.system.cache import get_cache_factory
+        from typing import Any
```
akiva10b: please move to top - also "Any" might be more appropriate for the Union you had earlier for the raw_date vars

sefaria/views.py
```diff
@@ -1196,6 +1197,182 @@ def index_sheets_by_timestamp(request):
...
+
+        # Get cache instance
+        cache_instance: Any = get_cache_factory("default")
+
+        # For Redis cache backend using django-redis
+        try:
+            if hasattr(cache_instance, "delete_pattern"):
```
akiva10b: based on this code I understand the following: we need to hold multiple versions of Strapi in cache because of timezones, different timezones will request different dateranges, the way we approach timezones is by holding multiple date ranges, correct?

sefaria/views.py
```diff
@@ -1196,6 +1197,182 @@ def index_sheets_by_timestamp(request):
...
+                return jsonResponse(
+                    {"success": f"Invalidated {deleted_count} cache entries"}
+                )
+
+            else:
+                # Fallback for other cache backends (FileBasedCache, etc.)
+                cache_backend = cache_instance.__class__.__name__
+                logger.info(f"Using fallback invalidation for {cache_backend}")
+
+                if hasattr(cache_instance, "clear"):
```
akiva10b: everyrthing here which is trying to clean the cache while managing different caches should be a utility function in the cache.py file. (excellent work!)

static/js/context.js
```diff
@@ -178,17 +178,17 @@ function StrapiDataProvider({ children }) {
           }
         }
         `;
-        const result = fetch(STRAPI_INSTANCE + "/graphql", {
+        // Use the new cache endpoint instead of calling Strapi directly to minimize API calls
+        const url = new URL("/api/strapi/graphql-cache", window.location.origin);
+        url.searchParams.append('start_date', startDate.split('T')[0]); // Only use date part
```
akiva10b, line 183: here we are losing the timezone when splitting the string. did you verify that our servers which are on UTC-0 will return the correct responses based on the date they recieve?

sefaria/system/tests/strapi_cache_test.py
relyks: This is true, but I thought using reflection would reflect (no pun intended) what the `CACHES` variable is at runtime. Given that the codebase doesn't override or modify the cache backend based on conditions, it's static after startup. Your suggestion is simpler :)

sefaria/views.py
relyks: Done

sefaria/views.py
relyks: Added it

sefaria/views.py
relyks: Done, but I rather keep the type for the raw_date vars because that is the actual type and I was able to resolve an issue with the code with `pyright` because I didn't have it correctly typed

sefaria/views.py
relyks: The original goal was for the client to cache banners/modals/sidebar ads that could show in the future, so that the client wouldn't have to make more than one API call after multiple page loads until after a period of time. The client could also store multiple content objects from Strapi. While those content objects are stored, they could, in theory, be changed live during usage based on context (time or an event) on the site without making any other API calls. 

Content created by MarCom or Dev is typically made before and published over a time period in the future. Multiple content objects are generally made and then added to Strapi for a campaign, but for sequential periods. For example, one banner ends right before midnight, while another banner starts right after midnight, but the previous banner was part of a campaign that was supposed to last 10 days. 

The date range enables the retrieval of already existing and future campaigns, while also having the side effect of handling time zones. If one client is in a different timezone on a different day, its date range will be the same as others, but adjusted by 1 day. The client will choose the right content based on its own time. In theory, there should only be 3 different cache entries at a time on the backend, because there are at most 3 different dates on the planet at any given time.

Let me know if anything doesn't make sense?

sefaria/views.py
relyks: Done, I moved it over

static/js/context.js
relyks: Yeah, I did verify that. From what I understand, context.js (the main part of the Strapi code) will be rendered on the server and client. The setup with hydration will happen with the context providers, but the key part in the useEffect hook will only run in the browser, including the API calls to Strapi. The Strapi API calls are never made while rendering in Node. Times are in UTC everywhere, and the user's timezone is only used on the client side to determine if a specific content object should be shown. Even though MarCom and Dev are only using English content right now, there can also be separate datetimes set for different locales

sefaria/system/cache.py
```diff
@@ -168,4 +202,95 @@ def reset_all(self):
...
+                    "backend": cache_backend,
+                    "count": deleted_count,
+                    "message": f"Invalidated {deleted_count} cache entries matching pattern '{pattern}'"
+                }
+
+            else:
+                # Fallback for other cache backends (FileBasedCache, etc.)
+                logger.info(f"Using fallback invalidation for {cache_backend}")
+
+                if hasattr(cache_instance, "clear"):
```
akiva10b: I am worried about this, killing cache in the middle for all SHARED_DATA_CACHE_ALIAS can be bad, can we find a pattern and delete it, or alternatively have an overt check that this isnt django cache

akiva10b: my worry is that if a syntax change happens in a django upgrade we can be screwed

sefaria/views.py
```diff
@@ -1196,6 +1198,149 @@ def index_sheets_by_timestamp(request):
...
+            return jsonResponse(
+                {"error": "GraphQL query required in request body"}, status=400
+            )
+
+        # Create cache key from the specified dates. The query structure will be static while using the dates in its body
+        cache_key: str = f"strapi_graphql_{start_date}_{end_date}"
+
+        # Try to get from cache first
+        # There should be at most 3 keys (date ranges in the cache) at the same time based on frontend usage
+        cached_result: Optional[str] = get_cache_elem(cache_key, cache_type="default")
```
akiva10b, line 1394: i believe there is a var for this?

static/js/context.js
akiva10b: so why do we need multiple cache keys? is it not always the same data no matter the date? And the client is the one that determines what to show?

relyks: It's not always same data. Different data will be returned from Strapi for different date ranges. That's why multiple cache keys are necessary. The client just chooses what to show within the date range.

sefaria/system/cache.py
relyks: Yeah, you're totally right. Good catch. Even though the function isn't intended for use in other scenarios besides mine right now, it'd be bad if the cache_type is changed to the shared kind during a future call. Clearing all of the cache when you can't match a pattern isn't a good idea, because if a pattern deletion fails, other data stored in the cache unrelated to what's desired will also be removed. In local development, it will even cause issues if redis isn't properly set up. I'm going to just return an error message instead

sefaria/views.py
relyks: There isn't? Unless do you want me to refactor so there's a single variable to represent `cached_result` and `result_json` and have a single return statement? There'd be less return statements but then the code will be less clear

PR #2586, opened 2025-07-24
fix(topics): Return null on link to sheet search if no title exists

yitzhakc:
## Description
Bugfix to prevent error on rendering the link to sheet search from a topic page where there is no topic title

## Code Changes
Added a validation conditional to `LinkToSheetsSearchComponent` in `TopicPage.jsx`


static/js/TopicPage.jsx
```diff
@@ -932,7 +932,10 @@ const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, set
 
 
   const LinkToSheetsSearchComponent = () => {
-
+    if (!topicTitle) {
+      // If topicTitle is not set, we cannot generate the search URLs
+      return null;
+    }
```
relyks, line 938: I'm pretty sure this wouldn't apply. I tried looking through the rest of the code. Both languages should be defined in the `topicTitle` object, but in case they are not, the conditional should be the following to be more defensive in case neither are defined:
```
if (!topicTitle || (!topicTitle.en && !topicTitle.he)) {
  return null;
}
```

static/js/TopicPage.jsx
```diff
@@ -932,7 +932,10 @@ const TopicSideColumn = ({ slug, links, clearAndSetTopic, parashaData, tref, set
...
     let searchUrlEn = `/search?q=${topicTitle.en}&tab=sheet&tvar=1&tsort=relevance&stopics_enFilters=${topicTitle.en}&svar=1&ssort=relevance`;
     let searchUrlHe = `/search?q=${topicTitle.he}&tab=sheet&tvar=1&tsort=relevance&stopics_heFilters=${topicTitle.he}&svar=1&ssort=relevance`;
```
relyks, line 940: Other places in the code use this similar fallback technique. If one of the other languages isn't defined, the other is used.
```
const enTitle = topicTitle.en || topicTitle.he || '';
const heTitle = topicTitle.he || topicTitle.en || '';
```
and then use those variables as part of the path in search query parameter. I don't think this is necessary though.

PR #2595, opened 2025-07-29
fix(local-settings): update SESSION_ENGINE to use cache_db 

yodem:
fix for - https://app.shortcut.com/sefaria/story/18766/web-users-are-logged-out-unexpectedly-and-have-to-login-again

Changed the session engine to be cached_db for persistent storage of cache data

### DIFF
--- a/helm-chart/sefaria/templates/configmap/local-settings-file.yaml
+++ b/helm-chart/sefaria/templates/configmap/local-settings-file.yaml
@@ -125,7 +125,7 @@ data:
     # This is the field that is actually used
     GOOGLE_OAUTH2_CLIENT_SECRET_FILEPATH = "/client-secret/client_secrets.json"
 
-    SESSION_ENGINE = "django.contrib.sessions.backends.cache"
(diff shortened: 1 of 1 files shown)


helm-chart/sefaria/templates/configmap/local-settings-file.yaml
```diff
@@ -125,7 +125,7 @@ data:
     # This is the field that is actually used
     GOOGLE_OAUTH2_CLIENT_SECRET_FILEPATH = "/client-secret/client_secrets.json"
 
-    SESSION_ENGINE = "django.contrib.sessions.backends.cache"
+    SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
```
akiva10b, line 128: this looks like the correct change - the way to check it (simply) would be to check if: a - the pg database has a record of the session, and b - you delete a web pod and redis pod, and the session still exists

yodem: It worked

PR #2596, opened 2025-07-29
Chore/sc 22611/change help links from sheets to zendesk

schreibersefaria:
Replaced all the old links of help and getting started with new links I was given


schreibersefaria:
> @schreibersefaria I think the main lesson here is we should convert the magic strings to vars. i think we have constants.js or something like this for them.

Do you we an equivalent in python?
And do you think in the tests I should also extract the path part of the consts or a duplication makes sense for tests?

PR #2604, opened 2025-08-04
Feature/sc 35573/achieve adherence to accessibility wcag 2 2

schreibersefaria:
PR for Cauldron


static/css/color-palette.css
```diff
@@ -31,5 +32,7 @@
   --reference-orange: #D4896C;
   --miscelaneous-green: #3B5849;
   --selected-option: #000000;
-  --select-blue: #0B71E7;
+  --select-blue: #0056b3;
```
schreibersefaria: Double check why this wasn't reverted

static/css/s2.css
```diff
@@ -15,6 +15,26 @@ body.dark {
 body:not(.user-is-tabbing) *:focus {
   outline: none;
 }
+
+/* Skip Link for Accessibility - Override Button positioning to hide off-screen by default */
+.sefaria-common-button.skip-link {
```
stevekaplan123: "!important" is considered bad in CSS.  it's really ideal to use it as rarely as possible

static/css/s2.css
```diff
@@ -5648,6 +5672,14 @@ h1.topic-landing-header {
 }
 .readerNavMenuCloseButton {
   font-family: var(--english-serif-font-family);
+  color: var(--darker-grey); /* Ensures 3:1 contrast ratio for accessibility */
+}
+
+/* Replacement for obsolete <big> tag - WCAG compliance */
+.big-text,
```
stevekaplan123, line 5680: What is the obsolete big tag?

static/css/s2.css
```diff
@@ -5648,6 +5672,14 @@ h1.topic-landing-header {
 }
 .readerNavMenuCloseButton {
   font-family: var(--english-serif-font-family);
+  color: var(--darker-grey); /* Ensures 3:1 contrast ratio for accessibility */
```
stevekaplan123: I think we should have a more centralized place to explain something like this.  Maybe at the top of the file?  I'm not sure where

static/css/common-components.css
```diff
@@ -18,11 +19,14 @@ button.sefaria-common-button {
     appearance: none; /* Remove default appearance */
 }
 
-button.sefaria-common-button:focus {
-    outline: none; /* Remove focus outline */
+body.user-is-tabbing button.sefaria-common-button:focus,
+body.user-is-tabbing a.sefaria-common-button:focus {
```
stevekaplan123, line 23: where does this "user-is-tabbing" class come from?  I'm confused

reader/templatetags/sefaria_tags.py
```diff
@@ -550,3 +550,12 @@ def subdomain_url(context, url_name, *args, **kwargs):
         return reverse(url_name, args=args, kwargs=kwargs)
     except NoReverseMatch:
         return '#'
+
+
+@register.simple_tag(takes_context=True)
+def interface_lang_code(context):
```
stevekaplan123: (1) Why are we setting the language like this to begin with?  (2) And why do so in so many pages?  Why isn't it sufficient to just do it in base.html?

static/css/s2.css
```diff
@@ -4641,7 +4665,7 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
 }
 .mobileSearchFiltersHeader .readerNavMenuCloseButton {
   font-size: 45px;
-  color: var(--medium-grey);
+  color: var(--darker-grey); /* Override with accessible color for 3:1 contrast */
```
stevekaplan123, line 4668: Again, for me personally I don't know if this is the best place for comments like this.  Curious what others think

static/css/s2.css
```diff
@@ -8697,8 +8730,9 @@ But not to use a display block directive that might break continuous mode for ot
 .emptyDesktop {
   margin-top: 200px;
   text-align: center;
-  color: #999;
+  color: var(--medium-grey);
 }
+/* DEPRECATED: Use Button component from static/js/common/Button.jsx instead */
```
stevekaplan123, line 8736: Can we remove the deprecated code?  Or create a story to remove it eventually?

static/js/BookPage.jsx
```diff
@@ -169,11 +169,19 @@ class BookPage extends Component {
...
+          href={"/" + Sefaria.normRef(Sefaria.lastPlaceForText(title).ref)}
+          onKeyDown={(e) => { if (e.key === ' ') { e.preventDefault(); window.location.href = e.target.href; } }}
+        >
           <InterfaceText>Continue Reading</InterfaceText>
         </a>
         :
-        <a className="button small readButton" href={"/" + Sefaria.normRef(this.state.indexDetails["firstSectionRef"])}>
+        <a 
+          className="button small readButton" 
+          href={"/" + Sefaria.normRef(this.state.indexDetails["firstSectionRef"])}
```
stevekaplan123, line 183: The onKeyDown function is repetitive and shouldn't be handled as a lambda function but as a const function within the component (or if it occurs elsewhere, in util.js or sefaria.js)

static/js/AddToSourceSheet.jsx
```diff
@@ -312,16 +351,99 @@ class AddToSourceSheetBox extends Component {
...
+                    }); 
+                  } 
+                }
+                if (e.key === 'Enter' || e.key === ' ') { 
+                  e.preventDefault(); 
+                  if (sheets && sheets[this.state.focusedSheetIndex]) { 
+                    this.selectSheet(sheets[this.state.focusedSheetIndex]); 
+                  } 
+                }
+              }}
```
stevekaplan123: A lot of these checks for various keys look like things we'll be doing many times in other places right?  If so, the logic can be moved out of here and handled elsewhere.

static/js/common/Button.jsx
```diff
@@ -7,22 +11,161 @@
...
+        {...(!!targetModule ? { 'data-target-module': targetModule } : {})}
+        {...(!!activeModule ? { 'data-active-module': activeModule } : {})}
+      >
+        {icon && (<img src={`/static/icons/${icon}.svg`} className="button-icon" alt={alt} />)}
+        {children}
+      </a>
+    );
+  }
+
+  const handleKeyDown = (e) => {
```
stevekaplan123: There shouldn't be two functions in one file called handleKeyDown

static/js/common/Button.jsx
```diff
@@ -7,22 +11,161 @@
...
       {icon && (<img src={`/static/icons/${icon}.svg`} className="button-icon" alt={alt} />)}
       {children}
     </button>
   );
 };
 
-export default Button;
\ No newline at end of file
+// Custom PropTypes validator for targetModule
+const targetModuleValidator = (props, propName, componentName) => {
```
stevekaplan123: What is the targetModuleValidator for?  What problem does it solve?

static/js/common/Button.jsx
```diff
@@ -7,22 +11,161 @@
...
+ * // For elements that should trigger their own click behavior:
+ * const handleKeyDown = handleKeyboardClick();
+ * <button onKeyDown={handleKeyDown}>Click me</button>
+ *
+ * @example
+ * // For elements with custom click handlers:
+ * const handleKeyDown = handleKeyboardClick(myCustomClickHandler);
+ * <div onKeyDown={handleKeyDown}>Custom element</div>
+ */
+const handleKeyboardClick = (onClick) => {
```
stevekaplan123: This is very confusing -- handleKeyboardClick + 2 handleKeyDown functions

static/js/common/DropdownMenu.jsx
```diff
@@ -122,12 +136,65 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
+    };
+
     return (
         <div className={positioningClass} ref={wrapperRef}>
-           <div className="dropdownLinks-button" onClick={handleButtonClick}>
-              {buttonComponent}
+           <div
+             className="dropdownLinks-button"
+           >
+              {React.cloneElement(buttonComponent, {
```
stevekaplan123, line 169: I've never seen cloneElement in our codebase.  Why are we using it?  What's it do?

static/js/sheets.js
```diff
@@ -805,19 +805,9 @@ $(function() {
 
 	// ---------- Save Sheet --------------
```
stevekaplan123, line 806: Is sheets.js used in our code?  What does it do?

static/js/ConnectionsPanel.jsx
```diff
@@ -924,8 +925,8 @@ class WebPagesList extends Component {
       });
       sites = Object.values(sites).sort(this.webSitesSort);
       content = sites.map(site => {
-        return (<div className="website" role="button" tabindex="0" onKeyUp={(event) => event.key==='Enter' && this.setFilter(site.name)} onClick={() => this.setFilter(site.name)} key={site.name}>
-          <img className="icon" src={site.faviconUrl} />
+        return (<div className="website" role="button" tabindex="0" onKeyDown={handleKeyboardClick(() => this.setFilter(site.name))} onClick={() => this.setFilter(site.name)} key={site.name}>
```
stevekaplan123: Why do you pass the function like this and call it at the same time?  Normally in React I see `onKeyDown={() => handleKeyBoardClick(...}`

static/js/Header.jsx
```diff
@@ -663,20 +696,20 @@ const HelpButton = () => {
...
-        <InterfaceText>Sign Up</InterfaceText>
-      </a>
+    <Button href="/register" targetModule={Sefaria.LIBRARY_MODULE}>
+      <InterfaceText>Sign Up</InterfaceText>
     </Button>
   )
 }
 
 const CreateButton = () => {
+  const handleCreate = () => {
```
stevekaplan123: handleCreate doesn't get called anywhere?

static/js/ReaderApp.jsx
```diff
@@ -2358,11 +2356,14 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       <StrapiDataProvider>
         <AdContext.Provider value={this.getUserContext()}>
           <div id="readerAppWrap">
+            <Button href="#main" className="skip-link">{Sefaria._("Skip to main content")}</Button>
             <InterruptingMessage />
             <Banner onClose={this.setContainerMode} />
             <div className={classes} onClick={this.handleInAppLinkClick}>
               {header}
-              {panels}
+              <main id="main" role="main">
```
stevekaplan123, line 2362: What's the main tag for?

static/js/Misc.jsx
```diff
@@ -2443,28 +2525,92 @@ class Dropdown extends Component {
...
+          <div
+            className={`dropdownMain noselect${this.state.selected ? " selected":""}`}
+            onClick={this.toggle}
+            role="button"
+            tabIndex="0"
+            aria-haspopup="listbox"
+            aria-expanded={this.state.optionsOpen}
+            aria-controls={`${this.props.name}-listbox`}
+            ref={(el) => { this.triggerRef = el; }}
+            onKeyDown={(e) => {
```
stevekaplan123: This logic should be moved out to actual functions not lambda functions

static/css/s2.css
schreibersefaria: Removed. I think the top of this file is as good as not there.
But hopefully the deciders of colors will start taking accessibility into account

static/css/s2.css
schreibersefaria: It's an HTML tag that we were using:
"The <big> tag in HTML was an inline element used to make text appear one font size larger than the surrounding text."

static/css/s2.css
schreibersefaria: We have it as a task.
Or more accuratly, we have a task to create a task because replacing all of buttons is a big many tasks.
I think inline comments like this are very helpful for agents so they don't keep using old styling we are hoping to deprecate.

static/css/common-components.css
schreibersefaria: It's in static/js/Header.jsx lines 173-178.
I know see it's Skylers recent code.

reader/templatetags/sefaria_tags.py
schreibersefaria: 2 - It's set in pages that don't extend base - except in one case where it was redundant and now removed.
1 - Good question, I replaced the use of this function with {{ request.LANGUAGE_CODE }}

schreibersefaria: Unless you mean why are we setting `lang` in all the pages?
That is a WCAG requirment

static/css/s2.css
stevekaplan123: Why did you remove it if you think the comment is necessary?  Maybe you're right.  I'm not sure how comments should work in CSS.

static/js/AddToSourceSheet.jsx
schreibersefaria: I refactored it out to utils. The dropdown key handeling was in two places.

schreibersefaria: There are similar patterns for handeling non <Button> clickable components. I'll refactor them as well. 
Also, form submitions

static/js/common/Button.jsx
schreibersefaria: Woops!

static/css/s2.css
schreibersefaria: I don't think it's necessary, I made a hundred such changes and won't comment them all.
My note is that we should somewhere have a general comment saying we should adhere to WCAG color contrast. I'm not sure where that should be, but likely not in the code at all.

static/js/common/Button.jsx
schreibersefaria: Removed one and renamed the other

schreibersefaria: Removed the other one as well as part of another refactor

static/js/common/Button.jsx
schreibersefaria: Added a comment. I want to make sure the targetModule isn't used when the button is used for <button>. So it's only used for <a>.
Do you think it's overkill?

static/js/common/DropdownMenu.jsx
schreibersefaria: It's needed to add accessibility logic to the button component that is sent by the parent component.
We could refactor to force parent components to send a specific type of component or do maybe do a complex validation, but I think this is the simplest implementation.
I added a clear comment

static/js/ConnectionsPanel.jsx
schreibersefaria: It seams this was solved as part of another fix

static/js/sheets.js
schreibersefaria: It's of the old editor.
I've reverted any change to it

static/js/Header.jsx
schreibersefaria: Thanks, removed

static/js/ReaderApp.jsx
schreibersefaria: It's a marker for accessibility.
Screen users can use it to navigate and it's used in the 'skip to main content' button

templates/static/design-system.html
```diff
@@ -1,10 +1,6 @@
-{% load sefaria_tags %}
```
stevekaplan123: what is design-system and why remove sefaria_tags?

static/js/common/DropdownMenu.jsx
```diff
@@ -122,12 +136,76 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
         };
     }, []);
 
+    useEffect(() => {
```
stevekaplan123, line 139: Try to re-use utility functions as much as possible or at least clean it up

static/js/common/RadioButton.jsx
```diff
@@ -2,21 +2,62 @@ import React from "react";
...
 
-export default function RadioButton  ({isActive, onClick, value, name, label, id, ...rest}) {
+export default function RadioButton  ({isActive, onClick, value, name, label, id, onKeyDown, ...rest}) {
+    const handleChange = (e) => {
+        e.stopPropagation();
+        onClick();
+    };
+
+    const handleKeyDown = (e) => {
+        // Handle arrow keys for radio group navigation
```
stevekaplan123: Is it possible to re-use utility functions (or create new ones) or clean this up?

static/js/Misc.jsx
```diff
@@ -440,21 +436,59 @@ class TabView extends Component {
...
+      <div 
+        className={classNames({active: isActive, justifyright: tab.justifyright})} 
+        key={tab.id} 
+        data-tab-index={index} 
+        role="tab"
+        tabIndex={isActive ? "0" : "-1"}
+        aria-selected={isActive}
+        onClick={(e) => {this.onClickTab(e, tab.clickTabOverride)}}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' || e.key === ' ') {
```
stevekaplan123: ArrowLeft/ArrowRight logic again.  Maybe we can re-use some of this in utility functions or at least clean it up

static/js/AddToSourceSheet.jsx
```diff
@@ -312,16 +351,61 @@ class AddToSourceSheetBox extends Component {
...
+              tabIndex="0"
+              role="listbox"
+              aria-label="Your sheets list"
+              id="user-sheets-listbox"
+              ref={(el) => { this.listboxRef = el; }}
+              aria-activedescendant={`user-sheet-option-${Math.min(Math.max(this.state.focusedSheetIndex, 0), (sheets ? sheets.length - 1 : 0))}`}
+              onKeyDown={Util.handleListboxKeyDown({
+                currentIndex: this.state.focusedSheetIndex,
+                maxIndex: sheets ? sheets.length - 1 : 0,
+                onNavigate: (newIndex) => {
```
stevekaplan123: onNavigate and onSelect are hard to understand.

templates/static/design-system.html
schreibersefaria: I don't know what it is but it inherits from base so they aren't needed

static/js/common/DropdownMenu.jsx
schreibersefaria: Fixed, but I still didn't go over the implementation so wait with checking

static/js/AddToSourceSheet.jsx
schreibersefaria: Here too, I still need to review the code

schreibersefaria:
@stevekaplan123 After more analysis I think the cloneElement is a decent option.
The best alternative I can think about is to create a new component and force all calls to DropdownMenu to send this component.
We can also send a function to buttonComponent that creates a component and receives `...props`
```
// Before:
buttonComponent={<img src="/static/icons/ellipses.svg" alt="Options"/>}

// After:
buttonComponent={(props) => (
  <button {...props} className="icon-button">
    <img src="/static/icons/ellipses.svg" alt="Options"/>
  </button>
)}
```

static/js/common/DropdownMenu.jsx
schreibersefaria: Looks solid enough

static/js/common/RadioButton.jsx
schreibersefaria: Note sure when you wrote this but it's now using `Util.handleRadioKeyDown`

static/js/AddToSourceSheet.jsx
schreibersefaria: looks good

static/js/Misc.jsx
schreibersefaria: refactored to use an existing util

schreibersefaria:
@stevekaplan123 I believe I went through and fixed all your comments.

static/js/CollectionPage.jsx
```diff
@@ -350,7 +350,7 @@ const CollectionAbout = ({collection, isAdmin, toggleLanguage}) => (
 
 
 const EditCollectionButton = ({slug}) => (
-  <a className="button small white" href={`/sheets/collections/${slug}/settings`} data-target-module={Sefaria.SHEETS_MODULE}>
+  <a className="button small white" href={`/sheets/collections/${slug}/settings`} data-target-module={Sefaria.activeModule}>
```
stevekaplan123, line 333: Why did you change this?  Collections only exist in the sheets/voices module

static/js/Misc.jsx
```diff
@@ -447,28 +447,19 @@ class TabView extends Component {
...
-            // This ensures the DOM has been updated before attempting to focus
-            requestAnimationFrame(() => {
-              const newTabElement = document.querySelector(`[data-tab-index="${newIndex}"]`);
-              if (newTabElement) {
-                newTabElement.focus();
+          Util.handleTabKeyDown(e, {
+            currentIndex: index,
+            tabCount: this.props.tabs.length,
+            onNavigate: (newIndex) => {
+              const newTab = this.props.tabs[newIndex];
```
stevekaplan123: you do this a few times where you define the function when passing it to a util function.  i think this makes it harder to debug.  it should just be a function onNavigate in the component, so that you can say "onNavigate: onNavigate"

stevekaplan123 (changes requested):
Looks like a lot more DRY and modular, but I dont like defining functions inside function calls, which you did a few times.  Other than that, this looks great.

static/js/Misc.jsx
schreibersefaria: refactored

static/js/ReaderDisplayOptionsMenu.jsx
```diff
@@ -12,6 +12,14 @@ const ReaderDisplayOptionsMenu = () => {
     const {language, setOption, panelMode, aliyotShowStatus, textsData, vowelsAndCantillationState, punctuationState, width, panelPosition} = useContext(ReaderPanelContext);
     const menuRef = useRef(null);
 
+    const onClose = React.useCallback(() => {
```
stevekaplan123: Why useCallback?

schreibersefaria: No good reason.
I guess I wasn't familiar enough with the syntax to notice and correct AI.
Thanks

PR #2606, opened 2025-08-04
Feature/sc 35390/rebased render navigational links per module

yitzhakc:
## Description

## Code Changes

## Notes


static/js/TopicsPage.jsx
```diff
@@ -18,8 +18,7 @@ const TopicsPage = ({setNavTopic, multiPanel}) => {
   let categoryListings = Sefaria.topic_toc.map((topic, i) => <TopicTOCCard topic={topic} setNavTopic={setNavTopic} key={i}/>);
   const letter = Sefaria.interfaceLang === "hebrew" ? "א" : "a";
   const description = {"en": "Browse or search our complete list of topics.", "he": Sefaria._("Browse or search our complete list of topics.")};
-  const prefix = Sefaria.activeModule === "sheets" ? "/sheets/" : "/"
-  const topicsA_Z = <Card cardTitleHref={`${prefix}topics/all/${letter}`}
+  const topicsA_Z = <Card cardTitleHref={`/topics/all/${letter}`}
```
yitzhakc, line 21: revert this change

reader/views.py
```diff
@@ -218,6 +218,7 @@ def base_props(request):
         "multiPanel":  not request.user_agent.is_mobile and not "mobile" in request.GET,
         "initialPath": request.get_full_path(),
         "interfaceLang": request.interfaceLang,
+        "domainModules": json.loads(DOMAIN_MODULES) if isinstance(DOMAIN_MODULES, str) else DOMAIN_MODULES, # if DOMAIN_MODULES is a string, it is a json string, so we need to load it
```
akiva10b: can you add to the docstring the expected format

sefaria/local_settings_ci.py
```diff
@@ -22,7 +22,10 @@
     "http://hebrew.example.org": "hebrew",
     "http://english.example.org": "english",
 }
-
+DOMAIN_MODULES = {
+    "library": "http://localhost:8000",
+    "sheets": "http://localhost:8000/sheets",
+}
```
akiva10b, line 28: 👍

sefaria/sheets.py
```diff
@@ -655,7 +655,7 @@ def test():
 	for s in ss:
 		lang = get_sheet_language(s)
 		if lang == "some hebrew":
-			print("{}\thttps://www.sefaria.org/sheets/{}".format(strip_tags(s["title"]).replace("\n", ""), s["id"]))
+			print("{}\thttps://sheets.sefaria.org/sheets/{}".format(strip_tags(s["title"]).replace("\n", ""), s["id"]))
```
akiva10b, line 661: so to be clear - here we are maintaining **sheets**.sefaria.org/**sheets**

which would internally map to localhost/sheets/sheets unless otherwise specified - can you verify this or explain in the PR intro what are the different URLs we intend to see

sefaria/urls.py
```diff
@@ -353,20 +353,21 @@
...
-    url(r'^password/reset/confirm/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$', sefaria_views.CustomPasswordResetConfirmView.as_view(), name='password_reset_confirm'),
-    url(r'^password/reset/complete/$', sefaria_views.CustomPasswordResetCompleteView.as_view(), name='password_reset_complete'),
-    url(r'^password/reset/done/$', sefaria_views.CustomPasswordResetDoneView.as_view(), name='password_reset_done'),
-    url(r'^api/register/$', sefaria_views.register_api),
-    url(r'^api/login/$', TokenObtainPairView.as_view(), name='token_obtain_pair'),
-    url(r'^api/login/refresh/$', TokenRefreshView.as_view(), name='token_refresh'),
-    url(r'^api/account/delete$', reader_views.delete_user_account_api),
-]
+# Registration and Login need to support both the main site and the sheets prefix.
+for prefix in ('', 'sheets/'):
```
akiva10b: this should be pulled from either the keys of the variable in settings.py or some other var - but not a magic string as it is a pretty important string. Also an Enum could work, whatever centralizes it

static/js/AddToSourceSheet.jsx
```diff
@@ -321,10 +321,9 @@ class AddToSourceSheetBox extends Component {
             </div>
             <div className="newSheet noselect">
               <input className="newSheetInput noselect" placeholder={Sefaria._("Name New Sheet")}/>
-              <div className="button small noselect" onClick={this.createSheet} >
-                <span className="int-en">Create</span>
-                <span className="int-he">יצירה</span>
-              </div>
+              <Button size="fillwidth" className="small noselect" onClick={this.createSheet} activeModule="sheets">
```
akiva10b: this would be another location would could use a Enum as opposed to magic string

static/js/AddToSourceSheet.jsx
```diff
@@ -369,14 +368,14 @@ class ConfirmAddToSheet extends Component {
     return (<div className="confirmAddToSheet addToSourceSheetBox">
               <div className="message">
                 <span className="int-en">
-                  <a href={sref}>{srefTitles["en"]}</a>
+                  <a href={sref} data-attr-module={!!this.props.nodeRef ? "sheets" : "library"}>{srefTitles["en"]}</a>
```
akiva10b: in all of these locations an Enum is warranted

static/js/EditCollectionPage.jsx
```diff
@@ -119,12 +119,18 @@ class EditCollectionPage extends Component {
...
-          window.location = "/sheets/collections/" + data.collection.slug;
+          window.location = "/sheets/collections/" + data.collection.slug; 
         }
     }.bind(this)).fail(function() {
         alert(Sefaria._("Unfortunately an error occurred saving your collection."));
     });
   }
+  getCancelLink() {
+    if (this.props.initialData) {
+      return `${this.props.domainModules.sheets}/sheets/collections/${this.state.slug}`;
```
akiva10b: I dont think that domainModules should be propped down but also globally imported - I see that later on you use Sefaria.domainModules so why would it be different in the static/js/EditCollectionPage.jsx component?

static/js/ReaderApp.jsx
```diff
@@ -1100,35 +1100,42 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
...
       return;
     }
-    const href = linkTarget.getAttribute('href');
+
+    let href = linkTarget.getAttribute('href');
     if (!href) {
       return;
     }
+    
+    const moduleTarget = linkTarget.getAttribute('data-attr-module');
```
akiva10b: can you add a docstring to this function explaining it

akiva10b (commented):
Overall love the concept of data-attr-module and targetModule, things I would change:
- Use a Enum for ass "sheets" and "library" uses (Enum can be on the Sefaria object), no reason to discover a typo in testing, the IDE catches variable names and will prevent errors + all other reasons why magic strings are bad.
- Move all uses of domainModules to the global Sefaria.domainModules and don't prop them down from parent component

static/js/TopicsPage.jsx
yitzhakc: @stevekaplan123 said he added the logic for modules to the `Card` object itself

sefaria/sheets.py
yitzhakc: Correct.
In cauldron-land/production the nginx maps that internally to `http://web/sheets` (which is why we it I don't think we added `sheets.sefaria.org` to `ALLOWED_HOSTS`, iirc)
Locally, everything will just show up as `http://localhost/sheets`.

I will update the PR description.

static/js/EditCollectionPage.jsx
stevekaplan123: EditCollectionPage is not rendered by ReaderApp so it doesnt have access to the Sefaria module.  It's rendered directly by edit_collection.html (which ultimately uses base.html template, which has access to domainModules because the backend function `base_props` is called when rendering base.html)

sefaria/urls.py
stevekaplan123: In solving this, we should take into account code in middleware.py     

```
MODULE_ROUTES = {
        '/sheets/': 'sheets',
    }
```

static/js/Footer.jsx
```diff
@@ -62,7 +62,7 @@ class Footer extends Component {
                 <Link href="/daf-yomi" en="Daf Yomi" he="דף יומי" />
                 <Link href="/torah-tab" en="Torah Tab" he="תורה טאב" />
                 <Link href="/people" en="Authors" he="מחברים" />
-                <Link href="/sheets/collections" en="Collections" he="אסופות" data-attr-module="sheets" />
+                <Link href="/sheets/collections" en="Collections" he="אסופות" data-target-module={Sefaria.SHEETS_MODULE} />
```
stevekaplan123: Wrong syntax

sefaria/local_settings_ci.py
```diff
@@ -26,6 +26,10 @@
     "library": "http://localhost:8000",
     "sheets": "http://localhost:8000/sheets",
 }
+MODULE_ROUTES = {
```
yitzhakc, line 29: Should this be the default in the template files as well?

static/js/EditCollectionPage.jsx
akiva10b: why can't we import Sefaria and use Sefaria.getModuleURL ?

sefaria/system/middleware.py
```diff
@@ -269,9 +265,9 @@ def __call__(self, request):
             return self.get_response(request)
 
         # Find the matching route prefix
-        for route_prefix, module_name in self.MODULE_ROUTES.items():
+        for module_name, route_prefix in json.loads(MODULE_ROUTES).items():
```
yitzhakc, line 268: I don't see any import of `MODULE_ROUTES`...
Shouldn't we be doing 
`from django.conf import settings`
and then here use `settings.MODULE_ROUTES`? It should already be an object too, no need to do json.loads?

sefaria/urls.py
```diff
@@ -353,20 +353,22 @@
...
-    url(r'^password/reset/confirm/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$', sefaria_views.CustomPasswordResetConfirmView.as_view(), name='password_reset_confirm'),
-    url(r'^password/reset/complete/$', sefaria_views.CustomPasswordResetCompleteView.as_view(), name='password_reset_complete'),
-    url(r'^password/reset/done/$', sefaria_views.CustomPasswordResetDoneView.as_view(), name='password_reset_done'),
-    url(r'^api/register/$', sefaria_views.register_api),
-    url(r'^api/login/$', TokenObtainPairView.as_view(), name='token_obtain_pair'),
-    url(r'^api/login/refresh/$', TokenRefreshView.as_view(), name='token_refresh'),
-    url(r'^api/account/delete$', reader_views.delete_user_account_api),
-]
+# Registration and Login need to support both the main site and the sheets prefix.
+for prefix in json.loads(MODULE_ROUTES).values():
```
yitzhakc, line 357: Isn't `MODULE_ROUTES` already a dictionary?

static/js/common/Button.jsx
```diff
@@ -10,11 +10,11 @@
  * @returns {JSX.Element}
  * @constructor
  */
-const Button = ({ variant = '', size = '', icon, children, onClick, disabled=false, className = '', activeModule }) => {
+const Button = ({ variant = '', size = '', icon, children, onClick, disabled=false, className = '', activeModule=null }) => {
   return (
     <button
       disabled={disabled}
-      {...(activeModule ? { 'data-active-module': activeModule } : {})}
+      {...(!!activeModule ? { 'data-active-module': activeModule } : {})}
```
yitzhakc, line 17: Why this modification?

static/js/common/Card.jsx
```diff
@@ -1,10 +1,13 @@
 import {InterfaceText} from "../Misc";
 import React from "react";
 const Card = ({cardTitle, cardTitleHref, oncardTitleClick, cardText, bottomLinkText, bottomLinkUrl, analyticsEventName, analyticsLinkType}) => {
+    cardTitleHref = Sefaria.activeModule === "sheets" ? `/sheets/${cardTitleHref}` : cardTitleHref;
```
yitzhakc: We need to introduce constants here

static/js/Header.jsx
```diff
@@ -96,16 +96,16 @@ const LoggedInDropdown = ({module}) => {
...
-                <DropdownMenuItem url={'/sheets/saved'}>
+                <DropdownMenuItem url={'/sheets/saved'} targetModule={Sefaria.SHEETS_MODULE}>
                   <InterfaceText>Saved</InterfaceText>
                 </DropdownMenuItem>
-                <DropdownMenuItem url={'/sheets/history'}>
+                <DropdownMenuItem url={'/sheets/history'} targetModule={Sefaria.SHEETS_MODULE}>
                   <InterfaceText>History</InterfaceText>
                 </DropdownMenuItem>
-                <DropdownMenuItem url={'/settings/account'}>
+                <DropdownMenuItem url={'/settings/account'} targetModule={Sefaria.LIBRARY_MODULE}>
```
yitzhakc, line 108: This is not working properly - it seems like it's still rendering the URL as `https://www.sheets.mod-subdomains.cauldron.sefaria.org/settings/account`, while it should be dropping the `sheets.`

static/js/HeaderAutocomplete.jsx
```diff
@@ -86,15 +86,15 @@ function sortByTypeOrder(array) {
 }
 
 const getURLForObject = function(type, key) {
-    if (type === "Collection") {
+    if (type === "Collection" && Sefaria.activeModule === "sheets") {
```
yitzhakc: Change these to use the constants

static/js/ReaderApp.jsx
```diff
@@ -798,16 +797,16 @@ class ReaderApp extends Component {
     return hist;
   }
   
-  modifyURLbasedOnModule(hist) {
-    if (Sefaria.activeModule === "sheets" && (!hist.url.startsWith("/sheets"))) {
+  modifyURLbasedOnModule(url) {
+    if (Sefaria.activeModule === "sheets" && (!url.startsWith("/sheets"))) {
```
yitzhakc: Use constants for both "sheets" and "/sheets"

static/js/Header.jsx
stevekaplan123: This is partially because DropdownMenu is preventing the propagation of the link click so that it is not getting handled by the ReaderApp handleInAppLinkClick.  I'm assuming that it might not be easy to allow propagation while still getting "DropdownMenu" functionality.  How do you think we should handle this?

stevekaplan123: One possibility is that DropdownMenu.jsx has a function that can calculate given a relative path and targetModule what the full URL should be.

stevekaplan123: And really since openURL in ReaderApp does that as well, we could have a utility function that does that -- that both DropdownMenu and ReaderApp use

sefaria/urls.py
stevekaplan123: It's a string in mod-subdomains.yaml:       `MODULE_ROUTES: '{"library": "/", "sheets": "/sheets/"}'`
It should be converted by local-settings-file.yaml's     `MODULE_ROUTES = json.loads(os.getenv("MODULE_ROUTES", "{}"))`
 but for some reason it didn't work and for the time being I just decided to forcibly convert it here.

static/js/Header.jsx
stevekaplan123: It also seems like the utility function could take a route parameter and create URLs to lead with "/" or "/sheets/" depending on MODULE_ROUTES

sefaria/system/middleware.py
stevekaplan123: `from sefaria.settings import *` imports it but maybe it should be explicit?

stevekaplan123: It's not an object because of: https://github.com/Sefaria/Sefaria-Project/pull/2606/files#r2277772115

static/js/Header.jsx
stevekaplan123: I decided to implement the function I discussed here in util.js called 'fullURL'.  It's now used in ReaderApp and DropdownMenu

sefaria/local_settings_ci.py
stevekaplan123: I didn't know what to do regarding treating MODULES_ROUTES and DOMAIN_MODULES as json objects or as strings given the fact that local-settings-file.yaml doesn't seem to convert the string properly

yitzhakc: Ok. We're going to merge this in, and we'll modify it as needed in the future.

PR #2607, opened 2025-08-05
Add sheet scoring integration

morganizzzm:
## Description
Adds full sheet-scoring integration so that every sheet receives a combined_score (ref relevance × title interest × creativity) generated by the LLM pipeline, stored in MongoDB, and exposed to the client. The “Sheets with Ref” UI now defaults to sorting by this relevance score.

## Code Changes
_sefaria/helper/llm/sheet_scoring.py_

- New helper module
- _prepare_update_data() strips None values
- save_sheet_scoring_output() writes scores back to db.sheets
- make_sheet_scoring_input() converts raw sheet dict → SheetScoringInput

_sefaria/helper/llm/tasks.py_

- Registers web.save_sheets_score Celery task
- Adds generate_and_save_sheet_scoring() chain (llm.score_sheet → save_sheet_scoring)

_sefaria/sheets.py_

-  Injects combined_score into each sheet API payload so the frontend can sort

_static/js/sheets/SheetsWithRefPage.jsx_ 

- Keeps full sortTypeArray (re-enables “relevance”)
- Default sort set to "relevance"
- Sorting function now orders by combined_score (desc)

## Notes

- **Behavior**: Users opening the “Sheets” sidebar will see sheets ranked by the new relevance metric instead of views/date.

- **Data**: Mongo documents now gain/refresh ref_scores, ref_levels, title_interest_level, creativity_score, and processed_at.


sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
+import logging
+from typing import Any,Dict,Optional
```
YishaiGlasner: spaces (everywhere)

sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
...
+
+from bson import ObjectId
+from sefaria.system.database import db
+from sefaria_llm_interface.scoring_io import (
```
YishaiGlasner: we're doing it in one line and without parentheses

sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        Filtered update data
+    """
+    update_data = {
+        "ref_scores": result.ref_scores,
+        "ref_levels": result.ref_levels,
+        "language": result.language,
+        "title_interest_level": result.title_interest_level,
+        "title_interest_reason": result.title_interest_reason,
+        "processed_at": result.processed_at,
```
YishaiGlasner: shouldn't it be `processed_datetime`?

sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Prepare update data for database, filtering out None values.
+
+    Args:
+        result: Result dictionary from processing
+
+    Returns:
+        Filtered update data
+    """
+    update_data = {
```
YishaiGlasner: you can do:
```
asdict(result)
```
and then pop `sheet_id`

sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    sheet_id = result.sheet_id
+
+    try:
+        object_id = ObjectId(sheet_id)
+        update_data = _prepare_update_data(result)
+
+        update_result = db.sheets.update_one(
+            {"_id": object_id},
+            {"$set": update_data}
```
YishaiGlasner: i think it would be better to put all these as an object, we can call it `llm_scoring` or if you have a better name.
and then of course you should also change it anywhere

sefaria/helper/llm/sheet_scoring.py
```diff
@@ -0,0 +1,83 @@
...
+
+        return success
+
+    except Exception as e:
+        logger.error(f"Error updating sheet {sheet_id}: {e}")
+        return False
+
+
+def make_sheet_scoring_input(
+        sheet_content: Dict[str,Any]) -> SheetScoringInput:
```
YishaiGlasner: newline isn't needed

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
+        for sheet_id in collection.sheets:
+            sheet_id_to_collections[sheet_id].append(collection)
+
+    for sheet in sheets:
+        collections = sheet_id_to_collections[int(sheet["id"])]
+        sheet["collections"] = [{'name': collection.name, 'slug': collection.slug} for collection in collections]
+    return sheets
+
+
+import math
```
YishaiGlasner: move imports to head

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
+from typing import Iterable,Sequence,Union,List
+
+Number = Union[int,float]
+
+
+def percentile(data: Iterable[Number], q: Union[Number, Sequence[Number]]) -> \
```
YishaiGlasner: @nsantacruz it seems to me that adding numpy to our requirements is preferred on doing it ourselves. what do you think? we can import numpy in function and doing it optional dependency for dev.
(if this fundtion remains we should find the right place to put it)

Margo, chat gpt claims this should work. It's shorter, but i didn't check. i guess it lies:
```
def percentile(data, percent):
    """
    Calculate the percentile of a list of values.
    `percent` should be between 0 and 100.
    """
    if not data:
        return None
    data = sorted(data)
    k = (len(data) - 1) * (percent / 100)
    f = int(k)
    c = min(f + 1, len(data) - 1)
    if f == c:
        return data[int(k)]
    return data[f] + (data[c] - data[f]) * (k - f)
```

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
+            return float(xs[lo])
+        frac = r - lo
+        return xs[lo] * (1.0 - frac) + xs[hi] * frac
+
+    if isinstance(q, (int, float)):
+        return one(float(q))
+    return [one(float(p)) for p in q]
+
+
+def calculate_combined_score(sheets, pasuk_ref, title_weight=0.3, ref_weight=0.3,
```
YishaiGlasner: it isn't necessarily pasuk, so `ref` is better

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
+    Calculate combined score for sheets based on title interest, ref relevance and creativity.
+    Returns sheets with added combined_score field.
+
+    """
+    if not sheets:
+        return sheets
+
+    try:
+        # Convert pasuk_ref to Ref object and get all segments
+        target_ref = model.Ref(pasuk_ref) if isinstance(
```
YishaiGlasner: we're doing it in one line

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
+            sheet."""
+            ref_map = sheet.get("ref_scores", {})
+            return max((ref_map.get(seg, 0) for seg in segments), default=0)
+
+        # Extract raw ref scores
+        ref_scores_raw = [get_sheet_ref_score(sheet) for sheet in sheets]
+
+        # Extract and clip creativity scores
+        creativity_scores = [
+            max(0, min(1, sheet.get("creativity_score", 0)))
```
YishaiGlasner: can `creativity_score` be a negative number?
maybe it will be better to ensure that it isn't when saving

sefaria/sheets.py
```diff
@@ -294,1014 +293,1129 @@ def annotate_displayed_collections(sheets):
...
-			}
-			results.append(sheet_data)
-	return results
+    """
+    Returns a list of sheets that include ref,
+    formating as need for the Client Sidebar.
+    If `uid` is present return user sheets, otherwise return public sheets.
+    If `in_collection` (list of slugs) is present, only return sheets in one of the listed collections.
+    """
+    try:
```
YishaiGlasner: why did you add this try-except? was it only for debugging? if so, please remove it

YishaiGlasner (changes requested):

sefaria/sheets.py
morganizzzm: no, it will necessarily be positive -- this is a percentage of user generated text in a sheet, it's range is from 0 to 1

sefaria/helper/llm/sheet_scoring_tools.py
```diff
@@ -0,0 +1,92 @@
...
+    Find all sheets (public + private) that cite the tref in includedRefs OR expandedRefs.
+    """
+    segments = _segment_list_for_ref(tref)
+    query = {"$or": [{"includedRefs": {"$in": segments}},
+                     {"expandedRefs": {"$in": segments}}]}
+    logger.debug("Querying sheets for tref=%r with segments count=%d", tref, len(segments))
+    return db.sheets.find(query)
+
+
+def _find_all_sheets() -> Iterable[Dict[str, Any]]:
```
YishaiGlasner: we want only published sheets

sefaria/sheets.py
```diff
@@ -261,1048 +261,1162 @@ def change_sheet_owner(sheet_id, new_owner_id):
...
+        if sheet["status"] == "public" and not scored_by_llm:
+            generate_and_save_sheet_scoring(sheet)
+            scored_by_llm = True
+
+    else:
+        sheet.pop("_id", None)
+        db.sheets.find_one_and_replace({"id": sheet["id"]}, sheet)
+
+    if not scored_by_llm and not new_sheet and sheet["status"] == "public":
+        generate_and_save_sheet_scoring(sheet)
```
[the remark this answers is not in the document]
YishaiGlasner: it only fires celery task. the celery task can fail and it won't prevent saving.

sefaria/sheets.py
```diff
@@ -261,1048 +261,1162 @@ def change_sheet_owner(sheet_id, new_owner_id):
...
+        return vs[f]
+    return vs[f] + (vs[c] - vs[f]) * (k - f)
+
+
+def _p5_p95(values):
+    return _percentile(values, 5), _percentile(values, 95)
+
+
+def calculate_combined_score(
+    sheets, ref, title_weight=0.3, ref_weight=0.3, creativity_weight=0.4
```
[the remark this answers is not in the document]
YishaiGlasner: 0.3+0.3+0.4=1...

sefaria/sheets.py
```diff
@@ -261,1048 +261,1162 @@ def change_sheet_owner(sheet_id, new_owner_id):
...
+        sheets_ids = [sheet for sublist in sheets_list for sheet in sublist]
+        query["id"] = {"$in": sheets_ids}
+
+    projection = {"id": 1, "title": 1, "owner": 1, "viaOwner":1, "via":1, "dateCreated": 1, "includedRefs": 1, "expandedRefs": 1,
+         "views": 1, "topics": 1, "status": 1, "summary":1, "attribution":1, "assigner_id":1, "likes":1,
+         "displayedCollection":1, "options":1, "llm_scoring":1}
+    sheetsObj = db.sheets.find(query, projection)
+    sheetsObj.hint("expandedRefs_1")
+    sheets = [s for s in sheetsObj]
+    sheets = calculate_combined_score(sheets, str(oref))
```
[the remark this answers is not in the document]
YishaiGlasner: i'm not sure i understand this - the score is different for every ref, so it should be calculated for any ref

static/js/SearchPage.jsx
```diff
@@ -26,6 +27,9 @@ class SearchPage extends Component {
 
   render () {
     const classes = classNames({readerNavMenu: 1, compare: this.props.compare});
+    const showAiBadge =    this.props.type === 'sheet' &&
+        this.props.searchState?.sortType?.toLowerCase?.() === 'relevance';
+    const aiBadgeTexts = 'These sheet results are ranked by AI relevance.';
```
stevekaplan123: Since SearchPage can also display non-sheets, it seems confusing to have this text in SearchPage

static/js/sheets/SheetsWithRefPage.jsx
```diff
@@ -17,7 +17,9 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // they are never changed.
 
     const [refs, setRefs] = useState(srefs);
-    const sortTypeArray = SearchState.metadataByType['sheet'].sortTypeArray.filter(sortType => sortType.type !== 'relevance');
+    const sortTypeArray = SearchState.metadataByType['sheet'].sortTypeArray;
+
+    // const sortTypeArray = SearchState.metadataByType['sheet'].sortTypeArray.filter(sortType => sortType.type !== 'relevance');
```
stevekaplan123: We can remove this comment

static/js/SearchPage.jsx
stevekaplan123: Since we may use AI in other Searchpage contexts eventually, we may want to have wrapper components for AiInfoTooltip based on whether we're in voices or library module

PR #2614, opened 2025-08-13
helm: remove linker

BrendanGalloway:
## Description

## Code Changes

## Notes

### DIFF
--- a/helm-chart/sefaria/templates/configmap/nginx.yaml
+++ b/helm-chart/sefaria/templates/configmap/nginx.yaml
@@ -175,6 +175,17 @@ data:
           proxy_pass https://storage.googleapis.com/sefaria-sitemaps$request_uri;
         }
 
+        location /api/find-refs {
+          if ($request_method = OPTIONS) {
+            add_header Access-Control-Allow-Origin "*" always;
+            add_header Access-Control-Allow-Methods "POST, OPTIONS" always;
+            add_header Access-Control-Allow-Headers "Content-Type" always;
+            add_header Content-Length 0;
+            add_header Content-Type  text/plain;
+            return 204;
+          }
+        }
+
       } # server
 
       types {
--- a/helm-chart/sefaria/templates/rollout/task.yaml
+++ b/helm-chart/sefaria/templates/rollout/task.yaml
@@ -1,4 +1,4 @@
-{{- if or .Values.tasks.enabled .Values.tasks.gpuServer }}
+{{- if or .Values.tasks.enabled .Values.gpuServer }}
 {{- $QueueConfig := (fromYaml (include "sefaria.tasks.queues" . )) -}}
 ---
 # apiVersion: apps/v1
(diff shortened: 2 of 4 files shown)


helm-chart/sefaria/templates/configmap/nginx.yaml
```diff
@@ -175,18 +175,6 @@ data:
           proxy_pass https://storage.googleapis.com/sefaria-sitemaps$request_uri;
         }
 
-        {{- if .Values.linker.enabled }}
```
nsantacruz, line 178: We actually still need this clause. It adds the correct access control headers for linker API requests since the linker can be embedded on third party websites

helm-chart/sefaria/templates/service/task.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if or .Values.tasks.enabled .Values.tasks.gpuServer }}
```
nsantacruz: should be `.Values.gpuServer`

helm-chart/sefaria/templates/rollout/task.yaml
```diff
@@ -1,5 +1,5 @@
-{{- if .Values.tasks.enabled }}
- {{- $QueueConfig := (fromYaml (include "sefaria.tasks.queues" . )) -}}
+{{- if or .Values.tasks.enabled .Values.tasks.gpuServer }}
```
nsantacruz: should be `.Values.gpuServer`

helm-chart/sefaria/templates/rollout/task.yaml
```diff
@@ -79,6 +79,12 @@ spec:
             value: "varnish-{{ .Values.deployEnv }}-{{ .Release.Revision }}"
           - name: HELM_REVISION
             value: "{{ .Release.Revision }}"
+          {{- with .Values.gpuServer }}
+          - name: GPU_SERVER_HOST
+            value: "{{ . }}"
```
nsantacruz, line 84: shouldn't this be `.Values.gpuServer`?

helm-chart/sefaria/templates/configmap/nginx.yaml
BrendanGalloway: Ok, so we always want cors  to allow all on /api/find-refs? regardless of linker enabled or not?  In cauldrons where there is no task/ner pods that URL will just fail?

helm-chart/sefaria/templates/rollout/task.yaml
BrendanGalloway: `with` is a conditional scope change within the template.  If the `with` value exists, the block is scoped to it, so {{ . }} resolves to {{ .Values.gpuServer}} within the block

helm-chart/sefaria/templates/configmap/nginx.yaml
nsantacruz: ok, that sounds fine

PR #2615, opened 2025-08-13
Profile Navigation Functionality From Collections

yodem:
## Browser History Navigation Fix
[sc-21372]
### Description
Fixed broken browser back button for collection pages. When navigating from collection → profile → back button, users would get a blank tab instead of returning to the collection page. The issue was caused by `this.replaceHistory` flag not being reset after use, causing all subsequent navigations to use `history.replaceState` instead of `history.pushState`, effectively flattening browser history.

### Code Changes
- **ReaderApp.jsx**: Added centralized `replaceHistory` flag reset in `setPanelState()` method
- **ReaderPanel.jsx**: Removed unnecessary `oldReplaceHistory` variable in `setTab()` method
- **Key fix**: `setPanelState()` now resets `this.replaceHistory = false` after using the flag

### Notes
- Root cause: `updateCollectionName()` and other ReaderPanel methods set `replaceHistory = true` but ReaderApp never reset it
- Solution ensures proper `history.pushState` vs `history.replaceState` behavior
- I tested other navigation scenarios and they work correctly (tested: basic nav, search, text refs, collections, multi-level flows)

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -278,6 +278,7 @@ class ReaderApp extends Component {
     }
 
     this.setContainerMode();
+    // Pass this.replaceHistory flag to updateHistoryState - it will be reset there
     this.updateHistoryState(this.replaceHistory);
   }
 
@@ -794,9 +795,17 @@ class ReaderApp extends Component {
 
     return hist;
   }
-  updateHistoryState(replace) {
-    // Always reset replaceHistory flag first, before any early returns
-    const shouldReplace = replace;
+  updateHistoryState(shouldReplace) {
+    // IMPORTANT: reset replaceHistory flag first, before any early returns such as !this.shouldHistoryUpdate() or return;
+    // 
+    // The this.replaceHistory system works as follows:
+    // 1. Various ReaderPanel methods set this.replaceHistory = true when they want
+    //    the NEXT history update to use replaceState instead of pushState
+    //    (e.g., tab changes, collection name updates, connection focus changes)
+    // 2. componentDidUpdate calls updateHistoryState(this.replaceHistory)
+    // 3. We MUST reset this.replaceHistory = false immediately to prevent it from
+    //    "sticking" and affecting ALL future history updates
+    // 4. We use the shouldReplace parameter for this specific update
     this.replaceHistory = false;
     
     if (!this.shouldHistoryUpdate()) {
@@ -1284,6 +1293,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     });
   }
   setPanelState(n, state, replaceHistory) {
+    // Set replaceHistory flag - this will be consumed and reset by updateHistoryState
     this.replaceHistory  = Boolean(replaceHistory);
     //console.log(state)
     // When the driving panel changes language, carry that to the dependent panel


yodem:
@YishaiGlasner  I updated the description and the code.
Your comment was very helpful XD

static/js/ReaderApp.jsx
```diff
@@ -1301,6 +1300,10 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       new_state["defaultPanelSettings"] = Sefaria.util.clone(state.settings);
     }
     this.setState(new_state);
+    // Reset replaceHistory to false to avoid breaking subsequent navigation
```
stevekaplan123: I don't understand 1303-1306 relation to 1284: `this.replaceHistory = Boolean(replaceHistory)`.  We first set replaceHistory in 1284 and then set it to false in 1303.  What is this accomplishing?

stevekaplan123: Ah I see replaceHistory is used outside this function so you're setting it globally and then unsetting.

stevekaplan123 (approved):

yodem:
I asked cursor to trace through the exact flow when you enter `http://localhost:8000/collections/OcDM4trS?tab=sheets`:

## 🌊 **The Complete Flow:**

### **1. Browser Navigation**
```
User types/enters: http://localhost:8000/collections/OcDM4trS?tab=sheets
Browser creates initial history entry with this URL
```

### **2. Django Backend Routes**
```python
# In sefaria/urls.py
url(r'^collections/(?P<slug>[^.]+)$', reader_views.collection_page)
# Matches and serves the React app
```

### **3. React App Initialization**
```javascript
// ReaderApp componentDidMount()
componentDidMount() {
  this.updateHistoryState(true); // ← REPLACE initial entry
  // ...
}
```

### **4. Collection Component Loads**
```javascript
// CollectionPage renders, calls ReaderPanel methods
// ReaderPanel.updateCollectionName() gets called
updateCollectionName(name) {
  this.replaceHistory = true;  // ← Sets flag
  this.conditionalSetState({collectionName: name});
}
```

### **5. State Updates Trigger History**
```javascript
// conditionalSetState → setPanelState → setState
componentDidUpdate() {
  this.updateHistoryState(this.replaceHistory); // ← Uses the flag
}
```

## 🎯 **Why We Need `this.replaceHistory = false`:**

### **Without the reset:**
```javascript
updateHistoryState(replace) {
  if (this.replaceHistory) {  // ← Still true from step 4
    history.replaceState(...); // ← Always replaces!
  }
  // No reset means this.replaceHistory stays true forever
}
```

**Result**: Every navigation after the collection page uses `replaceState`:
- ✅ Collections page loads
- ❌ Click profile → **replaces** collections entry (no new entry)  
- ❌ Click anything else → **replaces** profile entry
- ❌ Back button → No entries to go back to!

### **With the reset:**
```javascript
updateHistoryState(replace) {
  const shouldReplace = this.replaceHistory;
  this.replaceHistory = false;  // ← Reset for future use
  
  if (shouldReplace) {
    history.replaceState(...); // ← Only replaces when intended
  } else {
    history.pushState(...);    // ← Normal navigation pushes
  }
}
```

**Result**: Proper history stack:
- ✅ Collections page loads (replaces initial entry)
- ✅ Click profile → **pushes** new entry
- ✅ Back button → Goes back to collections ✅

## 🔑 **The Key Insight:**

`this.replaceHistory = true` is meant for **one-time use** (when setting collection name), not as a permanent setting. Without the reset, it becomes a "sticky" flag that breaks all future navigation.

The reset ensures each navigation decision is independent! 🎯

static/js/ReaderApp.jsx
```diff
@@ -795,6 +795,10 @@ class ReaderApp extends Component {
     return hist;
   }
   updateHistoryState(replace) {
+    // Always reset replaceHistory flag first, before any early returns
```
stevekaplan123: Can you just clarify this comment, as the early return only occurs in the shouldHistoryUpdate case?

yodem: also for       if (currentUrl === hist.url) { return; } // Never push history with the same URL

I wanted to do it in the end of the function but it might not get there so I put it in the beginning

stevekaplan123: I actually meant can you add to the comment itself.  The 'replaceHistory' variable is so confusing that it would be helpful to know more about why we're resetting it.   Also, why are you adding 'shouldReplace'?  Isn't 'replace' enough?

stevekaplan123: Or maybe the parameter of the function should be called 'shouldReplace"

PR #2619, opened 2025-08-14
Add deprecation notification for divine name substitution tool

yodem:
Added Message to Notify the users that the divine name substitution tool will no longer be supported in the source sheet.
Supported message in hebrew and english with links and guide how to do without the source sheet

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -23,7 +23,7 @@ import {SourceEditor} from "./SourceEditor";
 import {EditTextInfo} from "./BookPage";
 import ReactMarkdown from 'react-markdown';
 import TrackG4 from "./sefaria/trackG4";
-import * as constants from './constants';
+
 function useOnceFullyVisible(onVisible, key) {
   const targetRef = useRef(null);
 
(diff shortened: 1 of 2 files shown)


static/css/s2.css
```diff
@@ -8185,6 +8185,22 @@ But not to use a display block directive that might break continuous mode for ot
   line-height: 19px;
   color: #666666;
 }
+.divineNameDepricationNotification strong {
+  font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
```
YishaiGlasner: if there is no good reason, wehould not define `font-family` (i know it's common in our css):
```
  --english-font: var(--english-sans-serif-font-family);
  --hebrew-font: var(--hebrew-sans-serif-font-family);
```
this will also do the job for the hebrew which i guess you need also

static/js/Misc.jsx
```diff
@@ -2914,6 +2913,32 @@ const SheetMetaDataBox = (props) => (
   </div>
 );
 
+const DivineNameDepricationNotification = () => {
```
YishaiGlasner, line 2916: It seems like a temp banner, and we have components for that, i think. Skyler should know more (and i guess also others, but i've never used it)

static/css/s2.css
yodem: After trying this solution - there is an issue with the fonts regards to the "strong" tag.
i wanted it to be as the component that wrap my banner and its fixed to the font family I wrote.
when I tried to remove my css, it fallback to some other font and to solve the problem from its root wouldn't be so meaningful because we are about to delete it anyways...
So I think that I'll leave it as the font I wrote.

static/js/Misc.jsx
relyks: Next time @ me so I get notified lol. I actually haven't made something that would work in this particular scenario, since this needs to be an inline component within another one. The other components I've made or worked with only show in specific locations like the top level of the resources panel on Topics pages or on the top of the page. They can't be injected arbitrarily anywhere

static/css/s2.css
relyks: What @YishaiGlasner is right. You shouldn't be redefining the font stack. We should always be using the font family classes (`.sans-serif` or `.serif`) in case there needs to be a change in font everywhere (and that actually could be a possibility in the future).

You can remove all the font stack declarations you've made for your `p`, `strong`, and `a` elements within your component and either use `font-family: inherit` in your CSS declarations for each of those elements or add the `.sans-serif` class to the component's top level class declaration. Either approach will assign the right fonts.

I couldn't understand why the fonts are messed up by default though 🤷‍♂️

static/js/Misc.jsx
```diff
@@ -2914,6 +2913,32 @@ const SheetMetaDataBox = (props) => (
...
+
+  const lang = Sefaria.interfaceLang === "hebrew" ? "he" : "en";
+  const messages = constants.DEPRECATION_MESSAGES[lang];
+  const links = constants.DEPRECATION_LINKS[lang];
+
+  return (
+    <div className="divineNameDepricationNotification">
```
relyks: Remove the CSS declarations for the font stack for your elements and add `.sans-serif` at the top level here.
```suggestion
    <div className="divineNameDepricationNotification sans-serif">
```

static/css/s2.css
```diff
@@ -8185,6 +8185,22 @@ But not to use a display block directive that might break continuous mode for ot
...
+}
+
+.divineNameDepricationNotification p a {
+  font-size: 16px;
+  font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
```
relyks:
```suggestion
  font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
```
Remove this

static/css/s2.css
relyks:
```suggestion
  font-family: "Roboto", "Helvetica Neue", "Helvetica", sans-serif;
```
Remove this

static/js/Misc.jsx
```diff
@@ -2914,6 +2913,32 @@ const SheetMetaDataBox = (props) => (
...
+        </a>
+        {" "}{messages.andText}
+        <a href={links.extension}>
+          {messages.extensionText}
+        </a>
+        {messages.period}
+      </p>
+    </div>
+  );
+};
```
relyks: image

I'm not sure the best way of fixing this based on what you did, but there is a missing space before the last link. I think this should do the trick: `{" "}{messages.andText}{" "}`, but that also might add an unnecessary space in the Hebrew?

If that doesn't work, you might need to add extra space to the string in your definitions for English and add a `white-space: preserve` rule to your `.divineNameDepricationNotification` declaration so that the spaces show up for all children elements of your component.

static/js/constants.js
```diff
@@ -11,3 +11,39 @@ export const layoutLabels = {
...
+    }
+};
+
+export const DEPRECATION_MESSAGES = {
+    en: {
+    notice: "Please note:",
+    mainMessage: `The divine name substitution tool will no longer be available in the Sefaria Sheet Editor after ${DEPRECATION_DATE}.`,
+    continuationMessage: "If you would like to continue making changes to how the divine name appears in your sheets prior to printing,",
+    exportText: "export your sheet to Google Docs",
+    andText: "and use the 'Transform Divine Names' feature in the",
```
relyks:
```suggestion
    andText: "and use the 'Transform Divine Names' feature in the ",
```
Might need to add an extra space here, but I'm not sure

static/js/Misc.jsx
yodem: fixed

static/css/s2.css
yodem: fixed

static/css/s2.css
yodem: fixed

static/js/Misc.jsx
yodem: I added spaces in the constants as you said

static/js/constants.js
yodem: fixed

PR #2623, opened 2025-08-17
Chore/sc 22597/redirect for help sheets to new zendesk links

stevekaplan123:
## Description
Add HELP_CENTER_REDIRECTS in site_settings.py
There are pros and cons with this approach.  We could store a 'redirect' field in each sheet in the DB allowing us to avoid cluttering up the site_settings.py file.  However, this approach seems more straightforward as it allows us to have one place where all the redirects live, which allows us to edit this data more easily.  Also, this approach allows us to delete the sheets in the DB at some point.

### DIFF
--- a/sourcesheets/views.py
+++ b/sourcesheets/views.py
@@ -173,8 +173,9 @@ def view_sheet(request, sheet_id, editorMode = False):
     """
     help_center_redirects = SITE_SETTINGS.get('HELP_CENTER_REDIRECTS', {})
     lang_code = request.LANGUAGE_CODE if request.LANGUAGE_CODE in help_center_redirects else 'en'
-    if lang_code in help_center_redirects and str(sheet_id) in help_center_redirects[lang_code]:
-        return redirect(help_center_redirects[lang_code][str(sheet_id)])
+    redirect_url = help_center_redirects.get(lang_code, {}).get(str(sheet_id))
+    if redirect_url:
+        return redirect(redirect_url)
     
     embed = request.GET.get('embed', '0')
     if embed != '1' and editorMode is False:


sites/sefaria/site_settings.py
```diff
@@ -24,5 +24,65 @@
 	"WHAT_ARE_VOICES_PATHS": {
 		"en": "/sheets/674324",
 		"he": "/sheets/674327"
+	},
+	"HELP_CENTER_REDIRECTS": {
```
YishaiGlasner, line 28: i think i'd move it to another file, maybe a csv, so we can give other teams to edit (but it can also json or just python file for this dict).

stevekaplan123: @yitzhakc What do you think?  It might be that we never need to change this, or if we do need to change it, it might be rare enough that this current format is fine.

yitzhakc: Agreed, let's leave it for now. If a more concrete need arises, we can reconsider, but let's not prematurely optimize...

sourcesheets/views.py
```diff
@@ -165,19 +166,21 @@ def make_sheet_class_string(sheet):
...
     return " ".join(classes)
 
-
 @ensure_csrf_cookie
 def view_sheet(request, sheet_id, editorMode = False):
     """
     View the sheet with sheet_id.
     """
+    if sheet_id in SITE_SETTINGS['HELP_CENTER_REDIRECTS'][request.LANGUAGE_CODE]:
+        return redirect(SITE_SETTINGS['HELP_CENTER_REDIRECTS'][request.LANGUAGE_CODE][sheet_id])
```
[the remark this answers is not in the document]
yitzhakc: @stevekaplan123 ?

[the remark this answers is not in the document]
yitzhakc: @stevekaplan123 ?

PR #2637, opened 2025-08-31
use dvh rather than vh

YishaiGlasner:
that will calculate also things like url bar, etc.
Remain vh for older browsers.


yitzhakc (approved):
Probably.
Perhaps as part of the SCSS?

YishaiGlasner:
> Probably. Perhaps as part of the SCSS?

There is nothing between this world and the days of Messiah, but only scss

PR #2639, opened 2025-09-02
Bug/sc 36100/subs cauldron logging in from library module

stevekaplan123:
## Description
Templates and login were broken. 

## Code Changes
The change is to create our own tag 'subdomain_url' (in sefaria_tags.py) instead of Django's url tag to make sure we route to '/sheets/login/' vs '/login' depending on the module we're in

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -3444,7 +3444,6 @@ const LangSelectInterface = ({callback, defaultVal, closeInterface}) => {
       ))}
     </div>
   );
-
 }
 
 
--- a/static/js/sefaria/strings.js
+++ b/static/js/sefaria/strings.js
@@ -31,7 +31,7 @@ const Strings = {
     "Users": "משתמשים",
     "Categories": "קטגוריות",
 
-    // Texts Page
+    // Landing Pages
     "Learn More": "למדו עוד",
     "Learning Schedules": "לוח לימוד יומי",
     "Teach with Sefaria": "מלמדים עם ספריא",
@@ -56,6 +56,9 @@ const Strings = {
     "Sponsor A Day of Learning": "תרמו יום לימוד בספריא",
     "With your help, we can add more texts and translations to the library, develop new tools for learning, and keep Sefaria accessible for Torah study anytime, anywhere.": "בעזרת התרומה שלכם נוכל להוסיף מקורות, תרגומים, לפתח כלים חדשים ללימוד ולשמור על גישה חופשית ללימוד תורה, בכל מקום ובכל זמן.",
     "Sponsor A Day": "הקדישו יום לימוד",
+    "Community-Powered Jewish Learning": "חיבורים חדשים למקורות ולקהילה",
+    "Share. Discover. Join the Conversation.": "המקום ללמידה, שיתוף רעיונות, וכתיבת חומרים מקוריים.",
+    "Browse Topic Categories": "חיפוש מקורות לפי נושא",
 
     // Text Category Pages
     "All Portions": "פרשות השבוע",
@@ -160,9 +163,6 @@ const Strings = {
 
     // Sheets
     "Sheet": "דף מקורות",
-    "Community-Powered Jewish Learning": "חיבורים חדשים למקורות ולקהילה",
-    "Share. Discover. Join the Conversation.": "המקום ללמידה, שיתוף רעיונות, וכתיבת חומרים מקוריים.",
-    "Browse Topic Categories": "חיפוש מקורות לפי נושא",
     "About this Sheet": "אודות דף המקורות",
     "Tags": "תוויות",
     "Search for a text...": "חפש מקור בספריא",
--- a/static/js/sheets/SheetsHomePage.jsx
+++ b/static/js/sheets/SheetsHomePage.jsx
@@ -1,7 +1,6 @@
 import React  from 'react';
 import {NavSidebar} from "../NavSidebar";
 import {SheetsTopicsCalendar, SheetsTopicsTOC} from "./SheetsHomePageTopicsTOC";
-import Sefaria from "../sefaria/sefaria";
 const SheetsHeroBanner = ({title, message, videoOptions, posterImg}) => {
     /*
      * `title` and `message` are shown on top of the video. `posterImg` is shown while video is downloaded,


reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +523,30 @@ def date_string_to_date(dateString):
...
+    """
+    Custom template tag that resolves URLs based on the current subdomain.
+    For sheets subdomain, it will try to resolve to /sheets/ prefixed URLs first.
+    Falls back to regular URL resolution if sheets URL doesn't exist.
+    """
+    # Get the current request from the template context
+    request = kwargs.pop('request', None)
+    
+    # If we're on the sheets subdomain, try to resolve with /sheets/ prefix first
+    if request and hasattr(request, 'active_module') and request.active_module == 'sheets':
```
stevekaplan123: Remove magic strings from this and add legacy routes

reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +523,30 @@ def date_string_to_date(dateString):
...
+        # Try to resolve with /sheets/ prefix first
+        try:
+            sheets_url_name = f'sheets_{url_name}'
```
stevekaplan123: Another magic string

reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +524,25 @@ def date_string_to_date(dateString):
...
+
+@register.simple_tag
+def subdomain_url(url_name, *args, **kwargs):
+    """
+    Custom template tag that resolves URLs based on the current subdomain.
+    Falls back to regular URL resolution if subdomain URL doesn't exist.
+    """
+    request = kwargs.pop('request', None)
+    for prefix in MODULE_ROUTES.values():
+        if request and hasattr(request, 'active_module') and request.active_module == prefix.replace('/', ''):
```
YishaiGlasner: you can do it with no `and`s
```
if getattr(request, 'active_module', None) == prefix.replace('/', ''):
```

reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +524,25 @@ def date_string_to_date(dateString):
...
 		'<a href="/sheets/{}">a sheet</a>'.format(sheet_id)))
+
+
+@register.simple_tag
+def subdomain_url(url_name, *args, **kwargs):
+    """
+    Custom template tag that resolves URLs based on the current subdomain.
+    Falls back to regular URL resolution if subdomain URL doesn't exist.
+    """
+    request = kwargs.pop('request', None)
```
YishaiGlasner: i don't like it. my best freind claims you can add request to context and do:
```
@register.simple_tag(takes_context=True)
def subdomain_url(context, url_name, *args, **kwargs):
    request = context['request']
```
and then i guess you can get rid of `args` and `kwargs` but you know if they have some imfuence.
i don't familiar with all these django things so i'm not sure this is better.

sefaria/system/middleware.py
```diff
@@ -110,10 +110,12 @@ def process_request(self, request):
                 for domain in DOMAIN_LANGUAGES:
                     if DOMAIN_LANGUAGES[domain] == interface:
                         redirect_domain = domain
+                logger.info(f"Redirect domain: {redirect_domain}")
```
YishaiGlasner: is this debugging thing? (this is for many `logger`s here)

sefaria/system/middleware.py
```diff
@@ -255,21 +259,64 @@ class ModuleMiddleware(MiddlewareURLMixin):
...
+            current_host = request.get_host()
+            
+            # Remove port if present (for localhost:8000)
+            if ':' in current_host:
+                current_host = current_host.split(':')[0]
+            
+            # Check DOMAIN_MODULES to find which module this host belongs to
+            for module_name, module_domain in DOMAIN_MODULES.items():
+                # Remove protocol and port from module_domain for comparison
+                domain_to_check = module_domain
```
YishaiGlasner: why this?

sefaria/system/middleware.py
```diff
@@ -255,21 +259,64 @@ class ModuleMiddleware(MiddlewareURLMixin):
...
+                if '://' in domain_to_check:
```
YishaiGlasner: i don't loike this, because it seems as reinventing the wheel, so i've asked my freind again:
```
from urllib.parse import urlparse

def clean_domain(domain):
    parsed = urlparse(domain)
    # If scheme is missing, urlparse treats it as path, so add a dummy scheme
    if not parsed.netloc:
        parsed = urlparse('//' + domain)
    return parsed.hostname  # returns domain without scheme or port
```

sefaria/system/middleware.py
```diff
@@ -255,21 +259,64 @@ class ModuleMiddleware(MiddlewareURLMixin):
...
+        host_based_module = self._get_module_from_host(request)
+        if host_based_module:
+            active_module = host_based_module
+            logger.info(f"Active module from host: {active_module}")
+        else:
+            # Fallback to route-based detection
+            for module_name, route_prefix in MODULE_ROUTES.items():
+                logger.info(f"Processing request, route_prefix: {route_prefix}, request.path: {request.path}")
+                route_base_path = route_prefix.removesuffix('/')
+                if len(route_base_path) > 0 and (request.path.startswith(route_prefix) or request.path == route_base_path):
```
YishaiGlasner: i think this is better:
```
if route_base_path and re.search(f'^{route_base_path}(?:$|/)', request.path):
```

but i don't understand - we have:
```
"library": "/"
```
so when it's `/`, it won't work, isn't it?

sefaria/system/middleware.py
```diff
@@ -255,21 +259,64 @@ class ModuleMiddleware(MiddlewareURLMixin):
...
             return self.get_response(request)
 
-        # Find the matching route prefix
-        for module_name, route_prefix in MODULE_ROUTES.items():
-            route_base_path = route_prefix.removesuffix('/')
-            if len(route_base_path) > 0 and (request.path.startswith(route_prefix) or request.path == route_base_path):
-                active_module = module_name
-                break
-
+        # First, try to determine module based on host/subdomain from DOMAIN_MODULES
```
YishaiGlasner, line 288: i don't understand why we need these 2 steps

sefaria/urls.py
```diff
@@ -351,23 +351,22 @@
 ]
 
 
-# Registration and Login need to support both the main site and the sheets prefix.
 for prefix in MODULE_ROUTES.values():
-    prefix = prefix.lstrip('/') # remove the first slash because it's not necessary in urlpatterns; see MODULE_ROUTES in local_settings.py
+    prefix = prefix.replace('/', '')
```
YishaiGlasner: i'd use `strip` in case we'll have a path with slash in the middle

reader/templatetags/sefaria_tags.py
yitzhakc: In any case, we shouldn't be `pop` ing since then it's not available to any other function.

sefaria/system/middleware.py
yitzhakc: Do we still need the logging statements to debug?

sefaria/urls.py
```diff
@@ -26,6 +26,12 @@
 handler404 = 'reader.views.custom_page_not_found'
 
 
+
+def get_module_url_name(prefix, url_name):
+    """Generate URL name for module-specific routes, handling empty prefix case."""
```
yitzhakc: What is the empty prefix case?

templates/registration/password_reset_email.html
```diff
@@ -1,12 +1,12 @@
...
+{% load i18n sefaria_tags %}
 
 <html>
 <p>{% trans "A request has been made to reset the password for this email address on Sefaria.org." %}</p>
 
 <p>{% trans "Click this link, or copy it to your browser, to choose a new password:" %}</p>
 
-<a href="{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}">
-	{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}
+<a href="{{ protocol }}://{{ domain }}{% subdomain_url 'password_reset_confirm' uidb64=uid token=token %}">
```
yitzhakc, line 8: Have you tested sending a password reset email from the subdomain, and making sure that the link shows up correctly?

sefaria/urls.py
stevekaplan123: Good point.  I changed the comment and it should hopefully be clearer.

templates/registration/password_reset_email.html
stevekaplan123: Password reset doesn't work although other templates seem to work.  I think password reset should be dealt with in its own story

reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +525,90 @@ def date_string_to_date(dateString):
...
+            url = _resolve_subdomain_url(active_module, url_name, *args, **kwargs)
+            if url:
+                return url
+
+    # Fallback to default URL resolution
+    return _get_default_url(url_name, *args, **kwargs)
+
+
+@register.simple_tag
+def subdomain_url_for_domain(domain, url_name, *args, **kwargs):
```
yitzhakc: Could we possibly make this another use-case of `subdomain_url`? And make `domain` an optional parameter, that would pull the activemodule from the domain var and not from the context.request

sefaria/system/middleware.py
```diff
@@ -255,22 +255,55 @@ class ModuleMiddleware(MiddlewareURLMixin):
     def __init__(self, get_response):
         self.get_response = get_response
 
-    def __call__(self, request):
-        active_module = 'library'  # default active_module
+    def _get_module_from_host(self, request):
```
yitzhakc, line 258: I _really_ would like you to write some pytests on this please.

reader/templatetags/sefaria_tags.py
```diff
@@ -522,3 +525,90 @@ def date_string_to_date(dateString):
...
+    Returns the URL if found, '#' if not found.
+    """
+    try:
+        return reverse(url_name, args=args, kwargs=kwargs)
+    except NoReverseMatch:
+        return '#'
+
+
+@register.simple_tag(takes_context=True)
+def subdomain_url(context, url_name, *args, **kwargs):
```
yitzhakc, line 531: I _really_ would like you to write some pytests on this please.

PR #2641, opened 2025-09-03
Bug/sc 36093/ensure login remains across subdomains3

stevekaplan123:
## Description
Login across subdomains is now working due to the additions to sandbox-values.yaml.  I also solved added some explicit URLs (i.e. including the full domain) to the about page and cookies notification so we don't get into trouble there.


sefaria/system/middleware.py
```diff
@@ -110,10 +110,12 @@ def process_request(self, request):
                 for domain in json.loads(DOMAIN_LANGUAGES):
                     if json.loads(DOMAIN_LANGUAGES)[domain] == interface:
                         redirect_domain = domain
+                logger.info(f"Redirect domain: {redirect_domain}")
```
yitzhakc: Do we want to leave these logging statements?

build/ci/sandbox-values.yaml
```diff
@@ -69,3 +69,12 @@ localSettings:
   GLOBAL_WARNING: false
   GLOBAL_WARNING_MESSAGE: "Sefaria will be in <b>Read-Only</b> mode for scheduled maintenance from 4:45pm-6:45pm Pacific time. Edits will <b>not</b> be saved during that time."
   CRM_TYPE: "NONE"
+  # Session cookie settings for cross-subdomain support
```
yitzhakc, line 72: Do we set the relevant production values in a an adjacent yaml file? Perhaps we should do it as part of this PR? If not we should write up a story reminding us to do it.

yitzhakc (changes requested):

PR #2642, opened 2025-09-03
chore: refactor sheet.jsx and dont pass editorSaveState and editable …

stevekaplan123:
## Description
"Saved" indicator in sheets should not show up if I'm not editing a sheet and looking at someone else's sheet!

## Code Changes
1. Don't render EditorSaveStateIndicator if we're not editing right now!
2. Re-organize code in Sheet.jsx so that memory is used efficiently -- previously we were loading up SheetContent and Editor even though we are only using one of them at a time.  I also think the way I organized it makes it easier to read.

### DIFF
--- a/static/js/sheets/Sheet.jsx
+++ b/static/js/sheets/Sheet.jsx
@@ -133,8 +133,6 @@ class Sheet extends Component {
       const usingEditor = shouldUseEditor(sheet?.id);
       content = usingEditor ? this.getEditor(sheet) : this.getReader(sheet);
     }
-    return <div className={classes}>{content}</div>
-
   }
 }
 
--- a/static/js/sheets/Sheet.jsx
+++ b/static/js/sheets/Sheet.jsx
@@ -63,27 +63,8 @@ class Sheet extends Component {
      Sefaria.sheets._loadSheetByID[this.props.id].collections = Sefaria.getUserCollectionsForSheetFromCache(this.props.id);
      this.forceUpdate();
   }
-
-  render() {
-    const classes = classNames({sheetsInPanel: 1});
-    const sheet = this.getSheetFromCache();
-    const editable = Sefaria._uid === sheet?.owner;
-    let content, editor;
-    if (!sheet) {
-      content = (<LoadingMessage />);
-      editor = (<LoadingMessage />);
-    }
-    else {
-      const sidebar = <SheetContentSidebar
-                                  authorStatement={sheet.ownerName}
-                                  authorUrl={sheet.ownerProfileUrl}
-                                  authorImage={sheet.ownerImageUrl}
-                                  collections={sheet.collections}
-                                  toggleSignUpModal={this.props.toggleSignUpModal}
-                                  topics={sheet.topics}
-                                  editorSaveState={this.props.editorSaveState}
-                              />;
-        editor = <div className="sidebarLayout">
+  getEditor = (sheet) => (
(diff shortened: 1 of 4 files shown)


static/js/sheets/Sheet.jsx
```diff
@@ -103,41 +84,62 @@ class Sheet extends Component {
...
+          sheetID={sheet.id}
+          authorStatement={sheet.ownerName}
+          authorID={sheet.owner}
+          authorUrl={sheet.ownerProfileUrl}
+          authorImage={sheet.ownerImageUrl}
+          summary={sheet.summary}
+          toggleSignUpModal={this.props.toggleSignUpModal}
+          historyObject={this.props.historyObject}
+      />
+      <SheetContentSidebar
```
YishaiGlasner: this is a duplication. you can use `getSheetContentSidebar` (or maybe a shorter name, i don't know what's the meaning of content here, this is again old weird names)

static/js/sheets/Sheet.jsx
```diff
@@ -63,27 +63,8 @@ class Sheet extends Component {
...
-                                  authorStatement={sheet.ownerName}
-                                  authorUrl={sheet.ownerProfileUrl}
-                                  authorImage={sheet.ownerImageUrl}
-                                  collections={sheet.collections}
-                                  toggleSignUpModal={this.props.toggleSignUpModal}
-                                  topics={sheet.topics}
-                                  editorSaveState={this.props.editorSaveState}
-                              />;
-        editor = <div className="sidebarLayout">
+  getEditor = (sheet) => (
```
YishaiGlasner, line 66: i don't like the names `getEditor` and `getContent`. Im not sure there is something better, for the old components' names are not so clear.

static/js/sheets/Sheet.jsx
```diff
@@ -103,41 +84,62 @@ class Sheet extends Component {
...
+            toggleSignUpModal={this.props.toggleSignUpModal}
+            topics={sheet.topics}
+      />;
+    </div>
+  );
+  render() {
+    const classes = classNames({sheetsInPanel: 1});
+    const sheet = this.getSheetFromCache();
+    if (!sheet) {
+      return <div className={classes}><LoadingMessage /></div>
```
YishaiGlasner: this and line 138 duplicate the div with `className={classes}`.
I think it will be better to do the conditions for some variable (i'd name it `content`, but it's already use content (in a misleading way), so you can change the name of content of find another name), and then one return statement.

static/js/sheets/Sheet.jsx
```diff
@@ -103,41 +84,57 @@ class Sheet extends Component {
...
+  );
+  getSidebar = (sheet, editor) => (
+    <SheetSidebar
+          authorStatement={sheet.ownerName}
+          authorUrl={sheet.ownerProfileUrl}
+          authorImage={sheet.ownerImageUrl}
+          collections={sheet.collections}
+          toggleSignUpModal={this.props.toggleSignUpModal}
+          topics={sheet.topics}
+          editorSaveState={editor &&this.props.editorSaveState}
```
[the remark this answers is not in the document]
YishaiGlasner: copilot is right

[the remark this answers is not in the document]
stevekaplan123: Committed it (and also addressed the issues you mentioned previously).  Any other issues left?

YishaiGlasner (approved):

PR #2646, opened 2025-09-04
Feature/sc 36270/remove sheets

stevekaplan123:
## Description

## Code Changes
1.  Remove MODULE_ROUTES throughout since we no longer append "/sheets".  Likewise, lots and lots of front end changes but they just remove "/sheets" or change "sheets.sefaria.org" to "voices.sefaria.org".
2. I also changed the active_module options from being "sheets"/"library" to "voices"/"library".  For consistency, I thought we should change the sqllite DB so "sheets" topic pool is called "voices".  Therefore, django topic migration "0012" renames the sheets topics pool to voices but this turned out to be a problem because we only have one sqllite DB for all cauldrons, so all cauldrons' sheets topics pool end up being called 'voices'.  Doing this breaks the modularization cauldron, so I had to add to add django topic migration "0013".  It seems like we could just get rid of these for the time being but I wasn't sure if my local DB and the cauldrons DB required the history of the migrations files being present.  Now, for the time being, the topic pool is called 'sheets', which means I had to add code `get_topic_pool_name_for_module` in django_topics/utils.py and `getTopicPoolNameForModule` in sefaria.js to provide a mapping between 'sheets' and 'voices'.  Near the end of modularization project, we should consider removing "0013" so that "0012" renames the topics pool from being called 'sheets' to 'voices', and we can remove these mapping functions.
3. This PR contains a new helm chart but the helm chart needs modifications.  I discussed this in a comment in local-settings.yaml.
4. sefaria_tags.py: Now that we don't append "/sheets", the Django tag I added, context_url, isn't necessary for the Sefaria templates.  I changed it back to how it was on master whereby we call the built-in Django tag 'url' instead of the new 'context_url' tag.  However, I only commented out "context_url" rather than removing it.  I did this because though MOST templates work with 'url', the password reset template is broken, and can't be tested locally, and I wasn't sure what to do about it.  Out of confusion, I just wasn't sure if I should delete the 'context_url' tag just yet.  I created a story for the password reset template.
5. DOMAIN_MODULES has a new structure.  Instead of keys for 'voices' and 'library', there is a first level of keys for 'en' and 'he', each of which in turn has 2 keys 'voices' and 'library':
`DOMAIN_MODULES = {"en": {'voices': 'en domain for voices', 'library': 'en domain for library'}, 'he': {"voices": "he domain for voices", "library": "he domain for library"}}`
 I use `get_language_specific_domain_modules` in the backend to return a simpler structure.  If the interface lang is 'en', then  `get_language_specific_domain_modules` returns {'voices': 'en domain for voices', 'library': 'en domain for library'}.  That way, the front end is unchanged.
6. urls.py has been split into urls_library, urls_shared, and urls_sheets -- changed merged in from a PR made by @YishaiGlasner.  This helps us avoid supporting "voices.sefaria.org/texts" or "voices.sefaria.org/Genesis 3"
7. 'sheets-saved', 'texts-saved', 'sheets-history', and 'texts-history' were left as separate menuOpen modes, although the back-end was changed somewhat to handle the fact that these are almost identical cases.  This needs to be re-factored and re-thought, but I left it as is because there is a history/saved bug in "MDL bugs" and that is a better place to handle this re-factor.
## Notes
Sorry that this PR is huge...

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -3317,11 +3317,11 @@ def topic_pool_api(request, pool_name):
     
     # Map the pool_name to the correct topic pool based on active_module
     # If pool_name is 'voices', we need to use 'sheets' pool
-    actual_pool_name = get_topic_pool_name_for_module(pool_name)
+    expected_pool_name = get_topic_pool_name_for_module(pool_name)
     
     n_samples = int(request.GET.get("n"))
     order = request.GET.get("order", "random")
-    topic_slugs = DjangoTopic.objects.sample_topic_slugs(order, actual_pool_name, n_samples)
+    topic_slugs = DjangoTopic.objects.sample_topic_slugs(order, expected_pool_name, n_samples)
     response = [Topic.init(slug).contents() for slug in topic_slugs]
     return jsonResponse(response, callback=request.GET.get("callback", None))
 
@@ -3686,7 +3686,6 @@ def user_profile(request, username):
 
 
 @catch_error_as_json
-@csrf_exempt
 def profile_api(request, slug=None):
     """
     API for user profiles.
--- a/sefaria/helper/topic.py
+++ b/sefaria/helper/topic.py
@@ -15,6 +15,7 @@ from django_topics.utils import get_topic_pool_name_for_module
 import structlog
 from sefaria import tracker
 from sefaria.helper.descriptions import create_era_link
+from sefaria.constants.model import LIBRARY_MODULE
 logger = structlog.get_logger(__name__)
 
 def get_topic(v2, topic, lang, with_html=True, with_links=True, annotate_links=True, with_refs=True, group_related=True, annotate_time_period=False, ref_link_type_filters=None, with_indexes=True):
@@ -236,18 +237,18 @@ def annotate_topic_link(link: dict, link_topic_dict: dict) -> Union[dict, None]:
 
 
 @django_cache(timeout=24 * 60 * 60)
-def get_all_topics(limit=1000, displayableOnly=True, activeModule='library'):    
+def get_all_topics(limit=1000, displayableOnly=True, active_module=LIBRARY_MODULE):
     query = {"shouldDisplay": {"$ne": False}, "numSources": {"$gt": 0}} if displayableOnly else {}
     topic_list = TopicSet(query, limit=limit, sort=[('numSources', -1)])
     
-    # Get the actual pool name that should be used for this activeModule
-    expected_pool_name = get_topic_pool_name_for_module(activeModule)
+    # Get the actual pool name that should be used for this active_module
+    expected_pool_name = get_topic_pool_name_for_module(active_module)
     
     return [t for t in topic_list if expected_pool_name in t.get_pools()]
 
 @django_cache(timeout=24 * 60 * 60)
 def get_num_library_topics():
-    all_topics = DjangoTopic.objects.get_topic_slugs_by_pool('library')
+    all_topics = DjangoTopic.objects.get_topic_slugs_by_pool(LIBRARY_MODULE)
     return len(all_topics)
 
 


static/js/UserHistoryPanel.jsx
```diff
@@ -60,8 +60,8 @@ const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNa
     "history": "/texts/history"
   };
   const sheetsURLs = {
```
stevekaplan123, line 62: Re-factor without sheetsURLs since they are the same now?

static/js/ReaderApp.jsx
```diff
@@ -1186,13 +1185,13 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     } else if (path === "/texts/history") {
       this.showHistory();
 
-    } else if (path === "/sheets/history") {
+    } else if (path === "/history") {
       this.showSheetsHistory();
```
stevekaplan123, line 1185: this.showSheetsHistory and this.showSheetsSaved need to be re-thought

static/js/ReaderApp.jsx
```diff
@@ -827,9 +827,8 @@ class ReaderApp extends Component {
       hist.url += window.location.hash;
     }
     
-    console.log("Updating History - " + hist.url + " | " + currentUrl);
-    hist.url = Sefaria.util.modifyRelativePathbasedOnModule(hist.url); 
-    console.log("Updating History2 - " + hist.url + " | " + currentUrl);
+    // console.log("Updating History - " + hist.url + " | " + currentUrl);
+    // hist.url = Sefaria.util.modifyRelativePathbasedOnModule(hist.url);
```
stevekaplan123: should this function be removed entirely?

static/js/ReaderApp.jsx
```diff
@@ -584,22 +584,22 @@ class ReaderApp extends Component {
...
           case "texts-saved":
             hist.title = Sefaria._("My Saved Content");
-            hist.url = "texts/saved";
+            hist.url = "saved";
             hist.mode = "textsSaved";
             break;
           case "sheets-saved":
             hist.title = Sefaria._("My Saved Content");
-            hist.url = "sheets/saved";
+            hist.url = "saved";
```
stevekaplan123, line 591: do we need two cases for saved/history in sheets vs texts?

reader/views.py
```diff
@@ -1021,21 +1021,23 @@ def calendars(request):
...
+    Unified saved content view that works for both library and sheets modules
+    """
     title = _("My Saved Content")
     desc = _("See your saved content on Sefaria")
     profile = UserProfile(user_obj=request.user)
     props = {"saved": {"loaded": True, "items": profile.get_history(saved=True, secondary=False, serialized=True, annotate=True, limit=20)}}
-    return menu_page(request, props, page="sheets-saved", title=title, desc=desc)
+    
+    # Determine page name based on active module
+    active_module = getattr(request, 'active_module', 'library')
```
stevekaplan123: One function could maybe determine page name in saved_content, user_history_content and for the templates (subdomain_url)?

helm-chart/sefaria/conf/nginx.template.conf.tpl
```diff
@@ -255,7 +255,6 @@ http {
       opentracing on;
       opentracing_propagate_context;
       {{- end }}
-      rewrite ^/(.*)$ /{{ $v }}/$1 break;
```
stevekaplan123, line 272: should this be put back in?

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-        try:
-            from urllib.parse import urlparse
-            
-            current_host = request.get_host()
-            parsed_current = urlparse(f"http://{current_host}")
-            current_hostname = parsed_current.hostname
-            
-            for module_name, module_domain in DOMAIN_MODULES.items():
+        
+        from urllib.parse import urlparse
```
stevekaplan123, line 265: Removed code dealing with "/sheets" as it's no longer appended

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-        Determine and set the active module for the request.
-        First checks host-based module, then falls back to route-based detection.
-        """
-        # First, try to determine module based on host/subdomain from DOMAIN_MODULES
-        host_based_module = self._get_module_from_host(request)
-        if host_based_module:
-            active_module = host_based_module
-        else:
-            # Check each module route to see if path matches
-            for module_name, route_prefix in MODULE_ROUTES.items():
```
stevekaplan123, line 294: Removed this because we don't use MODULE_ROUTES anymore

build/startup/primeAll.bash
```diff
@@ -7,7 +7,7 @@ export TARGET_HOSTNAME=${TARGET_HOSTNAME:-localhost} # export makes the variable
 timeout ${TIMEOUT:-900} bash -c 'while [[ "$(curl -s -k -o /dev/null -w ''%{http_code}'' http://$TARGET_HOSTNAME/healthz-rollout)" != "200" ]]; do sleep 10; done'
 
 
-./primeLocalCache.bash
+# ./primeLocalCache.bash
```
yitzhakc, line 10: What does this do?

sefaria/system/context_processors.py
```diff
@@ -87,9 +87,10 @@ def base_props(request):
 
 @user_only
 def module_context(request):
+    from sefaria.utils.util import get_language_specific_domain_modules
```
YishaiGlasner: why no in the head of the file? we anyway import this file

sefaria/utils/util.py
```diff
@@ -523,6 +524,11 @@ def short_to_long_lang_code(code):
         code = "english"
     return code
 
+def get_language_specific_domain_modules(interfaceLang):
```
YishaiGlasner, line 527: i'm not sure this is the right place of this functiom. as a matter of fact, i'm not sure what exactly we have in this file, but that probably means it is the right place for onthing...
i don't think views helper should be here with some sort of pther helpers.

sefaria/views.py
```diff
@@ -120,7 +120,8 @@ def form_valid(self, form):
             from_email=self.from_email,
             extra_email_context=self.extra_email_context,
         )
-        return super().form_valid(form)
+        # Don't call super().form_valid(form) as it would send the email again
+        return HttpResponseRedirect(self.get_success_url())
```
YishaiGlasner, line 124: i don't understand this or how it is related to the pr

static/js/TopicPage.jsx
```diff
@@ -751,7 +751,7 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
 
       // Only add author works tab if in library module and has indexes
       if (indexes?.length && Sefaria.activeModule === Sefaria.LIBRARY_MODULE) {
-        tabs.push({
+        tabs.unshift({
```
YishaiGlasner, line 754: is this related to subdomains?

static/js/sefaria/sefaria.js
```diff
@@ -530,8 +531,15 @@ Sefaria = extend(Sefaria, {
     }
   },
   isSefariaURL: function(url) {
-    //change this to just check Sefaria.org not domain modules
-    return url.hostname.includes('sefaria.org');
+    if (!Sefaria.domainModules) {
```
YishaiGlasner: i dont understand this cindition

static/js/sefaria/sefaria.js
```diff
@@ -530,8 +531,15 @@ Sefaria = extend(Sefaria, {
...
+      return false;
+    }
+    for (const [name, href] of Object.entries(Sefaria.domainModules)) {
```
```
return Object.values(Sefaria.domainModules).some(href => url.href.startsWith(href));
```

static/js/sefaria/sefaria.js
```diff
@@ -2085,6 +2093,15 @@ _media: {},
         key:   poolName + queryString,
         store: this._TopicsByPool
     });
+  },
+  getTopicPoolNameForModule: function(activeModule) {
+    // Maps active_module to the correct topic pool name
+    // When active_module is 'voices', use 'sheets' pool
+    const moduleToPoolMapping = {
```
YishaiGlasner, line 2092: why not simply
```
return activeModule === 'library' ? 'library' : 'sheets';
```

reader/templatetags/sefaria_tags.py
```diff
@@ -527,26 +526,76 @@ def sheet_via_absolute_link(sheet_id):
...
+    """
+    Template tag that resolves URLs based on a specific domain, which is necessary in email templates where request context is not available.
+    Determines the module from the domain and generates the appropriate full URL.
+    Primarily used in email templates where request context is not available.
+    Falls back to regular URL resolution if module-specific URL doesn't exist.
+    """
+    module_name = None
+    
+    # Parse domain to get hostname if it's a full URL
+    from urllib.parse import urlparse
```
YishaiGlasner: move to head

reader/templatetags/sefaria_tags.py
```diff
@@ -527,26 +526,76 @@ def sheet_via_absolute_link(sheet_id):
 		'<a href="/sheets/{}">a sheet</a>'.format(sheet_id)))
 
 
-@register.simple_tag(takes_context=True)
```
YishaiGlasner, line 530: i don't understand this code, but ut bothers me that you say you changed it back to how it was on master - how can it be that it was much complicated where we had no subdomains? i wonder if we keep it more simple.

reader/views.py
```diff
@@ -3117,9 +3123,15 @@ def topic_page(request, slug, test_version=None):
     """
     Page of an individual Topic
     """
+    from django_topics.utils import get_topic_pool_name_for_module
```
YishaiGlasner, line 3131: move to head

sefaria/helper/topic.py
```diff
@@ -236,9 +236,15 @@ def annotate_topic_link(link: dict, link_topic_dict: dict) -> Union[dict, None]:
 
 @django_cache(timeout=24 * 60 * 60)
 def get_all_topics(limit=1000, displayableOnly=True, activeModule='library'):
+    from django_topics.utils import get_topic_pool_name_for_module
```
YishaiGlasner: move to head

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-        '/api/',
         '/interface/',
+        '/api/',
         '/apple-app-site-association',
         '/static/',
     }
 
     def __init__(self, get_response):
-        self.get_response = get_response
+        self.get_response = get_response
```
YishaiGlasner: remove spaces

sefaria/system/middleware.py
YishaiGlasner: move to head

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-            
-            for module_name, module_domain in DOMAIN_MODULES.items():
+        
+        from urllib.parse import urlparse
+        
+        current_host = request.get_host()
+        parsed_current = urlparse(f"http://{current_host}")
+        current_hostname = parsed_current.hostname
+        
+        for lang, module in DOMAIN_MODULES.items():
```
YishaiGlasner:
```
for module in DOMAIN_MODULES.values():
```

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
+            for module_name, module_domain in module.items():
                 parsed_domain = urlparse(module_domain)
```
YishaiGlasner: i think 
```
if current_hostname == urlparse(module_domain). hostname:
```
is clear, so no need for 2 extra vars

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-            for module_name, route_prefix in MODULE_ROUTES.items():
-                if request.path.startswith(route_prefix):
-                    active_module = module_name
-                    break
-        
-        return active_module
 
     def __call__(self, request):
+        request.active_module = 'library'
         if not self.should_process_request(request):
```
YishaiGlasner, line 275: what's the benefit here? it seems `_get_module_from_host` is very light so why skip it?

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
-            # Check each module route to see if path matches
-            for module_name, route_prefix in MODULE_ROUTES.items():
-                if request.path.startswith(route_prefix):
-                    active_module = module_name
-                    break
-        
-        return active_module
 
     def __call__(self, request):
+        request.active_module = 'library'
```
YishaiGlasner: i think this is not so clear here. i prefer the previous version

sefaria/system/middleware.py
```diff
@@ -246,64 +247,45 @@ def process_response(self, request, response):
...
         if not self.should_process_request(request):
-            request.active_module = 'library'
             return self.get_response(request)
 
-        request.active_module = self._set_active_module(request)
+        active_module_from_host = self._get_module_from_host(request)
```
YishaiGlasner: after prefering the previous version here you can do
```
request.active_module = self._set_active_module(request) or 'librart'
```

django_topics/migrations/0012_rename_sheets_pool_to_voices.py
```diff
@@ -0,0 +1,40 @@
...
+
+def rename_sheets_to_voices(apps, schema_editor):
+    """Rename the 'sheets' pool to 'voices' in the database."""
+    TopicPool = apps.get_model('django_topics', 'TopicPool')
+    
+    try:
+        sheets_pool = TopicPool.objects.get(name='sheets')
+        sheets_pool.name = 'voices'
+        sheets_pool.save()
+        print("Successfully renamed 'sheets' pool to 'voices'")
```
yitzhakc: Better to use logging commands, and not prints

django_topics/migrations/0013_revert_voices_to_sheets.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
yitzhakc, line 1: As discussed, there should be a story changing this to have no operations (`operations = []`)

django_topics/models/pool.py
```diff
@@ -4,7 +4,7 @@
 
 class PoolType(Enum):
     LIBRARY = "library"
-    SHEETS = "sheets"
+    SHEETS = "sheets"
```
yitzhakc: Extra whitespace?

django_topics/utils.py
```diff
@@ -0,0 +1,45 @@
+"""
+Utility functions for django_topics app.  The module_to_pool_mapping and pool_to_module_mapping can be removed once we rename the sheets pool to voices.
```
yitzhakc, line 2: Make sure to add this to the story mentioned above.

reader/templatetags/sefaria_tags.py
```diff
@@ -17,8 +17,7 @@
 from django.contrib.sites.models import Site
 from django.utils.translation import ugettext as _
 
-from sefaria.local_settings import MODULE_ROUTES
-from sefaria.urls import get_module_url_name
+from sefaria.local_settings import DOMAIN_MODULES
```
yitzhakc: You should not be importing directly from settings files.
Please use 
```
from django.conf import settings

... settings.DOMAIN_MODULES ...
```

reader/templatetags/sefaria_tags.py
```diff
@@ -527,26 +526,76 @@ def sheet_via_absolute_link(sheet_id):
...
+        base_domain = base_domain.rstrip('/')
+        if not relative_path.startswith('/'):
+            relative_path = '/' + relative_path
+        return base_domain + relative_path
+    
+    # Fallback: return relative path (for cases where module not in DOMAIN_MODULES)
+    return relative_path
+
+
+# TODO: we may not need this anymore
```
yitzhakc: Can you please expand on why we may not need it, and how we know if we don't need it (and when can we remove it :-) )

reader/templatetags/sefaria_tags.py
```diff
@@ -527,26 +526,76 @@ def sheet_via_absolute_link(sheet_id):
...
-                    pass    
-
-    # Default URL resolution (used when request is not available or no subdomain match)
     try:
         return reverse(url_name, args=args, kwargs=kwargs)
     except NoReverseMatch:
         return '#'
+
+
+def _resolve_url_with_module(module_name, url_name, *args, **kwargs):
```
yitzhakc: Please add this to the list of functions to test in our test story

reader/views.py
```diff
@@ -287,11 +286,15 @@ def reader_redirect(uref):
...
         return response
 
+    active_module = getattr(request, "active_module", "library")
+
     for version in ['ven', 'vhe']:
         if request.GET.get(version) and '|' not in request.GET.get(version):
             return _reader_redirect_add_languages(request, tref)
 
     if sheet is None:
+        if active_module != 'library':
```
yitzhakc: Do we not have `library` as a constant somewhere, e.g. `LIBRARY_MODULE`? If we don't, please add it, and change all the places where it's hardcoded to use the constant.

reader/views.py
yitzhakc: Definitely - you could probably have a dictionary that maps the page name based on the key, with the function just getting the dictionary based on the module.

reader/views.py
```diff
@@ -3128,7 +3144,7 @@ def topic_page(request, slug, test_version=None):
     if request.active_module == "library":
         title = short_title + " | " + _("Texts from Torah, Talmud and Sefaria's library of Jewish sources.")
         desc = _("Jewish texts about %(topic)s from Torah, Talmud and other sources in Sefaria's library.") % {'topic': short_title}
-    elif request.active_module == "sheets":
+    elif request.active_module == "voices":
```
yitzhakc: Here too, we should have a constant for this hardcoded value of `voices`

sefaria/system/middleware.py
```diff
@@ -14,6 +14,7 @@
 from sefaria.site.site_settings import SITE_SETTINGS
 from sefaria.model.user_profile import UserProfile
 from sefaria.utils.util import short_to_long_lang_code, get_lang_codes_for_territory
+from sefaria.settings import DOMAIN_MODULES
```
yitzhakc: As above, it is not recommended to directly import from settings. Rather, use `from django.conf import settings`, and then `settings.DOMAIN_MODULES`

sefaria/utils/util.py
```diff
@@ -8,6 +8,7 @@
 from functools import wraps
 from itertools import zip_longest
 from sefaria.constants.model import ALLOWED_TAGS_IN_ABSTRACT_TEXT_RECORD
+from sefaria.settings import DOMAIN_MODULES
```
yitzhakc: As above.

sefaria/local_settings_ci.py
```diff
@@ -19,21 +19,14 @@
...
     "library": "http://localhost:8000",
-    "sheets": "http://localhost:8000",
-}
-MODULE_ROUTES = {
-    "library": "/",
-    "sheets": "/sheets/",
+    "voices": "http://voices.localhost:8000",
 }
-
+ALLOWED_HOSTS = ['127.0.0.1', "0.0.0.0", '[::1]'] + ["localhost", "voices.localhost"]
```
yitzhakc: No need to separate these lists, you can make them a single list

sefaria/local_settings_example.py
```diff
@@ -38,27 +38,23 @@
     }
 }"""
 
+
+################ These are things you can change! ###########################################################################
```
yitzhakc: Maybe you could write this a bit clearer? :-) 

`User-defined settings`?

reader/templatetags/sefaria_tags.py
YishaiGlasner: @yitzhakc  we are doing in many places
```
from sefaria.settings import
```
(but not
```
from sefaria.local_settings import
```
)

sefaria/model/autospell.py
```diff
@@ -344,8 +344,8 @@ def _type_to_active_module(self, co):
             "ref": ["library"],
             "Term": ["library"],
             "TocCategory": ["library"],
-            "User": ["sheets"],
-            "Collection": ["sheets"],
+            "User": ["voices"],
```
yitzhakc: Use a constant here please (for both `voices` and `library`)

sefaria/model/notification.py
```diff
@@ -353,7 +353,7 @@ def _build_query_with_scope(self, uid, read=None, is_global=None, suspected_spam
             query["is_global"] = is_global
         if suspected_spam is not None:
             query["suspected_spam"] = suspected_spam
-        query["type"] = {"$in" if scope == 'sheets' else "$nin": Notification.sheets_notification_types}
+        query["type"] = {"$in" if scope == 'voices' else "$nin": Notification.sheets_notification_types}
```
yitzhakc: Use a constant here please

static/js/sefaria/sefaria.js
```diff
@@ -2085,6 +2093,15 @@ _media: {},
         key:   poolName + queryString,
         store: this._TopicsByPool
     });
+  },
+  getTopicPoolNameForModule: function(activeModule) {
```
yitzhakc, line 2089: Add this to the story of renaming the `sheets` pool that we've discussed please.

static/js/sefaria/sefaria.js
```diff
@@ -3536,7 +3555,7 @@ _media: {},
     }
   },
   getLogoutUrl: () => {
-    const next = Sefaria.activeModule === 'sheets' ? 'sheets' : 'texts';
+    const next = Sefaria.activeModule === Sefaria.SHEETS_MODULE ? 'sheets' : 'texts';
```
yitzhakc: Can we double check this? I thought that when we're in `SHEETS_MODULE` there is no longer a `/sheets` url, and it's just `/`.

static/js/sefaria/util.js
```diff
@@ -293,21 +293,9 @@ class Util {
...
-      else if (Sefaria.activeModule !== Sefaria.SHEETS_MODULE && path.startsWith(sheetsPrefix)) {
-        // If we are not in the sheets module, remove /sheets from the beginning of the URL
-        return path.replace(sheetsPrefix, "");
-      }
-      return path;
-    }
     static fullURL(relativePath, moduleTarget) {
       if (relativePath.startsWith("/")) { // if the path is relative, prepend the module URL
-        const moduleURL = Sefaria.getModuleURL(moduleTarget); // derive the host URL from the module target (e.g. 'https://sheets.sefaria.org' or 'https://www.sefaria.org')
+        const moduleURL = Sefaria.getModuleURL(moduleTarget); // derive the host URL from the module target (e.g. 'https://voices.sefaria.org' or 'https://www.sefaria.org')
```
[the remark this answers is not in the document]
yitzhakc: Not sure why this comment still exists ;-)

static/js/Header.jsx
```diff
@@ -134,16 +134,14 @@ const LoggedInDropdown = ({module}) => {
...
   return (
       <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
           <div className='dropdownLinks-options'>
-              <DropdownMenuItem url={libraryURL} newTab={Sefaria.activeModule !== Sefaria.LIBRARY_MODULE} targetModule={Sefaria.LIBRARY_MODULE}>
+              <DropdownMenuItem url={"/"} newTab={Sefaria.activeModule !== Sefaria.LIBRARY_MODULE} targetModule={Sefaria.LIBRARY_MODULE}>
                   <DropdownMenuItemWithIcon icon={'/static/icons/library_icon.svg'} textEn={"Library"}/>
               </DropdownMenuItem>
               <DropdownMenuSeparator/>
-              <DropdownMenuItem url={sheetsURL} newTab={Sefaria.activeModule !== Sefaria.SHEETS_MODULE} targetModule={Sefaria.SHEETS_MODULE}>  
+              <DropdownMenuItem url={"/sheets"} newTab={Sefaria.activeModule !== Sefaria.SHEETS_MODULE} targetModule={Sefaria.SHEETS_MODULE}>
```
yitzhakc: Is this supposed to be `/`? Or we're leaving `/sheets`? (related to previous comment as well)

static/js/HeaderAutocomplete.jsx
```diff
@@ -87,15 +87,15 @@ function sortByTypeOrder(array) {
 
 const getURLForObject = function(type, key) {
```
yitzhakc, line 88: Should this function have a catch-all error handling in case it's not any of these cases?

static/js/ReaderApp.jsx
yitzhakc: Perhaps not, but it should be a separate story 😄

static/js/ReaderApp.jsx
yitzhakc: As above... We can do it in a separate story

static/js/UserHistoryPanel.jsx
yitzhakc: Separate story 😉

templates/email/notification.html
```diff
@@ -27,7 +27,7 @@
 
 	{% elif notification.type == "discuss" %}
 		{{ notification.content.adder|user_link|absolute_link }} {% trans "added a note to a" %}
-        <a href="https://www.sefaria.org{{ notification.content.discussion_path|url_safe }}">
+        <a href="{{ domain_modules.library }}{{ notification.content.discussion_path|url_safe }}">
```
yitzhakc, line 30: Discussions are on library?

templates/registration/logged_out.html
```diff
@@ -10,7 +10,7 @@
 <div id="loggedOut" >
 
 	<p id="message">
-		You are now logged out.  <a href="{% subdomain_url "login" %}">Log back in</a>.
+		You are now logged out.  <a href="{% url "login" %}">Log back in</a>.
```
yitzhakc, line 13: Why isn't this a subdomain_url? We should be able to know from the host of the request which login page to redirect to

templates/registration/login.html
```diff
@@ -14,7 +14,7 @@
                 <span class="int-en">You are already logged in as </span>
                 <span class="int-he">הנכם מחוברים כבר בתור</span>
                 {{ user.email }}.
-                <a href="{% subdomain_url "logout" %}">
+                <a href="{% url "logout" %}">
```
yitzhakc, line 17: As above

yitzhakc (changes requested):

reader/templatetags/sefaria_tags.py
stevekaplan123: What I meant was that the tag subdomain_url was not used on master.  I had added it this summer to modularization but eventually realized we didn't need it.

build/startup/primeAll.bash
stevekaplan123: Akiva added this line a few weeks ago

reader/views.py
stevekaplan123: I actually re-factored saved and history in another PR and think we should deal with these issues there.  Closing this for now.

static/js/TopicPage.jsx
stevekaplan123: Not exactly.  unshift adds the element to the beginning of the array.  This is a UI requirement for authors pages.

sefaria/utils/util.py
stevekaplan123: I'm not sure I follow.  Which file should contain the helper functions?

sefaria/views.py
stevekaplan123: I shouldn't have included this in the PR, but for some reason the password reset email would get sent twice upon calling form_valid(form).  form.save() already sends the email so here a redirect is all that should be returned.  It was somewhat relevant to the PR because changing all the routes was affecting templates like the password reset template and so I found this bug in QA.

templates/email/notification.html
stevekaplan123: I genuinely don't know.  I'm making a story to investigate this.

templates/registration/logged_out.html
stevekaplan123: We don't need the subdomain_url tag anymore because teh template URLs are now in urls_shared.py and behave the same in either module. For example,
    url(fr'^login/?$', sefaria_views.CustomLoginView.as_view(), name='login'),

static/js/HeaderAutocomplete.jsx
stevekaplan123: I don't think so because it's being called inside a mapping function.  One simple approach would be to filter out objects with null URL values if none of these conditions apply.

sefaria/system/middleware.py
stevekaplan123: If "should_process_request" returns false, we knew we were in the library module.  Now, with your changes to urls.py we may not need this anymore but I don't think it should be part of this PR

reader/views.py
```diff
@@ -3296,9 +3313,15 @@ def topic_graph_api(request, topic):
 @catch_error_as_json
 def topic_pool_api(request, pool_name):
     from django_topics.models import Topic as DjangoTopic
+    from django_topics.utils import get_topic_pool_name_for_module
+    
+    # Map the pool_name to the correct topic pool based on active_module
+    # If pool_name is 'voices', we need to use 'sheets' pool
+    actual_pool_name = get_topic_pool_name_for_module(pool_name)
```
yitzhakc: I don't fully understand what the flow is here. What could the possible variance be between `pool_name` and `actual_pool_name`?

In addition, elsewhere in this file you used the expression `expected_pool_name`. Would that be correct here? Or perhaps `module_pool_name`? I feel like `actual_pool_name` vs `pool_name` is still a bit confusing..

reader/views.py
```diff
@@ -3663,6 +3686,7 @@ def user_profile(request, username):
 
 
 @catch_error_as_json
+@csrf_exempt
```
yitzhakc: Since this view can receive a `POST` request, it's best to not exempt it from `csrf`.

sefaria/helper/topic.py
```diff
@@ -235,10 +236,14 @@ def annotate_topic_link(link: dict, link_topic_dict: dict) -> Union[dict, None]:
 
 
 @django_cache(timeout=24 * 60 * 60)
-def get_all_topics(limit=1000, displayableOnly=True, activeModule='library'):
+def get_all_topics(limit=1000, displayableOnly=True, activeModule='library'):
```
yitzhakc: Remove the whitespace?

Also, `library` should be a constant please

yitzhakc: I know you didn't make this as part of this PR, but I since we're touching it, please change the `activeModule` to be `active_module` (since in python we name variables `snake_case` and not `camelCased`)

sefaria/utils/util.py
yitzhakc: I think it's fine, unless @YishaiGlasner there's a specific place you wanted it to go?

yitzhakc (changes requested):

reader/views.py
stevekaplan123: I'll change it to expected_pool_name.  The issue is we get passed 'voices' as the 'pool_name', but currenly the actual pool_name is 'sheets'.  There is a dedicated story to clean this up so that we will have 'voices' across the board and wont need these mapping functions.

PR #2654, opened 2025-09-08
chore: re-factor SheetsHomePage

stevekaplan123:
## Description
CSS and ordering of sheets home page on mobile were buggy.

## Code Changes
Two issues:  
1. NavSidebar includes SidebarFooter by default.  However, in this case, we actually want NavSidebar to display in the middle of the page, so I added a parameter 'includeFooter' to NavSidebar. 
2. The sheets home page is meant to look very differently in desktop vs mobile, but we only have one component, SheetsHomepage, which gets passed a parameter 'multiPanel' to determine how to display.  This was already awkward before, but with the change in (1) above (i.e. passing 'includeFooter' to NavSidebar), SheetsHomepage would become very hard to read, so I decided to break it into two components, one for desktop and one for mobile.  'multiPanel' now simply determines which component gets displayed.

### DIFF
--- a/static/js/NavSidebar.jsx
+++ b/static/js/NavSidebar.jsx
@@ -8,7 +8,7 @@ import { Promotions } from './Promotions'
 import {SignUpModalKind} from "./sefaria/signupModalContent";
 import Button from "./common/Button";
 
-const NavSidebar = ({sidebarModules}) => {
+const NavSidebar = ({sidebarModules, includeFooter = true}) => {
   return <div className="navSidebar sans-serif">
     {sidebarModules.map((m, i) =>
       <SidebarModules
@@ -16,7 +16,7 @@ const NavSidebar = ({sidebarModules}) => {
         props={m.props || {}}
         key={i} />
     )}
-    <SidebarFooter />
+    {!!includeFooter && <SidebarFooter />} 
   </div>
 };
 
--- a/static/js/sheets/SheetsHomePage.jsx
+++ b/static/js/sheets/SheetsHomePage.jsx
@@ -1,5 +1,5 @@
 import React  from 'react';
-import {NavSidebar} from "../NavSidebar";
+import {NavSidebar, SidebarFooter} from "../NavSidebar";
 import {SheetsTopicsCalendar, SheetsTopicsTOC} from "./SheetsHomePageTopicsTOC";
 const SheetsHeroBanner = ({title, message, videoOptions, posterImg}) => {
     /*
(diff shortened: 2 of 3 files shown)


static/js/NavSidebar.jsx
```diff
@@ -8,15 +8,15 @@ import { Promotions } from './Promotions'
 import {SignUpModalKind} from "./sefaria/signupModalContent";
 import Button from "./common/Button";
 
-const NavSidebar = ({sidebarModules}) => {
+const NavSidebar = ({sidebarModules, includeFooter = true}) => {
```
YishaiGlasner, line 12: great opportunity to add proptypes

static/js/sheets/SheetsHomePage.jsx
```diff
@@ -19,36 +19,68 @@ const SheetsHeroBanner = ({title, message, videoOptions, posterImg}) => {
...
+                                  sheetsTopicsTOC={sheetsTopicsTOC} />;
+  }
+  else {
+    return <SheetsHomePageMobile sheetsHeroBanner={sheetsHeroBanner} 
+                                 sheetsTopicsCalendar={sheetsTopicsCalendar} 
+                                 sheetsTopicsTOC={sheetsTopicsTOC} />;
+  }     
+}
+ 
+const SheetsHomePageMobile = ({sheetsHeroBanner, sheetsTopicsCalendar, sheetsTopicsTOC}) => {
```
YishaiGlasner: please add proptypes

static/js/sheets/SheetsHomePage.jsx
```diff
@@ -19,36 +19,68 @@ const SheetsHeroBanner = ({title, message, videoOptions, posterImg}) => {
...
-                    {!multiPanel && sheetsTopicsTOC}
+                    <SheetsHomePageSidebar includeFooter={false} />
+                    {sheetsTopicsTOC}
+                    <div className="sans-serif"><SidebarFooter /></div>
                 </div>
             </div>
         </div>
 }
+
+const SheetsHomePageDesktop = ({sheetsHeroBanner, sheetsTopicsCalendar, sheetsTopicsTOC}) => {
```
YishaiGlasner: this is almost complete duplication. it seems there are only 3 different lines. i think one component will be better. it would have `{multiPanel && sheetsTopicsTOC}` in one place, `{!multiPanel && sheetsTopicsTOC}` in the other and `includeFooter={!multiPanel}`

stevekaplan123: Technically more lines of code are required, but I personally think that the next engineer who has to deal with the sheets home page will appreciate the two components as there's no logic in either component and they are easy to understand.  What do you think?

YishaiGlasner: i don't care of lines. but the next engineer that will want to add a classname, for example. will have to do it twice. and if someone will forget to do it in both, we'll have undesired difference. this is the reason for DRY.
but if you think it's probable we will want to have other differences, we should think of a way to do 2 components with DRY.

stevekaplan123: For now, I'll make it one component, but I think if the product design changes, we should break it up.

yitzhakc: I'm inclined to agree with @YishaiGlasner that ideally we want to keep this a single component, so long as the _functionality_ remains the same (or a subset). But I also agree with @stevekaplan123 that as the design changes, we need to explore how better to break that down.

static/js/sheets/SheetsHomePage.jsx
```diff
@@ -20,36 +21,70 @@ const SheetsHeroBanner = ({title, message, videoOptions, posterImg}) => {
...
+  const sheetsTopicsTOC = <SheetsTopicsTOC handleClick={setNavTopic}/>;
+
+  return (
+    <div className="readerNavMenu sheetsHomepage" key="0">
+      <div className="content">
+        {sheetsHeroBanner}
+        <div className="sidebarLayout">
+          <div className="contentInner">
+            <div className="sheetsTopics">
+              {multiPanel ? (
```
YishaiGlasner: i think that can be cleaner:
```
<>
{sheetsTopicsCalendar}
{multiPanel  && sheetsTopicsTOC}
</>
```

stevekaplan123: Fair enough.  Changed it

PR #2655, opened 2025-09-08
Updates 5 strings and added hebrew support

schreibersefaria:
I updated 5 strings in the voices landing page.
Currently there was no Hebrew support so I added the string to `strings.js` and used `Sefaria._()`
Two of the strings existed in `strings.js` so I used them as is.

I checked the En and He work.

### DIFF
--- a/static/js/sefaria/strings.js
+++ b/static/js/sefaria/strings.js
@@ -160,6 +160,9 @@ const Strings = {
 
     // Sheets
     "Sheet": "דף מקורות",
+    "Community-Powered Jewish Learning": "חיבורים חדשים למקורות ולקהילה",
+    "Share. Discover. Join the Conversation.": "המקום ללמידה, שיתוף רעיונות, וכתיבת חומרים מקוריים.",
+    "Browse Topic Categories": "חיפוש מקורות לפי נושא",
     "About this Sheet": "אודות דף המקורות",
     "Tags": "תוויות",
     "Search for a text...": "חפש מקור בספריא",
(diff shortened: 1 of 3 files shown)


stevekaplan123 (approved):

static/js/sheets/SheetsHomePage.jsx
```diff
@@ -1,6 +1,7 @@
 import React  from 'react';
 import {NavSidebar} from "../NavSidebar";
 import {SheetsTopicsCalendar, SheetsTopicsTOC} from "./SheetsHomePageTopicsTOC";
+import Sefaria from "../sefaria/sefaria";
```
YishaiGlasner, line 4: i think we don't need this (and maybe it also undesired). i think we have Sefaria in the global with some magic

static/js/sefaria/strings.js
```diff
@@ -160,6 +160,9 @@ const Strings = {
 
     // Sheets
     "Sheet": "דף מקורות",
+    "Community-Powered Jewish Learning": "חיבורים חדשים למקורות ולקהילה",
+    "Share. Discover. Join the Conversation.": "המקום ללמידה, שיתוף רעיונות, וכתיבת חומרים מקוריים.",
+    "Browse Topic Categories": "חיפוש מקורות לפי נושא",
```
YishaiGlasner, line 165: is this unique to sheets?

schreibersefaria: For now it is.
I think ordering the Strings.js may deserve it's own task because some of the strings already existed and they were in Topics or other such titles.
Where would you put them? If we have an idea I'll make an update

PR #2658, opened 2025-09-08
chore: only want _CAT_REF_LINK filtering in library

stevekaplan123:
## Description

## Notes

### DIFF
--- a/static/js/TopicPage.jsx
+++ b/static/js/TopicPage.jsx
@@ -704,7 +704,7 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
       return displayTabs;
     }
 
-    const setupAdditionalTabs = () => {
+    const setupAdditionalTabs = (displayTabs) => {
       // Setup additional tabs that are not related to sources: such as filter, language toggle, and author works on Sefaria (this last one is only relevant if we're in the library module and the topic is an author)
       // Finally, return 'onClickFilterIndex' and 'onClickLangToggleIndex' to be used by 'TabView' component so that it knows which tab in 'displayTabs'
       // corresponds to the filter and lang toggle
@@ -712,7 +712,8 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
       let onClickLangToggleIndex = 2; // lang toggle tab defaults to the 3rd tab
       const indexes = topicData?.indexes;
 
-      if (indexes?.length) {
+      // Only add author works tab if in library module and has indexes
+      if (indexes?.length && Sefaria.activeModule === Sefaria.LIBRARY_MODULE) {
         displayTabs.push({
           title: {en: "Works on Sefaria", he: Sefaria.translation('hebrew', "Works on Sefaria")},
           id: 'author-works-on-sefaria',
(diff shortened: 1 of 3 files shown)


static/js/TopicPage.jsx
```diff
@@ -765,34 +766,37 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
...
-                                      renderItem={renderWrapper(toggleSignUpModal, topicData, topicTestVersion, langPref)}
-                                      onSetTopicSort={onSetTopicSort}
-                                      topicSort={topicSort}
-                                    />
-                                  );
-                               }});
+    let [onClickLangToggleIndex, onClickFilterIndex] = setupAdditionalTabs(displayTabs);
+    
+    const tabContent = [];
+    displayTabs.forEach(tabObj => {
```
YishaiGlasner: why did you change `map` to `forEach`? `map` seems to be the exact thing you want to do.

also, i don't understand the way you changed the handling of author items here. it seems to much different than before.

static/js/TopicPage.jsx
```diff
@@ -810,10 +814,14 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
                                     [onClickFilterIndex]: ()=>setShowFilterHeader(!showFilterHeader),
                                     [onClickLangToggleIndex]: ()=>{setShowLangSelectInterface(!showLangSelectInterface)}
                                   }}>
-                          {authorIndices}{tabsWithSources}
+                          {tabContent}
                         </TabView>;
 
-    // Currently, there are data inconsistencies in the DB, so we should show LoadingMessage if there are no tabs to display.
+    // If there are no tabs with content, don't render anything
```
YishaiGlasner: maybe you should explain in what case it can happen

static/js/sheets/Sheet.jsx
```diff
@@ -133,6 +133,7 @@ class Sheet extends Component {
       const usingEditor = shouldUseEditor(sheet?.id);
       content = usingEditor ? this.getEditor(sheet) : this.getReader(sheet);
     }
+    return <div className={classes}>{content}</div>;
```
YishaiGlasner, line 136: i don't understand this - we had no return before?! so how did it work?

yitzhakc: Steve wrote in the notes of the PR that it was removed by accident in a previous PR

static/js/TopicPage.jsx
```diff
@@ -704,15 +704,16 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
       return displayTabs;
     }
 
-    const setupAdditionalTabs = () => {
+    const setupAdditionalTabs = (displayTabs) => {
```
yitzhakc: Did this also exist previously and was removed accidentally at some point?
Did it work when `displayTabs` wasn't passed in here as a parameter?

stevekaplan123: It actually worked even though it wasn't passed.  But it seemed worthwhile to restore it.

static/js/TopicPage.jsx
```diff
@@ -656,6 +656,39 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
     }
     const currentLang = getCurrentLang(langPref);
 
+    const generateDisplayTabs = (tabMetaDataList) => {
+      let displayTabs = [];
+      tabMetaDataList.forEach(tabObj => {
```
YishaiGlasner, line 661: why not using `map`? making new empty array and pushing into is what map does, isn't it?

stevekaplan123: 'map' would keep all elements.  I don't want all of them.  2 elements of the tab metadata list are filter and language buttons.  this is due to the weirdness of TabView.

stevekaplan123: I'll add a comment

YishaiGlasner: ok, you're saying that the previous implementation was wrong, for it has `undefined`s?

YishaiGlasner (approved):

static/js/TopicPage.jsx
stevekaplan123: Right.  At some point they need to be filtered out.

static/js/TopicPage.jsx
```diff
@@ -810,11 +822,17 @@ const TopicPageTabView = ({topic, topicData, tab, setTab, translationLanguagePre
...
-                          {authorIndices}{tabsWithSources}
+                          {displayTabs}
                         </TabView>;
 
-    // Currently, there are data inconsistencies in the DB, so we should show LoadingMessage if there are no tabs to display.
-    return (!topicData.isLoading && displayTabs.length) ? topicTabView : <LoadingMessage />; 
+    if (topicData.isLoading) {
+      return <LoadingMessage />;
+    }
+    else if (tabMetaDataList.length === 0) {
```
yitzhakc, line 831: What is the new logic here? It's not the same as previously.

Before, we would display loading unless both `isLoading` is false AND there are tabs.
Was that a bug? We're purposely splitting that? Is returning `null` here something that has been handled and that there's a string that says "No works", or something?

yitzhakc (changes requested):

static/js/TopicPage.jsx
stevekaplan123: The previous logic seemed convoluted and wrong.  We have author pages that have no books or sheets and therefore shouldn't display "Loading"

PR #2660, opened 2025-09-09
Update existing 2 sidebar sections

YishaiGlasner:
## Description
Swith order of 'create a sheet' and the about secions.
Change texts in both.
In mobile, different texts for create a sheet and removes the button.


static/js/NavSidebar.jsx
```diff
@@ -793,9 +793,11 @@ const StayConnected = () => {
 };
 
 const GetStartedButton = () => {
-    const href = Sefaria._v({"en": "/sheets/393695", "he": "/sheets/399333"})
+    const href = Sefaria._v({"en": "/sheets/674324", "he": "/sheets/674327"})
```
schreibersefaria: I'm not sure what the task was exactly so this may not be relevant.
In production we are using a param here.
If you are already making a change to this maybe it's good to fix this.

Production line: https://github.com/Sefaria/Sefaria-Project/blob/17090b93ee3498fb6400010f8714b4aead327351/static/js/NavSidebar.jsx#L200

static/js/NavSidebar.jsx
```diff
@@ -793,9 +793,11 @@ const StayConnected = () => {
...
     return <Button variant="secondary" className="getStartedSheets">
-      <a href={href} data-target-module={Sefaria.SHEETS_MODULE}>Get Started</a>
+      <a href={href} data-target-module={Sefaria.SHEETS_MODULE}>
+          <InterfaceText text={{'en': 'Learn More', 'he': 'למידע נוסף'}} />
```
schreibersefaria, line 799: Would `Sefaria._` be a better solution so we can then support more languages if needed.

static/js/NavSidebar.jsx
```diff
@@ -807,18 +809,28 @@ const CreateSheetsButton = () => {
...
-);
+const CreateASheet = () => {
+    let enText, heText;
+    if (Sefaria.multiPanel) {
+        enText = 'Mix and match sources from the Sefaria Library, along with outside sources, images, videos, and your own commentary, to share digitally.';
+        heText = 'כל דף הוא בגדר לוח חלק, מרחב בו תוכלו לחבר בין הרעיונות שלכם לבין המקורות ולהוסיף מקורות מדיה מגוונים. השתמשו בכלים המוצעים כדי לכתוב מחשבות ותובנות אודות המקורות שבספרייה, להוסיף קטעי וידאו ואודיו, לצטט מהמקורות שבספרייה.';
+    } else {
+        enText = 'Use a computer to mix and match sources from the Sefaria Library, along with outside sources, images, videos, and your own commentary. The Voices Editor is not supported on mobile devices.';
+        heText = 'כל דף הוא בגדר לוח חלק, מרחב בו תוכלו לחבר בין הרעיונות שלכם לבין המקורות ולהוסיף מיני מדיה מגוונים. לא ניתן ליצור או לערוך דפים באמצעות המכשיר הנייד. על מנת לגשת לעורך הדפים של ספריא, יש להיכנס לאתר האינטרנט ״חיבורים״ במחשב.';
+    }
```
schreibersefaria, line 820: Here too maybe we should use `Sefaria._`.
It's a very long text so it's a bit funny to use it but it should functionally be the same.

schreibersefaria (approved):
Left a few comments but overall it looks straight forward so you can fix or not, whatever you think.

static/js/NavSidebar.jsx
YishaiGlasner: we already have 'Learn More' with different Hebrew...
another reason not to use the strings as keys

static/js/NavSidebar.jsx
YishaiGlasner: here we're using TitledText that demands a `text` object. I'm not sure why this is was done is that way, but this is a code written for MDL, so i assumed it should be ok

static/js/NavSidebar.jsx
YishaiGlasner: good point!
i really don't know why we're using here sheets and not the help center

static/js/NavSidebar.jsx
schreibersefaria: Potentially you could use different keys within strings.
But if we don't have a good solution we like let's flag this @yitzhakc

stevekaplan123 (commented):
Why aren't we using ZenDesk for these help sheets? @YishaiGlasner

PR #2665, opened 2025-09-10
Make Cookie banner cookie work across modules.

schreibersefaria:
Added logic to save the cookie for the banner to a cookie domain.
So for example, in both  https://modularization.cauldron.sefaria.org/ and https://sheets.modularization.cauldron.sefaria.org/ the cookie will be saved in `.cauldron.sefaria.org`

I have added a util func to findLongestCommonSuffix

### DIFF
--- a/static/js/sefaria/util.js
+++ b/static/js/sefaria/util.js
@@ -421,6 +421,82 @@ class Util {
         
         return commonSuffix;
     }
+    
+    /**
+     * Determines the appropriate cookie domain for cross-subdomain cookie sharing.
+     * 
+     * This function analyzes Sefaria.domainModules to find the common parent domain
+     * that all configured modules share, then uses that as the cookie domain.
+     * 
+     * Why this approach is better:
+     * - Uses the actual configured domains rather than guessing from current hostname
+     * - Automatically adapts to any domain configuration without hardcoded logic
+     * - Leverages the existing domain module system that's already working
+     * 
+     * How it works:
+     * 1. Extract hostnames from all domain modules (e.g., "www.sefaria.org", "sheets.sefaria.org")
+     * 2. Use Sefaria.util.findLongestCommonSuffix() to find the longest common suffix that all hostnames share
+     * 3. Use that as the cookie domain (e.g., ".sefaria.org")
+     * 
+     * Examples:
+     * - Production: ["www.sefaria.org", "sheets.sefaria.org"] → ".sefaria.org"
+     * - Cauldron: ["modularization.cauldron.sefaria.org", "sheets.modularization.cauldron.sefaria.org"] → ".modularization.cauldron.sefaria.org"
+     * - Development: ["localhost:8000", "localhost:8000"] → null (no domain set)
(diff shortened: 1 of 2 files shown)


static/js/sefaria/util.js
```diff
@@ -381,6 +381,122 @@ class Util {
...
+        
+        // IP addresses don't support subdomains, so no cross-subdomain cookie sharing possible
+        if (hostnames.some(hostname => /^\d+\.\d+\.\d+\.\d+/.test(hostname))) {
+            return null;
+        }
+        
+        // Browsers don't allow setting cookies with domain ".localhost" - this is a security feature
+        // For localhost development, we need to dismiss the cookie banner on each module
+        if (hostnames.some(hostname => hostname === 'localhost' || hostname.includes('.localhost'))) {
+            return null;
```
stevekaplan123, line 480: Is there a way to avoid all of this special casing?

static/js/sefaria/util.js
```diff
@@ -381,6 +381,122 @@ class Util {
...
+        }
+        
+        // Find the longest common suffix
+        const commonSuffix = Util.findLongestCommonSuffix(hostnames);
+        
+        if (commonSuffix && commonSuffix.length > 0) {
+            // Special handling for domain suffixes that don't start with "."
+            // This happens when we have a mix of bare domains and subdomains:
```
stevekaplan123, line 488: I don't understand this case.  Does this apply to cauldrons or only production?

stevekaplan123: Does this code assume DOMAIN_MODULES['library'] is 'sefaria.org' rather than 'www.sefaria.org'?

static/js/sefaria/util.js
```diff
@@ -381,6 +381,122 @@ class Util {
             index++;
         return str1.substring(0, index);
     }
+    
+    /**
+     * Finds the longest common suffix among an array of strings
```
stevekaplan123, line 386: We already have a function in util.js:
```
static commonSubstring(str1, str2){
        const length = Math.min(str1.length, str2.length);
        let index = 0;
        while(index<length && str1[index] === str2[index])
            index++;
        return str1.substring(0, index);
    }
```

static/js/sefaria/util.js
schreibersefaria: Yes that is a good example of when this could happen.

static/js/sefaria/util.js
schreibersefaria: That function checks the prefix and not suffix

static/js/sefaria/util.js
schreibersefaria: Well we could do one big `if` but it will be less readable.

schreibersefaria: I could combine some and extract others to functions as well.

schreibersefaria: How is that?

yitzhakc (approved):

PR #2667, opened 2025-09-11
Fix settings page can't save

schreibersefaria:
The problem: [Account Settings Page can't be saved](https://app.shortcut.com/sefaria/story/36630)
Added the missing CSRF JavaScript to the account settings template
This is an area of the code I'm not familiar so please review carefully. Try see this as more of me asking if this solution looks good than a PR. Happy to huddle and discuss.
Cauldron for testing https://www.mdl-setting-save-fix.cauldron.sefaria.org/texts


schreibersefaria:
Turns out this isn't a bug https://app.shortcut.com/sefaria/story/36711

PR #2668, opened 2025-09-11
Notification Badge

yodem:
## Description
Update notification count retrieval to consider active module, improve notification read marking, and refine CSS for header notifications display

## Code Changes
- Added Scope as active module for the notification retrieval in NotificationPanel and in the backend.
- Restored CSS of old notification icon to use badge

## Notes
- I suggest to store the pagination data in the url.
- I think that adding popup for the notifications so the user wont have to go to the notifications page, will be a better UX.


static/css/s2.css
```diff
@@ -1597,6 +1598,9 @@ a.interfaceLinks-option.active, a.interfaceLinks-option.inactive {
 .bookPage.compare .content {
   padding: 30px 10px;
 }
+.singlePanel .readerNavMenu .content .contentInner {
+  min-height: revert;
+}
```
yodem, line 1696: This isnt my code, its from steve's branch that relates to - [removeshh.cauldron.sefaria.org](http://removeshh.cauldron.sefaria.org/)

static/js/Header.jsx
```diff
@@ -186,6 +185,10 @@ const Header = (props) => {
     }
   }, []);
 
+  const unread = props.headerMode ? ((Sefaria.notificationCount > 0) ? 1 : 0) : Sefaria.notificationCount > 0 ? 1 : 0
```
stevekaplan123: I think this line's logic could definitely be simplified

static/js/Header.jsx
```diff
@@ -286,7 +280,18 @@ const Header = (props) => {
...
-        {Sefaria._uid && (props.module === Sefaria.LIBRARY_MODULE ? librarySavedIcon : sheetsNotificationsIcon)}
+        {Sefaria._uid && (props.module === Sefaria.LIBRARY_MODULE ? librarySavedIcon :
+          <div className='sheetsNotificationsHeaderIcon'>
+            <a
+              href="/notifications"
+              data-target-module={Sefaria.VOICES_MODULE}
+              onKeyDown={(e) => Util.handleKeyboardClick(e)}
+              className={notificationsClasses}
+            >
+              <img src='/static/icons/notification.svg' alt={Sefaria._("Notifications")} />
```
stevekaplan123: I think it would be easier to read if it were " ? librarySavedIcon : sheetsSavedIcon" rather than a whole div here.

stevekaplan123 (approved):

PR #2674, opened 2025-09-14
Chore/sc 36593/update favicons everywhere

relyks:
## Description

## Code Changes

## Notes


sefaria/system/middleware.py
```diff
@@ -264,8 +264,9 @@ def _get_module_from_host(self, request):
             from urllib.parse import urlparse
             
             current_host = request.get_host()
-            parsed_current = urlparse(f"http://{current_host}")
-            current_hostname = parsed_current.hostname
+            # Extract hostname directly
```
stevekaplan123: Why re-invent the wheel to parse a URL?

sefaria/system/middleware.py
```diff
@@ -284,23 +285,30 @@ def _set_active_module(self, request):
...
         First checks host-based module, then falls back to route-based detection.
+        Sets is_subdomain_access flag.
         """
-        # First, try to determine module based on host/subdomain from DOMAIN_MODULES
-        host_based_module = self._get_module_from_host(request)
-        if host_based_module:
-            active_module = host_based_module
-        else:
-            # Check each module route to see if path matches
+        # Start with hostname-based detection, fall back to default
```
stevekaplan123: Why is it necessary to change the request and add 'is_subdomain_accesss' to match both "/sheets" and "/sheets/"?  Can't regexes in urls.py handle that?

sefaria/system/middleware.py
relyks: This is a mistake. I was messing with the code here to see how that worked. I'm going to revert what you had before because it worked fine

sefaria/system/middleware.py
relyks: I decided to create a view to dynamically generate web manifests that web browsers use frequently. Chrome-based browsers attempt to retrieve the web manifest on every first page load during a session. `is_subdomain_access` isn't relevant here, but I use it later on in that view. `is_subdomain_access` is used to determine what URLs to provide in the manifest. However, I no longer need to have it as part of the context processors for templates, so I will remove it from there. Also, I think the hostname isn't available when the URL route is processed?

I technically don't need to have it part of the request object, and I could determine it there in the view function, so passing it through the request object isn't completely necessary. I thought it'd be better to determine it here, since it's already in a place where it can be defined, and it'd be faster than determining it in the view. When I'm in the view, I'd essentially be doing similar logic there. If it's something that is only needed in that view, do you think I should just do it there?

Are you okay with the other changes I made to how I determine the `active_module`? When I have both `DOMAIN_MODULES` set to `http://localhost:8000`, the previous code doesn't work to determine the correct `active_module`.

Maybe we or I can make some tests to ensure that this works correctly? I think this will be important to ensure the user gets the right branding

Let me know if anything I said doesn't make sense?

stevekaplan123: I don't fully understand the above comment and probably we should find a time to huddle.  When do you think is an overlap between our working times?  I work standard 9-5 hours.  As for middleware.py, we are in the midst of a major re-factor in this branch: feature/sc-36270/remove-sheets (see https://app.shortcut.com/sefaria/story/36270/remove-sheets)
In this branch, we decided to remove /sheets that is at the beginning of the route and therefore removed MODULE_ROUTES entirely.  "/sheets/{sheet_id}" and "/sheets/new" are the only major cases where we use "/sheets" now.  This approach hasn't been tested as cauldrons were down yesterday.  But you should know about it for this project, especially because middleware.py code is in the process of being changed in feature/sc-36270/remove-sheets

reader/views.py
```diff
@@ -4825,3 +4857,40 @@ def is_database_reachable():
...
+    """
+    active_module = module
+    lang = getattr(request, 'interfaceLang', 'english')
+
+    # Set start_url to ensure users land on the correct module entry point when launching PWA
+    if active_module == DEFAULT_MODULE or active_module not in MODULE_ROUTES:
+        start_url = "/"
+    else:
+        # For modules like sheets, use their route path as the entry point
+        # Ensures users go to /sheets/ rather than just / when launching the PWA
```
yitzhakc: With the changes being made in #2646 , the path to sheets will also be `/`, but at the relevant subdomain, and we will be removing `MODULE_ROUTES`.

templates/static/annualreport.html
```diff
@@ -14,6 +16,14 @@
             position:absolute; left: 0; right: 0; bottom: 0; top: 0px;
         }
     </style>
+
+    <link rel="shortcut icon" href="{{ STATIC_PREFIX }}icons/library/favicon.ico?v=2" sizes="48x48">
```
yitzhakc, line 20: Why `?v=2`? Won't this just update eventually based on `Cache-Control`?

yitzhakc (commented):
We should also double check we're not planning on changing the `key` of the module` to `voices`

templates/static/annualreport.html
relyks: Yeah, it should, but some clients are stupid and don't follow the headers for favicons once they've been downloaded. It's a good idea for the future. If we change the favicons again, that URL parameter acting as a cache buster should be changed to `?v=3`, so that they'll change as soon as there is a new deploy.

reader/views.py
relyks: I only use the `MODULE_ROUTES` for dynamically creating the manifest. Maybe I should separate this part into a different PR and get back to this once you guys finalize how the middleware will work? I'm still confused about the paths

relyks:
> We should also double check we're not planning on changing the `key` of the module`to`voices`

Yeah, unfortunately, I think that might need to be known upfront, because I use the key as part of the file path. I can determine that dynamically based on a constant, but that seems wasteful to check it every time?

Maybe creating a new enum class for module names/keys might be a good idea? If you define a new enum class, you can have it so:
```
MODULES.VOICES  == "voices"   # True
MODULES.VOICES  == "sheets"   # True
MODULES.LIBRARY == "sheets"    # False
```

reader/views.py
relyks: Will `voices.sefaria.org` be the same as `voices.sefaria.org/sheets`?

yitzhakc: @stevekaplan123 Can you confirm this?

yitzhakc:
> > We should also double check we're not planning on changing the `key` of the module`to`voices`
> 
> Yeah, unfortunately, I think that might need to be known upfront, because I use the key as part of the file path. I can determine that dynamically based on a constant, but that seems wasteful to check it every time?
> 
> Maybe creating a new enum class for module names/keys might be a good idea? If you define a new enum class, you can have it so:
> 
> ```
> MODULES.VOICES  == "voices"   # True
> MODULES.VOICES  == "sheets"   # True
> MODULES.LIBRARY == "sheets"    # False
> ```

Ok, we definitely **are** changing the key of the module to `voices` (see #2646).
Would you like to wait until we've merged that into `modularization-main`? Or change the target merge branch of the PR?

templates/base.html
```diff
@@ -48,17 +48,24 @@
...
+    <link rel="shortcut icon" href="{{ STATIC_PREFIX }}icons/{{ active_module }}/favicon.ico?v=2" sizes="48x48">
+    <link rel="icon" type="image/svg+xml" href="{{ STATIC_PREFIX }}icons/{{ active_module }}/favicon.svg?v=2" sizes="any">
+    <link rel="icon" type="image/png" href="{{ STATIC_PREFIX }}icons/{{ active_module }}/favicon.png?v=2" sizes="96x96">
+    <link rel="apple-touch-icon" sizes="180x180" href="{{ STATIC_PREFIX }}icons/{{ active_module }}/apple-touch-icon.png?v=2">
+    <link rel="manifest" href="/site.webmanifest?v=2">
+    {# In the future in case there are more than 2 modules, define a custom filter with a dictionary that maps module names to colors/strings #}
+    <meta name="theme-color" content="{% if active_module == 'voices' %}#5D956F{% else %}#18345D{% endif %}">
+    <meta name="apple-mobile-web-app-title"
+          content="{% if active_module == 'voices' %}
+                    {% if lang == 'hebrew' or lang == 'he' %}חיבורים{% else %}Voices{% endif %}
```
yitzhakc: Are we able to use the Django-translated strings in templates, instead of hardcoding these?

reader/views.py
```diff
@@ -4758,6 +4758,38 @@ def apple_app_site_association(request):
...
+    
+    # Get the active module and default to the same default as middleware
+    active_module = getattr(request, 'active_module', DEFAULT_HOST)
+    
+    # Construct the path to the module-specific icon file
+    # Filename comes from the capture group in the URL pattern
+    icon_path = os.path.join(STATICFILES_DIRS[0], 'icons', active_module, filename)
+    
+    # Set appropriate content type based on file extension
+    extensions_to_content_types_map = {
```
yitzhakc: This is a constant, it could be moved out of the function

templates/base.html
relyks: Changed, though I don't like the whole setup for internationalization in Django lol

reader/views.py
relyks: Yes

templates/base.html
yitzhakc: It's a bit heavy handed, that's for sure..

yitzhakc (approved):

PR #2675, opened 2025-09-14
Removed the scaling of the sheet page - fixing font size issue

schreibersefaria:
The font size was too small.
This was because a patch that reduced the font size to `font-size: 62.5%`
It seems like the core issue was solved so removing the scaling should be OK.

The commit that introduced the [font-size problem](https://github.com/Sefaria/Sefaria-Project/commit/dab6fa7520d1fd3fd9a4a136beddb78f3d04e188)

Fix language-toggle alignment (broke after the initial change) by vertically centering sheet header items.


yodem (approved):
LGTM

PR #2678, opened 2025-09-15
Account settings save and cancel redirect

yodem:
## Description
Converted legacy HTML elements to semantic, accessible components and consolidated duplicate JavaScript handlers for better maintainability

## Code Changes
**HTML/Template Changes:**
- **Cancel buttons**: Converted `<a href="/my/profile">` links to `<button type="button" onclick="history.back()">` for proper browser history navigation
- **Save buttons**: Converted `<div>` elements with `role="button"` to semantic `<button type="button">` elements
- **Class consolidation**: Replaced duplicate IDs (`saveAccountSettings`, `saveAccountSettingsBottom`) with shared class `saveAccountSettingsBtn`

**JavaScript Changes:**
- **Bottom Save Button Fixed**: The bottom save button didn't work, probably because it has the same ID as the top button, so the "onclick" functionality didn't work. Instead for duplicate code with different ID, I assigned the functionality on class name instead of ID selector.
- **Single event handlers**: Consolidated two separate click/keyup handlers into one using class selector `$(".saveAccountSettingsBtn")`
- **Improved maintainability**: Save logic now exists in one place instead of two identical functions
- NavigateBack function - if the history length is greater than 2 (default for browsers) - go back. Otherwise go to /texts. it looks like reading the history url is a security issue so you cant do that.

## Note: React Migration Needed
This page should be converted to React for consistency with the rest of the Sefaria codebase. The current jQuery-based approach with manual DOM manipulation should be replaced with:
- React components for form controls and buttons
- State management for form data
- React Router for navigation instead of `history.back()`
- Event handlers as component methods instead of jQuery selectors

The semantic HTML structure implemented here will translate well to JSX components.

### DIFF
--- a/templates/account_settings.html
+++ b/templates/account_settings.html
@@ -12,7 +12,7 @@
             <span class="int-he">הגדרות חשבון</span>
         </h1>
         <div class="end">
-            <button type="button" class="button small transparent control-elem" onclick="handleCancel()">
+            <button type="button" class="button small transparent control-elem" onclick="navigateBack()">
                 <span class="int-en">Cancel</span>
                 <span class="int-he">ביטול</span>
             </button>
@@ -159,7 +159,7 @@
                 <span class="int-en">Save</span>
                 <span class="int-he">שמירה</span>
             </button>
-            <button type="button" class="button transparent small control-elem" onclick="handleCancel()">
+            <button type="button" class="button transparent small control-elem" onclick="navigateBack()">
                 <span class="int-en">Cancel</span>
                 <span class="int-he">ביטול</span>
             </button>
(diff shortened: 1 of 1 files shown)


templates/account_settings.html
```diff
@@ -12,14 +12,14 @@ <h1>
             <span class="int-he">הגדרות חשבון</span>
         </h1>
         <div class="end">
-            <a class="button small transparent control-elem" href="/my/profile">
+            <button type="button" class="button small transparent control-elem" onclick="history.back()">
```
YishaiGlasner: what will happen if i came directly to this url?
definitiley rare scenario, but maybe we should handle it.

yodem: Fixed - go to texts if history.length is less than 2 - the default of browsers (as far as I saw)
Regards to the 40 lines - I changed the description it wasnt clear/true. basically the submit button code didnt work for the bottom button because they had similar IDs. so fixing it was either duplicate the 40 lines of code or implement it by classname selector. I did the second solution

YishaiGlasner (dismissed):

PR #2682, opened 2025-09-16
Fix globe switching language

YishaiGlasner:
## The problem
The switching of language from the globe popover goes to the homepage (while it isn't happening in the profile menu language toggle).

## Description
1. First step was to refactor both toggles to the same component (both has the same title, and same 2 buttons, but with different styles).
2. Turns out that the probelm was that the `next` param in the toggle got the current path from a static function, so it isn't changing after it's mounted (it didn't happen in the other toggle, because its parent has a hook that triggers re-render). For solving it I made a new custom hook `useNextParam`, and refactor the code to use it in any place we need to use next param.

## Code Changes
For #1 I've changed `InterfaceLanguageMenu` (in Misc.jsx) to use `DropdownLanguageToggle`, and added the prop `currentLang` to it. This prop adds `active` classname to the active language (which needed in the InterfaceLanguageMenu). Now I had to move logic in the css, for getting two different styles menus from the same component.
For #2 I added   `useNextParam` hook and removed `getCurrentPage` static function (Misc.jsx). I refactored the code to use it in 2 cases:
1. Where it has used `getCurrentPage` (some places in Header.jsx and `DropdownLanguageToggle` in Dropdown.jsx).
2. Where we had duplicate logic for getting the `next` param other places in Header.jsx).

## Notes
Unfortunately, we have our own router. That means we hardly can track it, so the new hook has no dependencies and it triggers too many re-renders (this was also the situation in some places of Header.jsx).
We can (a) work on the new hook to find a way to reduce it, (b) return to our logic in master branch where menus are re-rendered when they are open and closed, or (c) live with it, it seems not to be so bad.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -714,6 +714,10 @@ div:has(#bannerMessage) + .readerApp.singlePanel .mobileNavMenu {
   padding-block-start: 5px;
   padding-block-end: 2px;
 }
+.globeLanguageToggle .languageHeader {
+  padding: 12px;
+  border-bottom: 1px solid #CCCCCC;
+}
 
 .header .interfaceLinks {
   display: flex;
@@ -805,7 +809,9 @@ div:has(#bannerMessage) + .readerApp.singlePanel .mobileNavMenu {
   padding: 4px;
   max-width: 220px;
 }
-.header .interfaceLinks .interfaceLinks-option {
+.header .interfaceLinks .interfaceLinks-option,
+.globeLanguageToggle .englishLanguageLink,
+.globeLanguageToggle .hebrewLanguageLink {
   display: flex;
   text-decoration: none;
   font-style: normal;
@@ -819,14 +825,12 @@ div:has(#bannerMessage) + .readerApp.singlePanel .mobileNavMenu {
 .header .interfaceLinks .interfaceLinks-option:hover {
   background-color: var(--lighter-grey);
 }
-.header .languageFlex .interfaceLinks-option:hover {
-  background-color: var(--lighter-grey);
-  text-decoration: none;
-}
 .header .interfaceLinks .interfaceLinks-menu.open {
   display: block;
 }
-.interface-hebrew .header .interfaceLinks .interfaceLinks-option.int-bi {
+.interface-hebrew .header .interfaceLinks .interfaceLinks-option.int-bi,
+.interface-hebrew .globeLanguageToggle .englishLanguageLink,
+.interface-hebrew .globeLanguageToggle .hebrewLanguageLink {
   direction: rtl;
 }
 .interface-english .header .interfaceLinks .interfaceLinks-option.int-bi {
@@ -888,7 +892,9 @@ div.interfaceLinks-row a {
 .header .profile-menu-he {
   color: #666666;
 }
-.header .interfaceLinks-option::before {
+.header .interfaceLinks-option::before,
+.globeLanguageToggle .hebrewLanguageLink::before,
+.globeLanguageToggle .englishLanguageLink::before {
   content: "";
   font-family: FontAwesome;
   color: #999;
@@ -905,11 +911,15 @@ a.interfaceLinks-option.int-bi.inactive {
 a.interfaceLinks-option.active, a.interfaceLinks-option.inactive {
   padding: 7px;
 }
-.header .interfaceLinks-option.active::before {
+.header .interfaceLinks-option.active::before,
+.globeLanguageToggle .hebrewLanguageLink.active::before,
+.globeLanguageToggle .englishLanguageLink.active::before {
   content: "\f00c";
   padding: 0 8px;
 }
-.header .interfaceLinks-option.inactive::before {
+.header .interfaceLinks-option.inactive::before,
+.globeLanguageToggle .hebrewLanguageLink::before,
+.globeLanguageToggle .englishLanguageLink::before {
   content: "";
   padding: 0 15px;
 }
@@ -15057,6 +15067,9 @@ span.ref-link-color-3 {color: blue}
 .header .headerDropdownMenu .dropdownLinks-menu a {
   color: #000000;
 }
+.header .headerDropdownMenu .dropdownLinks-menu .globeLanguageToggle a {
+  color: #666666;
+}
 .sheetsInPanel .headerDropdownMenu a.headerDropdownMenu::after {
   display: inline-block;
   height: 10px;
@@ -15428,6 +15441,19 @@ span.ref-link-color-3 {color: blue}
   height: 23px;
   padding-inline: 15px;
 }
+.globeLanguageToggle .dropdownLanguageToggle {
+  flex-direction: column;
+  align-items: unset;
+  height: unset;
+  padding: 4px 0;
+}
+.interface-hebrew .globeLanguageToggle .dropdownLanguageToggle {
+  flex-direction: column-reverse;
+}
+.globeLanguageToggle .englishLanguageLink:hover,
+.globeLanguageToggle .hebrewLanguageLink:hover {
+  background-color: var(--lighter-grey);
+}
 .interface-english .dropdownLanguageToggle a.hebrewLanguageLink,
 .interface-hebrew .dropdownLanguageToggle a.englishLanguageLink {
   color: #666666;
@@ -15438,11 +15464,14 @@ span.ref-link-color-3 {color: blue}
   padding: 10px 2px;
 }
 
-
-
 .englishLanguageButton::after {
   content: "•";
   padding: 6px;
+}
+
+.globeLanguageToggle .englishLanguageButton::after {
+  content: none;
+  padding: unset;
 }
 
 .profilePicAndButtonContainer {
@@ -15549,11 +15578,6 @@ span.ref-link-color-3 {color: blue}
 .librarySavedIcon img{
   width: 100%;
   height: 100%;
-}
-.languageFlex {
-  display: flex;
-  flex-direction: column;
-  direction:ltr;
 }
 
 @-webkit-keyframes load5 {
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -10,7 +10,6 @@ import {
   GlobalWarningMessage,
   InterfaceLanguageMenu,
   InterfaceText,
-  getCurrentPage,
   LanguageToggleButton,
   DonateLink,
   useOnceFullyVisible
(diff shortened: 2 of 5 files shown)


static/js/Header.jsx
```diff
@@ -19,25 +18,18 @@ import {ProfilePic} from "./ProfilePic";
 import {HeaderAutocomplete} from './HeaderAutocomplete'
 import { DropdownMenu, DropdownMenuSeparator, DropdownMenuItem, DropdownMenuItemWithIcon, DropdownLanguageToggle } from './common/DropdownMenu';
```
stevekaplan123: Is 'isClient' being used anymore in either component?

static/js/common/DropdownMenu.jsx
```diff
@@ -139,28 +140,31 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
+  const hebrewLink = Sefaria.util.fullURL(`/interface/hebrew?next=${next}`);
   return (
     <>
       <div className="languageHeader">
         <InterfaceText>Site Language</InterfaceText>
       </div>
       <div className='dropdownLanguageToggle'>
       <span className='englishLanguageButton'>
-        <a className="englishLanguageLink" href={englishLink}>
+        <a className={`englishLanguageLink ${(currentLang === 'english') ? 'active': ''}`} href={englishLink}>
```
stevekaplan123: I think we should use Sefaria._v to determine 'active' status instead of currentLang == 'english'

static/js/Hooks.jsx
```diff
@@ -273,6 +273,13 @@ function usePaginatedLoad(fetchDataByPage, setter, identityElement, numPages, re
   }, [fetchPage]);
 }
 
+function useNextParam() {
+  const [next, setNext] = useState(null);
```
stevekaplan123: can you add a description for this hook please?

static/js/common/DropdownMenu.jsx
YishaiGlasner: @stevekaplan123 you're right,. any reason not to use simply Sefaria.interfaceLang?

static/js/Hooks.jsx
```diff
@@ -273,6 +273,17 @@ function usePaginatedLoad(fetchDataByPage, setter, identityElement, numPages, re
   }, [fetchPage]);
 }
 
+function useNextParam() {
```
yodem: Because we use the native web history API, we can add a functionality to its functions. so in order to avoid the huge amount of re-renders, I tried to add a functionality to update the next param once the history changes, and only if there was a change.

also, consider to add a param to the hook of prefix for the nextParam, then you could just return the fixed url all at once. not sure if neccesary.
```
function useNextParam(prefix = null) {
  // Initialize with current path, URL-encoded for safe use in query parameters
  const [nextParam, setNextParam] = useState(() => 
    encodeURIComponent(Sefaria.util.currentPath())
  );

  useEffect(() => {
    // Function to update the nextParam when path changes
    const updateNextParam = () => {
      const newPath = encodeURIComponent(Sefaria.util.currentPath());
      // Only update state if path actually changed to prevent unnecessary re-renders
      setNextParam(prev => prev !== newPath ? newPath : prev);
    };

    // Handle browser back/forward navigation
    const handlePopState = () => updateNextParam();
    
    // Store original methods to restore on cleanup
    const originalPushState = history.pushState;
    const originalReplaceState = history.replaceState;

    // Override pushState to detect programmatic navigation
    history.pushState = function(...args) {
      originalPushState.apply(this, args);
      // Use setTimeout to ensure URL has been updated before checking
      setTimeout(updateNextParam, 0);
    };

    // Override replaceState to detect URL replacements
    history.replaceState = function(...args) {
      originalReplaceState.apply(this, args);
      // Use setTimeout to ensure URL has been updated before checking
      setTimeout(updateNextParam, 0);
    };

    // Listen for browser navigation events
    window.addEventListener('popstate', handlePopState);

    // Cleanup: restore original methods and remove event listeners
    return () => {
      window.removeEventListener('popstate', handlePopState);
      history.pushState = originalPushState;
      history.replaceState = originalReplaceState;
    };
  }, []); // Empty dependency array - only run on mount/unmount

  // If prefix is provided, return the complete URL with next parameter
  if (prefix) {
    const separator = prefix.includes('?') ? '&' : '?';
    return `${prefix}${separator}next=${nextParam}`;
  }

  // Otherwise, return just the encoded path
  return nextParam;
}
```

static/css/s2.css
```diff
@@ -805,7 +809,9 @@ div:has(#bannerMessage) + .readerApp.singlePanel .mobileNavMenu {
   padding: 4px;
   max-width: 220px;
 }
-.header .interfaceLinks .interfaceLinks-option {
+.header .interfaceLinks .interfaceLinks-option,
+.globeLanguageToggle .englishLanguageLink,
+.globeLanguageToggle .hebrewLanguageLink {
```
yodem, line 849: isnt it better to create a new classname?

static/js/Header.jsx
```diff
@@ -601,8 +582,8 @@ const ProfilePicMenu = ({len, url, name}) => {
                 <InterfaceText>Account Settings</InterfaceText>
               </a></div>
               <div className="interfaceLinks-row languages">
-                <a className={`${(Sefaria.interfaceLang == 'hebrew') ? 'active':''}`} href={`/interface/hebrew?next=${getCurrentPage()}`} id="select-hebrew-interface-link">עברית</a>
-                <a className={`${(Sefaria.interfaceLang == 'english') ? 'active':''}`} href={`/interface/english?next=${getCurrentPage()}`} id="select-english-interface-link">English</a>
+                <a className={`${(Sefaria.interfaceLang == 'hebrew') ? 'active':''}`} href={`/interface/hebrew?next=${currentUrl}`} id="select-hebrew-interface-link">עברית</a>
+                <a className={`${(Sefaria.interfaceLang == 'english') ? 'active':''}`} href={`/interface/english?next=${currentUrl}`} id="select-english-interface-link">English</a>
```
yodem: Maybe its out of scope but consider the following: 
- Make it the same experience for the user. either its side by side (from the avatar element) or one after the other (globe)
- Disable the current lang. why would the user click on the current lang?

static/css/s2.css
YishaiGlasner: i have created new classname - `globeLanguageToggle`.
in cases where i was sure the old classname isn't used somewhere else i've removed it.

static/js/Header.jsx
YishaiGlasner: it's more than out of scope - it demands product decisions...

static/js/Hooks.jsx
YishaiGlasner: i'll do 2 hooks, one for the current path as is, and one using it to generate param string.
regarding your added functionality, i afraid there are issues here, and i'm not sure we should work on it at this point. in this suggestion we're overriding history methods and using setTimeout. i'm not sure what that can cause, so i had a discussion with chatGPT. it raises some concerns: if multiple parts of the app also override these methods, race conditions if something immediately inspects state after navigation, it's used in multiple components so it will patch history multiple times (can be fixed with singleton), unexpected interactions of listeners when using by more than one component.
so i'm not sure about this. in the bottom line the component that re-render are not big so it isn't so bug issue.
@yitzhakc what do you think about it?

yodem: 1. if its for generating param string you dont need another hook you can create a util function.
2. we do not **override** we **add** functionality. this way anyone can add its own and it would not affect us.
3. regards to the timeout - its to ensure that the update state will happen only after updating the history.

static/js/Hooks.jsx
```diff
@@ -273,16 +273,23 @@ function usePaginatedLoad(fetchDataByPage, setter, identityElement, numPages, re
...
   });
-  return next;
+  return currentPath;
+}
+
+function useNextParam() {
+  /*
+  Returns a url next param with the current path (e.g. "next=/Genesis 1".
+   */
+  return `next=${encodeURIComponent(useCurrentPath())}`;
```
yodem: if it happens everytime you dont need a seperate function for it, i think its redundent. you can return an array where the first item is the text with "next=" and the second is only the currentPath and then support couple of functionalities. another solution is to return an object and deconstruct only the relevant data.

Option 1:
hook returns:
return ['next=${currentPath}',currentPath]

usage:
const [next] = useNextParam()

Option 2:

return {
currentPath,
next: 'next=${currentPath}'
}

usage:
const {next} = useNextParam()

YishaiGlasner: a util function that uses hook is a hook.
returning an array or object is not clean (especially an array, but also an object). one function should do one thing. otherwise, things go worse and worse.

yodem: A hook is basically a function that implement react features, so if there is a function that doesnt use react features - its a function.
regards to the returning way - you can see useState hooks that returns an array and tanstack(for example) uses hooks with deconstruct so its acceptable. I dont think we need that because the desired functionality is to return the full path with the next param so I think it would be better to get the login path (for example) as a param and return the full path with the next param appended to it.
However, if you would like to have both functionalities: return the next param by itself and the full path with the param appended - the currect way is to use an array or an object.

YishaiGlasner: as i've said any function calls hook is a hook - it uses react features.
returning array is acceptable but not clean. we don't want the full path, for it can have different params we don't know, it's too board. we want the string `next=...`. i thnk you idea to add a hook for not only getting the current path but also for getting the 'next=' is good for it reduce duplication so we need 2 hooks.

yodem: what other function are you talking about? i think the hook is enouth to determine the next param.
Again, I think returning array is ok as react decided to do with useState.
I also disagree about what we want, because if every time you append what you return to a string, it would be cleaner to pass the suffix as a param. However, It doesnt really matter, so I would go with whatever you like.
I just didnt get what 2 hooks are we talking about?

static/js/common/DropdownMenu.jsx
```diff
@@ -34,6 +34,22 @@ const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose
...
+    window.location.href = `${url}?next=${encodeURIComponent(Sefaria.util.currentPath())}`;
+  }
+  return (
+    <a className={className || 'interfaceLinks-option int-bi dropdownItem'}
+       href='#'
+       onClick={(e) => onClick(e)}
+       onKeyDown={(e) => Util.handleKeyboardClick(e, onClick)}
+    >
+      {children}
+    </a>
```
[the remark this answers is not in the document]
YishaiGlasner: copilot, didn't you read my description?

PR #333, opened 2017-06-08
add nodes to haamek davar 

stevekaplan123:

PR #2687, opened 2025-09-16
Tooltip for module switcher 

yodem:
## Description
Added an interactive tooltip component to inform users about recent platform changes and guide them to specialized areas for learning, creating, and extending digital Torah content. The tooltip appears positioned relative to navigation elements and can be dismissed permanently via localStorage.

## Code Changes
The following changes were made to the files below
static/js/common/ModuleSwitcherTooltip.jsx - New component implementing Tippy.js tooltip with multilingual markdown content, interactive buttons, and smart positioning logic
static/js/Header.jsx - Integrated tooltip component into header navigation, added ref management for target elements, and passed responsive layout props
static/css/Header.css - Added custom styling for tooltip appearance including light blue background, black border, button styling, and responsive width adjustments
package.json - Added @tippyjs/react and tippy.js dependencies for tooltip functionality

## Notes
Uses localStorage to permanently remember dismissal state

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -19,7 +19,8 @@ import {ProfilePic} from "./ProfilePic";
 import {HeaderAutocomplete} from './HeaderAutocomplete'
 import { DropdownMenu, DropdownMenuSeparator, DropdownMenuItem, DropdownMenuItemWithIcon, DropdownLanguageToggle } from './common/DropdownMenu';
 import Button from './common/Button';
-  
+import ModuleSwitcherTooltip from './common/ModuleSwitcherTooltip';
+
 const LoggedOutDropdown = ({module}) => {
   const [isClient, setIsClient] = useState(false);
   const [next, setNext] = useState("/");
@@ -134,28 +135,33 @@ const LoggedInDropdown = ({module}) => {
 }
 
 const ModuleSwitcher = () => {
+  const dropdownRef = useRef(null);
   const libraryURL = Sefaria.moduleRoutes[Sefaria.LIBRARY_MODULE];
   const sheetsURL = Sefaria.moduleRoutes[Sefaria.SHEETS_MODULE];
   return (
-      <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
-          <div className='dropdownLinks-options'>
-              <DropdownMenuItem url={libraryURL} newTab={Sefaria.activeModule !== Sefaria.LIBRARY_MODULE} targetModule={Sefaria.LIBRARY_MODULE}>
-                  <DropdownMenuItemWithIcon icon={'/static/icons/library_icon.svg'} textEn={"Library"}/>
-              </DropdownMenuItem>
-              <DropdownMenuSeparator/>
-              <DropdownMenuItem url={sheetsURL} newTab={Sefaria.activeModule !== Sefaria.SHEETS_MODULE} targetModule={Sefaria.SHEETS_MODULE}>  
-                  <DropdownMenuItemWithIcon icon={'/static/icons/sheets_icon.svg'} textEn={'Sheets'}/>
-              </DropdownMenuItem>
-              <DropdownMenuSeparator/>
-              <DropdownMenuItem url={'https://developers.sefaria.org'} newTab={true}>
-                  <DropdownMenuItemWithIcon icon={'/static/icons/developers_icon.svg'} textEn={'Developers'}/>
-              </DropdownMenuItem>
-              <DropdownMenuSeparator/>
-              <DropdownMenuItem url={'/products'} newTab={true}>
-                <InterfaceText text={{'he': 'לכל המוצרים שלנו', 'en': 'See all products ›'}}/>
-              </DropdownMenuItem>
-          </div>
-      </DropdownMenu>
+      <ModuleSwitcherTooltip targetRef={dropdownRef}>
+        <div ref={dropdownRef}>
+          <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={<img src='/static/icons/module_switcher_icon.svg'/>}>
+              <div className='dropdownLinks-options'>
+                  <DropdownMenuItem url={libraryURL} newTab={Sefaria.activeModule !== Sefaria.LIBRARY_MODULE} targetModule={Sefaria.LIBRARY_MODULE}>
+                      <DropdownMenuItemWithIcon icon={'/static/icons/library_icon.svg'} textEn={"Library"}/>
+                  </DropdownMenuItem>
+                  <DropdownMenuSeparator/>
+                  <DropdownMenuItem url={sheetsURL} newTab={Sefaria.activeModule !== Sefaria.SHEETS_MODULE} targetModule={Sefaria.SHEETS_MODULE}>
+                      <DropdownMenuItemWithIcon icon={'/static/icons/sheets_icon.svg'} textEn={'Sheets'}/>
+                  </DropdownMenuItem>
+                  <DropdownMenuSeparator/>
+                  <DropdownMenuItem url={'https://developers.sefaria.org'} newTab={true}>
+                      <DropdownMenuItemWithIcon icon={'/static/icons/developers_icon.svg'} textEn={'Developers'}/>
+                  </DropdownMenuItem>
+                  <DropdownMenuSeparator/>
+                  <DropdownMenuItem url={'/products'} newTab={true}>
+                    <InterfaceText text={{'he': 'לכל המוצרים שלנו', 'en': 'See all products ›'}}/>
+                  </DropdownMenuItem>
+              </div>
+          </DropdownMenu>
+        </div>
+      </ModuleSwitcherTooltip>
 );
 }
 
@@ -256,8 +262,7 @@ const Header = (props) => {
                 setTranslationLanguagePreference={props.setTranslationLanguagePreference} /> : null}
 
         { Sefaria._uid && (props.module ===Sefaria.LIBRARY_MODULE ? librarySavedIcon : sheetsNotificationsIcon) }
-
-          <ModuleSwitcher />
+           <ModuleSwitcher />
 
           { Sefaria._uid ?
             <LoggedInDropdown module={props.module}/>
--- a/static/js/common/ModuleSwitcherTooltip.jsx
+++ b/static/js/common/ModuleSwitcherTooltip.jsx
@@ -1,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import Tippy from '@tippyjs/react';
+import 'tippy.js/dist/tippy.css';
+import Sefaria from '../sefaria/sefaria';
+import { InterfaceText } from '../Misc';
+import '../../../static/css/Header.css';
+
+const ModuleSwitcherTooltip = ({ targetRef, children }) => {
+  const [isTooltipVisible, setTooltipVisible] = useState(false);
+
+  useEffect(() => {
+    const dismissed = localStorage.getItem('sefaria.moduleSwitcherTooltipDismissed');
+    if (dismissed !== 'true') {
+      setTooltipVisible(true);
+    }
+  }, []);
+
+  const hideTooltip = () => {
+    setTooltipVisible(false);
+    localStorage.setItem('sefaria.moduleSwitcherTooltipDismissed', 'true');
+  };
+
+  const tooltipContent = (
+    <div className="module-switcher-tooltip">
+        <InterfaceText markdown={{
+            en: `We made some changes to the structure of the Sefaria web platform. To better serve our different missions we present specialized areas for **learning** ("Library"), **creating** ("Voices"), and **extending** ("Developers") digital Torah. Welcome to the __Next 10 Years__ of Sefaria!`,
+            he: `ערכנו כמה שינויים במבנה פלטפורמת האינטרנט של ספריא. כדי לשרת טוב יותר את שלל המשימות שלנו, אנו מציגים אזורים ייעודיים ל**לימוד** ("ספרייה"), **יצירה** ("קולות"), ו**הרחבה** ("מפתחים") של התורה הדיגיטלית. ברוכים הבאים ל־__עשר השנים הבאות__ של ספריא!`
+        }} />
+      <div className="module-switcher-tooltip-actions">
+        <a href="https://example.com" target="_blank" rel="noopener noreferrer">
+          <InterfaceText text={{en: "Learn more", he: "למידע נוסף"}} />
+        </a>
+        <button onClick={hideTooltip} className="module-switcher-tooltip-button">
+          <InterfaceText text={{en: "Got it!", he: "הבנתי"}} />
+        </button>
+      </div>
+    </div>
+  );
+
+  return (
+    <>
+      {children}
+      <Tippy
+        content={tooltipContent}
+        visible={isTooltipVisible}
+        placement="top"
+        interactive={true}
+        arrow={true}
+        appendTo={() => document.body}
+        reference={targetRef}
+        className="module-switcher-tippy"
+      />
+    </>
+  );
+};
+
+export default ModuleSwitcherTooltip;
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
       "dependencies": {
         "@playwright/test": "^1.42.1",
         "@sentry/react": "^7.57.0",
+        "@tippyjs/react": "^4.2.6",
         "body-parser": "^1.15.1",
         "buffer": "^6.0.3",
         "cheerio": "^0.22.0",
@@ -58,6 +59,7 @@
         "source-list-map": "^2.0.0",
         "striptags": "^2.1.1",
         "sync-exec": "^0.6.2",
+        "tippy.js": "^6.3.7",
         "webpack-node-externals": "^3.0.0",
         "webpack-sources": "^3.2.1",
         "winston": "^3.3.3"
@@ -2202,6 +2204,16 @@
         "node": ">=16"
       }
     },
+    "node_modules/@popperjs/core": {
+      "version": "2.11.8",
+      "resolved": "https://registry.npmjs.org/@popperjs/core/-/core-2.11.8.tgz",
+      "integrity": "sha512-P1st0aksCrn9sGZhp8GMYwBnQsbvAWsZAX44oXNNvLHGqAOcoVxmjZiohstwQ7SqKnbR47akdNi+uleWD8+g6A==",
+      "license": "MIT",
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/popperjs"
+      }
+    },
     "node_modules/@sentry-internal/tracing": {
       "version": "7.61.0",
       "resolved": "https://registry.npmjs.org/@sentry-internal/tracing/-/tracing-7.61.0.tgz",
@@ -2300,6 +2312,19 @@
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/@socket.io/component-emitter/-/component-emitter-3.1.0.tgz",
       "integrity": "sha512-+9jVqKhRSpsc591z5vX+X5Yyw+he/HCB4iQ/RYxw35CEPaY1gnsNE43nf9n9AaYjAQrTiI/mOwKUKdUs9vf7Xg=="
+    },
+    "node_modules/@tippyjs/react": {
+      "version": "4.2.6",
+      "resolved": "https://registry.npmjs.org/@tippyjs/react/-/react-4.2.6.tgz",
+      "integrity": "sha512-91RicDR+H7oDSyPycI13q3b7o4O60wa2oRbjlz2fyRLmHImc4vyDwuUP8NtZaN0VARJY5hybvDYrFzhY9+Lbyw==",
+      "license": "MIT",
+      "dependencies": {
+        "tippy.js": "^6.3.1"
+      },
+      "peerDependencies": {
+        "react": ">=16.8",
+        "react-dom": ">=16.8"
+      }
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.0",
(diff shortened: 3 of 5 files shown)


static/css/s2.css
```diff
@@ -9612,6 +9604,7 @@ body.interface-english .publishBox .react-tags__suggestions ul {
 .interface-hebrew .editorSaveStateIndicator {
   flex-direction: row-reverse;
 }
+
```
schreibersefaria: You removed a bunch of new lines in-between CSS rules, maybe don't add this one

static/css/s2.css
```diff
@@ -89,8 +89,10 @@ a, a:hover {
   --reference-orange: #D4896C;
   --miscelaneous-green: #3B5849;
   --selected-option: #000000;
+  --module-switcher-tooltip-background: #1F2937;
```
schreibersefaria: Was this chosen by product?
If yes, where is the spec, I'd love to see
If not, maybe use the existing sefaria blue? Also if not, product will very likely want to choose a specific color

static/css/s2.css
```diff
@@ -16031,5 +16017,47 @@ span.ref-link-color-3 {color: blue}
   direction: rtl;
 }
 
+/* Floating UI tooltip styles */
+.floating-ui-tooltip {
+  background-color: #A9BDDB;
```
schreibersefaria: You defined `module-switcher-tooltip-background` in `/* colors */` section (line 92) but you don't use it.
Note, you use this hardcoded value twice

static/css/s2.css
```diff
@@ -16031,5 +16017,47 @@ span.ref-link-color-3 {color: blue}
...
+  border-radius: 8px;
+  box-shadow: 0 0px 6px 2px rgba(0, 0, 0, 0.2);
+  z-index: 9999;
+  color: #222;
```
schreibersefaria: Here too, try not to hardcode a color

static/css/s2.css
schreibersefaria: Note, also in the new css you added you have spacing.

static/css/s2.css
schreibersefaria: Maybe --darkest-grey (#333)

static/css/s2.css
```diff
@@ -16031,5 +16017,47 @@ span.ref-link-color-3 {color: blue}
...
+  color: white;
+  border: none;
+}
+.floating-ui-arrow {
+  position: absolute;
+  width: 8px;
+  height: 8px;
+  background-color: #A9BDDB;
+  transform: rotate(45deg);
+}
```
schreibersefaria: You put this inside the GUIDE OVERLAY COMPONENT STYLES section (see line 16063)
Could you move it out and create a new section.
If you think it should be inside, create a sub section or something like that

static/js/Header.jsx
```diff
@@ -138,9 +139,12 @@ const LoggedInDropdown = ({ module }) => {
   );
 }
 
-const ModuleSwitcher = () => {
+const ModuleSwitcher = ({multiPanel, mobileMenuButtonRef}) => {
```
schreibersefaria: The comment I got from Yishai last week: "don't add multipanel as prop; rather get it from Sefaria.multiPanel"

static/js/Header.jsx
```diff
@@ -295,9 +302,8 @@ const Header = (props) => {
             translationLanguagePreference={props.translationLanguagePreference}
             setTranslationLanguagePreference={props.setTranslationLanguagePreference} /> : null}
 
-        {Sefaria._uid && (props.module === Sefaria.LIBRARY_MODULE ? librarySavedIcon : sheetsNotificationsIcon)}
-
-        <ModuleSwitcher />
+        { Sefaria._uid && (props.module === Sefaria.LIBRARY_MODULE ? librarySavedIcon : sheetsNotificationsIcon) }
+           <ModuleSwitcher multiPanel={props.multiPanel} mobileMenuButtonRef={mobileMenuButtonRef} />
```
schreibersefaria: Why do you send the `mobileMenuButtonRef` here?

static/js/Header.jsx
```diff
@@ -308,13 +314,17 @@ const Header = (props) => {
...
-    <>
-      <div>
-        <button onClick={props.onMobileMenuButtonClick} aria-label={Sefaria._("Menu")} className="menuButton">
-          <i className="fa fa-bars"></i>
-        </button>
-      </div>
+    const mobileHeaderContent = (
+      <>
+        <div>
+          <ModuleSwitcherTooltip targetRef={null} multiPanel={props.multiPanel} mobileTargetRef={mobileMenuButtonRef}>
```
schreibersefaria: Maybe omit `targetRef` for simplicity

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria, line 1: Why did you put this in `common`. Intuitively I'd have thought to put it outside common because it's specific for the ModuleSwitcher

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+const PLACEMENT_CONFIG = {
+  desktop: 'bottom',
+  mobileHebrew: 'right',
+  mobileEnglish: 'left',
+};
+
+const ARROW_OFFSET_OUTSIDE = '-4px';
+const TOOLTIP_OFFSET = 10;
+const TOOLTIP_SHIFT_PADDING = 8;
+const TOOLTIP_Z_INDEX = 9999;
```
schreibersefaria: In CSS I think you write `9999` again, You likely do it also for other `const`s, it's just one I remember.

I'd think to put CSS constants in the CSS file but I don't really know what is נהוג

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+    return conditions.find(([condition]) => condition)?.[1];
+  };
+
+  const configKey = getArrowConfigKey();
+  const arrowStyleFactory = ARROW_STYLE_CONFIG[configKey];
+  const arrowStyle = arrowStyleFactory(middlewareData, placement);
+
+  const tooltipContent = (
+    <div>
+      <InterfaceText markdown={{
```
schreibersefaria, line 127: You should put the He text in `Strings.js` (using `Sefaria._` to retrieve it)

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+    <div>
+      <InterfaceText markdown={{
+        en: `PLACEHOLDER TEXT: We made some changes to the structure of the Sefaria web platform. To better serve our different missions we present specialized areas for **learning** ("Library"), **creating** ("Voices"), and **extending** ("Developers") digital Torah. Welcome to the __Next 10 Years__ of Sefaria! PLACEHOLDER & Draft Styles`,
+        he: `טיוטה: ערכנו כמה שינויים במבנה פלטפורמת האינטרנט של ספריא. כדי לשרת טוב יותר את שלל המשימות שלנו, אנו מציגים אזורים ייעודיים ל**לימוד** ("ספרייה"), **יצירה** ("קולות"), ו**הרחבה** ("מפתחים") של התורה הדיגיטלית. ברוכים הבאים ל־__עשר השנים הבאות__ של ספריא!`
+      }} /> 
+      <div className="tooltip-actions">
+        <a href="https://example.com" target="_blank" rel="noopener noreferrer">
+          <InterfaceText text={{en: "Learn more", he: "למידע נוסף"}} />
+        </a>
+        <button onClick={hideTooltip} className="tooltip-button">
```
schreibersefaria: If creating a new button, ideally use Button

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+};
+
+const ARROW_OFFSET_OUTSIDE = '-4px';
+const TOOLTIP_OFFSET = 10;
+const TOOLTIP_SHIFT_PADDING = 8;
+const TOOLTIP_Z_INDEX = 9999;
+const STORAGE_KEY = 'sefaria.moduleSwitcherTooltipDismissed';
+
+// Arrow style configuration based on placement and language
+const ARROW_STYLE_CONFIG = {
```
schreibersefaria: I'm not sure what this does.
Could you add some commenting explaining it + why you define styling here and not in the CSS file?

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+    return {
+      position: 'absolute',
+      ...(middlewareData.arrow?.x != null && { left: `${middlewareData.arrow.x}px` }),
+      ...(middlewareData.arrow?.y != null && { top: `${middlewareData.arrow.y}px` }),
+      [staticSide]: ARROW_OFFSET_OUTSIDE,
+    };
+  },
+};
+
+const ModuleSwitcherTooltip = ({ targetRef, children, multiPanel, mobileTargetRef }) => {
```
schreibersefaria: It would be great to have a JSdoc here to clarify they use of `ModuleSwitcherTooltip`

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+  const [isTooltipVisible, setTooltipVisible] = useState(false);
+  const isMobile = !multiPanel;
+  const isHebrew = Sefaria.interfaceLang === "hebrew";
+  const arrowRef = useRef(null);
+
+  // Get target element
+  const targetElement = isMobile ? (mobileTargetRef || targetRef) : targetRef;
```
schreibersefaria: Maybe unify `mobileTargetRef` and `targetRef` each time just sending the one you need. Then you won't need this.

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,182 @@
...
+  const isHebrew = Sefaria.interfaceLang === "hebrew";
+  const arrowRef = useRef(null);
+
+  // Get target element
+  const targetElement = isMobile ? (mobileTargetRef || targetRef) : targetRef;
+
+  // Setup floating-ui
+  const { x, y, refs, strategy, middlewareData, update } = useFloating({
+    placement: isMobile 
+      ? (isHebrew ? PLACEMENT_CONFIG.mobileHebrew : PLACEMENT_CONFIG.mobileEnglish)
```
schreibersefaria, line 35: Have you already seen how hebrew css differences is normally done?
You may have a good reason for doing it like this, I just wanted to make sure

schreibersefaria: And mobile*

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria, line 1: Tabbing to the buttons is not in the correct order, it's at the end of the page. Try see if that can be fixed

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,207 @@
...
+      document.addEventListener('touchstart', handleClickOutside);
+    }
+
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+      document.removeEventListener('touchstart', handleClickOutside);
+    };
+  }, [isTooltipVisible, targetElement, refs.floating]);
+
+  const hideTooltip = () => {
```
schreibersefaria, line 97: Why did you choose localStorage and not cookies?
I'm asking primarily because modals in Sefaria use cookies. Although I see both are used in the codebase

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,207 @@
...
+    setTooltipVisible(false);
+    localStorage.setItem(STORAGE_KEY, 'true');
+  };
+
+  // Calculate arrow position based on placement
+  const placement = middlewareData.placement || (isMobile 
+    ? (isHebrew ? PLACEMENT_CONFIG.mobileHebrew : PLACEMENT_CONFIG.mobileEnglish) 
+    : PLACEMENT_CONFIG.desktop);
+
+  // Calculate arrow position using modern functional approach
```
schreibersefaria: I think you can remove "modern functional approach" form the comment

static/js/common/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,207 @@
...
+          <InterfaceText text={{en: "Got it!", he: "הבנתי"}} />
+        </button>
+      </div>
+    </div>
+  );
+
+  const tooltipElement = isTooltipVisible && targetElement?.current && (
+    <div
+      ref={refs.setFloating}
+      className={`floating-ui-tooltip${isHebrew ? ' interface-he' : ''}`}
```
schreibersefaria: I believe .interface-hebrew is automatically added if you are in He

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,207 @@
...
+  const arrowStyle = arrowStyleFactory(middlewareData, placement);
+
+  const tooltipContent = (
+    <div>
+      <InterfaceText markdown={{
+        en: `PLACEHOLDER TEXT: We made some changes to the structure of the Sefaria web platform. To better serve our different missions we present specialized areas for **learning** ("Library"), **creating** ("Voices"), and **extending** ("Developers") digital Torah. Welcome to the __Next 10 Years__ of Sefaria! PLACEHOLDER & Draft Styles`,
+        he: `טיוטה: ערכנו כמה שינויים במבנה פלטפורמת האינטרנט של ספריא. כדי לשרת טוב יותר את שלל המשימות שלנו, אנו מציגים אזורים ייעודיים ל**לימוד** ("ספרייה"), **יצירה** ("קולות"), ו**הרחבה** ("מפתחים") של התורה הדיגיטלית. ברוכים הבאים ל־__עשר השנים הבאות__ של ספריא!`
+      }} /> 
+      <div className="tooltip-actions">
+        <a href="https://example.com" target="_blank" rel="noopener noreferrer">
```
schreibersefaria, line 132: There is a place for hardcoded strings. Not sure now where it is, I'll try remember but you can ask in the group

schreibersefaria (changes requested):
Some of the comments I put in one place but are general.
Happy to show you how to do HE strings and CSS etc - Although I'm really not the professional.
Overall it seems well done.

static/css/s2.css
yodem: Unfortunately there is no spec, just me talking with Mickey.
After the demo tomorrow and talking with Penina they would decide on a color.
So it's basically temporary but it also might be the actual color.

static/css/s2.css
yodem: fixed

static/css/s2.css
yodem: removed the color, I think it should be dark and its dark by default.

static/js/ModuleSwitcherTooltip.jsx
yodem: Moved to js.

static/js/common/ModuleSwitcherTooltip.jsx
yodem: Removed

static/js/ModuleSwitcherTooltip.jsx
yodem: It doesn't relate to how we use our css, but how floating-ui useFloating hook determine its values based on the required placement. and the placement is being determined based on isMobile and isHebrew

static/js/ModuleSwitcherTooltip.jsx
yodem: As far as i know, you need to use cookies for small amounts of data that need to be sent to the server, such as for session management and user tracking. And use local storage for user data that only need to be accessed by the browser, like user preferences that persist across sessions.

And the tooltip got nothing to do with the server so I chose localStorage

static/js/ModuleSwitcherTooltip.jsx
yodem: The reason I didnt to it is that its a temporary component, so I didnt want it to be next to the persistent data.

static/js/ModuleSwitcherTooltip.jsx
yodem: Same as the previous comment, I'm absolutely down for using consts in a centralized place, but it's a temporary component. Also, the URL for - https://example.com wasn't given to me yet...

static/js/common/ModuleSwitcherTooltip.jsx
yodem: fixed

static/js/Header.jsx
yodem: removed it

static/js/Header.jsx
yodem: because I used it, but now its deleted

static/js/ModuleSwitcherTooltip.jsx
yodem: Didnt understand the comment

static/css/s2.css
schreibersefaria: Makes sense. Just make sure to put all the colors as variables and use them.

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: Ahh I didn't realise that, interesting.
I'd comment that very clearly.
Maybe even put it in the component/file name. And of course make sure there is a task already to remove it

static/js/common/ModuleSwitcherTooltip.jsx
schreibersefaria: This is less critical if It's a temp component that will be used once and deleted

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: Try reaching the new buttons using the keyboard.
You will find that you have to go through the whole page with Tab before you can reach the new buttons. 
It's not the end of the world, but it would be great if we can fix this.

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: A good reason.
The only disadvantage I can think of is that there is no TTL (because it's a temp component), but I guess it's reasonable.

yodem: You right, I think thats the wanted behavior but I"ll ask Mickey

static/css/s2.css
```diff
@@ -89,8 +89,10 @@ a, a:hover {
   --reference-orange: #D4896C;
   --miscelaneous-green: #3B5849;
   --selected-option: #000000;
+  --module-switcher-tooltip-background: #FFFFFF;
```
schreibersefaria, line 92: Ideally put this in `color-palette.css`

static/css/s2.css
```diff
@@ -6488,12 +6488,6 @@ big {
   padding-inline-start: 30px;
   display: inline-block;
 }
-
-.segment .segmentText .poetry.indentAll {
-  padding-inline-start: 30px;
-  display: inline-block;
-}
```
schreibersefaria, line 6495: Just making sure you didn't remove this by mistake because it isn't clearly to do with your current work

static/js/Header.jsx
```diff
@@ -134,8 +135,11 @@ const LoggedInDropdown = ({ module }) => {
 }
 
 const ModuleSwitcher = () => {
+  const dropdownRef = useRef(null);
   const logoPath = Sefaria.interfaceLang === "hebrew" ? "/static/img/logo-hebrew.png" : "/static/img/logo.svg";
-  return (
+  return (    
+  <ModuleSwitcherTooltip targetRef={dropdownRef}>
+    <div ref={dropdownRef}>
     <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={
```
schreibersefaria, line 143: I think you missed an indentation

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: Note, this is still not accessible. 
We can make a decision to be OK with that but it should be a decision.

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: I slacked you. Not sure what you mean here by temporary.
But in any case, I wouldn't have the strings here in the middle. I'd put them in a const or something.

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,174 @@
...
+import Sefaria from './sefaria/sefaria';
+
+// Configuration constants
+const PLACEMENT_CONFIG = {
+  desktop: 'bottom',
+  mobileHebrew: 'right',
+  mobileEnglish: 'left',
+};
+
+const TOOLTIP_OFFSET = 10;
```
schreibersefaria, line 22: Why is the offset here and not in CSS?

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,174 @@
...
+const getTooltipShiftPadding = (isMobile) => isMobile ? 10 : 50;
```
schreibersefaria, line 23: Here too, wouldn't this be better in the CSS

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,174 @@
...
+    if (isTooltipVisible) {
+      document.addEventListener('mousedown', handleHeaderDropdownClick);
+      document.addEventListener('touchstart', handleHeaderDropdownClick);
+    }
+
+    return () => {
+      document.removeEventListener('mousedown', handleHeaderDropdownClick);
+      document.removeEventListener('touchstart', handleHeaderDropdownClick);
+    };
+  }, [isTooltipVisible]);
```
schreibersefaria, line 95: Wouldn't you typically include hideTooltip in the dependencies if using it in the effect?

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,174 @@
...
+  }, [targetRef, refs]);
+
+  // Auto-update positioning
+  useEffect(() => {
+    if (isTooltipVisible && targetRef?.current && refs.floating.current) {
+      return autoUpdate(targetRef.current, refs.floating.current, update);
+    }
+  }, [isTooltipVisible, targetRef, refs.floating.current, update]);
+
+  // After setTooltipVisible(true) is called, force an update in a useEffect
```
schreibersefaria, line 67: Wouldn't the previous effect also catche this because isTooltipVisible is a dependency?

static/js/ModuleSwitcherTooltip.jsx
schreibersefaria: Generally it would be good to define the role of the component and ask an agent what else needs to be done for accessibility.

static/js/ModuleSwitcherTooltip.jsx
```diff
@@ -0,0 +1,174 @@
...
+        className={getArrowClassName()}
+        style={arrowDynamicStyle}
+      />
+    </div>
+  );
+
+  return (
+    <>
+      {children}
+      {typeof document !== 'undefined' && ReactDOM.createPortal(
```
schreibersefaria, line 166:
```suggestion
      {isTooltipVisible &&  typeof document !== 'undefined' && ReactDOM.createPortal(
```

schreibersefaria (changes requested):
Overall, I agree with your question about if a tooltip is the right way to go, being as this would typically not be seen.
I think another review ideally from Yishai, or maybe Steve would be relevant here.

PR #2691, opened 2025-09-17
Fix deleting of sheets not updating component in user profile

schreibersefaria:
Update the data fetching affect in SheetsList to trigger on refreshData - which was already being changed when a sheet was deleted.

Changed the request for sheets to be post and not get - as the backend expects
Fixed `class` to be `className` in SheetsList - Removing a console warning

### DIFF
--- a/static/js/UserProfile.jsx
+++ b/static/js/UserProfile.jsx
@@ -357,7 +357,7 @@ const SheetsList = ({profile, handleSheetDelete, handleCollectionsChange, toggle
   }
 
   return (
-    <div class="sheetsProfileList">
+    <div className="sheetsProfileList">
       {sheets.map(sheet => (
         <SheetListing
           key={sheet.id}
--- a/static/js/UserProfile.jsx
+++ b/static/js/UserProfile.jsx
@@ -283,7 +283,8 @@ class UserProfile extends Component {
                     <SheetsList profile={this.props.profile}
                                   handleSheetDelete={this.handleSheetDelete}
                                   handleCollectionsChange={this.handleCollectionsChange}
-                                  toggleSignUpModal={this.props.toggleSignUpModal}/>}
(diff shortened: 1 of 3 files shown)


static/js/UserProfile.jsx
```diff
@@ -328,36 +329,24 @@ UserProfile.propTypes = {
   profile: PropTypes.object.isRequired,
 };
 
-const SheetsList = ({profile, handleSheetDelete, handleCollectionsChange, toggleSignUpModal}) => {
+const SheetsList = ({profile, handleSheetDelete, handleCollectionsChange, toggleSignUpModal, refreshData}) => {
```
YishaiGlasner, line 332: if you already understand this component so it will be easy to add proptypes, please do it

static/js/UserProfile.jsx
```diff
@@ -283,7 +283,8 @@ class UserProfile extends Component {
                     <SheetsList profile={this.props.profile}
                                   handleSheetDelete={this.handleSheetDelete}
                                   handleCollectionsChange={this.handleCollectionsChange}
-                                  toggleSignUpModal={this.props.toggleSignUpModal}/>}
+                                  toggleSignUpModal={this.props.toggleSignUpModal}
+                                  refreshData={this.state.refreshSheetData}/>}
```
YishaiGlasner, line 287: i'm not sure how this work. is `refreshData` a bool? if so, can it get true, and then again true, so it won't trigger the useEffect?

schreibersefaria: No, it uses Math.random()
This isn't my implementation, I just added it to the useEffect

static/js/UserProfile.jsx
schreibersefaria: I can't say I do understand it.
I really just added the refreshData

PR #2692, opened 2025-09-18
Django hosts

YishaiGlasner:
## Description
Use django-hosts to direct requests of different subdomain.

## Code Changes
1. Add library to requirements, and add settings for it.
2. Define different hosts for the subdomains.
3. Split urls.py to 3 files:
- urls_library.py
- urls_sheets.py, both defined as the urls of hosts
- urls_shared.py for urls that should be in both subdimains (mainly apis, but also urls that exist in both subdomains and uses `active_module`, and also many redirections (here i judt don't want to check what exactly it does - we can let the redirection to happen anyway))

## Notes
We should consider removing of the middleware the adds `active_module`, and add this as a param of the view, that will be added in each urls file.

### DIFF
--- a/sefaria/urls_sheets.py
+++ b/sefaria/urls_sheets.py
@@ -14,7 +14,6 @@ handler404 = 'reader.views.custom_page_not_found'
 
 urlpatterns = [
     url(r'^$', reader_views.community_page),
-    url(r'^community/?$', reader_views.community_page),
     url(r'sheets-with-ref/(?P<tref>.+)$', sheets_views.sheets_with_ref),
     url(r'^collections/?$', reader_views.public_collections),
     url(r'^collections/new$', reader_views.edit_collection_page),


sefaria/urls_sheets.py
```diff
@@ -0,0 +1,51 @@
...
+from sefaria.urls_shared import shared_patterms, maintenance_patterms
+from sefaria.settings import DOWN_FOR_MAINTENANCE
+
+admin.autodiscover()
+handler500 = 'reader.views.custom_server_error'
+handler404 = 'reader.views.custom_page_not_found'
+
+urlpatterns = [
+    url(r'^$', reader_views.community_page),
+    url(r'^community/?$', reader_views.community_page),
```
stevekaplan123: I'm not sure what's supposed to happen when going to community.  DId we ask Mickey recently?  I thought the page doesn't exist in modularization.

sefaria/urls_shared.py
```diff
@@ -0,0 +1,276 @@
...
+    url(r'^saved/?$', reader_views.saved_content),
+    url(r'^history/?$', reader_views.user_history_content),
+    url(r'^search/?$', reader_views.search),
+    url(r'^search-autocomplete-redirecter/?$', reader_views.search_autocomplete_redirecter),
+    url(r'^notifications/?$', reader_views.notifications),
+    url(r'^person/(?P<name>.+)$', reader_views.person_page_redirect),
+    url(r'^people/Talmud/?$', reader_views.talmud_person_index_redirect),
+    url(r'^people/?$', reader_views.person_index_redirect),
+
+    url(r'^api/profile/user_history$', reader_views.user_history_api),
```
stevekaplan123, line 23: This is confusing since the profile will only exist in voices module.  Do we need the user_history_api for user history in library module?  It seems like the others here should just be in voices module.

sefaria/urls_shared.py
```diff
@@ -0,0 +1,276 @@
...
+    url(r'^topics/category/(?P<topicCategory>.+)?$', reader_views.topics_category_page),
+    url(r'^topics/all/(?P<letter>.)$', reader_views.all_topics_page),
+    url(r'^topics/?$', reader_views.topics_page),
+    url(r'^topics/b/(?P<slug>.+)$', reader_views.topic_page_b),
+    url(r'^topics/(?P<slug>.+)$', reader_views.topic_page),
+
+    url(r'^new-home/?$', reader_views.new_home_redirect),
+    url(r'^account/?$', reader_views.my_profile),
+    url(r'^my/notes/?$', reader_views.my_notes_redirect),
+    url(r'^sheets/tags/?$', reader_views.topics_redirect),
```
stevekaplan123, line 40: Why are these sheets routes shared?

sefaria/urls_sheets.py
YishaiGlasner: i think you're right - we should have the legacy of sefaria.org/community redirected to voices home page (by nginx), but we don't need to support voices.sefaria.org/community (by django). i'll remove it.

sefaria/urls_shared.py
YishaiGlasner: i think all apis should be in both. we can or cannot use it. it isn't a page that the user will land and get weird behaviour

sefaria/urls_sheets.py
stevekaplan123: And the redirect in the helm chart just sends it to "voices.sefaria.org/" I think

sefaria/urls_sheets.py
```diff
@@ -0,0 +1,51 @@
...
+    url(r'^sheets/new/?$', sheets_views.new_sheet),
+    url(r'^sheets/(?P<sheet_id>\d+)$', sheets_views.view_sheet),
+    url(r'^sheets/visual/(?P<sheet_id>\d+)$', sheets_views.view_visual_sheet),
+    url(r'^sheets/(?P<tref>[\d.]+)$', reader_views.catchall, {'sheet': True}),
+
+    url(r'^my/profile', reader_views.my_profile),
+    url(r'^profile/(?P<username>[^/]+)/?$', reader_views.user_profile),
+    url(r'^settings/profile?$', reader_views.edit_profile),
+
+    url(fr'^login/?$', sefaria_views.CustomLoginView.as_view(), name='sheets_login'),
```
stevekaplan123, line 33: What benefit do we get now in calling this sheets_login etc?  That was what we did when we couldn't set the active_module based on request.get_host in middleware.py, but now that we can, why do we need 'sheets_' in the names?

sefaria/urls_shared.py
YishaiGlasner: i put here many redirections things, for it can be redirected in wither case, and then it will be discovered if the page exists or not in the given module

sefaria/urls_shared.py
YishaiGlasner: all of these are redirections, so i think we don't need to discover about any of them to what module it goes. we can leave it as is and after redirection django will discover if it has view in the specific module.

sefaria/urls_sheets.py
YishaiGlasner: i don't know. i don't see we had used this name before so i don't understand what it should do.

stevekaplan123: So where did the name come from then?

YishaiGlasner: i just copy-pasted from urls.py. isn't it you that added it?

stevekaplan123: If we remove it, does this still work?  If so, we should make that change and I think we can merge this into the remove-sheets branch.

stevekaplan123 (approved):

PR #2694, opened 2025-09-18
Return 404 for ref when active module isn't 'library'.

YishaiGlasner:


reader/views.py
```diff
@@ -286,6 +286,8 @@ def reader_redirect(uref):
         response['Location'] += "?%s" % params if params else ""
         return response
 
+    active_module = getattr(request, "active_module", "library")
+
```
stevekaplan123, line 290: I'm confused why this is necessary.  Wasn't this already working once we split the urls file into urls_library and urls_sheets?  I commented on what I think the relevant shortcut story is.

YishaiGlasner: this is used in line 300

reader/views.py
```diff
@@ -295,6 +297,8 @@ def reader_redirect(uref):
             oref = Ref.instantiate_ref_with_legacy_parse_fallback(tref)
         except InputError:
             raise Http404
+        if active_module != 'library':
```
stevekaplan123: Why instantiate the ref on line 297 if we don't know what module we're in yet?

PR #2699, opened 2025-09-21
Bug/sc 36292/sheets link in resources panel of library

schreibersefaria:
Added a secondary option for ToolsButton (Buttons in the side button)
Added a 'open in new tab' icon for the Sheets button
Set the icon button to 14px
Validating we get an altText if an icon is given

### DIFF
--- a/static/js/ConnectionsPanel.jsx
+++ b/static/js/ConnectionsPanel.jsx
@@ -663,7 +663,9 @@ const ResourcesList = ({ srefs, setConnectionsMode, counts }) => {
   // A list of Resources in addition to connection
   return (
     <div className="toolButtonsList">
-      <ToolsButton en="Sheets" he="דפי מקורות" image="sheet.svg" count={counts["sheets"]} urlConnectionsMode="Sheets" onClick={() => createSheetsWithRefURL(srefs)} secondaryIcon="open-panel.svg" secondaryIconAlt="Opens in new window" />
+      <ToolsButton en="Sheets" he="דפי מקורות" image="sheet.svg" count={counts["sheets"]} urlConnectionsMode="Sheets" onClick={() => createSheetsWithRefURL(srefs)}>
+        <ToolsButton.SecondaryIcon icon="open-panel.svg" alt="Opens in new window" />
+      </ToolsButton>
       <ToolsButton en="Web Pages" he="דפי אינטרנט" image="webpages.svg" count={counts["webpages"]} urlConnectionsMode="WebPages" onClick={() => setConnectionsMode("WebPages")} />
       <ToolsButton en="Topics" he="נושאים" image="hashtag-icon.svg" count={counts["topics"]} urlConnectionsMode="Topics" onClick={() => setConnectionsMode("Topics")} alwaysShow={Sefaria.is_moderator} />
       <ToolsButton en="Manuscripts" he="כתבי יד" image="manuscripts.svg" count={counts["manuscripts"]} urlConnectionsMode="manuscripts" onClick={() => setConnectionsMode("manuscripts")} />
@@ -1007,8 +1009,8 @@ AdvancedToolsList.propTypes = {
 
 const ToolsButton = ({ en, he, onClick, urlConnectionsMode = null, icon, image,
                        count, control = "interface", typeface = "system", alwaysShow = false,
-                       secondaryHe, secondaryEn, greyColor=false, highlighted=false, experiment=false, 
-                       secondaryIcon, secondaryIconAlt }) => {
+                       secondaryHe, secondaryEn, greyColor=false, highlighted=false, experiment=false,
+                       children }) => {
   const clickHandler = (e) => {
     e.preventDefault();
     gtag("event", "feature_clicked", {name: `tools_button_${en}`})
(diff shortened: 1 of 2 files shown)


static/js/ConnectionsPanel.jsx
```diff
@@ -1036,6 +1037,7 @@ const ToolsButton = ({ en, he, onClick, urlConnectionsMode = null, icon, image,
           {count ? (<span className="connectionsCount">({count})</span>) : null}
           {experiment ? <span className="experimentLabel">Experiment</span> : null}
         </span>
+        {secondaryIcon ? <img src={`/static/icons/${secondaryIcon}`} className="toolsButtonSecondaryIcon" alt={secondaryIconAlt} /> : null}
```
YishaiGlasner: you can do
```
{secondaryIcon && <img src={`/static/icons/${secondaryIcon}`} className="toolsButtonSecondaryIcon" alt={secondaryIconAlt} />}
```

static/js/ConnectionsPanel.jsx
```diff
@@ -1053,7 +1055,16 @@ ToolsButton.propTypes = {
   highlighted: PropTypes.bool,
   experiment: PropTypes.bool,
   secondaryEn: PropTypes.string,
-  secondaryHe: PropTypes.string
+  secondaryHe: PropTypes.string,
+  secondaryIcon: PropTypes.string,
+  secondaryIconAlt: function(props, propName, componentName) {
```
YishaiGlasner: this makes proptypes very hard to read, so rahter then help the reader to know it should be string, it gives him more things to understand.
i think it will be better to do the check inside the component, but ii think this kind of issue should be discussed in a wider circle.

YishaiGlasner (commented):
It seems we have a nice and simple component, and we're starting to make it complex (i'm saying starting for this kind of things always continue - there are more cases and simple button becomes monster).
i think we should do something cleaner - maybe another component for 2 images + moving the shared logic of the 2 components to a another component/function.

static/js/ConnectionsPanel.jsx
yodem: whats special about the secondaryIconAlt that you create a custom validate function? it seems to me that its supposed to be like the other props.

static/js/ConnectionsPanel.jsx
yodem: if you can apply what Yishai wrote to the rest of the conditional rendering it would be great

static/js/ConnectionsPanel.jsx
```diff
@@ -1007,7 +1007,8 @@ AdvancedToolsList.propTypes = {
 
 const ToolsButton = ({ en, he, onClick, urlConnectionsMode = null, icon, image,
                        count = null, control = "interface", typeface = "system", alwaysShow = false,
-                       secondaryHe, secondaryEn, greyColor=false, highlighted=false, experiment=false }) => {
+                       secondaryHe, secondaryEn, greyColor=false, highlighted=false, experiment=false, 
+                       secondaryIcon=null, secondaryIconAlt=null }) => {
```
yodem: you can drop the =null from the props. the conditional rendering wont render them because they are undefiend (if you use &&). maybe we should consider make default Alt as string.

static/js/ConnectionsPanel.jsx
schreibersefaria: done

static/js/ConnectionsPanel.jsx
schreibersefaria: Simplified it. Following the discussion on slack I may return something
@yodem - the difference is that I want to validate that the altText is given if and only if the icon is also given.

static/js/ConnectionsPanel.jsx
schreibersefaria: Removed.
I also removed some others

yodem (approved):
So now if I would like to use the SecondaryIcon, I need to use it that way:
`<ToolsButton props... >
     <ToolsButton.SecondaryIcon props... />
 </ToolsButton>`
correct?

schreibersefaria:
@yodem exactly

static/js/ConnectionsPanel.jsx
```diff
@@ -1033,15 +1036,26 @@ const ToolsButton = ({ en, he, onClick, urlConnectionsMode = null, icon, image,
         {iconElem}
         <span className="toolsButtonText">
           {control === "interface" ? <InterfaceText text={{ en: en, he: he }} /> : <ContentText text={{ en: en, he: he }} />}
-          {count ? (<span className="connectionsCount">({count})</span>) : null}
-          {experiment ? <span className="experimentLabel">Experiment</span> : null}
+        {count && (<span className="connectionsCount">({count})</span>)}
```
YishaiGlasner: indent

schreibersefaria: Fixed

schreibersefaria:
Claude Summary of the removing of the secondaryEn and secondaryHe:
Removed unused secondaryEn and secondaryHe props from ToolsButton component. These props were originally added in commit 489bd3773 to support buttons with multiple text states (used in the SheetToolsList component for copy and Google Drive export functionality). However, the entire SheetToolsList component was subsequently removed in commit 67b1bd489, leaving these props unused throughout the codebase. This cleanup removes the dead code and simplifies the ToolsButton interface.

YishaiGlasner (approved):

PR #2703, opened 2025-09-25
Changed guide cookie to set on load

schreibersefaria:
Updated the guide to set the cookie on load instead of on close.
This was people with small screens or people who just get confused won't see the guide every time.

* I moved the call to `setCookie` to be in the effect that manages loading
* I removed the call from `handleClose`
* I removed the `shouldSetCookie` param form `handleClose` because it's no longer needed - we never set cookie

Tested this locally by removing the cookie -> seeing a load -> refreshing => no guide loads (as expected)

### DIFF
--- a/static/js/GuideOverlay.jsx
+++ b/static/js/GuideOverlay.jsx
@@ -204,8 +204,8 @@ const GuideOverlay = ({
       // Auto view fires when guide shows without being forced (no cookie exists)
       trackGuideEvent("guide_view_auto", guideType);
       setIsVisible(true);
-      setCookie(); // Guide should only show once, so we set the cookie when it shows
       loadGuideData();
+      setCookie(); // Guide should only show once, so we set the cookie when it shows
     }
   }, []);
 


static/js/GuideOverlay.jsx
```diff
@@ -204,6 +204,7 @@ const GuideOverlay = ({
       // Auto view fires when guide shows without being forced (no cookie exists)
       trackGuideEvent("guide_view_auto", guideType);
       setIsVisible(true);
+      setCookie(); // Guide should only show once, so we set the cookie when it shows
```
yodem: Why didnt you put it both:
setIsVisible(true)
and setCookie() in the loadGuideData()?
If i understood it currectly, this is the function we want to run if the conditions met, and after we execute the function succesfully we want to:

1.  setIsVisible(true);
2.  setCookie();

schreibersefaria: The loadGuideData is in charge of loading data, not setting cookies etc' so I didn't think it correct to put them together.

Regarding the order I could think of arguments in both directions but I guess moving the setCookie to after the load data could work.
Regarding moving the order of setIsVisible, there is an analytics being triggered which I think is reason enough to not start changing things in this small ticket.

yodem: My only concern is if retrieving the guide data would fail - the cookie and the set would set either way.
so you might want to add try, catch.

PR #2713, opened 2025-09-30
Fix auto completions

YishaiGlasner:
## Background
The problem this PR addresses is the autocompletion suggestion in adding a source to a sheet (from the sheet).
For understanding the issue, one should know the history:
1. Auto completions are genereated by names api. They are used in some places in the site.
2. In master we have 2 relevent params - `type` for getting only one type of auto completions, and `topic_pool` for getting the topic for only one pool (library or sheets).
3. In MDL we change the auto completins of the main search, so any module will get only results that are related to the module. For that purpose, the `get_name_completions` uses the `active_module` (shich is added by middleware to requests).
The problem is that in when one wants to add a source to sheet, he's in the sheets module but needs the results from the library.
 
## API considareations
I think the adding of `active_module` param was wrong. Here are my considerations:
1. It has the above problem.
2. API public endpoint should not vary based on our subdomains. The API is an external-facing contract and should be independent of our internal routing.
3. The results are filtered base on types and topic pool, so we should use/refactor existing params to get the desired result. 

## Solution
I first reverted the api and its uses to what it is in master (not git revert for some weird git issue).
Then I refactored the `type` param to support multiple values per request.

## Code changes
1. Reverting of the code in both server and client sides.
2. Refactoring code to support multiple `type` values (and renaming the `type` to `types`) - also in both sides.
3. Add the desired types to the HeaderAutocomplete when being on sheets module.

## Note for reviewers
Most of the diff here is taken from master. Maybe it would be easier to read it against master rather than modularization-main (but I'm not sure how it can be done).

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -2663,13 +2663,13 @@ def terms_api(request, name):
     return jsonResponse({"error": "Unsupported HTTP method."})
 
 
-def get_name_completions(name, limit, topic_override=False, type=None, topic_pool=None, exact_continuations=False, order_by_matched_length=False):
+def get_name_completions(name, limit, topic_override=False, types=None, topic_pool=None, exact_continuations=False, order_by_matched_length=False):
     """
     Function to get completions (objects and titles) for a given name.
     :param name: string to get completions for
     :param limit: int number of items to return
     :param topic_override: bool
-    :param type: string - get only completions of objects of this specific type
+    :param types: list of strings - get only completions of objects of those specific types
     :param topic_pool: string - get completions of topic-objects of this specific topic pool
     :param exact_continuations: bool - if true get only completions of objects whose title contains an exact match to 'name'
     :param order_by_matched_length: bool - if true return completion objects by ascending order of length - such that shorter titles whose greater part is a match to 'name' will appear first.
@@ -2699,7 +2699,7 @@ def get_name_completions(name, limit, topic_override=False, type=None, topic_poo
             completion_objects = [o for n in completions for o in lexicon_ac.get_data(n)]
 
         else:
-            completions, completion_objects = completer.complete(name, limit, type=type, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
+            completions, completion_objects = completer.complete(name, limit, types=types, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
             object_data = completer.get_object(name)
 
     except DictionaryEntryNotFoundError as e:
@@ -2709,7 +2709,7 @@ def get_name_completions(name, limit, topic_override=False, type=None, topic_poo
         completions = list(OrderedDict.fromkeys(t))  # filter out dupes
         completion_objects = [o for n in completions for o in lexicon_ac.get_data(n)]
     except InputError:  # Not a Ref
-        completions, completion_objects = completer.complete(name, limit, type=type, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
+        completions, completion_objects = completer.complete(name, limit, types=types, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
         object_data = completer.get_object(name)
 
     return {
@@ -2730,11 +2730,11 @@ def name_api(request, name):
     name = name[1:] if topic_override else name
     # Number of results to return.  0 indicates no limit
     LIMIT = int(request.GET.get("limit", 10))
-    type = request.GET.get("type", None)
+    types = request.GET.getlist("type", None)
     topic_pool = request.GET.get("topic_pool", None)
     exact_continuations = bool(int(request.GET.get("exact_continuations", False)))
     order_by_matched_length = bool(int(request.GET.get("order_by_matched_length", False)))
-    completions_dict = get_name_completions(name, LIMIT, topic_override, type=type, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
+    completions_dict = get_name_completions(name, LIMIT, topic_override, types=types, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
     ref = completions_dict["ref"]
     topic = completions_dict["topic"]
     d = {
--- a/reader/views.py
+++ b/reader/views.py
@@ -2663,14 +2663,14 @@ def terms_api(request, name):
     return jsonResponse({"error": "Unsupported HTTP method."})
 
 
-def get_name_completions(name, limit, topic_override=False, type=None, active_module=None, exact_continuations=False, order_by_matched_length=False):
+def get_name_completions(name, limit, topic_override=False, type=None, topic_pool=None, exact_continuations=False, order_by_matched_length=False):
     """
     Function to get completions (objects and titles) for a given name.
     :param name: string to get completions for
     :param limit: int number of items to return
     :param topic_override: bool
     :param type: string - get only completions of objects of this specific type
-    :param active_module: string - filter out objects irrelevant for this module ('library' vs 'sheets').  If active_module is None, results for both modules are returned.
+    :param topic_pool: string - get completions of topic-objects of this specific topic pool
     :param exact_continuations: bool - if true get only completions of objects whose title contains an exact match to 'name'
     :param order_by_matched_length: bool - if true return completion objects by ascending order of length - such that shorter titles whose greater part is a match to 'name' will appear first.
     """
@@ -2699,7 +2699,7 @@ def get_name_completions(name, limit, topic_override=False, type=None, active_mo
             completion_objects = [o for n in completions for o in lexicon_ac.get_data(n)]
 
         else:
-            completions, completion_objects = completer.complete(name, limit, type=type, active_module=active_module, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
+            completions, completion_objects = completer.complete(name, limit, type=type, topic_pool=topic_pool, exact_continuations=exact_continuations, order_by_matched_length=order_by_matched_length)
             object_data = completer.get_object(name)
 
     except DictionaryEntryNotFoundError as e:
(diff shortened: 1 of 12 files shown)


reader/views.py
```diff
@@ -2730,11 +2730,11 @@ def name_api(request, name):
     name = name[1:] if topic_override else name
     # Number of results to return.  0 indicates no limit
     LIMIT = int(request.GET.get("limit", 10))
-    type = request.GET.get("type", None)
-    active_module = request.GET.get('active_module', None)
+    types = request.GET.getlist("type", None)
```
yodem, line 2744: isnt it supposed to be types?
```suggestion
    types = request.GET.getlist("types", None)
```

sefaria/model/autospell.py
```diff
@@ -201,7 +201,7 @@ def get_data(self, instring):
         except KeyError:
             return None
 
-    def complete(self, instring, limit=0, redirected=False, type=None, active_module=None, exact_continuations=False, order_by_matched_length=False):
+    def complete(self, instring, limit=0, redirected=False, types=None, topic_pool=None, exact_continuations=False, order_by_matched_length=False):
```
yodem, line 205: removing active_module has no effect on any other usage?

sefaria/model/autospell.py
```diff
@@ -334,46 +334,34 @@ def _candidate_order(self, c):
...
-        filtered_completions = []
-        for cs, co in zip(completion_strings, completion_objects):
-            allowed_modules = self._type_to_active_module(co)
-            if self.active_module and self.active_module not in allowed_modules:
-                continue
-            if self.type and self._type_norm_map[co["type"]] != self.type:
-                continue
-            filtered_completions.append((cs, co))
-
+    def _filter_completions_by_type(self, completion_strings, completion_objects):
```
yodem, line 338: is this code from master?

static/js/sefaria/sefaria.js
```diff
@@ -1351,15 +1351,22 @@ Sefaria = extend(Sefaria, {
...
-  getName: function(name, limit = undefined, type=undefined, exactContinuations=undefined, orderByMatchedLength=undefined) {
+  getName: function(name, limit = undefined, types=undefined, topicPool=undefined, exactContinuations=undefined, orderByMatchedLength=undefined) {
     const trimmed_name = name.trim();
-    let params = {active_module: this.activeModule};
+    let params = {};
     // if (refOnly) { params["ref_only"] = 1; }
     if (limit != undefined) { params["limit"] = limit; }
-    if (type != undefined) { params["type"] = type; }
+    if (types != undefined) { params["type"] = types; }
+    if (topicPool !== undefined) { params["topic_pool"] = topicPool; }
```
yodem: if you only want to know if these parameters contain value, perhaps you should use if(limit) / if(types)
without the != undefiend

static/js/HeaderAutocomplete.jsx
```diff
@@ -362,7 +362,8 @@ export const HeaderAutocomplete = ({onRefClick, showSearch, openTopic, openURL,
           return[];
         }
         try {
-        const d = await Sefaria.getName(inputValue);
+        let types = Sefaria.activeModule === Sefaria.SHEETS_MODULE ? ['Topic', 'User', 'Collection'] : undefined;
```
yodem: why not empty array?

```suggestion
        let types = Sefaria.activeModule === Sefaria.SHEETS_MODULE ? ['Topic', 'User', 'Collection'] : [];
```

yodem: this - ['Topic', 'User', 'Collection']
 is the core of the solution correct? the rest of the code is from master and you added the right categories?

reader/views.py
YishaiGlasner: no, the api param is `type` (as it's for one type), but it can show multiple times so the list named `types`

sefaria/model/autospell.py
YishaiGlasner: no, it was introduced in MDL for this purpose

sefaria/model/autospell.py
YishaiGlasner: yes
https://github.com/Sefaria/Sefaria-Project/blob/master/sefaria/model/autospell.py

static/js/sefaria/sefaria.js
YishaiGlasner: you're right but i'm not sure - the design here was default value of undefined (which seems to be redundant), and checking it isn't equals to undefined, so maybe there are cases it should be null. it's like this for all the params, not only the ones i've touched here.

static/js/HeaderAutocomplete.jsx
YishaiGlasner: i don't think we should use empty array to say "don't filter". using null ismore reasonable, but `Sefaria.getName` expects undefined.
the core solution is on server sides - supporting multiple types, and this is the core change in client side using it (the rest of code changes in client side are from master, and i just made slight changes - changing `type` string param to `types` list).

static/js/HeaderAutocomplete.jsx
```diff
@@ -362,7 +362,9 @@ export const HeaderAutocomplete = ({onRefClick, showSearch, openTopic, openURL,
...
+        console.log(1, types, Sefaria.activeModule)
```
yitzhakc: Remove console.log

static/js/HeaderAutocomplete.jsx
```diff
@@ -362,7 +362,9 @@ export const HeaderAutocomplete = ({onRefClick, showSearch, openTopic, openURL,
           return[];
         }
         try {
-        const d = await Sefaria.getName(inputValue);
+        let types = Sefaria.activeModule === Sefaria.SHEETS_MODULE ? ['Topic', 'User', 'Collection'] : undefined;
```
yitzhakc: I think that as of #2646 this should be `VOICES_MODULE`

yitzhakc (changes requested):
We don't happen to have any unit tests for these functions, do we?

static/js/HeaderAutocomplete.jsx
YishaiGlasner: @yitzhakc i'm not sure. 1. one of the reasons not to use magic strings is that it's difficult to change. so we use variable, and have to change it also? 2. i don't think it's a bad name. it's more precise than voices - voices is the english official name for sheets module.

static/js/sefaria/sefaria.js
yodem: there is a cleaner and more modern way to do it:

```
const params = {
    // Re-introducing the commented logic:
    // Assuming refOnly is a boolean flag. If truthy, sets "ref_only": 1
    ...(refOnly && { ref_only: 1 }), 
    
    // If limit is present, add { limit: limit } (using object property shorthand)
    ...(limit != null && { limit }), 
    
    // If types is present, add { type: types }
    // Note: key name 'type' is different from variable name 'types'
    ...(types != null && { type: types }), 
    
    // If topicPool is present, add { topic_pool: topicPool }
    // Note: key name 'topic_pool' is different from variable name 'topicPool'
    ...(topicPool != null && { topic_pool: topicPool }),
  };
```

static/js/sefaria/sefaria.js
```diff
@@ -1351,15 +1351,22 @@ Sefaria = extend(Sefaria, {
...
     if (orderByMatchedLength) { params["order_by_matched_length"] = 1; }
-    let queryString = Object.keys(params).map(key => key + '=' + params[key]).join('&');
+    let queryString = Object.keys(params).map(key => {
+        const value = params[key];
+        if (Array.isArray(value)) {
+            return value.map(v => `${encodeURIComponent(key)}=${encodeURIComponent(v)}`).join('&');
+        }
+        return `${encodeURIComponent(key)}=${encodeURIComponent(value)}`;
+    }).join('&');
     queryString = (queryString ? "?" + queryString : "");
```
yodem: 1. its better not to use let.
2. if you want to turn the array into a flatten array and manipulate its elements with based on some logic its better to use reduce.
3. I think this logic is complicated enough so I would suggets to export it to a different util function.
something like:
```
const toQueryString = (params) => {
  const querySegments = Object.entries(params).reduce((acc, [key, value]) => {
    const encodedKey = encodeURIComponent(key);
    
    if (Array.isArray(value)) {
      // If the value is an array, push multiple key=value segments
      value.forEach(v => {
        // Ensure values in the array are also properly encoded
        acc.push(`${encodedKey}=${encodeURIComponent(v)}`);
      });
    } else {
      // Otherwise, push a single key=value segment
      acc.push(`${encodedKey}=${encodeURIComponent(value)}`);
    }
    return acc;
  }, []); // Start with an empty array to collect all segments

  const queryString = querySegments.join('&');
  
  // Prepend '?' only if the query string is not empty
  return queryString ? `?${queryString}` : '';
};

```

static/js/sheets.js
```diff
@@ -160,7 +160,7 @@ $(function() {
 	$(document).on("click", "#inlineAddSourceOK", function() {
 		var $target = $("#addInterface").prev(".sheetItem");
         var ref = $("#inlineAdd").val();
-		Sefaria.getName(ref, undefined, 'ref').then(function(q) {
+		Sefaria.getName(ref, undefined, ['ref']).then(function(q) {
```
yodem, line 163: maybe use async await?

YishaiGlasner: i only added brackets

static/js/sefaria/sefaria.js
```diff
@@ -1349,18 +1349,28 @@ Sefaria = extend(Sefaria, {
...
     const trimmed_name = name.trim();
-    let params = {active_module: this.activeModule};
-    // if (refOnly) { params["ref_only"] = 1; }
-    if (limit != undefined) { params["limit"] = limit; }
-    if (type != undefined) { params["type"] = type; }
-    if (exactContinuations) { params["exact_continuations"] = 1; }
-    if (orderByMatchedLength) { params["order_by_matched_length"] = 1; }
-    let queryString = Object.keys(params).map(key => key + '=' + params[key]).join('&');
-    queryString = (queryString ? "?" + queryString : "");
+    let params = {
```
yodem: I think const would be better than let

static/js/sefaria/sefaria.js
```diff
@@ -1349,18 +1349,28 @@ Sefaria = extend(Sefaria, {
...
   },
   _lookups: {},
-
+  buildQueryString(params) {
+    // params is an object where value is string or array of strings
+    const encodePair = (key, value) =>
+      Array.isArray(value)
+        ? value.map(v => encodePair(key, v))
+        : `${encodeURIComponent(key)}=${encodeURIComponent(value)}`;
+    let queryString = Object.entries(params)
```
yodem: const is generally preferred instead of let

PR #2715, opened 2025-09-30
Slack notifications for when accounts are deleted

relyks:
## Description
This adds Slack notifications to the #engineering-signal channel when user accounts are deleted, providing visibility into account deletions for stakeholders. Accounts are manually reviewed and deleted by administrators in different sources like the email management platform and CRM besides the MongoDB and Postgresql database instances.

## Code Changes
* `sefaria/helper/slack/send_message.py`: Enhanced the module for sending messaging to Slack by using the modern blocks API. The function usage allows for backwards compatibility
* ` sefaria/utils/user.py`: Updated account deletion flow
  * Added `deletion_type` parameter ("self", "admin", "manual")
  * Added admin_user parameter to track who performed admin deletions
  * Sends Slack notifications with deletion context (user info, deletion source), so stakeholders are aware of how and why the account deletion occurred
* Updated deletion call sites to have additional context based on the `deletion_type`
    * `reader/views.py` - Self-service user deletion endpoint
    * `sefaria/views.py` - Admin deletion endpoint
    * `sefaria/utils/user.py` - User merge function

## Notes
By tracking `deletion_type` and `admin_user`, this implements a lightweight audit trail that helps with debugging and visibility (no emails are being sent)


sefaria/utils/user.py
```diff
@@ -5,24 +5,41 @@
...
+from sefaria.helper.slack.send_message import send_message
 
 logger = structlog.get_logger(__name__)
 
 
-def delete_user_account(uid, confirm=True):
+def delete_user_account(
+    uid: int,
+    confirm: bool = True,
+    deletion_type: Literal["self", "admin", "manual"] = "manual",
```
yitzhakc: Please don't use magic strings. Turn this into an enum or consts.

sefaria/utils/user.py
```diff
@@ -47,11 +64,41 @@ def delete_user_account(uid, confirm=True):
...
+            # Get admin profile for name and email
+            try:
+                admin_profile: model.UserProfile = model.UserProfile(id=admin_user.id)
+                admin_name: str = admin_profile.full_name or "No name given"
+                admin_email: str = str(admin_user.email)
+                deletion_source: str = f"Admin deleted by {admin_name} ({admin_email})"
+            except:
+                deletion_source: str = f"Admin deleted by {str(admin_user.email)}"
+        else:
+            deletion_source: str = {
```
yitzhakc, line 103: This looks like it should be hoisted out of the function, or at the very least initialized at the beginning.

sefaria/utils/user.py
relyks: Changed to enum

sefaria/utils/user.py
relyks: Done

yitzhakc (dismissed):

sefaria/utils/user.py
```diff
@@ -5,24 +5,62 @@
...
+    user_name: str = user.full_name or "No name given"
+
+    # Get admin info upfront if this is an admin deletion
+    admin_name: Optional[str] = None
+    admin_email: Optional[str] = None
+    if deletion_type == DeletionType.ADMIN and admin_user:
+        admin_email = str(admin_user.email)
+        try:
+            admin_profile: model.UserProfile = model.UserProfile(id=admin_user.id)
+            admin_name = admin_profile.full_name or "No name given"
```
yitzhakc, line 54: The user data isn't synced between the `UserProfile` in Mongo and the `User` model in Django? 🤯 
I would have thought that if we have `admin_user`, we can just do `f"{admin_user.first_name} {admin_user.last_name}"`

sefaria/utils/user.py
yitzhakc: I had meant the entire dictionary, not only the constants...

sefaria/utils/user.py
```diff
@@ -47,11 +64,41 @@ def delete_user_account(uid, confirm=True):
...
+            username='Account Manager',
+            pretext='User Account Deleted',
+            text=f'{deletion_source}\nUser: {user_name} ({user_email})\nUser ID: {uid}',
+            icon_emoji=':warning:'
+        )
+        if not success:
+            logger.warning(f"Failed to send Slack notification for user deletion: {uid}")
+    except Exception as e:
+        logger.error(f"Error sending Slack notification for user deletion {uid}: {e}")
+
```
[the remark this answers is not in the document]
yitzhakc: Do we have any test coverage on slack sending messages?
We can add a test and mock away the actual `send_message` function

sefaria/views.py
```diff
@@ -1215,7 +1215,7 @@ def delete_user_by_email(request):
             return jsonResponse({"failure": "incorrect password"})
         try:
             id_to_delete = UserProfile(email=email)
```
yitzhakc, line 1217: I would have thought that `UserProfile(email=email)` returns the entire `UserProfile`. Maybe weo don't need to do it manually on sefaria/utils/user.py:53...

yitzhakc (changes requested):

PR #2719, opened 2025-10-08
chore: put inside script tags

stevekaplan123:
## Description
Django-csrf.js needs to be inside script tags to actually run and not show as text!

### DIFF
--- a/templates/edit_profile.html
+++ b/templates/edit_profile.html
@@ -149,7 +149,7 @@
 {% block js %}
 	<script src="{% static 'ckeditor/ckeditor.js' %}"></script>
 	<script src="{% static 'ckeditor/adapters/jquery.js' %}"></script>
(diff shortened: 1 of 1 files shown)


templates/edit_profile.html
```diff
@@ -149,8 +149,8 @@ <h2>
 {% block js %}
 	<script src="{% static 'ckeditor/ckeditor.js' %}"></script>
 	<script src="{% static 'ckeditor/adapters/jquery.js' %}"></script>
-	{% include "js/django-csrf.js" %}
-	<script>
+	<script>{% include "js/django-csrf.js" %}</script>
```
YishaiGlasner: why not
```suggestion
	<script src="{% static 'js/django-csrf.js' %}"></script>
```

yitzhakc (approved):

PR #2726, opened 2025-10-16
Hosts.py Iteration over Domain Modules

stevekaplan123:
## Description
I added iteration to hosts.py and also fixed some links broken by a recent merge.

### DIFF
--- a/sefaria/hosts.py
+++ b/sefaria/hosts.py
@@ -1,5 +1,5 @@
 from django_hosts import patterns, host
-from sefaria import settings    
+from django.conf import settings
 from sefaria.constants.model import LIBRARY_MODULE, VOICES_MODULE
 from urllib.parse import urlparse
 import re
--- a/sefaria/client/util.py
+++ b/sefaria/client/util.py
@@ -6,7 +6,7 @@ from django.http import HttpResponse
 from django.core.mail import EmailMultiAlternatives
 from webpack_loader import utils as webpack_utils
 
-from sefaria import settings as sls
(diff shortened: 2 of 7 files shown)


sefaria/hosts.py
```diff
@@ -7,8 +7,16 @@
...
+# Build domain patterns, including Hebrew domains only if they're configured
+library_domains = [get_domain_reg(settings.DOMAIN_MODULES["en"][LIBRARY_MODULE])]
+sheets_domains = [get_domain_reg(settings.DOMAIN_MODULES["en"][VOICES_MODULE])]
+
+if "he" in settings.DOMAIN_MODULES:
+    library_domains.append(get_domain_reg(settings.DOMAIN_MODULES["he"][LIBRARY_MODULE]))
+    sheets_domains.append(get_domain_reg(settings.DOMAIN_MODULES["he"][VOICES_MODULE]))
+
+library_domain = '|'.join(library_domains)
+sheets_domain = '|'.join(sheets_domains)
```
yodem: We can iterate over the settings.DOMAIN_MODULES keys and append the values per iteration.

sefaria/hosts.py
yitzhakc, line 2: Django does not recommend importing the settings file directly.
We should be using it through
```
from django.conf import settings
```

yitzhakc (changes requested):

sefaria/hosts.py
stevekaplan123: Changed

PR #2727, opened 2025-10-16
feat: Add batch upload support for Workflowy outline files

yitzhakc:
## Description
Enables moderators to upload multiple Workflowy (.opml) files simultaneously, significantly improving workflow efficiency for bulk text imports.

## Code Changes
Backend changes (sefaria/views.py):
- Accept both single file (wf_file) and multiple files (workflowys[])
- Process files in a loop, collecting results and error messages
- Maintain backward compatibility for single file uploads
- Improved error handling with per-file error reporting
- Return appropriate responses based on single vs. batch upload

Frontend changes (static/js/ModeratorToolsPanel.jsx):
- Added multiple file selection support with file list display
- Dynamic upload button showing file count for batch uploads
- File state management and validation
- Automatic form reset after successful upload
- Enhanced user feedback during upload process

## Notes 
Extracted from the `kehati` branch, see #2527

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -1375,8 +1375,8 @@ def modtools_upload_workflowy(request):
     c_index = request.POST.get("c_index", "").lower() == "true"
     c_version = request.POST.get("c_version", "").lower() == "true"
     
-    delims = request.POST.get("delims", None) if len(request.POST.get("delims", "")) else None
-    term_scheme = request.POST.get("term_scheme", None) if len(request.POST.get("term_scheme", "")) else None
+    delims = request.POST.get("delims") if request.POST.get("delims", "") != "" else None
+    term_scheme = request.POST.get("term_scheme") if request.POST.get("term_scheme", "") != "" else None
 
     uid = request.user.id
 
--- a/static/js/ModeratorToolsPanel.jsx
+++ b/static/js/ModeratorToolsPanel.jsx
@@ -7,6 +7,7 @@ import Cookies from 'js-cookie';
 import { saveAs } from 'file-saver';
 import qs from 'qs';
 import { useState } from 'react';
+import { InterfaceText, EnglishText, HebrewText } from "./Misc";
 
 class ModeratorToolsPanel extends Component {
   constructor(props) {
@@ -218,20 +219,7 @@ class WorkflowyModeratorTool extends Component{
 
     this.setState({uploading: true, uploadMessage: `Uploading ${this.state.files.length} file${this.state.files.length > 1 && 's'}...`});
 
-    const data = new FormData();
-
-    // Add files 
-    if (this.state.files.length >= 1) {
-      this.state.files.forEach(file => {
-        data.append('workflowys[]', file);
-      });
-    }
-
-    // Add other form data
-    data.append('c_index', this.state.c_index ? true : false);
-    data.append('c_version', this.state.c_version ? true : false);
-    data.append('delims', this.state.delims || '');
-    data.append('term_scheme', this.state.term_scheme || '');
+    const data = new FormData(event.target);
 
     const request = new Request(
         '/modtools/upload_text',
@@ -325,7 +313,7 @@ class WorkflowyModeratorTool extends Component{
               Upload Workflowy file(s):
               <input
                 type="file"
-                name="workflowys"
+                name="workflowys[]"
                 ref={this.wfFileInput}
                 multiple
                 accept=".opml"
@@ -342,6 +330,7 @@ class WorkflowyModeratorTool extends Component{
               <input
                 name="c_index"
                 type="checkbox"
+                value="true"
                 checked={this.state.c_index}
                 onChange={this.handleInputChange} />
            </label>
@@ -350,7 +339,8 @@ class WorkflowyModeratorTool extends Component{
               <input
                 name="c_version"
                 type="checkbox"
-                checked={this.state.c_version || false}
+                value="true"
+                checked={this.state.c_version}
                 onChange={this.handleInputChange} />
            </label>
            <label>
@@ -377,14 +367,16 @@ class WorkflowyModeratorTool extends Component{
                type="submit"
                disabled={this.state.uploading || this.state.files.length === 0}
              >
-                <span className="int-en">
+              <InterfaceText>
+                <EnglishText className="int-en">
                   {this.state.uploading ? 'Uploading...' :
                    this.state.files.length > 1 ? `Upload ${this.state.files.length} Files` : 'Upload'}
-                </span>
-                <span className="int-he">
-                  {this.state.uploading ? 'מעלה...' : 
+                </EnglishText>
+                <HebrewText className="int-he">
+                  {this.state.uploading ? 'מעלה...' :
                    this.state.files.length > 1 ? `העלאת ${this.state.files.length} קבצים` : 'העלאה'}
-                </span>
+                </HebrewText>
+              </InterfaceText>
              </button>
          </form>
         <div id="wf-upl-msg" className="wf-upl-msg">{this.state.uploadMessage || ""}</div>


sefaria/views.py
```diff
@@ -1364,20 +1364,60 @@ def modtools_upload_workflowy(request):
...
+    elif 'workflowys[]' in request.FILES:
+        # Multiple file upload
+        files = request.FILES.getlist("workflowys[]")
+    else:
+        return jsonResponse({"error": "No files provided. Use 'wf_file' for single file or 'workflowys[]' for multiple files."})
+
+    # Handle checkbox parameters - support both boolean and string formats
     c_index = request.POST.get("c_index", False)
+    if isinstance(c_index, str):
+        c_index = c_index.lower() == 'true'
```
YishaiGlasner: why not

```suggestion
c_index = request.GET.get('c_index', '').lower() == 'true'
```
with no condition?
also, maybe we should have a util function for getting bool query param (and it seems that in most views we use 0/1 for booleans).

sefaria/views.py
```diff
@@ -1364,20 +1364,60 @@ def modtools_upload_workflowy(request):
...
+            results.append(res)
+            messages.append(f"Imported {file.name}")
+        except Exception as e:
+            if len(files) == 1:
+                # Single file - maintain backward compatibility by raising exception
+                raise e
+            else:
+                # Multiple files - return error with partial results
+                error_message = f"Failed to process {file.name}: {e}"
+                return jsonResponse({"error": error_message, "message": ' • '.join(messages)})
```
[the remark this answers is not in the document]
YishaiGlasner: what i wanted to say

static/js/ModeratorToolsPanel.jsx
```diff
@@ -191,18 +197,48 @@ class WorkflowyModeratorTool extends Component{
...
+      this.state.files.forEach(file => {
+        data.append('workflowys[]', file);
+      });
+    }
+
+    // Add other form data
+    data.append('c_index', this.state.c_index);
+    data.append('c_version', this.state.c_version || false);
+    data.append('delims', this.state.delims || '');
+    data.append('term_scheme', this.state.term_scheme || '');
```
YishaiGlasner: i think having the defaults in the constructor is better

sefaria/views.py
```diff
@@ -1364,20 +1364,60 @@ def modtools_upload_workflowy(request):
...
+            wfparser = WorkflowyParser(file, uid, term_scheme=term_scheme, c_index=c_index, c_version=c_version, delims=delims)
+            res = wfparser.parse()
+            if res.get("error"):
+                # If the parser returns a specific error, report it
+                raise InputError(f"Error in {file.name}: {res['error']}")
+            results.append(res)
+            messages.append(f"Imported {file.name}")
+        except Exception as e:
+            if len(files) == 1:
+                # Single file - maintain backward compatibility by raising exception
```
YishaiGlasner: why? it isn't complicated to refactor it to have the same response structure for both cases and it will be much more clean and clear

sefaria/views.py
yitzhakc: Added a util function

static/js/ModeratorToolsPanel.jsx
yitzhakc: What do you mean? I don't think we can add initial data to a `FormData()` constructor.

sefaria/views.py
yitzhakc: fixed

sefaria/utils/util.py
```diff
@@ -611,3 +611,11 @@ def decorated_function(*args, **kwargs):
             return return_value
         return decorated_function
     return argumented_decorator
+
+def parse_bool(val, default=False):
```
YishaiGlasner: i don't like this function:
1. it has no context here. it should go to somewhere realted to requests.
2. in the context of request i know we're using 0/1. why to have here string?
3. why not to handle 0/1?
4. how can `isinstance(val, bool)` be true if it came from a request?
5. conclusion - we should try to be consistent regarding our boolean qeury params and document our assumption sabout them in this function which should be in a place related to views.
I think that if you're not willing to this refactor now, you should omit this function and return to the previous parsing of `c_index` and the other params.

static/js/ModeratorToolsPanel.jsx
```diff
@@ -177,8 +177,14 @@ class WorkflowyModeratorTool extends Component{
...
     super(props);
     this.handleWfSubmit = this.handleWfSubmit.bind(this);
+    this.handleInputChange = this.handleInputChange.bind(this);
+    this.handleFileChange = this.handleFileChange.bind(this);
     this.wfFileInput = React.createRef();
-    this.state = {c_index: true};
+    this.state = {
+      c_index: true,
+      files: [],
+      multipleFiles: false
```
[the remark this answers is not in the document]
YishaiGlasner: that's true

static/js/ModeratorToolsPanel.jsx
```diff
@@ -177,8 +177,14 @@ class WorkflowyModeratorTool extends Component{
     constructor(props) {
     super(props);
     this.handleWfSubmit = this.handleWfSubmit.bind(this);
+    this.handleInputChange = this.handleInputChange.bind(this);
```
YishaiGlasner: can you use arrow function and get rif of these lines?

static/js/ModeratorToolsPanel.jsx
```diff
@@ -191,18 +197,48 @@ class WorkflowyModeratorTool extends Component{
...
-    this.setState({uploading: true, uploadMessage:"Uploading..."});
-    const data = new FormData(event.target);
-    console.log(data);
+
+    if (this.state.files.length === 0) {
+      this.setState({uploadMessage: "Please select at least one file", error: true, errorIsHTML: false});
+      return;
+    }
+
+    this.setState({uploading: true, uploadMessage: `Uploading ${this.state.files.length} file${this.state.files.length > 1 ? 's' : ''}...`});
```
YishaiGlasner:
```suggestion
    this.setState({uploading: true, uploadMessage: `Uploading ${this.state.files.length} file${this.state.files.length > 1 && 's'}...`});
```

static/js/ModeratorToolsPanel.jsx
```diff
@@ -191,18 +197,48 @@ class WorkflowyModeratorTool extends Component{
...
+
+    const data = new FormData();
```
YishaiGlasner: why not using `event.target`, and then only handle the files fields?

static/js/ModeratorToolsPanel.jsx
```diff
@@ -221,10 +257,45 @@ class WorkflowyModeratorTool extends Component{
...
+
+                const successes = resp_json.successes || [];
+                const failures = resp_json.failures || [];
+
+                let uploadMessage = "";
+                let uploadResult = "";
+
+                // Build summary message
+                if (failures.length === 0) {
+                    uploadMessage = `Successfully imported ${successes.length} file${successes.length > 1 ? 's' : ''}`;
```
YishaiGlasner:
```suggestion
                    uploadMessage = `Successfully imported ${successes.length} file${successes.length > 1 && 's'}`;
```
and so on

static/js/ModeratorToolsPanel.jsx
```diff
@@ -289,9 +372,20 @@ class WorkflowyModeratorTool extends Component{
...
-             <button className="modtoolsButton" name="wf-submit" type="submit">
-                <span className="int-en">Upload</span>
-                <span className="int-he">Upload</span>
+             <button
+               className="modtoolsButton"
+               name="wf-submit"
+               type="submit"
+               disabled={this.state.uploading || this.state.files.length === 0}
+             >
+                <span className="int-en">
```
YishaiGlasner: use `InterfaceText`

YishaiGlasner (changes requested):

static/js/ModeratorToolsPanel.jsx
yitzhakc: Yes!

static/js/ModeratorToolsPanel.jsx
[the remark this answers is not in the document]
yitzhakc: Fixed

static/js/ModeratorToolsPanel.jsx
yitzhakc: Fixed

static/js/ModeratorToolsPanel.jsx
yitzhakc: Because I'm not trying to refactor the entire tool 🙃

YishaiGlasner: but previously we used `event.target`. you cahnged it

static/js/ModeratorToolsPanel.jsx
YishaiGlasner: i refer to the occurrences of `this.state.term_scheme ||`.

sefaria/views.py
```diff
@@ -1364,20 +1364,40 @@ def modtools_upload_workflowy(request):
...
+    if 'workflowys[]' in request.FILES:
+        files = request.FILES.getlist("workflowys[]")
+    else:
+        return jsonResponse({"error": "No files provided. Use 'workflowys[]' for multiple files."})
+
+    # Handle checkbox parameters
+    c_index = request.POST.get("c_index", "").lower() == "true"
+    c_version = request.POST.get("c_version", "").lower() == "true"
+    
+    delims = request.POST.get("delims", None) if len(request.POST.get("delims", "")) else None
```
YishaiGlasner: if len is 0, it\s falsey, so you can do it simplier:
```suggestion
    delims = request.POST.get("delims", None) or None
```

static/js/ModeratorToolsPanel.jsx
yitzhakc: Right. Now that I've fixed the default there, I'll remove this.

sefaria/views.py
yitzhakc: Alas, `delims` and `term_scheme` are parsed in Django as empty strings which are not Falsy. I've made it more explicit though.

static/js/ModeratorToolsPanel.jsx
yitzhakc: Whoops! You are 100% correct. Fixed.

static/js/ModeratorToolsPanel.jsx
yitzhakc: Done

YishaiGlasner (approved):

PR #2740, opened 2025-10-20
Bug/sc 36465/secondtry

stevekaplan123:
## Description
Re-factors saved and history pages and simplifies logic to avoid several bugs.

## Code Changes
1. Consolidates routes (instead of /texts/saved and /sheets/saved, there's just /saved).  Consolidates menuOpen values (e.g. texts-saved → saved, texts-history → history) and removes unnecessary wrapper components for sheets vs library cases.
2. Removes filter logic and dedupe logic from front-end and moves it to the back-end.  By moving this logic to the back-end, the front-end can focus on display issues instead of convoluted logic to figure out what is supposed to be displayed depending on what page we're on (sheets, history or notes) and what module we're in (library or voices).  Specifically, the logic is removed from the UserHistoryPanel.  The PR adds a `sheets_only` flag to the user_history API based on active_module, and also uses request.active_module for the same purpose in the views for "/saved" and "/history".  This allows us to return sheets in the voices module and return texts in the library module.  In addition, dedupe logic now exists in the back-end for history pages to make sure multiple items in a row aren't from the same book or same sheet.

### DIFF
--- a/static/js/UserHistoryPanel.jsx
+++ b/static/js/UserHistoryPanel.jsx
@@ -123,11 +123,11 @@ const dedupeItems = (items) => {
   */
   const deduped = [];
   const key = Sefaria.activeModule === Sefaria.VOICES_MODULE ? 'sheet_id' : 'book';  
-  let prev;
+  let prevValue;
   for (const item of items) {
-    if (item[key] !== prev) {
+    if (item[key] !== prevValue) { // item[key] is the name of the sheet or book
       deduped.push(item);
-      prev = item[key];
+      prevValue = item[key];
     }
   }
   return deduped;


static/js/UserHistoryPanel.jsx
```diff
@@ -17,27 +17,13 @@ import {
...
+const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNav, compare, toggleSignUpModal}) => {
 
   const [notes, setNotes] = useState(null);
-  const [dataStore, setDataStore] = useState(initialStore);
-
   const contentRef = useRef();
 
   useEffect(() => {
-    Sefaria.allPrivateNotes((data) => {
+    menuOpen === 'notes' && Sefaria.activeModule === 'library' && Sefaria.allPrivateNotes((data) => {
```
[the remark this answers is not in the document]
stevekaplan123: activeModule won't change unless module is changed which would re-render the page and component

reader/views.py
```diff
@@ -189,8 +189,8 @@ def base_props(request):
             "calendars": get_todays_calendar_items(**_get_user_calendar_params(request)),
             "notificationCount": profile.unread_notification_count(),
             "notifications": profile.recent_notifications(scope=active_module).client_contents(),
-            "saved": {"loaded": False, "items": profile.get_history(saved=True, secondary=False, serialized=True, annotate=False)}, # saved is initially loaded without text annotations so it can quickly immediately mark any texts/sheets as saved, but marks as `loaded: false` so the full annotated data will be requested if the user visits the saved/history page
-            "last_place": profile.get_history(last_place=True, secondary=False, sheets=False, serialized=True)
+            "saved": {"loaded": False, "items": profile.get_history(saved=True, secondary=False, serialized=True, annotate=False, sheets_only=active_module == VOICES_MODULE)}, # saved is initially loaded without text annotations so it can quickly immediately mark any texts/sheets as saved, but marks as `loaded: false` so the full annotated data will be requested if the user visits the saved/history page
```
YishaiGlasner: doing `sheets_only = active_module == VOICES_MODULE` and then `sheets_only=sheets_only` in the call will be more readable

reader/views.py
```diff
@@ -1035,19 +1035,17 @@ def saved_content(request):
...
+    return menu_page(request, props, page="saved", title=title, desc=desc)
+
 
 
 def get_user_history_props(request):
     if request.user.is_authenticated:
         profile = UserProfile(user_obj=request.user)
-        uhistory =  profile.get_history(secondary=False, serialized=True, annotate=True, limit=20) if profile.settings.get("reading_history", True) else []
+        sheets_only = request.active_module == VOICES_MODULE
+        uhistory =  profile.get_history(secondary=False, serialized=True, annotate=True, limit=20, sheets_only=sheets_only) if profile.settings.get("reading_history", True) else []
```
YishaiGlasner: this is hard to read. please move `profile.settings.get("reading_history", True)` to a var. also i think that a full if (not one liner) would be better in this case

sefaria/model/user_profile.py
```diff
@@ -208,14 +208,30 @@ def get_user_history(uid=None, oref=None, saved=None, secondary=None, last_place
...
+        query["is_sheet"] = sheets_only
         if secondary is not None:
             query["secondary"] = secondary
         if last_place is not None:
             query["last_place"] = last_place
         if serialized:
-            return [uh.contents(for_api=True, annotate=annotate) for uh in UserHistorySet(query, proj={"uid": 0, "server_time_stamp": 0}, sort=[("time_stamp", -1)], limit=limit, skip=skip)]
+            items = [uh.contents(for_api=True, annotate=annotate) for uh in UserHistorySet(query, proj={"uid": 0, "server_time_stamp": 0}, sort=[("time_stamp", -1)], limit=limit, skip=skip)]
+            if not saved:
+                # Deduplicate consecutive items with the same book or sheet_id for history mode
```
YishaiGlasner: move to a separate function

sefaria/model/user_profile.py
```diff
@@ -208,14 +208,30 @@ def get_user_history(uid=None, oref=None, saved=None, secondary=None, last_place
...
+                        deduped_items.append(item)
+                    else:
+                        prev = deduped_items[-1]
+                        # Check if current item is different from previous
+                        if item.get('is_sheet') and item.get('sheet_id') == prev.get('sheet_id'):
+                            continue  # Skip duplicate sheet
+                        elif not item.get('is_sheet') and item.get('book') == prev.get('book'):
+                            continue  # Skip duplicate book
+                        else:
+                            deduped_items.append(item)
```
YishaiGlasner:
```suggestion
                        key = 'sheet_id' if item.get('is_sheet') else 'book'
                        if item.get(key) != prev.get(key)::
                            deduped_items.append(item)
```

static/js/sefaria/sefaria.js
```diff
@@ -2659,7 +2659,8 @@ _media: {},
   userHistory: {loaded: false, items: []},
   loadUserHistory: function (limit, callback) {
       const skip = Sefaria.userHistory.items.length;
-      const url = `/api/profile/user_history?secondary=0&annotate=1&limit=${limit}&skip=${skip}`;
+      const sheets_only = Sefaria.activeModule === Sefaria.VOICES_MODULE ? 1 : 0;
```
```suggestion
      const sheets_only = Number(Sefaria.activeModule === Sefaria.VOICES_MODULE);
```

static/js/UserHistoryPanel.jsx
YishaiGlasner: i think using `&&` is not the best practice. it's clear when you want to assign (or inline jsx rendering), but not when you want to call a function.

static/js/UserHistoryPanel.jsx
```diff
@@ -48,58 +34,37 @@ const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNa
...
 
   const title = (
     <span className="sans-serif">
-      <a
-        href={ dataSource === 'library' ?  libraryURLs.saved : sheetsURLs.saved }
-        data-target-module={dataSource === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
-        className={"navTitleTab" + (menuOpen === 'texts-saved' || menuOpen === 'sheets-saved' ? ' current' : '') }
-        onKeyDown={(e) => Util.handleKeyboardClick(e)}
-      >
+      <a href="/saved" data-target-module={Sefaria.activeModule === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
```
YishaiGlasner: indentation and newlines should be as before

static/js/UserHistoryPanel.jsx
```diff
@@ -48,58 +34,37 @@ const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNa
...
         <img src="/static/icons/bookmark.svg" alt={Sefaria._("Saved")} />
         <InterfaceText>Saved</InterfaceText>
       </a>
-      <a
-        href={ dataSource === "library" ?  libraryURLs.history : sheetsURLs.history }
-        data-target-module={dataSource === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
-        className={"navTitleTab" + (menuOpen === 'texts-history' || menuOpen === 'sheets-history' ? ' current' : '')}
-        onKeyDown={(e) => Util.handleKeyboardClick(e)}
-      >
+      <a href="/history" data-target-module={Sefaria.activeModule === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
```
YishaiGlasner: indentation and newlines

static/js/UserHistoryPanel.jsx
```diff
@@ -48,58 +34,37 @@ const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNa
...
         <InterfaceText>History</InterfaceText>
       </a>
-      { dataSource === "library" &&
-        <a
-          href="/texts/notes"
-          className={"navTitleTab" + (menuOpen === 'notes' ? ' current' : '')}
-          onKeyDown={(e) => Util.handleKeyboardClick(e)}
-        >
+      { Sefaria.activeModule === "library" &&
+        <a href="/texts/notes" className={"navTitleTab" + (menuOpen === 'notes' ? ' current' : '')}
```
YishaiGlasner: indentation and newlines

static/js/UserHistoryPanel.jsx
```diff
@@ -48,58 +34,37 @@ const UserHistoryPanel = ({menuOpen, toggleLanguage, openDisplaySettings, openNa
...
-  const libraryURLs = {
-    "saved": "/texts/saved",
-    "history": "/texts/history"
-  };
-  const sheetsURLs = {
-    "saved": "/saved",
-    "history": "/history"
+  const currentDataStore = menuOpen === 'saved' ? Sefaria.saved : Sefaria.userHistory;
+  const store = {
+    'loaded': currentDataStore?.loaded || false,
```
YishaiGlasner: previously it was always true, isn't it?

static/js/UserHistoryPanel.jsx
```diff
@@ -145,52 +109,29 @@ UserHistoryPanel.propTypes = {
...
 
-const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal, dataSource}) => {
+const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal}) => {
   const [items, setItems] = useState(store.loaded ? store.items : null);
 
   // Store changes when switching tabs, reset items
   useEffect(() => {
     setItems(store.loaded ? store.items : null);
-  }, [store]);
+  }, [menuOpen]);
```
YishaiGlasner: is this `useEffect` is needed? i if understand correctly it was for switching tabs, but now there are no tabs

static/js/UserHistoryPanel.jsx
```diff
@@ -145,52 +109,29 @@ UserHistoryPanel.propTypes = {
...
+  
+  const savedParam = menuOpen === 'saved' ? '&saved=1' : '';
+  const sheetsOnlyParam = Sefaria.activeModule === Sefaria.VOICES_MODULE ? '&sheets_only=1' : '';
```
YishaiGlasner: i think

```suggestion
  const saved = Number(menuOpen === 'saved');
  const sheetsOnly = Number(Sefaria.activeModule === Sefaria.VOICES_MODULE);
```
and then
```
    url: `/api/profile/user_history?secondary=0&annotate=1&saved=${saved}&sheets_only=${sheetsOnly},
```
is cleaner.
is there reason not to have the param as 0?

sefaria/model/user_profile.py
YishaiGlasner:
```suggestion
                prev = object()
                for item in items:
                        # Check if current item is different from previous
                        key = 'sheet_id' if item.get('is_sheet') else 'book'
                        if item.get(key) != prev.get(key):
                            deduped_items.append(item)
                        prev = item
```

sefaria/model/user_profile.py
```diff
@@ -208,14 +208,30 @@ def get_user_history(uid=None, oref=None, saved=None, secondary=None, last_place
...
-            query["is_sheet"] = sheets
+        query["is_sheet"] = sheets_only
         if secondary is not None:
             query["secondary"] = secondary
         if last_place is not None:
             query["last_place"] = last_place
         if serialized:
-            return [uh.contents(for_api=True, annotate=annotate) for uh in UserHistorySet(query, proj={"uid": 0, "server_time_stamp": 0}, sort=[("time_stamp", -1)], limit=limit, skip=skip)]
+            items = [uh.contents(for_api=True, annotate=annotate) for uh in UserHistorySet(query, proj={"uid": 0, "server_time_stamp": 0}, sort=[("time_stamp", -1)], limit=limit, skip=skip)]
+            if not saved:
```
YishaiGlasner: is there a situation where we don't want to dedupe? if not, we can do it always.
(if yes, maybe we should keep do it in client side)

stevekaplan123: We don't dedupe in 'saved' mode because we want exactly what the user chose.  I felt like it made sense to do it on the backend because the frontend ideally should be rendering data from the backend.

static/js/UserHistoryPanel.jsx
stevekaplan123: Well, it may have been that way in this component but I don't think that makes sense as it is initialized to false in sefaria.js.

static/js/UserHistoryPanel.jsx
stevekaplan123: I don't see what is wrong with the newlines and indentation in my PR.  Can you explain?

sefaria/model/user_profile.py
YishaiGlasner: so it isn't really deduption, it's filtering, isn't it?
and this is a presntation concern and not data integrity concern, so i think it should be done by client side.

static/js/UserHistoryPanel.jsx
YishaiGlasner: so the behaviour will change? are you sure that it's correct now?

static/js/UserHistoryPanel.jsx
YishaiGlasner: you have here 2 attrs in a line, but not all of them. that's very confusing for me. i would do:

```suggestion
        <a 
           href="/texts/notes" 
           className={"navTitleTab" + (menuOpen === 'notes' ? ' current' : '')}
           onKeyDown={(e) => Util.handleKeyboardClick(e)}
        >
```

sefaria/model/user_profile.py
stevekaplan123: You have a point.  Deduping is now in front end.

static/js/UserHistoryPanel.jsx
stevekaplan123: I don't know for sure.  But my code does fix the bugs it was supposed to.  I don't see how it could be bad for the code to also allow for 'loaded' to be false.

static/js/sefaria/sefaria.js
```diff
@@ -2668,7 +2668,9 @@ _media: {},
   userHistory: {loaded: false, items: []},
   loadUserHistory: function (limit, callback) {
       const skip = Sefaria.userHistory.items.length;
-      const url = `/api/profile/user_history?secondary=0&annotate=1&limit=${limit}&skip=${skip}`;
+      const saved = 0;
+      const sheetsOnly = Number(Sefaria.activeModule === Sefaria.VOICES_MODULE);
+      const url = `/api/profile/user_history?secondary=0&annotate=1&limit=${limit}&skip=${skip}&saved=${saved}&sheets_only=${sheetsOnly}`;
```
```suggestion
const params = new URLSearchParams({
  secondary: 0,
  annotate: 1,
  limit,
  skip: Sefaria.userHistory.items.length,
  saved: 0,
  sheets_only: +(Sefaria.activeModule === Sefaria.VOICES_MODULE);,
});

const url = `/api/profile/user_history?${params.toString()}`;
```
(i didn't know about this `+(bool)` syntax)

static/js/UserHistoryPanel.jsx
```diff
@@ -17,89 +17,59 @@ import {
...
-    const newDataStore = menuOpen === 'texts-saved' || menuOpen === 'sheets-saved' ? Sefaria.saved : Sefaria.userHistory;
-    setDataStore(newDataStore);
-  }, [menuOpen, Sefaria.saved, Sefaria.userHistory]);
+    if (menuOpen === 'notes' && Sefaria.activeModule === 'library') {
+      Sefaria.allPrivateNotes((data) => {
+        if (Array.isArray(data)) {
+          const flattenedNotes = data.map(note => ({
+            ref: note.ref,
+            text: note.text
+          }));
```
YishaiGlasner:
```suggestion
          const flattenedNotes = data.map(({ref, text}) => ({ref, text}));
```

static/js/UserHistoryPanel.jsx
```diff
@@ -17,89 +17,59 @@ import {
...
 
   const title = (
     <span className="sans-serif">
-      <a
-        href={ dataSource === 'library' ?  libraryURLs.saved : sheetsURLs.saved }
-        data-target-module={dataSource === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
-        className={"navTitleTab" + (menuOpen === 'texts-saved' || menuOpen === 'sheets-saved' ? ' current' : '') }
-        onKeyDown={(e) => Util.handleKeyboardClick(e)}
-      >
+      <a href="/saved" data-target-module={Sefaria.activeModule === 'library' ? Sefaria.LIBRARY_MODULE : Sefaria.VOICES_MODULE}
```
YishaiGlasner: newlines

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +114,59 @@ UserHistoryPanel.propTypes = {
...
-  return (
-    <UserHistoryPanel menuOpen={menuOpen}
-                      toggleLanguage={toggleLanguage}
-                      openDisplaySettings={openDisplaySettings}
-                      openNav={openNav}
-                      compare={compare}
-                      toggleSignUpModal={toggleSignUpModal}
-                      dataSource={'sheets'} />
-  )
+const dedupeItems = (items, saved) => {
```
YishaiGlasner: I think it's more reasonable to ommit `saved` here, and to call the function only when you want to dedupe

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +114,59 @@ UserHistoryPanel.propTypes = {
...
+  }
+  
+  const deduped = [];
+  let prev = {};
+  
+  for (const item of items) {
+    const key = item.is_sheet ? 'sheet_id' : 'book';
+    if (item[key] !== prev[key]) {
+      deduped.push(item);
+      prev = item;
```
YishaiGlasner: i don't know if for first ittem it can be that `item[key]` will be undefined, but if so it will be equal to `prev[key]`. that can solve it:

```suggestion
let prevKey;
for (const item of items) {
  const key = item.is_sheet ? 'sheet_id' : 'book';
  if (item[key] !== prevKey) {
    deduped.push(item);
    prevKey = item[key];
```
(that also saves some memory, for `prevKey` is lighter than the whole `prev`)

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +114,59 @@ UserHistoryPanel.propTypes = {
...
+const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal}) => {
+  const [items, setItems] = useState(store.loaded ? store.items : null);
+  
+  const saved = Number(menuOpen === 'saved');
+  const sheetsOnly = Number(Sefaria.activeModule === Sefaria.VOICES_MODULE);
 
   useScrollToLoad({
     scrollableRef: scrollableRef,
-    url: "/api/profile/user_history?secondary=0&annotate=1" + ((menuOpen === 'sheets-saved' || menuOpen === 'texts-saved') ? '&saved=1' : ''),
+    url: `/api/profile/user_history?secondary=0&annotate=1&saved=${saved}&sheets_only=${sheetsOnly}`,
```
YishaiGlasner: use `URLSearchParams`, and `+` rather than `Number`

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +113,65 @@ UserHistoryPanel.propTypes = {
...
+const dedupeItems = (items) => {
+  /*
+  Deduplicates consecutive items with the same book or sheet_id.  
+  Essentially, we don't want two history items in a row that are of the same book (or if we're in voices, the same sheet).
+  :param items: list of UserHistory objects to deduplicate
+  :return: list of deduplicated items
+  */
+  const deduped = [];
+  const key = Sefaria.activeModule === Sefaria.VOICES_MODULE ? 'sheet_id' : 'book';  
+  let prev;
```
YishaiGlasner: i think `prevKey` is more clear

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +113,65 @@ UserHistoryPanel.propTypes = {
...
 };
 
-const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal, dataSource}) => {
-  const [items, setItems] = useState(store.loaded ? store.items : null);
 
-  // Store changes when switching tabs, reset items
-  useEffect(() => {
-    setItems(store.loaded ? store.items : null);
-  }, [store]);
+const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal}) => {
```
YishaiGlasner:
```suggestion
const userHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal}) => {
```
also, all out hooks should be in Hooks.jsx

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +113,65 @@ UserHistoryPanel.propTypes = {
...
+  // Store raw items in state (never deduped)
+  const [rawItems, setRawItems] = useState(store.loaded ? store.items.slice() : null);
+  
+  // Compute display items: dedupe for history, keep as-is for saved
+  // This automatically recalculates whenever rawItems or menuOpen changes
+  const items = useMemo(() => {
```
YishaiGlasner: `useMemo` is recommended only for expensive functions. are you sure this is the case?

static/js/UserHistoryPanel.jsx
```diff
@@ -145,62 +113,65 @@ UserHistoryPanel.propTypes = {
...
-const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal, dataSource}) => {
-  const [items, setItems] = useState(store.loaded ? store.items : null);
 
-  // Store changes when switching tabs, reset items
-  useEffect(() => {
-    setItems(store.loaded ? store.items : null);
-  }, [store]);
+const UserHistoryList = ({store, scrollableRef, menuOpen, toggleSignUpModal}) => {
+  // Store raw items in state (never deduped)
+  const [rawItems, setRawItems] = useState(store.loaded ? store.items.slice() : null);
```
YishaiGlasner, line 139: why to slice here, in the first time?

static/js/UserHistoryPanel.jsx
stevekaplan123: But it's not the key that it holds but the value of the key

static/js/UserHistoryPanel.jsx
stevekaplan123: The reason is because I want rawItems to point to a different reference of the data because I don't want to accidentally change the store's items.  'slice' keeps the store separate from the component's display data.

static/js/UserHistoryPanel.jsx
stevekaplan123: removed it

static/js/UserHistoryPanel.jsx
YishaiGlasner: my bad. what about `prevValue`? i think `prev` sounds like prev item

YishaiGlasner (approved):

PR #2747, opened 2025-10-21
Use seasonal-topic api rather than hebcal

YishaiGlasner:
## Code changes
urls.py - remove '_api/topics/seasonal-topic', and change 'api/calendars/topics/holiday' to have its view.
views.py - remove `next_holiday`, and use `request.interfaceLang` rather than sending it as query param in `seasonal_topic_api`.
sefaria.js - use `getUpcomingDay` in `getSeasonalTopic` and remove `_seasonalTopic`.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -3371,7 +3371,7 @@ def topic_ref_bulk_api(request):
 def seasonal_topic_api(request):
     from django_topics.models import SeasonalTopic
 
-    lang = request.interfaceLang[:2]
+    lang = request.LANGUAGE_CODE
     cb = request.GET.get("callback", None)
     diaspora = request.GET.get("diaspora", False)
 


reader/views.py
```diff
@@ -1671,24 +1671,8 @@ def find_holiday_in_hebcal_results(response):
...
-    response = requests.get(f"https://www.hebcal.com/hebcal?v=1&cfg=json&maj=on&start={current_date}&end={date_in_three_months}")
-    if response.status_code == 200:
-        topic = find_holiday_in_hebcal_results(response)
-        if topic:
-            return jsonResponse(topic)
-        else:
-            return jsonResponse({"error": "Couldn't find any topics corresponding to HebCal results"})
-    else:
-        return jsonResponse({"error": "Couldn't establish connection with HebCal API"})
+    return redirect(f'/_api/topics/seasonal-topic?lang={request.interfaceLang[:2]}')
```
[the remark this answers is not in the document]
YishaiGlasner: decorator removed

reader/views.py
```diff
@@ -1669,26 +1669,9 @@ def find_holiday_in_hebcal_results(response):
...
```
stevekaplan123: You said you did this because the URLs for the voices case were more clear.  But then why not just change the URL in the library case, so that we have one place for this instead of adding a redirect?

stevekaplan123: As in, the topics landing page can just use:     url(r'^api/calendars/topics/parasha/?$', reader_views.parasha_data_api),
    url(r'^api/calendars/topics/holiday/?$', reader_views.next_holiday),

static/js/sefaria/sefaria.js
```diff
@@ -2973,13 +2973,8 @@ _media: {},
...
     });
   },
-  _seasonalTopic: {},
   getSeasonalTopic: function() {
-    return this._cachedApiPromise({
-        url: `${Sefaria.apiHost}/_api/topics/seasonal-topic?lang=${Sefaria.interfaceLang.slice(0, 2)}`,
-        key: (new Date()).toLocaleDateString(),
-        store: this._seasonalTopic,
-    });
+    return this.getUpcomingDay('holiday');
```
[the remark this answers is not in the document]
YishaiGlasner: added it to the key in `getUpcomingDay`

reader/views.py
```diff
@@ -3391,7 +3371,7 @@ def topic_ref_bulk_api(request):
 def seasonal_topic_api(request):
     from django_topics.models import SeasonalTopic
 
-    lang = request.GET.get("lang")
+    lang = request.interfaceLang[:2]
```
yitzhakc: Can we please validate that this is a valid value?

yitzhakc (changes requested):

reader/views.py
YishaiGlasner: this comes from a middleware so we know what it is. anyway, i have found now we have more explicit attribue so i change to it.

PR #2748, opened 2025-10-22
Feature/sc 37346/in mobile web sheets view there s no website

stevekaplan123:
## Description
This PR refactors the Header component to consolidate header visibility and styling logic that was previously computed in ReaderApp. The change moves the responsibility for determining header visibility, language toggle display, and box shadow styling from the parent component into the Header component itself.

## Code Changes
Simplified ReaderApp by removing header-related logic and passing only the firstPanel object instead of multiple derived props
Moved header visibility, language toggle, and box shadow logic into the Header component with improved inline documentation

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -189,9 +189,11 @@ const Header = (props) => {
   const mobile = !props?.multiPanel;
   
   const shouldHide = () => {
-    // Header visibility logic - on mobile, return null when we are viewing library content.  When we return null,
-    // we either display no component at the top of the screen or display ReaderControls at the top of the screen, essentially as the header.
-    // However, if the mobile nav menu is open, even when vieiwng a book, we still want the header to display, and therefore we need to also check mobileNavMenuOpen.
+    // Determines whether or not this component should be displayed or not. 
+    // When the component is hidden, there are two cases: (1) the ReaderControls component is displayed instead of this component, essentially
+    // functioning as the header.  This case occurs when viewing a library text in mode "Text"
+    // and (2) there is simply no header at all.  This case occurs when viewing a library text in mode "TextAndConnections".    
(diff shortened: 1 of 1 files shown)


static/js/Header.jsx
```diff
@@ -185,6 +185,17 @@ const Header = (props) => {
       window.removeEventListener('keydown', handleFirstTab);
     }
   }, []);
+  
+  const mobile = !props?.multiPanel;
+  
+  const shouldHide = () => {
+    // Header visibility logic - on mobile, return null when we are viewing library content.  When we return null,
```
yitzhakc: Update the docstring.

The `shouldHide` function returns a boolean; the short-circuiting of the render function on line234 if `shouldHide()` returns `true` is where a null is returned.

static/js/Header.jsx
```diff
@@ -299,6 +309,10 @@ const Header = (props) => {
     </>
   );
 
+  // Language toggle logic - show on mobile for specific menu pages
+  const languageToggleMenus = ["navigation", "saved", "history", "notes"];
+  const hasLanguageToggle = !Sefaria.multiPanel && Sefaria.interfaceLang !== "hebrew" && languageToggleMenus.includes(props?.firstPanel?.menuOpen);
```
yitzhakc: Use `mobile` instead of `!Sefaria.multiPanel`
See if we have a constant for `hebrew`

static/js/Header.jsx
```diff
@@ -353,7 +366,7 @@ Header.propTypes = {
   showSearch: PropTypes.func.isRequired,
   openTopic: PropTypes.func.isRequired,
   openURL: PropTypes.func.isRequired,
-  hasBoxShadow: PropTypes.bool.isRequired,
+  firstPanel: PropTypes.object,
```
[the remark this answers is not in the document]
yitzhakc: If we do this elsewhere, we can do it here.
Otherwise ignore.

PR #2753, opened 2025-10-22
Add title to body of voices homepage.

YishaiGlasner:
## Code Changes
SheetsHomePage.jsx - add the title
strings.js - add the strings for title
s2.css - 1. add title's selector to its sibling selector in the container for margin. 2. change the marging-top margin in mobile to have spave between title and the upper gif.


schreibersefaria (approved):
looks straight forward

PR #2757, opened 2025-10-27
Analytics for Module Switcher

saengel:
## Description
This PR adds analytics to the Module Switcher using a data-attribute approach. Analytics are captured for the following interactions:
- Module switcher toggle (both opening and closing)
- Clicking on module items (Library, Voices, Developers)
- Clicking on non-module links (About Sefaria and More)

### In `static/js/Header.jsx`:
- Added `analyticsFeatureName` prop to the `<DropdownMenu />` component
- Added `analyticsEventName` and `analyticsEventText` props to the `"About Sefaria"` and `"More"` dropdown items

### In `static/js/common/DropdownMenu.jsx`:
- Added `analyticsFeatureName` prop to `DropdownMenu` component
- Implemented data-attribute-based analytics using `data-anl-*` attributes:
- - `data-anl-feature_name` on the wrapper to identify the feature
- - `data-anl-event` to specify the event type (`modswitch_toggle:click` for open/close, `modswitch_item_click:click` for items)
- - `data-anl-from` and `data-anl-to` on the button to track state transitions (open ↔ closed)
- - `data-anl-text` on menu items to capture which option was clicked
- Updated `DropdownMenuItem` to accept optional `analyticsEventName` and `analyticsEventText` props
- Updated `DropdownModuleItem` to include analytics data attributes with the module name
- Added optional `onOpen` and `onClose` callback props to DropdownMenu

## Technical Approach
- The implementation uses declarative data attributes (`data-anl-*`) 
- The state variable in `DropdownMenu` remains as `isOpen` (the planned rename to dropdownState was reverted).

## Notes
Big thank you to @stevekaplan123 for setting the foundation in https://github.com/Sefaria/Sefaria-Project/tree/chore/sc-37111/analytics-for-module-switcher - much appreciated :)

### DIFF
--- a/static/js/common/DropdownMenu.jsx
+++ b/static/js/common/DropdownMenu.jsx
@@ -159,14 +159,20 @@ DropdownModuleItem.propTypes = {
 
 const DropdownMenu = ({children, buttonComponent, positioningClass, analyticsFeatureName = null, onOpen = null, onClose = null}) => {
     /**
-     * `buttonComponent` is a React component for the opening/closing of a button.
-     * `positioningClass` is a string for the positioning of the dropdown menu.  It defines a CSS class.
-     *  Currently, we have two possible classes: 'headerDropdownMenu' and 'readerDropdownMenu'.
-     *  The former is a more general case.  Historically, the former was used in the header
-     *  and the latter in the reader.  See s2.css for definition of these classes.
-     * the menu will be closed in click anywhere except in an element where data attribute data-prevent-close="true" is set.
-     * this class is using useRef for open/close rather than useState, for changing state triggers re-rendering of the
-     * component and all its children, so when clicking on children their onClick won't be executed.
+     * DropdownMenu - A reusable dropdown menu component with keyboard navigation and analytics support
+     *
+     * @param {React.ReactNode} children - The content to display inside the dropdown menu
+     * @param {React.ReactElement} buttonComponent - React component that triggers the dropdown (will be wrapped with click handler)
+     * @param {string} positioningClass - CSS class for dropdown positioning. Options: 'headerDropdownMenu', 'readerDropdownMenu' (see s2.css)
+     * @param {string} [analyticsFeatureName] - Optional feature name for analytics tracking (sets data-anl-feature_name)
+     * @param {Function} [onOpen] - Optional callback fired when dropdown opens
+     * @param {Function} [onClose] - Optional callback fired when dropdown closes
+     *
+     * Behavior:
+     * - Closes on: click outside, Escape key, Tab out, or clicking any item without data-prevent-close="true"
+     * - Analytics: When analyticsFeatureName is provided, adds data-anl-* attributes for tracking open/close and item clicks
+     * - Accessibility: Traps focus within dropdown when open, returns focus to button when closed
+     * - Uses useState for isOpen to properly trigger re-renders for analytics data attributes
      */
 
     const [isOpen, setIsOpen] = useState(false);


static/js/common/DropdownMenu.jsx
```diff
@@ -163,34 +165,56 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
-    const [isOpen, setIsOpen] = useState(false);
+    const [dropdownState, setDropdownState] = useState(false);
     const menuRef = useRef(null);
     const wrapperRef = useRef(null);
     const buttonRef = useRef(null);
 
     const handleButtonClick = (e) => {
       e.stopPropagation();
-      setIsOpen(isOpen => !isOpen);
+      setDropdownState(dropdownState => {
```
stevekaplan123: I'm confused.  Why is "dropdownState" clearer than isOpen if you have to then set isOpen to dropdownState?

saengel: Sure - 

Because otherwise to get the analytics logic to work you'd have to initialize a new boolean that is curState copying the value of isOpen, and it seemed cleaner to me to reverse it and check if `isOpen` vs checking `curState`. Really just a matter of readability in the code. If it seems more confusing happy to switch it back or come up with a more descriptive alternative name. 

What do you think? Let me know if you want me to send a snippet of how this would work if the name remained `isOpen`?

static/js/common/DropdownMenu.jsx
```diff
@@ -163,34 +165,56 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
+        const isOpen = !dropdownState;
```
stevekaplan123: This line implies that dropdownState is true when the menu not open.  But below you say:
`            className={`dropdownLinks-menu ${ dropdownState ? "open" : "closed"}`} 
`
which implies that dropdownState is true when the menu is open

static/js/common/DropdownMenu.jsx
stevekaplan123: (1) Why do you have to create a temp var at all?  Why not just check !isOpen?
(2) In my mind, a variable like dropdownState should hold the value of a string not a boolean.  isOpen clearly holds the value of a boolean.

saengel: 1. Because otherwise the callbacks would be based on the "old" state -- 

For example, if we wrote it like this:

```
const handleButtonClick = (e) => {
      e.stopPropagation();
      setIsOpen(isOpen => {
        if (!isOpen) {
           onClose && onClose();
        } else {
          onOpen && onOpen();
        }
        return !isOpen;
      });
    };
```

This is the step-by-step of what happens. Let's say the menu is currently closed, and someone is clicking to open it: 


```
const handleButtonClick = (e) => {
      e.stopPropagation();
      setIsOpen(isOpen => { # isOpen is passed as `False` (since it's currently closed)
        if (!isOpen) { # This evaluates to `True`
           onClose && onClose(); # The logic here executes as-if we're CLOSING the menu, but we're really opening it
        } else {
          onOpen && onOpen();
        }
        return !isOpen;
      });
    };
```

We need to save the 'flipped boolean' (as copilot calls it) explicitly in a variable for us to use for the conditionals to execute properly. Once I was doing that, I thought the renaming might help for readability - but...

2. I hear your point and agree that `dropdownState` doesn't sound boolean-ish enough. Maybe that is just an issue of renaming... vs logic. What do you think?

static/js/common/DropdownMenu.jsx
saengel: This is funny, let me check again why I did that.

static/js/Header.jsx
```diff
@@ -129,14 +129,39 @@ const LoggedInDropdown = ({ module }) => {
...
+  const trackClose = () => {
+    gtag("event", "modswitch_close", {
+      feature_name: "module_switcher",
+      action: "close"
+    });
+  };
+
+  const trackItemClick = (data) => {
+    gtag("event", "modswitch_item_click", {
+      feature_name: "module_switcher",
```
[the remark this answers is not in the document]
YishaiGlasner: i also wonder (and also what the meaning of `text` here

static/js/Header.jsx
```diff
@@ -129,14 +129,39 @@ const LoggedInDropdown = ({ module }) => {
...
+  };
+
   return (
-    <DropdownMenu positioningClass="headerDropdownMenu" buttonComponent={
-      <button className="header-dropdown-button" aria-label={Sefaria._("Library")}>
-        <img src='/static/icons/module_switcher_icon.svg' alt={Sefaria._("Library")} />
-      </button>
-    }>
+    <DropdownMenu positioningClass="headerDropdownMenu"
+                  buttonComponent={ <button className="header-dropdown-button" aria-label={Sefaria._("Library")}>
```
YishaiGlasner: can you move it outside (i know it's there before). creating a complexed variable within jsx means it's created in any re-render.
we're still doing in some short functions (and i think it's indeed not the best), but here that means the `buttonComponent` will be unmount and re-mount any re-render which can cause bugs and not only performance issues.

static/js/common/DropdownMenu.jsx
yitzhakc: I feel that `dropdownState` does not sound boolean-ish enough, but don't see a problem with creating a temporary variable to store the flipped boolean :-)

static/js/common/DropdownMenu.jsx
```diff
@@ -204,15 +228,15 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
             Util.focusFirstElement(menuRef.current);
         }
-    }, [isOpen]);
+    }, [dropdownState]);
 
     const handleMenuKeyDown = (e) => {
         Util.trapFocusWithTab(e, {
             container: menuRef.current,
-            onClose: () => setIsOpen(false),
+            onClose: () => setDropdownState(false),
```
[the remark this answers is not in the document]
yitzhakc: At first glance, seems like an apt comment?

static/js/Header.jsx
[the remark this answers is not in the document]
yitzhakc: Isn't it just the text on the clicked item in the module switcher , which corresponds to the requested module?

static/js/Header.jsx
yitzhakc: Since I am not so knowledgable on the ins-and-outs of React component rendering gotchas, I asked why in practice, wouldn't the `ModuleSwitcher` component be rendered just once.
The reply:
> Even if ModuleSwitcher seems rendered once, it’s re-rendered every time its state or parent changes. The parent can change on state or props update or modification.
Defining elements like buttonComponent inline means they’re recreated, and potentially remounted, on every re-render.
This can cause performance issues and subtle bugs (like loss of focus/state).

So I'd say let's do this :-)

static/js/Header.jsx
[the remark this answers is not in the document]
saengel: Yes, and marking this as resolved since the names are per spec.

static/js/Header.jsx
saengel: Great, thank you both! Just pushed.

static/js/common/DropdownMenu.jsx
saengel: Just pushed a fix using `isOpen` - appreciate everyone's feedback! Thank you!

static/js/common/DropdownMenu.jsx
saengel: This should be fixed now - good catch!

static/js/common/DropdownMenu.jsx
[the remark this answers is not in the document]
saengel: Yes, added - just a drop confused about how that function is even ever called if we're preferring the `data-anl` approach which should handle all of that for us?

static/js/common/DropdownMenu.jsx
```diff
@@ -163,34 +165,56 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
+      }
+
       const preventClose = e.target.closest('[data-prevent-close="true"]');
       // Only toggle if no preventClose element was found
       if (!preventClose) {
-        setIsOpen(false);
+        setDropdownState(false);
+        onClose && onClose();
       }
     }
```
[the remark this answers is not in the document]
saengel: Done!

static/js/common/DropdownMenu.jsx
```diff
@@ -26,6 +26,8 @@ const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose
        href={fullURL}
        target={newTab ? '_blank' : null}
        data-prevent-close={preventClose}
+       data-anl-event={analyticsText ? "modswitch_item_click:click" : null}
```
yitzhakc: What's the benefit of `modswitch_item_click:click` over `modswitch_item:click`?

static/js/common/DropdownMenu.jsx
```diff
@@ -151,7 +155,7 @@ DropdownModuleItem.propTypes = {
   }).isRequired
 };
 
-const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
+const DropdownMenu = ({children, buttonComponent, positioningClass, analyticsFeatureName, onOpen, onClose, onItemClick}) => {
```
yitzhakc: Is there a reason you didn't make these new additional parameters null by default?

static/js/common/DropdownMenu.jsx
```diff
@@ -170,19 +174,30 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
 
     const handleButtonClick = (e) => {
       e.stopPropagation();
-      setIsOpen(isOpen => !isOpen);
+      setIsOpen(isOpen => {
+        const curState = !isOpen;
+        if (curState) {
+          onOpen && onOpen();
```
YishaiGlasner:
```suggestion
          onOpen?.();
```
and also in other places here

static/js/common/DropdownMenu.jsx
```diff
@@ -12,7 +12,7 @@ const DropdownMenuSeparator = () => {
 
 }
 
-const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose = false, targetModule = null}) => {
+const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose = false, targetModule = null, analyticsText = null}) => {
```
YishaiGlasner: maybe name it `analyticsEventText`? that will make the use of it as bool for `data-anl-event` more clear.

static/js/common/DropdownMenu.jsx
```diff
@@ -170,19 +174,30 @@ const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
...
       e.stopPropagation();
-      setIsOpen(isOpen => !isOpen);
+      setIsOpen(isOpen => {
+        const curState = !isOpen;
+        if (curState) {
+          onOpen && onOpen();
+        } else {
+          onClose && onClose();
+        }
+        return curState;
```
YishaiGlasner, line 192: why you added a return?

static/js/common/DropdownMenu.jsx
YishaiGlasner: i don't see where those 3 functions are passed to DropdownMenu

static/js/common/DropdownMenu.jsx
saengel: Good idea!

static/js/common/DropdownMenu.jsx
saengel: `modswitch_item_click` was the spec for the text of the event

static/js/common/DropdownMenu.jsx
saengel: It needs to set the state to the new value.

saengel: >The return is necessary because this is using React's functional state update pattern. When you call setIsOpen with a function like this:
setIsOpen(isOpen => {
  // ... do stuff
  return curState;
});
React expects that function to return the new state value. The function receives the current state as a parameter (isOpen) and must return what the new state should be.

static/js/common/DropdownMenu.jsx
saengel: Cool! Didn't realize that optional chaining worked for function calls too!

static/js/common/DropdownMenu.jsx
saengel: This should all be addressed now.

static/js/common/DropdownMenu.jsx
```diff
@@ -153,7 +157,7 @@ DropdownModuleItem.propTypes = {
   }).isRequired
 };
 
-const DropdownMenu = ({children, buttonComponent, positioningClass}) => {
+const DropdownMenu = ({children, buttonComponent, positioningClass, analyticsFeatureName = null, onOpen = null, onClose = null}) => {
```
YishaiGlasner, line 160: add new params to docstring

saengel: Good catch thanks!

PR #2758, opened 2025-10-27
Bug/sc 37696/navigating between modules and languages

schreibersefaria:
Fixing the cauldron cookie setting (now that it's up we found a bug)
*Will un-draft after this*

# Description
The goal of this card is to fix the buggy behavior of the staging environment with regards to module switching and language retention for a logged-out user. 

The desired behavior enforced by this card is as follows:
| Start Module & Language  | Navigates to | End Module & Language |
| ------------- | ------------- | ------------- |
| Library EN | Hebrew | Library HE |
| Library EN | Voices | Voices EN |
| Library HE | English | Library EN |
| Library HE | Voices | Voices HE (Chiburim) |
| Voices EN | Hebrew | Voices HE (Chiburim) |
| Voices EN | Library | Library EN |
| Voices HE (Chiburim) | English | Voices EN |
| Voices HE (Chiburim) | Library | Library EN |

The main issues were the use of `DOMAIN_LANGUAGES` which had a flat structure only accounting for language mapping for the domain name, as opposed to module and language mapping. To resolve this, we moved over to use `DOMAIN_MODULES` which has a nested structure accounting for both language and module choice. This data already existed, we just made sure to comb through all remaining instances of `DOMAIN_LANGUAGES` and remove them, and adapt functions to use `DOMAIN_MODULES`. 

We also enabled a cross-module cookie set at the domain level, to ensure that language preferences were preserved across modules. 

Some smaller bugs and patches were included along the way and will be elaborated on below. 

## Claude's TLDR

This branch represents a critical bug fix on top of the modularization work. While modularization-main introduced the infrastructure for module-based routing, it had several bugs in the language/module switching logic that made navigation unreliable. 

# Code Changes

1. `sefaria/utils/util.py` - The backend fixes to ensure use of `DOMAIN_MODULES`

Five new utility functions that form the backbone of the fix:
a. `current_domain_lang(request)` -Detects which language the current domain is pinned to (i.e. staging il is pinned to HE)
b. `get_redirect_domain_for_language` - Gets target domain URL while preserving current module
c. `needs_domain_switch` - Prevents redirect loops by checking if redirect is actually needed 
d. `def add_query_param` - Safely adds/updates query params for set-language-cookie, which we pass on every module switch to ensure cookie preferences are maintained. 
e. `get_cookie_domain` - Returns shared cookie domain for cross-module cookie sharing, i.e. for voices.sefaria.org and www.sefaria.org will return '.sefaria.org' so cookies work on both subdomains.

2. `sefaria/system/middleware.py`

- `LanguageSettingsMiddleware` - Updated redirect logic to preserve module when switching languages, and always adds set-language-cookie param for cross-domain cookie setting

- `LanguageCookieMiddleware `-  Sets cookie on shared domain (found by looking for the domain suffix)

- `ModuleMiddleware` - removed '/interface/' from excluded_url_prefixes which was the 'old' way to toggle the language switcher, because of this some pages from getting proper module detection

3. `reader/views.py` - Rewrote `interface_language_redirect()` to take into account the module and preserve it, while also always adding the `set-language-cookie` param to the url on the current domain. Also added module filtering to  `saved_content()` & `user_history_content()`.

4. `static/js/sefaria/sefaria.js` - Cache all hostnames from domainModules for fast lookups and fixed module URL lookup, URL validation logic, and user history loading to account for the nested structure of `DOMAIN_MODULES` (bug in redundant seasonal topic API also seems to have been dealt with, as well as the sheet title helper). Also adapted `getCookieDomain` to use the cached `getDomainHostnames` which accounts for the nested structure of `DOMAIN_MODULES`. 

5. `static/js/Header.jsx` - Minor cleanup in module switching logic - removed redundant code.

6. `static/js/ReaderApp.jsx` - Updated to pass sheets_only filter to saved/history components.

7. `sefaria/local_settings_ci.py` & `sefaria/local_settings_example.py`:

  ```#Updated DOMAIN_MODULES to match new nested structure
  DOMAIN_MODULES = {
      'en': {
          'library': 'http://localhost:8000',
          'voices': 'http://sheets.localhost:8000'
      },
      'he': {
          'library': 'http://localhost:8000',
          'voices': 'http://sheets.localhost:8000'
      }
  }
```

8. `sefaria/system/context_processors.py` - adapt to pass full `DOMAIN_MODULES` structure

9. `sefaria/system/tests/test_language_module_switching.py`: Unit tests for the cases outlined in the table above. 

# Notes
- This does not account for module switching with language retention on `localhost` that is a separate task
- Additionally, this does not account for user saved languages, which is also a separate task.

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -518,14 +518,8 @@ Sefaria = extend(Sefaria, {
 
     return result;
   },
-  _domainHostnamesCache: null,
   getDomainHostnames: function() {
     // Returns a Set of all hostnames from domainModules.
-    // Cached for performance
-    if (this._domainHostnamesCache) {
-      return this._domainHostnamesCache;
-    }
-
     const hostnames = new Set();
     for (const langModules of Object.values(this.domainModules)) {
       for (const moduleUrl of Object.values(langModules)) {
@@ -533,9 +527,7 @@ Sefaria = extend(Sefaria, {
         hostnames.add(url.hostname);
       }
     }
-    
 
-    this._domainHostnamesCache = hostnames;
     return hostnames;
   },
   getModuleURL: function(module=null) {
@@ -3723,7 +3715,6 @@ Sefaria.setup = function(data, props = null, resetCache = false) {
     Sefaria.loadServerData(data);
     let baseProps = props !=null ? props : (typeof DJANGO_VARS === "undefined" ? undefined : DJANGO_VARS.props);
     Sefaria.unpackBaseProps(baseProps);
-    Sefaria.getDomainHostnames(); // Preemptively populate domain hostnames cache
     Sefaria.util.setupPrototypes();
     Sefaria.util.setupMisc();
     var cookie = Sefaria.util.handleUserCookie(Sefaria._uid);

### CODE REVIEW

File: helm-chart/sefaria/templates/configmap/local-settings-file.yaml [hunk 1]
```diff
@@ -43,7 +43,6 @@ data:
     GLOBAL_WARNING = os.getenv("GLOBAL_WARNING").lower() == "true"
     GLOBAL_WARNING_MESSAGE = os.getenv("GLOBAL_WARNING_MESSAGE")
 
-    DOMAIN_LANGUAGES = json.loads(os.getenv("DOMAIN_LANGUAGE"))
```
yitzhakc: Just confirming - you've searched for all instances of `DOMAIN_LANGUAGES`? (or the agent did ;-))

static/js/sefaria/util.js
```diff
@@ -833,12 +842,14 @@ class Util {
         
         // Extract hostnames from all domain modules
         const hostnames = [];
-        for (const [_, moduleUrl] of Object.entries(Sefaria.domainModules)) {
-            try {
-                const url = new URL(moduleUrl);
-                hostnames.push(url.hostname);
-            } catch (e) {
-                // Invalid URL - skip this module
+        for (const [_, langModules] of Object.entries(Sefaria.domainModules)) {
```
yitzhakc: If we're iterating over all the hostnames of the domain modules, maybe we could store it somewhere and use it at `static/js/sefaria/sefaria.js` in `isSefariaURL` instead of iterating each time over a nested object?

helm-chart/sefaria/templates/configmap/local-settings-file.yaml
schreibersefaria: yes. Both the agent and I did

static/js/sefaria/util.js
schreibersefaria: How is that?

sefaria/local_settings_ci.py
```diff
@@ -16,12 +16,6 @@
...
 
 
-# Map domain to an interface language that the domain should be pinned to.
-# Leave as {} to prevent language pinning, in which case one domain can serve either Hebrew or English
-DOMAIN_LANGUAGES = {
-    "http://hebrew.example.org": "hebrew",
-    "http://english.example.org": "english",
-}
 DOMAIN_MODULES = {
     "library": "http://localhost:8000",
```
YishaiGlasner: you should change this

reader/views.py
```diff
@@ -220,13 +221,19 @@ def base_props(request):
             "saved": {"loaded": False, "items": []},
             "last_place": []
         }
+    # TEMP: Debug logging for DOMAIN_MODULES
```
YishaiGlasner: don't forget to remove

reader/views.py
```diff
@@ -1322,19 +1329,20 @@ def interface_language_redirect(request, language):
...
-        if DOMAIN_LANGUAGES[domain] == language and not request.get_host() in domain:
-            next = domain + next
-            next = next + ("&" if "?" in next else "?") + "set-language-cookie"
-            break
+    # Look up the target domain based on current module + target language
+    target_domain = get_redirect_domain_for_language(request, language)
+
+    if needs_domain_switch(request, target_domain):
+        # Switching domains - preserve path and add set-language-cookie param
+        next = target_domain + next
```
YishaiGlasner: use `urljoin`

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
         raise InputError("Invalid language. Must be 'english' or 'hebrew'.")
     return "en" if language == "english" else "he"
 
+
+def current_domain_lang(request):
```
YishaiGlasner: i think that this util file should be dedicated to data utils (language code included in that). we should put views utils in sone other place.
also i think we already have tools in views files.

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+    Returns the pinned language for the current domain, or None if current domain is not pinned.
+    Uses DOMAIN_MODULES to detect which language the current domain belongs to.
+
+    If the hostname matches multiple languages (e.g., in local development where both languages
+    use localhost), returns None to indicate the domain is not language-pinned.
+
+    :param request: Django request object
+    :return: 'english', 'hebrew', or None
+    """
+    if not (hasattr(settings, 'DOMAIN_MODULES') and settings.DOMAIN_MODULES):
```
YishaiGlasner:
```suggestion
    if not getattr(settings, 'DOMAIN_MODULES', None):
```

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
         raise InputError("Invalid language. Must be 'english' or 'hebrew'.")
     return "en" if language == "english" else "he"
 
+
+def current_domain_lang(request):
+    """
+    Returns the pinned language for the current domain, or None if current domain is not pinned.
+    Uses DOMAIN_MODULES to detect which language the current domain belongs to.
+
+    If the hostname matches multiple languages (e.g., in local development where both languages
```
YishaiGlasner: I think returning None isn't clean. Think about an engineer that comes to a view and see a call to this function. he won't imagina it can return None.
I suggest rather to have 2 function: one that returns bool if the env support different domains for different language, and one returns the lang. maybe we already have function that tell us if we're in prod, cauldron or local.

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+    use localhost), returns None to indicate the domain is not language-pinned.
+
+    :param request: Django request object
+    :return: 'english', 'hebrew', or None
+    """
+    if not (hasattr(settings, 'DOMAIN_MODULES') and settings.DOMAIN_MODULES):
+        return None
+
+    current_hostname = urlparse(f"http://{request.get_host()}").hostname
```
YishaiGlasner: won't work locally. also `module_url` will probably have https and not http

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+    matched_langs = []
+
+    for lang_code, modules in settings.DOMAIN_MODULES.items():
+        if not isinstance(modules, dict):
```
YishaiGlasner: is this necessary? can it be something else?

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+            if urlparse(module_url).hostname == current_hostname:
+                matched_langs.append(lang_code)
+                break  # Only need to match once per language
+
+    # If we matched multiple languages, domain is ambiguous - not pinned. Happens on Local
+    if len(matched_langs) != 1:
+        return None
+
+    # Only return language if domain uniquely identifies it
+    return 'english' if matched_langs[0] == 'en' else 'hebrew'
```
YishaiGlasner:
```suggestion
    return short_to_long_lang_code(matched_langs[0])
```

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+
+
+def get_redirect_domain_for_language(request, interface_lang):
```
YishaiGlasner: rename `interface_lang` to something like `target_lang`

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+    Determine if switching to target_domain requires a domain change.
+
+    Compares the current request host with the target domain's hostname.
+    Returns False if domains are the same (prevents redirect loops in local dev).
+
+    :param request: Django request object
+    :param target_domain: Full domain URL (e.g., 'https://www.sefaria.org') or None
+    :return: Boolean indicating if domain switch is needed
+    """
+    if not target_domain:
```
YishaiGlasner: why is this needed?

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+        return False
+
+    current_host = request.get_host()
+    target_host = urlparse(target_domain).hostname
+    return target_host and current_host != target_host
```
YishaiGlasner: why is `target_host and` needed?

sefaria/utils/util.py
```diff
@@ -539,6 +535,85 @@ def get_short_lang(language):
...
+
+
+def add_set_language_cookie_param(url):
```
YishaiGlasner: i think it should be general and using urllib tools:
```
def add_query_param(url, param, value=''):
    parsed = urlparse(url)
    query = dict(parse_qsl(parsed.query))
    query["set-language-cookie"] = ""
    new_query = urlencode(query, doseq=True)
    return urlunparse(parsed._replace(query=new_query))
```
please check we don't already have something similar.

addition - now i see you also use this function with param string rather than url string. this is misleading and against the docstring and param name.

static/js/common/DropdownMenu.jsx
```diff
@@ -125,7 +127,14 @@ const DropdownMenuItemWithIcon = ({icon, textEn='', descEn='', descHe=''}) => {
  * />
  */
 const DropdownModuleItem = ({url, newTab, targetModule, dotColor, text}) => {
+  // TEMP: Debug logging
+  console.log('[TEMP] DropdownModuleItem rendering with:', { url, newTab, targetModule, dotColor, text });
+
+  // fullURL automatically appends set-language-cookie when switching modules to preserve language preference
   const fullURL = targetModule ? Sefaria.util.fullURL(url, targetModule) : url;
```
stevekaplan123, line 129: I'm not sure 'fullURL' is where this change should have been done.  Not all dropdown menu items change language.  What about DropdownLanguageToggle or NextRedirectAnchor?

sefaria/system/middleware.py
```diff
@@ -162,20 +161,33 @@ def process_request(self, request):
...
+            target_domain = get_redirect_domain_for_language(request, lang)
+            logger.info("[TEMP] Processing set-language-cookie", target_domain=target_domain)  # TEMP
+
             path = quote(request.path, safe='/')
             params = request.GET.copy()
             params.pop("set-language-cookie")
             params_string = params.urlencode()
             params_string = "?" + params_string if params_string else ""
-            response = redirect(domain + path + params_string)
+            final_url = target_domain + path + params_string
```
YishaiGlasner: i think this should also be TEMP. `response = redirect(domain + path + params_string)` is very clear.

sefaria/system/context_processors.py
```diff
@@ -88,7 +88,7 @@ def base_props(request):
 def module_context(request):
     return {
         'active_module': request.active_module,
-        'domain_modules': get_language_specific_domain_modules(request.interfaceLang)
+        'domain_modules': DOMAIN_MODULES
```
YishaiGlasner, line 84: why do we need to add `DOMAIN_MODULES` to a request? it's available in settings for whoever needs it

sefaria/utils/util.py
stevekaplan123: We probably have a function to convert 'en' to 'english' and 'he' to 'hebrew'

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+  _domainHostnamesCache: null,
+  getDomainHostnames: function() {
+    // Returns a Set of all hostnames from domainModules.
+    // Cached for performance
+    if (this._domainHostnamesCache) {
+      return this._domainHostnamesCache;
+    }
+
+    const hostnames = new Set();
+    if (this.domainModules && typeof this.domainModules === 'object') {
```
YishaiGlasner: is there a case it isn't an object?
anyway 
```suggestion
    if (this?.domainModules === 'object') {
```

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
 
     return result;
   },
+  _domainHostnamesCache: null,
+  getDomainHostnames: function() {
```
YishaiGlasner, line 521: i think this should be run one time when we unpack the props from SSR.

sefaria/system/middleware.py
```diff
@@ -101,20 +101,19 @@ def process_request(self, request):
...
-                if redirect_domain:
-                    # When detected language doesn't match current domain langauge, redirect
+                redirect_domain = get_redirect_domain_for_language(request, interface)
+
+                if needs_domain_switch(request, redirect_domain):
+                    # When detected language doesn't match current domain language, redirect
+                    # while preserving the current module
                     path = request.get_full_path()
-                    path = path + ("&" if "?" in path else "?") + "set-language-cookie"
+                    path = add_set_language_cookie_param(path)
```
stevekaplan123: It seems you almost have the same code above in reader/views.py.  Perhaps you can extract lines 110-116 and the reader/views.py code into a function.  The reader/views.py code is:
```

    # Look up the target domain based on current module + target language
    target_domain = get_redirect_domain_for_language(request, language)

    if needs_domain_switch(request, target_domain):
        # Switching domains - preserve path and add set-language-cookie param
        next = target_domain + next
        next = add_set_language_cookie_param(next)
```

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+    }
+
+    const hostnames = new Set();
+    if (this.domainModules && typeof this.domainModules === 'object') {
+      for (const langModules of Object.values(this.domainModules)) {
+        for (const moduleUrl of Object.values(langModules)) {
+          try {
+            const url = new URL(moduleUrl);
+            hostnames.add(url.hostname);
+          } catch (e) {
```
YishaiGlasner: is this necessary?

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+  getDomainHostnames: function() {
+    // Returns a Set of all hostnames from domainModules.
+    // Cached for performance
+    if (this._domainHostnamesCache) {
+      return this._domainHostnamesCache;
+    }
+
+    const hostnames = new Set();
+    if (this.domainModules && typeof this.domainModules === 'object') {
+      for (const langModules of Object.values(this.domainModules)) {
```
```suggestion
      for (const langModules of Object.values(this.domainModules)) {const allModuleUrls = Object.values(this.domainModules)
  .flatMap(langModules => Object.values(langModules));
```
(and filter invalids if needed

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+
+    // TEMP: Debug logging
+    console.log('[TEMP] getModuleURL called with module:', module);
+    console.log('[TEMP] Sefaria.activeModule:', Sefaria.activeModule);
+    console.log('[TEMP] Sefaria.interfaceLang:', Sefaria.interfaceLang);
+    console.log('[TEMP] Sefaria.domainModules:', Sefaria.domainModules);
+
     module = module || Sefaria.activeModule;
-    const href = Sefaria.domainModules?.[module] || Sefaria.apiHost;
+    const langCode = Sefaria.interfaceLang === "hebrew" ? "he" : "en";
```
```suggestion
    const langCode = Sefaria._getShortInterfaceLang();
```

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+    console.log('[TEMP] Looking up domainModules[' + langCode + '][' + module + ']'); // TEMP
+
+    const href = Sefaria.domainModules?.[langCode]?.[module] || Sefaria.apiHost;
+    console.log('[TEMP] href resolved to:', href); // TEMP
+    console.log('[TEMP] Sefaria.apiHost fallback:', Sefaria.apiHost); // TEMP
+
     try {
-      return new URL(href);
-    } catch {
+      const url = new URL(href);
```
YishaiGlasner: this is also TEMP.

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,64 @@ Sefaria = extend(Sefaria, {
...
+    } catch (e) {
+      console.error('[TEMP] Error creating URL:', e); // TEMP
       return false;
     }
   },
   isSefariaURL: function(url) {
-    return Object.values(Sefaria.domainModules).some(href => url.href.startsWith(href));
+    // Check if URL's hostname matches any of our domain hostnames
+    const hostnames = this.getDomainHostnames();
+    return hostnames.has(url.hostname);
```
YishaiGlasner, line 553: previously we checked `startsWith`. that seems to break here

static/js/sefaria/util.js
```diff
@@ -293,13 +293,30 @@ class Util {
       return (typeof window === "undefined" ) ? this._initialPath :
                 window.location.pathname + window.location.search;
     }
-    static fullURL(relativePath, moduleTarget) {
+    static fullURL(relativePath, moduleTarget, preserveLanguageOnModuleSwitch = true) {
+      // TEMP: Debug logging for URL construction
+      console.log('[TEMP] fullURL called with:', { relativePath, moduleTarget, preserveLanguageOnModuleSwitch });
+
+      let url = relativePath;
```
YishaiGlasner: why?

static/js/sefaria/util.js
```diff
@@ -826,27 +843,13 @@ class Util {
...
-        }
-        
+        const hostnames = Array.from(Sefaria.getDomainHostnames());
+
         // Skip domain setting for empty hostnames and local development.
         // IP addresses don't have subdomain support.
         // Browsers don't allow setting cookies with domain ".localhost"
         // For localhost development, we need to dismiss the cookie banner on each module if we are using sheets.localhost
-        if (!hostnames.length || hostnames.some(hostname => Util.isIPAddress(hostname) || hostname.includes('localhost'))) {
+        if (hostnames.length === 0 || hostnames.some(hostname => Util.isIPAddress(hostname) || hostname.includes('localhost'))) {
```
YishaiGlasner: why?

static/js/sefaria/util.js
YishaiGlasner: why

static/js/sefaria/util.js
YishaiGlasner: why?

static/js/sefaria/util.js
```diff
@@ -293,13 +293,30 @@ class Util {
...
+      console.log('[TEMP] fullURL called with:', { relativePath, moduleTarget, preserveLanguageOnModuleSwitch });
+
+      let url = relativePath;
+
       if (relativePath.startsWith("/")) { // if the path is relative, prepend the module URL
         const moduleURL = Sefaria.getModuleURL(moduleTarget); // derive the host URL from the module target (e.g. 'https://voices.sefaria.org' or 'https://www.sefaria.org')
-        return moduleURL.origin + relativePath;
+        console.log('[TEMP] getModuleURL returned:', moduleURL); // TEMP
+        console.log('[TEMP] moduleURL.origin:', moduleURL?.origin); // TEMP
+        url = moduleURL.origin + relativePath;
```
YishaiGlasner: i suggest to use new URL in the first place and manipulate it rather than the string

static/js/common/DropdownMenu.jsx
```diff
@@ -37,8 +37,10 @@ const DropdownMenuItem = ({url, children, newTab, customCSS = null, preventClose
 const NextRedirectAnchor = ({url, children, className}) => {
   const onClick = (e) => {
     e.preventDefault();
-    window.location.href = `${url}?next=${encodeURIComponent(Sefaria.util.currentPath())}`;
-  }
+    const currentPath = Sefaria.util.currentPath();
+    const targetUrl = `${url}?next=${encodeURIComponent(currentPath)}`;
```
YishaiGlasner: this line seems redundant to me

saengel:
> Since we're delving into the routing, and dealing with logic that modifies requests, I'd really like to add some unit tests here that will enable us to validate that our known examples (e.g. the staging server URLs, the production URLs) "work".
> 
> You can use [override_settings](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.override_settings) in unittests to modify objects, but note that you'll need to replace the direct import of `from sefaria.settings import ...` (which is against Django's recommendations!) to `from django.conf import settings` and then usage would be `settings.DOMAIN_MODULES`.

@yitzhakc I just pushed some unit tests for staging and prod. 

I was struggling to figure out how to mimic being on one module and switching to the other within the same language (i.e. going from English Library and selecting Voices, making sure I stay on English voices). Let me know if you have any suggestions. For now, the tests within the same language but different module basically test "assuming I clicked on `MODULE` with the language cookie set..." but that doesn't feel 100% ideal to me. Curious if there's a better way.

reader/views.py
schreibersefaria: 👍

static/js/common/DropdownMenu.jsx
saengel: @stevekaplan123 my understanding is that even though not all of the options change the language, we need to preserve the language across any module changes. So even if someone is in `English Library` lets say and they switch to `English Voices` we need to pass the cookie there as a backup to help make sure the language stays constant across module switches. Does that make sense?

saengel:
> I didn't understand why `set-language-cookie` is needed when switching modules

@YishaiGlasner - thank you for all of the thorough and super-helpful comments! Will go through one-by-one, but for now, here's the answer I just wrote to @stevekaplan123 - https://github.com/Sefaria/Sefaria-Project/pull/2758#discussion_r2478018255

sefaria/system/context_processors.py
schreibersefaria: I'm not actually sure. We didn't add it just change it to use the new object.
AI says: Django server-rendered templates (.html files) - Need the context processor to access {{ domain_modules.library }}
If you think this isn't needed I can dive into it

sefaria/utils/util.py
saengel: Good point, addressed.

sefaria/utils/util.py
saengel: @schreibersefaria are there instances where we don't pass a target_domain? @YishaiGlasner raises a good point, this seems redundant to me... given the conditional return.

sefaria/utils/util.py
saengel: In a case where it is `None`, but @schreibersefaria is that even possible? If not, then this check can be removed as well as the one [here](https://github.com/Sefaria/Sefaria-Project/pull/2758#discussion_r2476833147). Maybe better practice to always check though?

static/js/common/DropdownMenu.jsx
YishaiGlasner: @saengel not to me. switching module is always within the same language interface, so nothing should be changed. if there's a problem we should find it and solve its root

sefaria/system/context_processors.py
YishaiGlasner: yes, i think it's needed. your change makes this redundant.
in the first place the idea here (i guess) was that we need the domain modules of the specific language in many places, so let's add it to the request, but now it's very weird to add a local var to request.
(making things redundant is good if that was not clear...)

sefaria/utils/util.py
YishaiGlasner: i think over-checking is not better practice, when everything is in the same 'place'. I mean everything here is our server side code, we define its settings so we should trust them to be there in the right structure.

sefaria/system/context_processors.py
schreibersefaria: Cool thanks! Removed it
I did some testing where claude told me things could break and everything is working.
Any idea if anything else has to be changed if I remove this?

sefaria/system/middleware.py
schreibersefaria: Because they are a bit different in path vs next I think this may not be worth the effort.
You can see I extracted most the logic into helper functions already.
What do you think?

static/js/common/DropdownMenu.jsx
schreibersefaria: Well we don't really know that switching module is always with the same language. We have to enforce this somehow.
The idea behind doing this in fullURL is that any time the users moves from module to another module he will stay in the same language.
Specifically DropdownModuleItem, is only used for Module dropdown, BUT hopefully fullURL is used whenever there is a link in between modules.
Is there a better place to put this?

sefaria/system/context_processors.py
YishaiGlasner: wait a minute, i think i was wrong. i thought we're in middlewares.py.
but it's context_processors.py, so it's reasonable to have here a var from local settings. i guess we use it in our templates so we do need it here.

static/js/common/DropdownMenu.jsx
YishaiGlasner: i don't understand why you need to enforce it - how can the language can be changed spontaneously when switching module?

static/js/sefaria/util.js
schreibersefaria: how is that?

static/js/common/DropdownMenu.jsx
schreibersefaria: Say I have both modules in Eng. Then while I'm in Voices and change to Heb. The Library doesn't change.
Then I click on a link and go back to the Library, as A user, I'd expect it to be in Heb. But if we don't add set-language-cookie the Library would be in Eng

static/js/sefaria/util.js
YishaiGlasner: something like (GPT, I didn't check):
```
let url;

// Build the base URL safely
if (relativePath.startsWith("/")) {
  url = new URL(relativePath, Sefaria.getModuleURL(moduleTarget));
} else {
  // relativePath may already be an absolute or relative URL
  url = new URL(relativePath, window.location.origin);
}

// Append `set-language-cookie` param if switching modules
if (preserveLanguageOnModuleSwitch && moduleTarget && moduleTarget !== Sefaria.activeModule) {
  url.searchParams.append("set-language-cookie", "");
}

return url.toString();
```

static/js/common/DropdownMenu.jsx
YishaiGlasner: now i got you, but why would that happen? do we store different languages for library and voices?

schreibersefaria: Mmmm well currently the cookie is set separately for the library and voices. I'll check if they can be set together.
@saengel We discussed this at some point, do you remember why we didn't do it?

static/js/sefaria/util.js
schreibersefaria: How is that?

static/js/common/DropdownMenu.jsx
schreibersefaria: Right, it was to support local development where you can't set a cookie domain for the lang cookie.
So you need to send this param to make sure the cookie is set for both modules.
I'm going to try add the proper setting of a cookie in the shared domain (so one cookie for both www.sefaria.org and voices.sefaria.org).
Do you think we should stop using `set-language-cookie` and not have local development work like on a deployed env?

sefaria/utils/util.py
schreibersefaria: thanks

static/js/common/DropdownMenu.jsx
saengel: Just a note - @YishaiGlasner - we are going to have to implement localhost separately. So @schreibersefaria maybe it is worthwhile to take this out if it isn't relevant for the scope of this, which intentionally excludes localhost. What do you think? Is it necessary for the cauldron?

sefaria/utils/util.py
schreibersefaria: removing some of the excessive validations

sefaria/utils/util.py
schreibersefaria: This is the logic as implemented by Brett
This task is already taking a very long time so I'm thinking to not work on this refactor, what do you think?

sefaria/utils/util.py
schreibersefaria: Well this was originally in middleware. 
It is used by both views and middleware.
Any good idea where I'd put this? Maybe in a new file under utils?

static/js/common/DropdownMenu.jsx
schreibersefaria: No, I don't think it's needed for the Cauldron (can't test because it's not up yet)
For now I removed setting set-language-cookie, but localhost no longer works.

sefaria/utils/util.py
schreibersefaria: Yea, that wasn't a good line, thanks.
How is this?

sefaria/utils/util.py
schreibersefaria: Changed to a generic function.
Couldn't find an existing one
Not sure I understand what you mean by param string here?

sefaria/utils/util.py
YishaiGlasner: sounds reasonable

static/js/common/DropdownMenu.jsx
schreibersefaria: Inlined.

sefaria/utils/util.py
YishaiGlasner: by param string i mean something like `?name=daniel&great=1`, witout the part before the `?` so it isn't a url

static/js/sefaria/sefaria.js
schreibersefaria: Removed all the temp logs, 
Sorry, should have done that before the PR

static/js/sefaria/sefaria.js
schreibersefaria: removed

static/js/sefaria/sefaria.js
schreibersefaria: removed the try catch in getDomainHostnames

static/js/sefaria/util.js
schreibersefaria: Are you asking why I did length validation in the if in getCookieDomain or something else I missed?

static/js/sefaria/sefaria.js
schreibersefaria: Well I don't think this will break because it's comparing hostnames.
I read now that the new implementation would miss a difference in IP.
What do you think, is the IP going to be a problem?
I think the new func is more readable

static/js/sefaria/util.js
YishaiGlasner: i mean why to change the `!hostnames.length` to `hostnames.length === 0`. i don't think we have a rule about prefering it.

static/js/sefaria/sefaria.js
schreibersefaria: Why would you run it?
Isn't running it lazily as good being as it isn't a big compute?

static/js/sefaria/sefaria.js
YishaiGlasner: a.b.com startwith a.b.co
i'm not saying it won't work, just that it seems different, so you should be sure it's ok

static/js/sefaria/sefaria.js
YishaiGlasner: this is a basic property of the session, so it should be extracted like all those other properties.

static/js/common/DropdownMenu.jsx
yitzhakc: I think the goal is that if we're moving from one module to the other, we want the language to "continue" with us, irrespective of what language had been previously set (or not set).

schreibersefaria: I believe our change to use a cookie on the cookie domain (shared across modules) should solve this for all but local host (which is another task).
Resolving this conversation

static/js/sefaria/sefaria.js
schreibersefaria: So just to call it once at the time of unpacking the props?

YishaiGlasner: i think you shouldn't use a cache here. in the unpacking extract this data to something like Sefaria.domainHostnames (you can have a function for the extraction but not here)

static/js/sefaria/sefaria.js
schreibersefaria: I think this will work.
Actually your example is of an edge case that startWith returns a false positive.

sefaria/utils/util.py
schreibersefaria: I created domains_and_languages.py and moved several functions there

static/js/sefaria/sefaria.js
schreibersefaria: But we already have Sefaria.domainModules.
getDomainHostnames Just returns a set instead of a dict (maybe the name isn't as clear as could be)
We extracted this logic to a function and cached it at Yitzhak's recommendation

sefaria/system/context_processors.py
schreibersefaria: returned

sefaria/utils/util.py
schreibersefaria: I believe this comment was originally on `current_domain_lang`

sefaria/utils/util.py
schreibersefaria: I'm not sure I am using it with a param string, where did you see that?

sefaria/system/middleware.py
schreibersefaria: I inlined it
It's now `            response = redirect(urljoin(target_domain, path) + params_string)`

static/js/sefaria/util.js
schreibersefaria: This is referring to `static fullURL`

static/js/sefaria/sefaria.js
YishaiGlasner: so just like `Sefaria.domainModules` we should have another one with a proper name

sefaria/utils/util.py
YishaiGlasner: i'll have to look in the next iteration

sefaria/system/middleware.py
```diff
@@ -101,20 +102,19 @@ def process_request(self, request):
...
             if any([bot in request.META.get('HTTP_USER_AGENT', '') for bot in no_direct]):
                 interface = domain_lang
             else:
-                redirect_domain = None
-                for domain in settings.DOMAIN_LANGUAGES:
-                    if settings.DOMAIN_LANGUAGES[domain] == interface:
-                        redirect_domain = domain
-                if redirect_domain:
-                    # When detected language doesn't match current domain langauge, redirect
+                redirect_domain = get_redirect_domain_for_language(request, interface)
```
YishaiGlasner: something with the naming disturbs me here. i see the word redirect, and ask myself how did it decide to redirect. juat later it becomes that this only a potential.

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns the pinned language for the current domain, or None if current domain is not pinned.
+    Uses DOMAIN_MODULES to detect which language the current domain belongs to.
+
+    If the hostname matches multiple languages (e.g., in local development where both languages
+    use localhost), returns None to indicate the domain is not language-pinned.
+
+    :param request: Django request object
+    :return: 'english', 'hebrew', or None
+    """
+    if not getattr(settings, 'DOMAIN_MODULES', None):
```
YishaiGlasner, line 42: maybe i already asked that, but is this can happen? if not, i think we should remove it and not overcheck

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+        return None
+
+    current_hostname = request.get_host().split(':')[0]  # Strip port if present
```
YishaiGlasner: make function to get stripped host (think about this - if we'll use ipv6 we would have to change it). use it also in other places here

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+    allowing cookies to be shared across modules (library/voices) while respecting
+    language-specific domains.
+
+    If language is None (ambiguous domain - same domains used for multiple languages),
+    finds the common domain suffix across ALL languages and modules.
+
+    :param language: 'english', 'hebrew' (long form), or None for cross-language domains
+    :return: Cookie domain string (e.g., '.sefaria.org') or None if no domain should be set
+    """
+    if not getattr(settings, 'DOMAIN_MODULES', None):
```
YishaiGlasner, line 105: same question as above

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+    :param request: Django request object
+    :param target_domain: Full domain URL (e.g., 'https://www.sefaria.org') or None
+    :return: Boolean indicating if domain switch is needed
+    """
+    current_hostname = request.get_host().split(':')[0]  # Strip port if present
+    target_hostname = urlparse(target_domain).hostname if target_domain else None
+    return target_hostname is not None and current_hostname != target_hostname
+
+
+def get_cookie_domain(language):
```
YishaiGlasner, line 91: that name tells as it returns a cookie. maybe `get_domain_suffix`?

sefaria/utils/util.py
yitzhakc: Let's not refactor this now :-)

sefaria/system/context_processors.py
yitzhakc: We inject it into the template, as it's consumed by the Javascript and not passed in as a base prop so I believe it is needed

build/ci/sandbox-values.yaml
```diff
@@ -57,7 +57,8 @@ secrets:
     ref: slack-webhook-helm
 localSettings:
   DEBUG: false
-  DOMAIN_MODULES: '{"en": {"library": "https://www.${NAME}.cauldron.sefaria.org", "voices": "https://voices.${NAME}.cauldron.sefaria.org"}, "he": {"library": "https://www.${NAME}.cauldron.sefaria.org", "voices": "https://voices.${NAME}.cauldron.sefaria.org"}}'
+  DOMAIN_LANGUAGE: {}
+  DOMAIN_MODULES: '{"library": "https://${NAME}.cauldron.sefaria.org", "sheets": "https://sheets.${NAME}.cauldron.sefaria.org"}'
```
yitzhakc, line 61: This is very strange, it looks like this is an older revision of what the `DOMAIN_MODULES` object is supposed to look like.

sefaria/utils/util.py
```diff
@@ -536,82 +536,23 @@ def get_short_lang(language):
     return "en" if language == "english" else "he"
 
 
-def current_domain_lang(request):
+def add_query_param(url, param, value=""):
```
yitzhakc: Do you want to move this to the other utility file as well?

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+        module_urls = settings.DOMAIN_MODULES.get(get_short_lang(language), {}).values()
+    else:
+        # Cross-language: collect and deduplicate URLs since different languages may share domains
+        seen_urls = set()
+        module_urls = []
+        for lang_modules in settings.DOMAIN_MODULES.values():
+            for url in lang_modules.values():
+                if url not in seen_urls:
+                    seen_urls.add(url)
+                    module_urls.append(url)
```
YishaiGlasner:
```suggestion
        module_urls = {url for lang_modules in settings.DOMAIN_MODULES.values() for url in lang_modules.values()}
```

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Extract hostnames, filtering out localhost and IP addresses
+    hostnames = []
+    for url in module_urls:
+        hostname = urlparse(url).hostname
+        if hostname and 'localhost' not in hostname and not re.match(r'^\d+\.\d+\.\d+\.\d+$', hostname):
```
YishaiGlasner: take the regex patterm out to a variable with a meaningful name

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+        seen_urls = set()
+        module_urls = []
+        for lang_modules in settings.DOMAIN_MODULES.values():
+            for url in lang_modules.values():
+                if url not in seen_urls:
+                    seen_urls.add(url)
+                    module_urls.append(url)
+
+    # Extract hostnames, filtering out localhost and IP addresses
+    hostnames = []
```
YishaiGlasner: this loop is not so pythonic. the new and cool pythonic way is:
```
hostnames = [
    hostname
    for url in module_urls
    if (hostname := urlparse(url).hostname)
    and 'localhost' not in hostname
    and not re.match(r'^\d+\.\d+\.\d+\.\d+$', hostname)
]
```
anyway, i think it will be better to do it in a different `_get_filtered_hostnames`

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+    Example: ['www.sefaria.org', 'voices.sefaria.org'] -> '.sefaria.org'
+
+    :param hostnames: List of 2+ hostname strings
+    :return: Domain suffix starting with '.' (e.g., '.sefaria.org'), or None if no common suffix
+    """
+    common_suffix = hostnames[0]
+
+    for hostname in hostnames[1:]:
+        # Trim from the beginning until we find a common suffix at a domain boundary
+        while common_suffix and not (hostname.endswith(common_suffix) and common_suffix.startswith('.')):
```
YishaiGlasner: it seems that
```suggestion
        while common_suffix and (not hostname.endswith(common_suffix) or not common_suffix.startswith('.')):
```
is much more readable even though longer

sefaria/utils/domains_and_languages.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Returns the shared domain part starting with a dot.
+    Example: ['www.sefaria.org', 'voices.sefaria.org'] -> '.sefaria.org'
+
+    :param hostnames: List of 2+ hostname strings
+    :return: Domain suffix starting with '.' (e.g., '.sefaria.org'), or None if no common suffix
+    """
+    common_suffix = hostnames[0]
+
+    for hostname in hostnames[1:]:
```
YishaiGlasner, line 146: i got this suggestion from GPT
```
common_suffix = next(
    (common_suffix[i:] for i in range(len(common_suffix))
     if hostname.endswith(common_suffix[i:]) and common_suffix[i:].startswith('.')),
    ''
)
```
i think it's more readable (and obviously more pythonic), because it find it positively

sefaria/utils/util.py
```diff
@@ -539,6 +535,26 @@ def get_short_lang(language):
         raise InputError("Invalid language. Must be 'english' or 'hebrew'.")
     return "en" if language == "english" else "he"
 
+
+def add_query_param(url, param, value=""):
```
YishaiGlasner: i think this should go to another place. even to views.py, but maybe to new views_util

sefaria/utils/util.py
```diff
@@ -539,6 +535,26 @@ def get_short_lang(language):
...
+    parameters (including duplicates) are preserved.
+
+    :param url: URL string
+    :param param: Query parameter name
+    :param value: Query parameter value (defaults to empty string)
+    :return: URL string with updated query parameters
+    """
+    parsed = urlparse(url)
+    pairs = [(k, v) for k, v in parse_qsl(parsed.query, keep_blank_values=True) if k != param]
+    pairs.append((param, "" if value is None else value))
```
YishaiGlasner: can we remove the if? value is default to '' so there is no reason to pass it None.
if not use `value or ""`

sefaria/utils/util.py
```diff
@@ -539,6 +535,26 @@ def get_short_lang(language):
...
+    Existing occurrences of the same parameter are replaced, and other query
+    parameters (including duplicates) are preserved.
+
+    :param url: URL string
+    :param param: Query parameter name
+    :param value: Query parameter value (defaults to empty string)
+    :return: URL string with updated query parameters
+    """
+    parsed = urlparse(url)
+    pairs = [(k, v) for k, v in parse_qsl(parsed.query, keep_blank_values=True) if k != param]
```
YishaiGlasner: what if i want to add another occurrence of existing param? that's legitimate (and we also use it).
i think. a general `add_query_param` should not remove it.
if this behaviour needed for this PR we should think how to do it in a different way.

static/js/sefaria/sefaria.js
```diff
@@ -518,20 +518,48 @@ Sefaria = extend(Sefaria, {
...
-    const href = Sefaria.domainModules?.[module] || Sefaria.apiHost;
+    const langCode = Sefaria._getShortInterfaceLang();
+
+    const href = Sefaria.domainModules?.[langCode]?.[module] || Sefaria.apiHost;
+
     try {
-      return new URL(href);
-    } catch {
+      const url = new URL(href);
+      return url;
```
```suggestion
      return new URL(href);
```

static/js/sefaria/sefaria.js
yitzhakc: I think it's fine to leave it in this function, and not refactor it away.
Perhaps we could just call the function preemptively as part of `Sefaria.setup`, since we know that this function will be called down the line during render...

sefaria/utils/domains_and_languages.py
yitzhakc: In a "regular" running of the server, no.
But we have various scripts that run and initialize django, and since we haven't extensively tested them all, this is a reasonable short-circuit.

sefaria/utils/domains_and_languages.py
yitzhakc: See my comment above.

sefaria/utils/util.py
yitzhakc: Can you explain your example please?

sefaria/system/middleware.py
yitzhakc: I mean, the function returns `None` if it's not meant to redirect.. I think that's pretty clear 🤷

build/ci/sandbox-values.yaml
schreibersefaria: Where did you see this?
Github didn't show me this diff.
It is very strange, probably a botched revert or something. I reverted the change to DOMAIN_MODULES

sefaria/utils/util.py
schreibersefaria: Well, it's now a general util so I thought it less suitable for the specific "domains" util file.
Do you think it's more suitable there?
If so, should I make the name more general, something like URLs?

build/ci/sandbox-values.yaml
schreibersefaria: Yes this was something Yotam told me to revert and it looks like you found a specific commit (the revert)

sefaria/system/middleware.py
schreibersefaria: Do you think `target_domain` would be more clear?
Not sure I understood your second comment.
I added a comment to clarify the need for `needs_domain_switch`

sefaria/utils/domains_and_languages.py
schreibersefaria: This is a name already used on the frontend so I think keeping it makes sense.
Also, I do think cookie domain is a term used so it should be clear.

sefaria/utils/domains_and_languages.py
schreibersefaria: You commented on this and I did remove a lot of overchecking.
I returned some of them as I worked

sefaria/utils/util.py
YishaiGlasner: `api/name/something?type=ref&type=topic`
i'm nost saying that such a csae will raise here, but that if the function name says it adds param it should remove nothing

sefaria/utils/domains_and_languages.py
schreibersefaria: Created _get_hostname_without_port and it deals with IPv6 just in case

sefaria/system/middleware.py
YishaiGlasner: maybe, i had no second comment, i just explain that when i came this line i thought "how are they sure it should be redirected?"

sefaria/utils/domains_and_languages.py
schreibersefaria: Thanks

sefaria/utils/domains_and_languages.py
schreibersefaria: Never saw the walrus operator, very cool

sefaria/utils/util.py
schreibersefaria: @yitzhakc pulling you into this conversation because you also had a comment about the location of this function.
Does a new `utils/views_utils.py` make sense to you?

sefaria/utils/util.py
schreibersefaria: Yishai had another idea of where to put this, I tagged you in the conversation

sefaria/utils/util.py
schreibersefaria: How about this flag letting the user chose the functions behaviour?
Next time I will think twice about your comments to make a function more generic ;-)

sefaria/system/middleware.py
schreibersefaria: Renamed

sefaria/utils/util.py
schreibersefaria: Moved add_query_param to the new utils/views_utils.py

sefaria/utils/util.py
schreibersefaria: OK, I refactored to only support adding after a chat with Yishai

static/js/sefaria/sefaria.js
schreibersefaria: After a conversation with Yishai we thought it best to leave the function but remove the cache.
It isn't a big calculation and isn't called often.
Yishai felt strongly that if we want the value to be saved then it should be a saved to the Sefaria obj in setup. I thought that was a duplication.
We both thought this is the most correct solution.

yitzhakc (approved):

PR #2764, opened 2025-10-28
fix(SheetsHomePage): Extract topic from holiday API response

schreibersefaria:
Fixed the getUpcomingDay function to deal with the nested structure of the data.
I'm not sure what broke now!

### DIFF
--- a/static/js/sheets/SheetsHomePageTopicsTOC.jsx
+++ b/static/js/sheets/SheetsHomePageTopicsTOC.jsx
@@ -41,7 +41,7 @@ const SheetsHoliday = ({handleClick}) => {
   useEffect(() => {
     Sefaria.getUpcomingDay('holiday').then(data => {setHoliday(data?.topic || null)});
   }, []);
-  if (Object.keys(holiday).length === 0) {
+  if (Object.keys(holiday || {}).length === 0) {
     return <div className="navBlock">Loading...</div>
   }
   return <TopicTOCCard topic={holiday} setTopic={handleClick} showDescription={true}/>;


static/js/sheets/SheetsHomePageTopicsTOC.jsx
```diff
@@ -39,7 +39,11 @@ const SheetsParashah = ({handleClick}) => {
 const SheetsHoliday = ({handleClick}) => {
   const [holiday, setHoliday] = useState({});
   useEffect(() => {
-    Sefaria.getUpcomingDay('holiday').then(setHoliday);
+    Sefaria.getUpcomingDay('holiday').then(data => {
+      if (data && data.topic) {
```
YishaiGlasner:
```suggestion
      if (data?.topic) {
```
or maybe we can do it without condition `setHoliday(data?.topic || {});`?

schreibersefaria: @YishaiGlasner Would the `|| {}` be the same as setting `setHoliday`

YishaiGlasner: hmm, this will do the `setHoliday` if `data?.topic`, but it will do a redundant re-render if not.
it should be changed to
```
const [holiday, setHoliday] = useState(null);
...
setHoliday(data?.topic || null);
```
that will set holliday if there's `topic`, and won't do redundant re-render.

YishaiGlasner: but then you should change also `Object.keys(holiday)` to `Object.keys(holiday || {})`, but anyway `useState({});` isn't best practice

schreibersefaria: Noice. I think I fixed

PR #2769, opened 2025-10-29
chore: fix english link

stevekaplan123:
## Description
Fix two links


yitzhakc:
> i don't like having all of this hardcoded

@YishaiGlasner 
It's not ideal, but I feel that in this situation it is the correct solution, in the balance between over-engineering and keeping it simple.

PR #2770, opened 2025-10-29
Feature/sc 37593/mdl backend redirect changes

yodem:
## Description
Add Django redirects from the library to the voices module for `/settings/profile` and `/community`. Implements cross-module navigation by redirecting users from `sefaria.org` to `voices.sefaria.org` for these paths, with query parameters preserved.

## Code Changes
- `reader/views.py`
  - Added `settings_profile_redirect()` to redirect `/settings/profile` to the voices module.
  - Added `community_to_voices_redirect()` to redirect `/community` to the voices module root.
  - Both redirects are language-aware, return 301s, and preserve query parameters.
- `sefaria/urls_library.py`
  - Added URL pattern `r'^settings/profile?$'` mapping to `settings_profile_redirect`.
  - Added URL pattern `r'^community/?$'` mapping to `community_to_voices_redirect`.

## Notes
- Redirects only apply on `sefaria.org`; accessing these paths on `voices.sefaria.org` returns 404.
- Query parameters are preserved (e.g., `?tab=notes`).
- URLs return 301.
- Follows the existing redirect pattern.

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -4636,104 +4636,73 @@ def talmud_person_index_redirect(request):
     return redirect(iri_to_uri('/topics/category/talmudic-figures'), permanent=True)
 
 
-def settings_profile_redirect(request):
+def redirect_to_voices(request, target_path):
     """
-    Redirect /settings/profile from library module to voices module
+    Redirect from library module to voices module
     """
     # Get the voices domain from settings
     lang_code = get_short_lang(request.interfaceLang)
     voices_domain = DOMAIN_MODULES.get(lang_code, {}).get(VOICES_MODULE)
-    target_url = urllib.parse.urljoin(voices_domain, "/settings/profile/")
-    
+    target_url = urllib.parse.urljoin(voices_domain, target_path)
+
     # Preserve query parameters
-    params = request.GET.urlencode()
-    if params:
+    if params := request.GET.urlencode():
         target_url += f"?{params}"
-    
+
     return redirect(target_url, permanent=True)
+
+
+def settings_profile_redirect(request):
+    """
+    Redirect /settings/profile from library module to voices module
+    """
+    return redirect_to_voices(request, "/settings/profile/")
 
 
 def community_to_voices_redirect(request):
     """
     Redirect /community from library module to voices module root
     """
-    # Get the voices domain from settings
-    lang_code = get_short_lang(request.interfaceLang)
-    voices_domain = DOMAIN_MODULES.get(lang_code, {}).get(VOICES_MODULE)
-    target_url = urllib.parse.urljoin(voices_domain, "/")
-    
-    # Preserve query parameters
-    params = request.GET.urlencode()
-    if params:
-        target_url += f"?{params}"
-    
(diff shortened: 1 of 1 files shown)


reader/views.py
```diff
@@ -4623,6 +4623,48 @@ def talmud_person_index_redirect(request):
     return redirect(iri_to_uri('/topics/category/talmudic-figures'), permanent=True)
 
 
+def settings_profile_redirect(request):
+    """
+    Redirect /settings/profile from library module to voices module
+    """
+    # Only redirect if on library module to avoid redirect loops
+    if getattr(request, 'active_module', LIBRARY_MODULE) != LIBRARY_MODULE:
+        raise Http404
```
YishaiGlasner: i think this is over validating. we have different urls for library and voices.
so this only for a future mistake of an engineer that used the wrong view?

reader/views.py
```diff
@@ -4623,6 +4623,48 @@ def talmud_person_index_redirect(request):
...
+    
+    # Get the voices domain from settings
+    domain_modules = get_language_specific_domain_modules(request.interfaceLang)
+    voices_domain = domain_modules[VOICES_MODULE]
+    target_url = f"{voices_domain}/settings/profile"
```
YishaiGlasner: use `urljoin`

YishaiGlasner (commented):
1. there is something weird with the description here. you wrote you remove the things that you just added.
2.  my comments on `settings_profile_redirect` are also for `community_to_voices_redirect`

reader/views.py
```diff
@@ -4636,6 +4636,106 @@ def talmud_person_index_redirect(request):
...
+    Redirect /settings/profile from library module to voices module
+    """
+    # Get the voices domain from settings
+    lang_code = get_short_lang(request.interfaceLang)
+    voices_domain = DOMAIN_MODULES.get(lang_code, {}).get(VOICES_MODULE)
+    target_url = urllib.parse.urljoin(voices_domain, "/settings/profile/")
+    
+    # Preserve query parameters
+    params = request.GET.urlencode()
+    if params:
```
YishaiGlasner: i discovered the cool walrus operator:
```suggestion
    if params := request.GET.urlencode():
```
you can also use it in the next functions

reader/views.py
```diff
@@ -4636,6 +4636,106 @@ def talmud_person_index_redirect(request):
...
+        target_url += f"?{params}"
```
YishaiGlasner, line 4650: it seems that all the functions here are doing the same, so you can have a function
```
def redirect_to_voices(request, target_path):
```
that will return redirect, and use it for all 5 new functions.

reader/views.py
```diff
@@ -4636,6 +4636,106 @@ def talmud_person_index_redirect(request):
...
+    
+    return redirect(target_url, permanent=True)
```
YishaiGlasner, line 4652: it seems that all the functions here are doing the same, so you can have a function
```
def redirect_to_voices(request, target_path):
```
that will return redirect, and use it for all 5 new functions.

PR #2788, opened 2025-11-03
Update Brand Graphics and Nomenclature

saengel:
## Description
Updated the mobile hamburger menu for brand graphics and updated naming. 

## Code Changes
1. In `static/css/s2.css` - Padding for dropdown dots, and logic to flip the chevron direction for Hebrew
2. `static/icons/chevron-right.svg` - The new chevron icon
3.  `static/js/Header.jsx` - Shifting language (ensuring language for `<InterfaceText />` is defined in `Strings.js` to account for future changes across the site, moving out the icons in favor of the dots, shifting the products into a chevron icon with new language. 
4. `static/js/sefaria/strings.js` - Adding relevant language for i8n purposes.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -15118,7 +15118,7 @@ span.ref-link-color-3 {color: blue}
   margin-inline-end: 15px;
 }
 
-.interface-hebrew .mobileAccountLinks img[src="/static/icons/chevron-right.svg"] {
+.interface-hebrew .mobileAccountLinks .chevron {
   transform: rotate(180deg);
 }
 
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -535,7 +535,7 @@ const MobileNavMenu = ({ onRefClick, showSearch, openTopic, openURL, close, visi
         </a>
 
         <a href="/products" data-target-module={Sefaria.LIBRARY_MODULE}>
-          <img src="/static/icons/chevron-right.svg"/>
+          <img className="chevron" src="/static/icons/chevron-right.svg"/>
           <InterfaceText>More from Sefaria</InterfaceText>
         </a>
 


static/css/s2.css
```diff
@@ -15114,6 +15114,14 @@ span.ref-link-color-3 {color: blue}
   flex-shrink: 0;
 }
 
+.mobileAccountLinks .dropdownDot {
+  margin-inline-end: 15px;
+}
+
+.interface-hebrew .mobileAccountLinks img[src="/static/icons/chevron-right.svg"] {
```
YishaiGlasner: That coupled to the file path and mixes presentation logic with content logic. Also it can cause problems with caching.
Use classnames.

static/js/Header.jsx
```diff
@@ -517,26 +517,26 @@ const MobileNavMenu = ({ onRefClick, showSearch, openTopic, openURL, close, visi
 
         {module === Sefaria.LIBRARY_MODULE &&
           <a href="/" data-target-module={Sefaria.VOICES_MODULE}>
-            <img src="/static/icons/sheets-mobile-icon.svg" alt={Sefaria._("Sheets")} />
-            <InterfaceText>Sheets</InterfaceText>
+            <span className="dropdownDot" style={{backgroundColor: `var(--sheets-green)`}}></span>
```
YishaiGlasner, line 520: is that the same way we do it for desktop? it would be better to have one way so it would be simple to refactor it to one component in future

static/js/sefaria/strings.js
```diff
@@ -693,9 +693,11 @@ const Strings = {
     "Collection Logo": "לוגו אסופה",
     "Settings": "הגדרות",
     "Library": "ספריה",
+    "Sefaria Library": "ספריית ספריא",
```
YishaiGlasner, line 696: we have these strings working in desktop, isn't it?

static/css/s2.css
saengel: Good call, thank you - you should see the fix pushed now

static/js/Header.jsx
saengel: Yes, this is how we do it exactly for desktop - thank you for the future thinking, good call!

static/js/sefaria/strings.js
saengel: I searched the `strings.js` file and didn't see any of the ones I needed to add for this spec, what do you suspect?

YishaiGlasner: i thought we have the same strings in switcher for desktop, but no i see there we have shorter strings.

yitzhakc:
I would posit that this is quite a simple (as in, low complexity) task 😁

PR #2793, opened 2025-11-03
Fixed saving saved alignment in Heb interface

schreibersefaria:
Removed some CSS that was causing the problem and add `left: auto;` for the tooltip


saengel (approved):
Looks great on my local!

PR #2794, opened 2025-11-04
Fix dropdown menu

YishaiGlasner:
## Description
1. Move label before input.
2. Add css for hiding layout buttons' label without spoiling accessibility.

## Note
I added the selector `.sr-only`, although not using it, for future use, and also it makes the purpose of the block clear.


schreibersefaria (approved):
Testing it and it didn't seem to break anything accessibility related

PR #2803, opened 2025-11-05
Update Specific HTML Titles

saengel:
## Description
This PR introduces helper functions to the backend and frontend where we set the `<title>` of the page to help create consistent branding with the suffixes on pages (i.e. `page-title | Voices on Sefaria` and `page-title | Library on Sefaria`). 

## Code Changes
1. In `reader/views.py` - Create a helper function we invoke whenever setting the title that handles the generic suffix for the module and handles special cases. 
2. In `sourcesheets/views.py` - Invoke that function for sheet-specific pages
3. In `static/js/ReaderApp.jsx` - parallel for the frontend (sometimes was overriding the backend logic). Some special logic around shifting from `<Specific Ref with Connections>` to `<Specific Ref with Connections> | Sefaria Library`, needed to filter out cases of more generalized connection lists (i.e. list of all Talmud connections).

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -3584,37 +3584,29 @@ _media: {},
           voices: "Sheets from Voices on Sefaria",
           library: "Texts from the Sefaria Library"
         },
-        collections: "Voices on Sefaria",
-        collection: "Voices on Sefaria Collection",
+        collections: {
+          voices: "Voices on Sefaria"
+        },
+        collection: {
+          voices: "Voices on Sefaria Collection"
+        },
         default: {
           voices: "Voices on Sefaria",
           library: "Sefaria Library"
         }
       };
   
-      // Special case: Home pages return complete title (not base + suffix pattern)
-      if (pageType === "home") {
-        return Sefaria._(suffixes.home[module]);
-      }
-  
       // Special case: Sheet titles need default if empty
       if (pageType === "sheet" && !baseTitle) {
         baseTitle = "Untitled";
       }
-  
-      // Get appropriate suffix based on page type
-      let suffix;
-      if (pageType === 'collections' || pageType === 'collection') {
-        // Collections pages are always Voices
-        suffix = suffixes[pageType];
-      } else if (pageType === 'topic') {
-        // Topics have module-specific descriptive suffixes
-        suffix = suffixes.topic[module];
-      } else {
-        // Default suffix for all other pages (pageType === "" or anything else)
-        suffix = suffixes.default[module];
+
+      if (!pageType) {
+        pageType = "default";
       }
   
+      const suffix = suffixes[pageType][module];
+
       // Combine base title with suffix
       return baseTitle ? `${Sefaria._(baseTitle)} | ${Sefaria._(suffix)}` : Sefaria._(suffix);
     },


reader/views.py
```diff
@@ -95,6 +95,92 @@
...
+        >>> # Sheet page
+        >>> get_page_title("My Torah Sources", "sheet", request)
+        "My Torah Sources | Voices on Sefaria"  # English Voices
+
+    Note:
+        - Base titles should be translated BEFORE calling this function using _()
+        - This function only adds module-specific suffixes
+        - For base title translations, update locale/he/LC_MESSAGES/django.po
+    """
+    if not request:
```
YishaiGlasner: why do we need this (and the default value)?

reader/views.py
```diff
@@ -95,6 +95,92 @@
...
+        return base_title
+
+    is_hebrew = (request.interfaceLang == "hebrew")
+    is_voices = (getattr(request, 'active_module', LIBRARY_MODULE) == VOICES_MODULE)
```
YishaiGlasner: why those are needed? you can just have `lang` and `module`

reader/views.py
```diff
@@ -95,6 +95,92 @@
...
+        suffix = suffixes[page_type][lang]
+    elif page_type == 'topic':
+        # Topics have module-specific descriptive suffixes
+        suffix = suffixes['topic'][module][lang]
+    else:
+        # Standard suffix for all other pages
+        suffix = suffixes['default'][module][lang]
+
+    # Combine base title with suffix
+    return f"{base_title} | {suffix}" if base_title else suffix
```
YishaiGlasner: won't it throw NameError if `page_type == "sheet"`?

reader/views.py
```diff
@@ -95,6 +95,92 @@
...
+        'default': {
+            'voices': {'he': "חיבורים בספריא", 'en': "Voices on Sefaria"},
+            'library': {'he': "ספריית ספריא", 'en': "Sefaria Library"}
+        }
+    }
+
+    lang = 'he' if is_hebrew else 'en'
+    module = 'voices' if is_voices else 'library'
+
+    # Special case: Home pages return complete title (not base + suffix pattern)
```
YishaiGlasner: next lines has many duplicates. suggestion (i can't say it's perfect but you can fix it):
```
suffix = ''
if page_type != 'sheet':
  page_suffixes = suffixes.get(page_type, suffixes['default'])
  suffix = page_titles.get(module, page_titles)[lang]
else:
  base_title = strip_tags(base_title) if base_title else ("ללא כותרת" if is_hebrew else "Untitled")
return f"{base_title} | {suffix}" if (base_title and suffix) else f"{base_title}{suffix}"
```
also if we already have a dict, you can put there the untitled strings

reader/views.py
```diff
@@ -632,7 +718,8 @@ def text_panels(request, ref, version=None, lang=None, sheet=None):
         "initialNavigationTopicTitle":    None,
     }
     if sheet == None:
-        title = primary_ref.he_normal() if request.interfaceLang == "hebrew" else primary_ref.normal()
+        ref_text = primary_ref.he_normal() if request.interfaceLang == "hebrew" else primary_ref.normal()
+        title = get_page_title(ref_text, "text", request)
```
YishaiGlasner: that mean we are changing the behaviour (adding suffix)? i ask b/c i don't see it in the description.

reader/views.py
```diff
@@ -95,6 +95,92 @@
...
+        - For base title translations, update locale/he/LC_MESSAGES/django.po
+    """
+    if not request:
+        return base_title
+
+    is_hebrew = (request.interfaceLang == "hebrew")
+    is_voices = (getattr(request, 'active_module', LIBRARY_MODULE) == VOICES_MODULE)
+
+    # Page title suffix configuration
+    suffixes = {
```
YishaiGlasner, line 138: why do we need this if we have the `_`?

reader/views.py
```diff
@@ -95,6 +95,92 @@
 import structlog
 logger = structlog.get_logger(__name__)
 
+
+def get_page_title(base_title, page_type="", request=None):
```
YishaiGlasner: this function does 2 things:
1. the same thing as `_`. this is redundant, and also hardcoding strings.
2. something like ' | '.join(base_title, suffix). function for doing this can be nice as it will reduce duplications.
i wrote the next comments before understanding that - now i think that this is function is redundant (and we need only function for joining).

static/js/ReaderApp.jsx
```diff
@@ -432,11 +432,99 @@ class ReaderApp extends Component {
   clonePanel(panel, prepareForSerialization) {
     return Sefaria.util.clone(panel, prepareForSerialization);
   }
+  getPageTitle(baseTitle, pageType = "") {
```
YishaiGlasner: again, we have the `_` that seperate strings from code? why do we need a duplication that mixes them?

reader/views.py
yitzhakc: @saengel 
I'll say that I disagree slightly with @YishaiGlasner in that I think such a function has its place, as the centralized location for logic that handles injecting the module string into the HTML title, and not relying on a simple joining for potential outliers that are guaranteed to be there.

But I agree that we should be using string injection into the translation function (`_`) and not passing in already-translated strings modifying it with (more) hard coded strings.

See [examples in the docs](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#internationalization-in-python-code) on how to use placeholders with named-string interpolation to inject strings as part of the translation features.

yitzhakc (changes requested):

reader/views.py
saengel: I shifted to use a placeholder approach, but still need the hard coded strings for the suffixes. Does this combination approach work? Or would you suggest something different?

reader/views.py
saengel: It's an extra check - to enable failing gracefully in the event of a function call without the proper request context. If it's too much / overkill - I'll remove it. Curious your take.

reader/views.py
saengel: It's a cleaner way of writing the if/else below with fewer statements, since instead of checking the module we can just go to the correct suffix based on the module in the dictionary.

reader/views.py
saengel: Yes, titles on sheets also have a suffix. Now should be `<Name of Sheet> | Voices on Sefaria`

static/js/ReaderApp.jsx
saengel: Marked as resolved because edited since to utilize `_`

reader/views.py
saengel: There's definitely a way to make this less redundant, but in my opinion it makes it much less readable. I'm going to keep it as is for now.

reader/views.py
YishaiGlasner: it seems to me overkill, for this is a function for handling request, so how would it be called without request? moreover, when you have a default value, it hints that you can call it without this param!

reader/views.py
```diff
@@ -95,6 +95,94 @@
...
+        - Base titles should be translated BEFORE calling this function using _()
+        - This function adds module-specific suffixes and formats them
+        - The title format "%(title)s | %(suffix)s" is translatable, allowing translators
+          to customize the separator, spacing, and ordering for their language
+        - For translations, update locale/he/LC_MESSAGES/django.po
+    """
+    if not request:
+        return base_title
+
+    is_voices = (getattr(request, 'active_module', LIBRARY_MODULE) == VOICES_MODULE)
```
YishaiGlasner: move this to line 156 (=before l157) for more clarity

reader/views.py
```diff
@@ -95,6 +95,94 @@
...
+
+    # Page title suffix configuration - using _() for translations
+    suffixes = {
+        'home': {
+            'voices': _("Voices on Sefaria"),
```
YishaiGlasner: i think it will be better to omit _ in the dict, and use it in the return

reader/views.py
```diff
@@ -95,6 +95,94 @@
...
+        # Topics have module-specific descriptive suffixes
+        suffix = suffixes['topic'][module]
+    else:
+        # Standard suffix for all other pages
+        suffix = suffixes['default'][module]
+
+    # Combine base title with suffix using Django's named-string interpolation
+    # This allows translators to customize the format, separator, and ordering
+    if base_title:
+        return _("%(title)s | %(suffix)s") % {'title': base_title, 'suffix': suffix}
```
YishaiGlasner, line 167: you're not translating here a sentence, but two phrases, and the separator doesn't change, so it will be cleaner to do
```
f"{_(base_title)} | {_(suffix)}"
```

static/js/ReaderApp.jsx
```diff
@@ -432,11 +432,70 @@ class ReaderApp extends Component {
...
+     * Generate consistent, module-aware, bilingual page titles.
+     * Mirrors the Python get_page_title() function in reader/views.py
+     *
+     * @param {string} baseTitle - Main content of title (should already be translated)
+     * @param {string} pageType - Page type: "home", "topic", "collection", "collections", or "" for default
+     * @returns {string} Formatted title with module suffix
+     */
+
+    // Get current module (library or voices)
+    const activeModule = Sefaria.activeModule || 'library';
```
YishaiGlasner: i think lines 446 447 and and 467 can be one line
```
const module = (Sefaria.activeModule === 'voices') ? 'voices' : 'library';

File: static/js/ReaderApp.jsx [hunk 13]
```diff
@@ -432,11 +432,70 @@ class ReaderApp extends Component {
...
+      collection: Sefaria._("Voices on Sefaria Collection"),
+      default: {
+        voices: Sefaria._("Voices on Sefaria"),
+        library: Sefaria._("Sefaria Library")
+      }
+    };
+
+    const module = isVoices ? 'voices' : 'library';
+
+    // Special case: Home pages return complete title (not base + suffix pattern)
```
YishaiGlasner: same comment as on server side

static/js/ReaderApp.jsx
```diff
@@ -432,11 +432,70 @@ class ReaderApp extends Component {
   clonePanel(panel, prepareForSerialization) {
     return Sefaria.util.clone(panel, prepareForSerialization);
   }
+  getPageTitle(baseTitle, pageType = "") {
```
YishaiGlasner: i think this function should live on sefaria.js.

reader/views.py
yitzhakc: I think the underlying question is - would we _want_ to call this function without `request`. 
That is, the only use of `request` is to retrieve the `active_module`. One could argue that it would be more correct (in terms of simplifying dependencies and enabling a more flexible use of the function) to pass along `request.active_module` to the function, and not the entire request object. This would allow us to use the function even without the request object. But as I mentioned at the beginning of my comment, I'm not sure we _want_ to call it without `request` :-)

reader/views.py
```diff
@@ -93,6 +95,94 @@
...
+        },
+        'collections': _("Voices on Sefaria"),
+        'collection': _("Voices on Sefaria Collection"),
+        'default': {
+            'voices': _("Voices on Sefaria"),
+            'library': _("Sefaria Library")
+        }
+    }
+
+    module = 'voices' if is_voices else 'library'
```
yitzhakc: Do we not have constants for these? Also, why not just use the `active_module` from the `request`?

reader/views.py
yitzhakc: @saengel It seems like the `suffixes` dictionary is already full of *translated* strings. Why then do we pass it _again_ through the `_()` function?

reader/views.py
```diff
@@ -93,6 +95,94 @@
...
+    This function takes a base title (which should already be translated via Django's _())
+    and adds the appropriate module suffix based on active_module and interface language.
+    The title format itself (including separator) is translatable using Django's named-string
+    interpolation, allowing translators to customize the format for their language.
+
+
+    Args:
+        base_title (str): Main content of title - SHOULD BE ALREADY TRANSLATED via _()
+                         Examples: _("Topics"), ref.normal(), sheet["title"], user.full_name
+        page_type (str): Specifies if it's a specific page type requiring a specific suffix (i.e. home, sheet, topic or collection)
```
yitzhakc: Can we use an enum or constants and not a lot of new magic strings please? Just to stress the importance, the page_type `generic` doesn't even show up here!

static/js/ReaderApp.jsx
```diff
@@ -432,11 +432,70 @@ class ReaderApp extends Component {
...
+     * Mirrors the Python get_page_title() function in reader/views.py
+     *
+     * @param {string} baseTitle - Main content of title (should already be translated)
+     * @param {string} pageType - Page type: "home", "topic", "collection", "collections", or "" for default
+     * @returns {string} Formatted title with module suffix
+     */
+
+    // Get current module (library or voices)
+    const activeModule = Sefaria.activeModule || 'library';
+    const isVoices = (activeModule === 'voices');
```
yitzhakc: Please don't use magic strings. I'm pretty sure we have a voices constant that can be used

static/js/ReaderApp.jsx
```diff
@@ -432,11 +432,70 @@ class ReaderApp extends Component {
...
+      },
+      collections: Sefaria._("Voices on Sefaria"),
+      collection: Sefaria._("Voices on Sefaria Collection"),
+      default: {
+        voices: Sefaria._("Voices on Sefaria"),
+        library: Sefaria._("Sefaria Library")
+      }
+    };
+
+    const module = isVoices ? 'voices' : 'library';
```
yitzhakc: Can't we take this from the active module?

yitzhakc (changes requested):

reader/views.py
saengel: Good point

reader/views.py
saengel: @YishaiGlasner - @yitzhakc asked we use placeholders instead, this seemed to be what the documentation recommended. 

@yitzhakc - whoops, thank you for catching.

static/js/ReaderApp.jsx
saengel: Yes, that would be better. Thanks for catching

reader/views.py
saengel: In terms of the page type?

reader/views.py
saengel: I misunderstood - still think it is cleaner but fixed the double translation issue.

reader/views.py
YishaiGlasner: in my understading, this is recommended when you translate a senetence, not when you only join phrases

static/js/ReaderApp.jsx
saengel: Agree, thanks

static/js/ReaderApp.jsx
saengel: Which specifically - the repetition, or where the translation is happening or something else? If you could elaborate that would be great.

reader/views.py
saengel: @yitzhakc can you clarify what you meant here in terms of placeholders? I could use some clarifying re my misunderstanding.

yitzhakc: In a previous version, I believe we were passing in the _translated_ string to `base_title` from the calling functions. 
Now that is no longer the case, so my comment here is moot. Resolving.

reader/views.py
```diff
@@ -93,6 +95,90 @@
...
+    COLLECTION = "collection"
+    COLLECTIONS = "collections"
+    TEXT = "text"
+    PROFILE = "profile"
+    DEFAULT = ""
+
+
+def get_page_title(base_title, module, page_type=""):
+    """
+    Generate consistent, module-aware page titles with appropriate suffixes.
```
yitzhakc: I'd add a reference to the fact that these page titles are also translated, e.g. language-aware as well.

static/js/sefaria/sefaria.js
```diff
@@ -3580,6 +3565,59 @@ _media: {},
...
     return `/logout?next=/${next}`;
   },
+  getPageTitle: (baseTitle, pageType = "") => {
+      /**
+       * Generate consistent, module-aware, bilingual page titles.
+       * Mirrors the Python get_page_title() function in reader/views.py
+       */
+  
+      // Get current module (library or voices)
+      const module = (Sefaria.activeModule === 'voices') ? 'voices' : 'library';
```
yitzhakc: We have  `VOICES_MODULE` and `LIBRARY_MODULE` constants in `constants.js`, please use them.

saengel: Ah, I didn't realize! Will do!

saengel: Just pushed a fix

reader/views.py
```diff
@@ -95,6 +95,84 @@
...
+        PageTypes.COLLECTION: {'voices': "Voices on Sefaria Collection"},
+        'default': {
+            'voices': "Voices on Sefaria",
+            'library': "Sefaria Library"
+        }
+    }
+
+    # Special case: Home pages return complete title (not base + suffix pattern)
+    if page_type == PageTypes.HOME:
+        return _(suffixes[PageTypes.HOME][module])
```
YishaiGlasner:
```suggestion
        base_title = None
```
this will do the same without different return, which is less clean (disturb the flow).
if it should be falsey without it, and these 2 lines can be removed it's even better...

static/js/sefaria/sefaria.js
```diff
@@ -3565,6 +3565,59 @@ _media: {},
...
+      if (pageType === 'collections' || pageType === 'collection') {
+        // Collections pages are always Voices
+        suffix = suffixes[pageType];
+      } else if (pageType === 'topic') {
+        // Topics have module-specific descriptive suffixes
+        suffix = suffixes.topic[module];
+      } else {
+        // Default suffix for all other pages (pageType === "" or anything else)
+        suffix = suffixes.default[module];
+      }
```
YishaiGlasner: do it here in the same way as server side

saengel: Good call - just pushed

reader/views.py
saengel: 👍

YishaiGlasner (approved):

PR #2804, opened 2025-11-05
feat(linker): support citations like Mishnah in Beitzah 14b which will link to the mishnah on that daf

nsantacruz:
## Description
Allows for matching citations that cite a mishnah inside Bavli. E.g. Mishnah on Beitzah 14b. In the case there are multiple mishnayot on that daf, it will return an ambiguous resolved ref.

## Code Changes
We support this by allowing Passages to have match templates and adding a flag to Schema Nodes that indicates this node can have passage children. 

In practice, we would add the "mishnah" match template to all relevant Passages that are mishnayot in bavli. This would then allow the linker to query for these mishnayot as children of the given daf.

This also allows for arbitrary named children of a schema node. e.g. see Exodus 15 Shirat HaYam.

### DIFF
--- a/sefaria/model/linker/referenceable_book_node.py
+++ b/sefaria/model/linker/referenceable_book_node.py
@@ -463,6 +463,7 @@ class PassageNodeSet(ReferenceableBookNode):
     
 class PassageMatcher:
 
+    # make class into singleton that is thread-safe
     _instance = None
     _lock = Lock()
 


sefaria/model/linker/referenceable_book_node.py
```diff
@@ -431,6 +434,34 @@ def fuzzy_match_score(self, lang, raw_ref_part) -> DiburHamatchilMatch:
...
+    def get_match_template_trie(self, *args, **kwargs):
+        return self._passage.get_match_template_trie(*args, **kwargs)
+    
+    
+class PassageNodeSet(ReferenceableBookNode):
+    
+    def __init__(self, ref: text.Ref):
+        query = {
+            "$or": [
+                {'ref_list': {'$regex': r}} for r in ref.regex(as_list=True)
```
yonadavGit: why use regex in query?

sefaria/model/linker/resolved_ref_refiner_factory.py
```diff
@@ -54,6 +54,7 @@ def initialize_resolved_ref_refiner_factory() -> ResolvedRefRefinerFactory:
         (key(RefPartType.NAMED, node_class=NamedReferenceableBookNode), ResolvedRefRefinerForNamedNode),
         (key(RefPartType.NUMBERED, node_class=NamedReferenceableBookNode), ResolvedRefRefinerForNamedNode),
         (key(RefPartType.DH), ResolvedRefRefinerForDiburHamatchilPart),
+        (key(RefPartType.NAMED, node_class=PassageNodeSet), ResolvedRefRefinerForPassage),
```
yonadavGit, line 57: cool
just from looking at this code line the logic is clear-
a ReferenceableBookNode should be treated as a potential passage match if it's named and its type is PassageNodeSet.

one optional note, the term "node" is a bit general, perhaps it's possible to chage the node_class key to sth like ReferenceableBookNode_key so it's clearer

sefaria/model/schema.py
```diff
@@ -1108,6 +1108,9 @@ def serialize(self, **kwargs):
 
     def is_segment_level_dibur_hamatchil(self) -> bool:
         return getattr(self, 'isSegmentLevelDiburHamatchil', False)
+    
+    def has_passage_children(self) -> bool:
+        return getattr(self, 'hasPassageChildren', False)
```
yonadavGit: why hasPassageChildren is in the data?
why can't it be computed from the data on the object level?

sefaria/model/linker/referenceable_book_node.py
```diff
@@ -450,18 +454,43 @@ def get_match_template_trie(self, *args, **kwargs):
...
         
     def get_children(self, *args, **kwargs) -> List['PassageNode']:
         return [PassageNode(passage) for passage in self._passages]
-        
+    
+    
+class PassageMatcher:
+
+    _instance = None
+    _lock = Lock()
```
yonadavGit, line 468: maybe a small comment explaining thatn these vars are there to ensure the mathcer is a safe singleton

sefaria/model/linker/referenceable_book_node.py
```diff
@@ -450,18 +454,43 @@ def get_match_template_trie(self, *args, **kwargs):
...
+    def __new__(cls):
+        if cls._instance is None:
+            with cls._lock:
+                if cls._instance is None:
+                    cls._instance = super().__new__(cls)
+                    cls._instance._initialize()
+        return cls._instance
+
+    def _initialize(self):
+        self._by_segment = defaultdict(list)
```
yonadavGit, line 479: love this text-book case of encapsulation

PR #2807, opened 2025-11-06
feat(linker): move wrapping of named entities to MUTC

nsantacruz:
## Description
Turns out that named entity wrapping is conflicting with link wrapping. This NE wrapping happens first, the char locations for the links no longer match the original text. The same will happen if we reverse the order.

Instead, we can integrate named entities into MUTC and have links and named entities be wrapped at the same time. This is great! 

One downside is that while links can be regenerated when a segment is edited, named entities currently can't (technically, the linker can also link named entities but this functionality is not rigorously tested). 

The solution can be to only delete citation spans when an MUTC is edited. If the number of spans is zero after citations are deleted, delete the MUTC. Otherwise save the MUTC with only the named entities.

# Begin Copilot summary here!

This pull request refactors how span types are handled in the `MarkedUpTextChunk` model and its consumers, moving from string literals to an explicit Enum. It also enforces stricter validation for span attributes and updates related logic in the linker helper and text request adapter. The changes improve code clarity, validation robustness, and future extensibility.

**Model and Validation Improvements:**

* Introduced the `MarkedUpTextChunkSpanType` Enum to define allowed span types (`QUOTE`, `NAMED_ENTITY`, `CITATION`), replacing string literals throughout the codebase.
* Updated the validation logic in `MarkedUpTextChunk._validate` to require a `ref` for `CITATION` spans and a `topicSlug` for `NAMED_ENTITY` spans, raising an error if these are missing.
* Extended the allowed span types in the schema to use all values from the new Enum, and made `ref` and `topicSlug` optional depending on the span type.

**Helper and Task Logic Changes:**

* Updated `_extract_resolved_spans` in `linker/tasks.py` to use the Enum for the `"type"` field instead of the string `"citation"`.
* Modified `_replace_existing_chunk` to filter and retain only `NAMED_ENTITY` spans (using the Enum value) in existing chunks, deleting or saving the chunk accordingly.

**Text Request Adapter Refactor:**

* Removed the named entity wrapping logic in favor of using marked up text chunks and replaced the `link_wrapper` function with `mutc_wrapper`, ensuring spans are applied using the updated model. [[1]](diffhunk://#diff-d42ecaa55ca81f346b614f436c44fb69821f09763886a9187797c8b199e83f40L150-L170) [[2]](diffhunk://#diff-d42ecaa55ca81f346b614f436c44fb69821f09763886a9187797c8b199e83f40L183-R164) [[3]](diffhunk://#diff-d42ecaa55ca81f346b614f436c44fb69821f09763886a9187797c8b199e83f40L197-R178) [[4]](diffhunk://#diff-d42ecaa55ca81f346b614f436c44fb69821f09763886a9187797c8b199e83f40L219)

### DIFF
--- a/helm-chart/sefaria/templates/rollout/task.yaml
+++ b/helm-chart/sefaria/templates/rollout/task.yaml
@@ -51,15 +51,18 @@ spec:
                 values:
                 - mongo
             topologyKey: kubernetes.io/hostname
-          - labelSelector:
-              matchExpressions:
-              - key: app
-                operator: In
-                values:
-                - "tasks-{{ .Values.deployEnv }}"
-            topologyKey: kubernetes.io/hostname
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - "tasks-{{ .Values.deployEnv }}"
+              topologyKey: kubernetes.io/hostname
         nodeAffinity:
-          {{- include "sefaria.nodeAffinities" . | nindent 10 }}
+          {{- merge ( (include "sefaria.nodeAffinities" . ) | fromYaml) .Values.tasks.nodeAffinity | toYaml | nindent 10 }}
       containers:
       - name: tasks
         image: "{{ .Values.web.containerImage.imageRegistry }}:{{ .Values.web.containerImage.tag }}"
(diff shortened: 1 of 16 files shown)


sefaria/model/marked_up_text_chunk.py
```diff
@@ -2,10 +2,19 @@
...
 from html import escape
+from enum import Enum
+from abc import ABC, abstractmethod
+from bisect import bisect_right
 import structlog
 logger = structlog.get_logger(__name__)
 
 
+class MUTCSpanType(Enum):
+    QUOTE = "quote"
```
yonadavGit, line 13: may good to add a small comment explaining difference between quote and citation

sefaria/model/marked_up_text_chunk.py
```diff
@@ -43,17 +52,17 @@ class MarkedUpTextChunk(AbstractMongoRecord):
                     "text": {"type": "string", "required": True},
                     "type": {
                         "type": "string",
-                        "allowed": ["quote", "citation"],
+                        "allowed": [x.value for x in MUTCSpanType],
```
yonadavGit, line 55: wow! didn't know you could loop this way over fields of an unisntantiated object!

sefaria/model/marked_up_text_chunk.py
```diff
@@ -73,22 +82,55 @@ def _validate(self):
 
 
         for span in self.spans:
+            if span['type'] == MUTCSpanType.CITATION.value and 'ref' not in span:
```
yonadavGit, line 85: maybe the different erroring behavior can also be incorporated into the MUTCSpan claess (meaning each class will implement an erroring-out behavior), and so this if statement could be removed

PR #2808, opened 2025-11-06
Adjusting button size and text

saengel:
This PR was already merged, in QA was asked to resize the dot size and center the text.

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -15119,8 +15119,8 @@ span.ref-link-color-3 {color: blue}
   height: 16px;
 }
 
-.mobileModuleSwitcher{
-  display: flex !important;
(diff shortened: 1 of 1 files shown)


static/css/s2.css
```diff
@@ -15114,6 +15114,16 @@ span.ref-link-color-3 {color: blue}
   flex-shrink: 0;
 }
 
+.mobileAccountLinks .dropdownDot {
+  width: 16px;
+  height: 16px;
+}
+
+.mobileModuleSwitcher{
+  display: flex !important;
```
YishaiGlasner: `important` is generally not recommended. try to see if there is better option

stevekaplan123: Yes I would second taht.  !important can cause weird css behavior and should be avoided

saengel: Great, just pushed a fix!

PR #2818, opened 2025-11-10
Replace `/help` buttons with a direct link

schreibersefaria:
Replaced all the uses of the endpoint `/help` with a direct link to `HELP_CENTER_URLS`
Made sure that these links will open in a new tab. 
Add the comment not to remove `/help` because it is still used in the mobile app.

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -49,7 +49,10 @@ const LoggedOutDropdown = ({module}) => {
             <InterfaceText text={{ 'en': 'New Additions', 'he': 'חידושים בארון הספרים של ספריא' }} />
           </DropdownMenuItem>
         }
-        <DropdownMenuItem url={'/help'}>
+        <DropdownMenuItem url={Sefaria._v({
+          he: Sefaria._siteSettings.HELP_CENTER_URLS.HE,
+          en: Sefaria._siteSettings.HELP_CENTER_URLS.EN_US
+        })} newTab={true}>
           <InterfaceText text={{ 'en': 'Help', 'he': 'עזרה' }} />
         </DropdownMenuItem>
       </div>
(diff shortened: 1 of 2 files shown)


static/js/Header.jsx
```diff
@@ -49,7 +49,10 @@ const LoggedOutDropdown = ({module}) => {
             <InterfaceText text={{ 'en': 'New Additions', 'he': 'חידושים בארון הספרים של ספריא' }} />
           </DropdownMenuItem>
         }
-        <DropdownMenuItem url={'/help'}>
+        <DropdownMenuItem url={Sefaria._v({
```
saengel, line 52: Can you explain the `._v()` use here? What is it, and why are we doing this instead of directly referencing the link?

static/js/Header.jsx
```diff
@@ -115,7 +118,10 @@ const LoggedInDropdown = ({ module }) => {
           </DropdownMenuItem>
         }
 
-        <DropdownMenuItem preventClose={true} url={'/help'}>
+        <DropdownMenuItem url={Sefaria._v({
```
saengel, line 121: Same here

static/js/NavSidebar.jsx
```diff
@@ -284,8 +284,8 @@ const Resources = () => (
 );
 
 
-const getSidebarFooterData = () => [{'he': 'אודות','en': 'About', 'url': `${Sefaria.getModuleURL(Sefaria.LIBRARY_MODULE).origin}/about`}, 
-                                    {'he': 'עזרה','en':'Help', 'url': `${Sefaria.getModuleURL(Sefaria.LIBRARY_MODULE).origin}/help`}, 
+const getSidebarFooterData = () => [{'he': 'אודות','en': 'About', 'url': `${Sefaria.getModuleURL(Sefaria.LIBRARY_MODULE).origin}/about`},
+                                    {'he': 'עזרה','en':'Help', 'url': Sefaria._v({he: Sefaria._siteSettings.HELP_CENTER_URLS.HE, en: Sefaria._siteSettings.HELP_CENTER_URLS.EN_US})},
```
saengel, line 288: And here

static/js/Header.jsx
schreibersefaria: We changed to use separate he/en links so we must deal with that because `url={<>}` expects to get a vlue.

image

static/js/Header.jsx
schreibersefaria: See comment above

static/js/NavSidebar.jsx
schreibersefaria: See comment above

PR #2821, opened 2025-11-10
Fix the skip in infinit scroll of history

schreibersefaria:
What I did:
Set initial value of `skip` to `0`

why:
The infinit scroll would skip history items 20-40.
Init history load was 20 and then load would start from 40.

The problem was that the `skip` param initiated to `20` but it was updated before the load was triggered (the change in `skip` triggers the load).
Typically you would want it to update with `+=theAmountToLoad` (in the code `size`). But in the first run that brought us to 40.


YishaiGlasner (approved):
i didn't fully understand the function, but the fix seems ok...

PR #2822, opened 2025-11-10
chore/sc-38362/add-collection-to-store-debug-info-from-linker

nsantacruz:
## Description
Adds new collection for saving debug data from linker for later analysis


sefaria/model/linker_output.py
```diff
@@ -0,0 +1,95 @@
...
+                    "text": {"type": "string", "required": True},
+                    "type": {
+                        "type": "string",
+                        "allowed": [x.value for x in MUTCSpanType],
+                        "required": True
+                    },
+                    "ref": {"type": "string", "required": False, "nullable": True},
+                    "topicSlug": {"type": "string", "required": False, "nullable": True},
+                    "categoryPath": {"type": "list", "schema": {"type": "string"}, "required": False, "nullable": True},
+                    "inputRefParts": {"type": "list", "schema": {"type": "string"}, "required": False, "nullable": True},
```
yonadavGit: it's possible to use "dictionary compreshension" to add all these keys to the schema that share the same value

sefaria/helper/linker/tasks.py
```diff
@@ -124,29 +133,107 @@ def link_segment_with_worker(linking_args_dict: dict) -> dict:
...
+    if not spans:
+        return
+    query = {
+        "ref": tref,
+        "versionTitle": version_title,
+        "language": lang,
+    }
+    existing = LinkerOutput().load(query)
+    if existing:
+        existing.delete()
```
yonadavGit: use update of the abstract mongo record instead of delete and recreate

sefaria/helper/linker/tasks.py
```diff
@@ -124,29 +133,107 @@ def link_segment_with_worker(linking_args_dict: dict) -> dict:
...
+    query["spans"] = spans
+    LinkerOutput(query).save()
+
+
+def _extract_debug_spans(doc: LinkedDoc) -> list[dict]:
+    spans = []
+    for resolved in doc.all_resolved:
+        if isinstance(resolved, ResolvedNamedEntity):
```
yonadavGit: look into the text-book ways of avoiding these isinstance checks

sefaria/helper/linker/tasks.py
```diff
@@ -124,29 +133,107 @@ def link_segment_with_worker(linking_args_dict: dict) -> dict:
...
+        "failed": resolved.resolution_failed,
+        "ambiguous": ambig or resolved.is_ambiguous,
+    }
+    if isinstance(resolved, ResolvedNamedEntity):
+        span.update({"type": MUTCSpanType.NAMED_ENTITY.value, "topicSlug": obj.slug if obj else None})
+    elif isinstance(resolved, ResolvedCategory):
+        span.update({"type": MUTCSpanType.CATEGORY.value, "categoryPath": obj.path if obj else None})
+    elif isinstance(resolved, ResolvedRef):
+        span.update({
+            "type": MUTCSpanType.CITATION.value,
```
yonadavGit: push this logic into the objects themselves

sefaria/model/linker/tests/linker_test.py
```diff
@@ -670,3 +657,21 @@ def test_map_new_indices(crrd_params):
...
+    [crrd(['@שמות', '#א', '#ב'])],  # not ambiguous
+    [crrd(["@ירושלמי", "@ברכות", "#יג ע״א"])],  # ambiguous
+])
+def test_linker_resolutions_debug(resolver_data):
+    from sefaria.helper.linker.tasks import _extract_debug_spans
+    from cerberus import Validator
+    
+    matches = get_matches_from_resolver_data(resolver_data)
+    doc = LinkedDoc("", matches, [], [])
+    cerberus_span_schema = {'spans': LinkerOutput.attr_schemas['spans']}
```
yonadavGit: use validate instead of the cerberus schema

sefaria/model/text.py
```diff
@@ -5716,8 +5701,8 @@ def _build_named_entity_resolver(lang: str):
         from .linker.named_entity_resolver import TopicMatcher, NamedEntityResolver
 
         named_entity_types_to_topics = {
-            # "PERSON": {"ontology_roots": ['people'], "single_slugs": ['god', 'the-tetragrammaton']},
-            # "GROUP": {'ontology_roots': ["group-of-people"]},
+            "PERSON": {"ontology_roots": ['people'], "single_slugs": ['god', 'the-tetragrammaton']},
+            "GROUP": {'ontology_roots': ["group-of-people"]},
```
yonadavGit: merge groups into "person"

sefaria/model/linker/ref_resolver.py
```diff
@@ -56,6 +58,40 @@ def __init__(self, raw_entity: RawRef, resolved_parts: List[RawRefPart], node, r
...
+    def is_ambiguous(self) -> bool:
+        return False
+    
+    @property
+    def raw_entity(self) -> RawRef:
+        return self._raw_entity
+    
+    def get_debug_spans(self) -> list[dict]:
+        span = self._get_base_debug_span()
+        span.update({
```
yonadavGit, line 72: this seems fine, except that .update() can overwrite existing values. its possible to merge dicts into a new one without risking overwriting existing keys:
`{**new_items, **span}`

sefaria/model/linker/ref_resolver.py
```diff
@@ -191,17 +227,32 @@ def resolution_failed(self) -> bool:
...
+        
+    @property
+    def is_ambiguous(self) -> bool:
+        return True
+    
+    @property
+    def raw_entity(self) -> RawRef:
+        return self._raw_entity
+    
+    def get_debug_spans(self) -> list[dict]:
```
yonadavGit, line 249: related to the other comment:
I can see now that the pattern is that get_debug_spans() only **adds** new keys to the output dict. I think it's possible (not sure if "preferable" though) to "hard code" this assumption into the architecture, for example by:

`making the parent get_debug_spans() accept a list of payload dicts from _get_debug_span_additionl_payloads() and map each one onto the base skeleton (e.g., [base | payload for payload in payloads]).`

sefaria/model/linker/ref_resolver.py
nsantacruz: I'm not sure I understand. What you suggested would potentially overwrite my `new_items` with the existing keys in the span. I think it's preferable to say "I know what I'm adding is good. If there's any keys that conflict, keep what I'm adding since I know what that is". What do you think?

sefaria/model/linker/ref_resolver.py
nsantacruz: I think this over complicates the arch. I prefer something simpler.

yonadavGit: You can define a get_additional_span_keys on the parent, and have the children implement it accordingly 
The get_spans() will call the above method together with base spans method to create and return the merging of these two dicts

PR #2823, opened 2025-11-10
Added sheets back to urls_sheets and to the import in ReaderApp

schreibersefaria:
Added SheetsLandingPage to imports and exports
Added /sheets route to Voices module `urls_sheets.py`. Routes /sheets to serve_static with the sheets template

### DIFF
--- a/sefaria/urls_sheets.py
+++ b/sefaria/urls_sheets.py
@@ -20,7 +20,7 @@ urlpatterns = [
     url(r'^collections/(?P<slug>[^.]+)/settings$', reader_views.edit_collection_page),
     url(r'^collections/(?P<slug>[^.]+)$', reader_views.collection_page),
 
-    url(r'^sheets/?$', reader_views.serve_static, {'page': 'sheets'}, name='sheets'),
+    url(r'^getstarted/?$', reader_views.serve_static, {'page': 'sheets'}, name='sheets'),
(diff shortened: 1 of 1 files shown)


sefaria/urls_sheets.py
```diff
@@ -20,16 +20,15 @@
     url(r'^collections/(?P<slug>[^.]+)/settings$', reader_views.edit_collection_page),
     url(r'^collections/(?P<slug>[^.]+)$', reader_views.collection_page),
 
-    url(r'^sheets/?$', sheets_views.sheets_home_page, name='sheets'),
+    url(r'^sheets/?$', reader_views.serve_static, {'page': 'sheets'}, name='sheets'),
```
stevekaplan123: This seems fine, but I'm not sure.  I'm curious @yitzhakc what you think.  Normally we add static pages to the static_pages list, but maybe we should do it this way since this is voices specific.   Seems like static_pages at the very least should somehow differentiate between voices and library static pages.

schreibersefaria: Yes that was the problem with `static_pages`, it appears in both modules

schreibersefaria: @yitzhakc tagging you again just in case you missed it

yitzhakc: I'm fine with this.

PR #2829, opened 2025-11-11
Put defined list of auto complete types for library

schreibersefaria:
Changed the autocomplete in library to have a fixed list of types of searches.
It was undefined before so it defaulted to all.

### DIFF
--- a/static/js/HeaderAutocomplete.jsx
+++ b/static/js/HeaderAutocomplete.jsx
@@ -28,11 +28,6 @@ const type_title_map = {
   "User": "Users"
 };
 
-const MODULE_AUTOCOMPLETE_TYPES = {
-  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Collection', 'Term'],
-  [Sefaria.VOICES_MODULE]: ['Topic', 'User', 'Collection']
-};
-
 function type_icon(itemType, itemPic) {
     if (itemType === "User" && itemPic !== "") {
       return itemPic;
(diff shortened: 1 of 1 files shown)


static/js/HeaderAutocomplete.jsx
```diff
@@ -365,7 +365,7 @@ export const HeaderAutocomplete = ({onRefClick, showSearch, openTopic, openURL,
           return[];
         }
         try {
-        let types = Sefaria.activeModule === Sefaria.VOICES_MODULE ? ['Topic', 'User', 'Collection'] : undefined;
+        let types = Sefaria.activeModule === Sefaria.VOICES_MODULE ? ['Topic', 'User', 'Collection'] : ['Topic', 'ref', 'TocCategory', 'Collection', 'Term'];
```
YishaiGlasner: maybe we should have an object with a meaningful name that its keys are module names and values are these arrays.

static/js/HeaderAutocomplete.jsx
```diff
@@ -28,6 +28,11 @@ const type_title_map = {
   "User": "Users"
 };
 
+const MODULE_AUTOCOMPLETE_TYPES = {
```
YishaiGlasner: this is only for header autocomplete, so i think it should be in `HeaderAutocomplete`.
it's cinfusing, for in other places we need different types (including 'ref' in one place in voices.

schreibersefaria: You mean inside the component?
Not sure I understood your second sentence

YishaiGlasner: yes, that what i meant.
second sentence just saying why it can't be global - because it's uniqute to the header's autocomplete.

schreibersefaria: got it, thanks
done

PR #2832, opened 2025-11-13
Rework module switcher tests to account for new logic. 

saengel:
## Description
Reworking the tests to accommodate for the new logic. 
1. In conjunction with @yitzhakc - decided not to move to testing the cookie set on the subdomain with module switching, and remove module switching tests from the code
2. 4 language switching tests for the 3 environments, cauldron, staging and prod. These test that the middleware doesn't redirect but the language cookie is retained. 

## Note
All tests are currently passing.

### DIFF
--- a/sefaria/system/tests/test_language_module_switching.py
+++ b/sefaria/system/tests/test_language_module_switching.py
@@ -1,5 +1,20 @@
 """
 Tests for language and module switching middleware and views.
+
+Cookie Strategy:
+- Cookies are set on shared domain suffix (e.g., .sefaria.org, .sef-stage.org)
+- Language switches (cross-domain) use ?set-language-cookie parameter
+- Module switches (same language) inherit cookies automatically without parameter
+
+Test Coverage:
+- Staging/Production: Test LANGUAGE switching scenarios (English ↔ Hebrew) across
+  different domains, verifying redirects with ?set-language-cookie parameter
+- Testing environment: Test that NO redirects occur on non-language-pinned domains (same domain
+  for both languages). Language preference is handled purely via cookies without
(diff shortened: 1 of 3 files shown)


sefaria/system/tests/test_language_module_switching.py
```diff
@@ -1,5 +1,21 @@
 """
 Tests for language and module switching middleware and views.
+
+Cookie Strategy (as of commit a26e7846b):
```
YishaiGlasner: why do we need commit hash in our code?

saengel: Removing now :)

saengel:
> I didn't read line by line but overall seems ok. 2 notes:
> 
> 1. comments should not include history. no one will understand what do you mwan by "no longer", and usually references to other commits should be avoid.
> 2. our cauldron are not public. this code makes the url of staging public, and gives hints for finding cauldrons. i'm not sure this is aweful, but it should be re-considered.

Very very good points, I apologize for the oversight on my end. 

1. Agree, my bad, should be fixed now. 
2. Great catch - replaced with `testing env` throughout the code and comments, with a note to us that this a mock up to refer to our testing env. What do you think? Have any better ideas?

sefaria/system/tests/test_language_module_switching.py
```diff
@@ -12,6 +27,18 @@
...
+    },
+    "he": {
+        "library": "https://www.modularization.testing.sefaria.org",
+        "voices": "https://voices.modularization.testing.sefaria.org"
+    }
+}
+
 STAGING_CONFIG = {
     "en": {
         "library": "https://www.sefariastaging.org",
```
YishaiGlasner: my note was also on that, akthough it's not from this pr

PR #2833, opened 2025-11-13
Chore/sc 38487/fix the text display options keyboard

schreibersefaria:
## Description
Fixed the behaviour of the Text Display Options when navigating with a keyboard
Added localised aria-labels

## Code Changes
Kept the focus on the menu even after a change in the page caused a re-render
Prevent close on different actions that could close the menu
added localisation strings
Updated the color in the font size changing icons

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,3 +147,6 @@ dump/*
 /playwright/.cache/
 /playwright-profile/
 .env
+
+*storybook.log
+*storybook-static/
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,146 @@
+import { mergeConfig } from "vite";
+import path, { dirname, join } from "path";
+import { fileURLToPath } from "url";
+import { createRequire } from "module";
+
+const require = createRequire(import.meta.url);
+const configDirname = path.dirname(fileURLToPath(import.meta.url));
+
+const getAbsolutePath = (value) => {
+  try {
+    return dirname(require.resolve(join(value, "package.json")));
+  } catch (error) {
+    console.warn(
+      `[storybook] Could not resolve package "${value}", falling back to direct import.`,
+    );
+    return value;
+  }
+};
+
+/** @type { import('@storybook/react-vite').StorybookConfig } */
+const config = {
+  stories: [
+    "../stories/**/*.stories.@(js|jsx|mjs|ts|tsx)",
+  ],
+
+  addons: [
+    getAbsolutePath("@storybook/addon-essentials"),
+    getAbsolutePath("@storybook/addon-docs"),
+    getAbsolutePath("@storybook/addon-interactions"),
+  ],
+
+  framework: {
+    name: getAbsolutePath("@storybook/react-vite"),
+    options: {},
+  },
+
+  typescript: {
+    reactDocgen: false,
+  },
+
+  async viteFinal(baseConfig) {
+    return mergeConfig(baseConfig, {
+      plugins: [
+        {
+          name: "mock-sefaria-module",
+          enforce: "pre",
+          resolveId(source) {
+            if (/\/sefaria(\.js)?$/.test(source)) {
+              return path.resolve(configDirname, "./mocks/sefaria.js");
+            }
+            return null;
+          },
+        },
+        {
+          name: "mock-sefaria-jquery",
+          enforce: "pre",
+          resolveId(source) {
+            if (source.includes("sefaria/sefariaJquery")) {
+              return source;
+            }
+            return null;
+          },
+          load(id) {
+            if (id.includes("sefaria/sefariaJquery")) {
+              return `
+                const jQueryModule = await import("jquery");
+                const jQuery = jQueryModule.default ?? jQueryModule;
+
+                if (typeof window !== "undefined") {
+                  window.$ = jQuery;
+                  window.jQuery = jQuery;
+                  await import("jquery.cookie");
+                  await import("jquery-ui");
+                  await import("jquery.scrollto");
+                }
+
+                export default jQuery;
+              `;
+            }
+            return null;
+          },
+        },
+        {
+          name: "context-jsx-loader",
+          enforce: "pre",
+          async transform(code, id) {
+            if (id.endsWith("/static/js/context.js")) {
+              const esbuild = await import("esbuild");
+              const result = await esbuild.transform(code, {
+                loader: "jsx",
+                sourcemap: true,
+              });
+              return {
+                code: result.code,
+                map: result.map,
+              };
+            }
+            return null;
+          },
+        },
+      ],
+      resolve: {
+        alias: [
+          {
+            find: "@static",
+            replacement: path.resolve(configDirname, "../static"),
+          },
+          {
+            find: /static\/js\/sefaria\/sefaria(?:\.js)?$/,
+            replacement: path.resolve(configDirname, "./mocks/sefaria.js"),
+          },
+          {
+            find: "jquery",
+            replacement: "jquery/dist/jquery.js",
+          },
+        ],
+        extensions: [
+          ".mjs",
+          ".js",
+          ".jsx",
+          ".json",
+          ".ts",
+          ".tsx",
+          ".cjs",
+        ],
+      },
+      optimizeDeps: {
+        esbuildOptions: {
+          loader: {
+            ".js": "jsx",
+          },
+        },
+      },
+      esbuild: {
+        loader: "jsx",
+        include: [
+          /\.storybook\/.*\.[jt]sx?$/,
+          /stories\/.*\.[jt]sx?$/,
+          /static\/js\/.*\.[jt]sx?$/,
+        ],
+      },
+    });
+  }
+};
+
+export default config;
--- a/.storybook/mocks/sefaria.js
+++ b/.storybook/mocks/sefaria.js
@@ -1,0 +1,7 @@
+import sefariaStub from "../sefariaStub.js";
+
+if (typeof globalThis !== "undefined") {
+  globalThis.Sefaria = globalThis.Sefaria || sefariaStub;
+}
+
+export default globalThis?.Sefaria ?? sefariaStub;
--- a/.storybook/preview.jsx
+++ b/.storybook/preview.jsx
@@ -1,0 +1,71 @@
+import React from "react";
+import "@static/css/common.css";
+import "@static/css/s2.css";
+import {
+  ReaderPanelContext,
+  AdContext,
+  StrapiDataContext,
+} from "@static/js/context";
+import sefariaStub from "./sefariaStub";
+
+if (typeof globalThis !== "undefined") {
+  globalThis.Sefaria = sefariaStub;
+  globalThis.sa_event = globalThis.sa_event || (() => {});
+  globalThis.gtag = globalThis.gtag || (() => {});
+}
+
+const mockStrapiValue = {
+  dataFromStrapiHasBeenReceived: true,
+  strapiData: {},
+  modal: null,
+  banner: null,
+};
+
+const mockAdContext = {
+  feature_name: "storybook",
+};
+
+export const globalTypes = {
+  interfaceLang: {
+    name: "Interface language",
+    description: "Toggle between English and Hebrew interface strings",
+    defaultValue: "english",
+    toolbar: {
+      icon: "globe",
+      items: [
+        { value: "english", title: "English" },
+        { value: "hebrew", title: "עברית" },
+      ],
+    },
+  },
+};
+
+/** @type { import('@storybook/react-vite').Preview } */
+const preview = {
+  parameters: {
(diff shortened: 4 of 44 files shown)


static/js/SourceTranslationsButtons.jsx
```diff
@@ -1,7 +1,8 @@
-import React, {useContext} from "react";
```
YishaiGlasner, line 1: why? and how it worked for you locally without it?

static/js/sefaria/strings.js
```diff
@@ -320,6 +320,12 @@ const Strings = {
     "Translations": "תרגומים",
     "Uncategorized": "לא מסווג",
     "Text display options": "אפשרויות תצוגת טקסט",
+    "Source-translation toggle": "מעבר בין מקור לתרגום",
```
YishaiGlasner, line 323: 1. i think we should have all the accessibility strings in one place
2. i don't think engineers should generate those strings (and i'm not sure that musculine הצג the wa we do things)

static/js/sefaria/sefaria.js
```diff
@@ -3458,6 +3458,11 @@ _media: {},
           return inputStr;
       }
   },
+  /**
+   * Translates English strings to current interface language.
+   * Add translations to strings.js.
+   * For JSX, use <InterfaceText> which calls this automatically.
```
YishaiGlasner: use this for jsx only for interface text!

static/js/common/ToggleSwitch.jsx
```diff
@@ -1,9 +1,8 @@
-import React from "react";
 import PropTypes from "prop-types";
 
 function ToggleSwitch({name, disabled, onChange, isChecked}) {
     return (
-        <div className="toggle-switch-container">
+        <div className="toggle-switch-container" data-prevent-close="true">
```
YishaiGlasner: why? did product as to change it?

static/js/common/ToggleSwitchLine.jsx
```diff
@@ -5,7 +5,7 @@ import {InterfaceText} from "../Misc";
 
 function ToggleSwitchLine({name, onChange, isChecked, text, disabled=false}) {
     return (
-        <div className={`toggle-switch-line ${disabled ? 'disabled' : ''}`}>
+        <div className={`toggle-switch-line ${disabled ? 'disabled' : ''}`} data-prevent-close="true">
```
YishaiGlasner: why? did product as to change it?

static/js/common/ToggleSwitch.jsx
```diff
@@ -1,9 +1,8 @@
-import React from "react";
```
YishaiGlasner, line 1: ?

static/js/common/RadioButton.jsx
```diff
@@ -17,12 +20,23 @@ export default function RadioButton  ({isActive, onClick, value, name, label, id
...
+    useEffect(() => {
+        if (isActive && !wasActiveRef.current) {
+            inputRef.current?.focus();
+        }
+        wasActiveRef.current = isActive;
+    }, [isActive]);
+
     return (
-        <div className='button'>
+        <div className='button' data-prevent-close="true">
```
YishaiGlasner: why? did product as to change it?

static/js/common/RadioButton.jsx
```diff
@@ -17,12 +20,23 @@ export default function RadioButton  ({isActive, onClick, value, name, label, id
         });
     };
 
+    // Text menu re-renders swap out the radio inputs; when that happens we
+    // explicitly restore focus to the newly active option so keyboard users
+    // don't get bumped back into the document body (which React would do by default).
```
YishaiGlasner: focus should not be lost by re-render (only by unmount and mount), so it seems the solution should be fixing the unmounting rather than adding all of this logic.

static/js/SourceTranslationsButtons.jsx
schreibersefaria: The import was greyed out (IDE telling me it's not used) so I removed it.
Should this not work?

YishaiGlasner: it should work for React 17 or later, but we have 16 where it should be necessary to explicit React where you have jsx.
i can understand that ide think we have 17 but not how it worked.

static/js/sefaria/strings.js
schreibersefaria: 1. Every image or icon in the site should have a text so I would think that putting them with the relevant functionality may be easier for reviewing and editing (`Reader Panel` in this case)
2. I agree but Yitzhak asked me to just do it last time I was working on accessibility so it doesn't take for ever. I think you are correct about הצג, I fixed that.

static/js/SourceTranslationsButtons.jsx
schreibersefaria: Very Interesting, I now returned it.
I don't know how it worked
This is what Codex says:
Webpack already compiles our JSX with the automatic runtime (node/webpack.config.js (line 61)), so Babel injects import {jsx} from 'react/jsx-runtime' behind the scenes whenever it sees JSX. Because npm has actually resolved react to 16.14.0 (see node_modules/react/package.json (line 27)), that react/jsx-runtime entry point exists even though package.json still lists ^16.8.6. So the components kept working without an explicit import React—the generated code never referenced a React identifier.

static/js/sefaria/sefaria.js
schreibersefaria: Not sure I understand you comment. How should `._` be used?

static/js/common/RadioButton.jsx
schreibersefaria: In practice we have a unmount because the content of the menue is rebuilt on changes.
Fixing that is a big refactor

static/js/common/RadioButton.jsx
schreibersefaria: Without this a keyboard user couldn't use the source/translation toggle.
Everytime you would use up/down arrow key the menu would close

static/js/common/ToggleSwitch.jsx
schreibersefaria: returned

static/js/common/ToggleSwitch.jsx
schreibersefaria: Same reason as before just for toggling switches.
In practice I see that toggling switches isn't causing a close in practice so for now I removed this.

schreibersefaria: OK so this wasn't exactly a duplicate but it did become redundant when I added it to ToggleSwitchLine.
I'll double check these with product. Unlike in `RadioButton` these could be seen as not a must for accessibility, I guess

static/js/common/ToggleSwitchLine.jsx
schreibersefaria: See comment in `ToggleSwitch`

schreibersefaria: I removed the change from there at the end because the line covers the switch.
I'm double checking with Mickey.
I added these for the same logic of RadioButton even if here they aren't as %100 necessary.

It's very common for accessibility users to click on a label of a toggle so I added this to the Line

static/js/common/RadioButton.jsx
YishaiGlasner: they can. when they are on source/translation option, up/down arrow changes this.
anyway, i think we can't change behaviour without approving of product team - they also know their are costs, but it's their decision.

static/js/sefaria/sefaria.js
YishaiGlasner: i'm saying that the comment "For JSX, use <InterfaceText>" is not exact - one should use the component only for interface text. if he wants something that is not interdace he should find another solution (this not theoretical, it did happen)

static/js/common/RadioButton.jsx
YishaiGlasner: why it's so big refactor?
this is kind of not small technical debt for not fixing a bug.

static/js/sefaria/sefaria.js
schreibersefaria: Ahh got it, thanks
So would this be more accurate:
`   * For displaying interface text you should use <InterfaceText> which calls this function automatically.`

static/js/common/RadioButton.jsx
schreibersefaria: Good point. Waiting for final confirmation from Pnina.
Up/down also closes the menu.

static/js/common/RadioButton.jsx
schreibersefaria: I'm not very familiar with the code, but what are you recommending?
Should I change the menu to always have all the sections and just hide them from being visible so the DOM structure doesn't change?

static/js/common/ToggleSwitchLine.jsx
schreibersefaria: Confirmed with product

static/js/common/RadioButton.jsx
schreibersefaria: Confirmed with product

schreibersefaria: Moved this to the whole menu instead of specific components

static/js/common/RadioButton.jsx
yitzhakc: @YishaiGlasner happy to hear your recommendation, and then we can decide if it's in the scope of this PR or not.
It seems to me that this issue pre-existed this story (we have an intentional [`Util.focusFirstElement`](https://github.com/Sefaria/Sefaria-Project/blob/28916ed1dedff79e3ed57a9a086127bf41f941ad/static/js/ReaderDisplayOptionsMenu.jsx#L102) in the code before this PR) and Daniel was just adding a comment explaining the reasoning.
So I'd want us to first understand the complexity cost of rewriting the way the menu is rendered before we decide to include it in the scope of this PR.

yitzhakc (changes requested):

static/js/common/RadioButton.jsx
YishaiGlasner: @yitzhakc my comment isn't about the doc string. i'm saying that we have a bug regarding the focus, and that this pr doesn't fix it, but patch it.

schreibersefaria: Yishai helped me find the bug. The problem was that the `TextColumn` was stealing the focus because of existing code (`scrollToHighlighted()`). So I'm now working to add the info about if the menu is open to the context and then add an exception to the focus stealing.

static/js/TextColumn.jsx
```diff
@@ -77,7 +77,7 @@ class TextColumn extends Component {
     } else if ((this.props.settings.language !== prevProps.settings.language) ||
         !Sefaria.areBothVersionsEqual(prevProps.currVersions, this.props.currVersions)) {
       // When the content the changes but we are anchored on a line, scroll to it
-      this.scrollToHighlighted();
+      this.scrollToHighlighted(false); // Don't change focus to not steal the focus from the menu where the text is changes from (e.g. ReaderDisplayOptionsMenu)
```
YishaiGlasner: // or from the other places that changed the text - translations box or about box

static/js/common/RadioButton.jsx
```diff
@@ -17,12 +20,23 @@ export default function RadioButton  ({isActive, onClick, value, name, label, id
         });
     };
 
+    // Text menu re-renders swap out the radio inputs; when that happens we
```
YishaiGlasner: remove

static/js/common/RadioButton.jsx
```diff
@@ -17,12 +20,23 @@ export default function RadioButton  ({isActive, onClick, value, name, label, id
...
+    //     wasActiveRef.current = isActive;
+    // }, [isActive]);
+
     return (
         <div className='button'>
             <label htmlFor={id}>
                 <InterfaceText>{label}</InterfaceText>
             </label>
             <input
+                // ref={inputRef}
```
YishaiGlasner: remove

static/js/common/RadioButton.jsx
```diff
@@ -1,9 +1,12 @@
-import React from "react";
+import React, {useRef, useEffect} from "react";
```
YishaiGlasner: isn't needed anymore

static/js/common/RadioButton.jsx
```diff
@@ -1,9 +1,12 @@
...
 import PropTypes from "prop-types";
 import {InterfaceText} from "../Misc";
 import Util from "../sefaria/util";
 
 export default function RadioButton  ({isActive, onClick, value, name, label, id, onKeyDown, ...rest}) {
+    // const inputRef = useRef(null);
```
YishaiGlasner: remove

schreibersefaria: Very embarrassing.

PR #2839, opened 2025-11-16
Speed up itag stripping2

nsantacruz:
## Description
Significantly speeds up itag stripping by moving from BeautifulSoup to an iterative process. This has a side benefit that it avoids issues where bs4 normalizes tags thereby making it hard to match text from before the itags were stripped.

## Code Changes
This pull request refactors the logic for normalizing and removing footnotes and related HTML tags across the codebase, replacing the previous `ITagNormalizer` implementation with a new, more robust `FootnoteNormalizer`. It updates the normalization pipeline, test cases, and the handling of footnotes in text processing and search. The changes improve accuracy and maintainability by using regular expressions for footnote detection and simplifying the removal of nested and variant footnote markers.

**Footnote and tag normalization improvements:**

* Replaced the old `ITagNormalizer` class with a new `FootnoteNormalizer` that uses regular expressions to identify and remove footnotes, handling nested footnotes and variants more accurately. (`sefaria/helper/normalization.py`, [sefaria/helper/normalization.pyL152-L218](diffhunk://#diff-b7ca68aa09633ed3137198aaea72e7cd4453c795e2ac98c936842133266d2786L152-L218))
* Updated the `NormalizerFactory` to use `FootnoteNormalizer` for footnotes and added normalization for other itag types and footnote marker variants. (`sefaria/helper/normalization.py`, [sefaria/helper/normalization.pyL369-R345](diffhunk://#diff-b7ca68aa09633ed3137198aaea72e7cd4453c795e2ac98c936842133266d2786L369-R345))

**Testing enhancements:**

* Refactored and expanded test cases to cover various footnote marker formats, nested footnotes, and the new normalization logic, ensuring correctness of the new approach. (`sefaria/helper/tests/normalization_tests.py`, [[1]](diffhunk://#diff-a09094ea08fbae28b1d138f5ed6ecc05fe424d74e011fca49fba5002837e62bbR5-R36) [[2]](diffhunk://#diff-a09094ea08fbae28b1d138f5ed6ecc05fe424d74e011fca49fba5002837e62bbL113-R135)
* Updated chunk test data to use new footnote marker formats and added cases for obscure and other itag types. (`sefaria/model/tests/chunk_test.py`, [[1]](diffhunk://#diff-c96ebee6b8010235b1d4b9ff122fe86a78e7af4706a6039e75910c8d5d0dd956L502-R507) [[2]](diffhunk://#diff-c96ebee6b8010235b1d4b9ff122fe86a78e7af4706a6039e75910c8d5d0dd956L475)

**Integration with text processing and search:**

* Refactored footnote extraction and tag stripping in `AbstractTextRecord` to use the new normalization logic, removing legacy BeautifulSoup-based code. (`sefaria/model/text.py`, [[1]](diffhunk://#diff-f40f2f9df8b67d126ae6fc7a1df82f68b3f50722d5f9fcc2273f86d4e37fddc4L1218-R1225) [[2]](diffhunk://#diff-f40f2f9df8b67d126ae6fc7a1df82f68b3f50722d5f9fcc2273f86d4e37fddc4L1248-R1240) [[3]](diffhunk://#diff-f40f2f9df8b67d126ae6fc7a1df82f68b3f50722d5f9fcc2273f86d4e37fddc4R35)
* Updated the search module to use the new footnote extraction method and improved how footnote content is appended to the main text. (`sefaria/search.py`, [sefaria/search.pyL596-R604](diffhunk://#diff-50caef9b129793c3a598628b4d78ab2a4aec68581c6eef5f5c26dcbeb0983e16L596-R604))

### DIFF
--- a/sefaria/helper/normalization.py
+++ b/sefaria/helper/normalization.py
@@ -169,16 +169,23 @@ class FootnoteNormalizer(AbstractNormalizer):
             i = i_open_end
             while depth > 0:
                 next_open = s.find('<i', i)
-                next_close = s.find('</i>', i)
-                if next_close == -1:
-                    # malformed HTML, bail out
+                next_close = s.find('</i', i)
+                
+                # determine where the </i> tag ends in case it's slightly malformed
+                next_close_end = -1
+                if s[next_close+3] == '>':
+                    next_close_end = next_close + 4
+                elif s[next_close+4] == '>':
+                    next_close_end = next_close + 5
+                if next_close == -1 or next_close_end == -1:
+                    # no </i> to match <i>, bail out
                     break
-                if next_open != -1 and next_open < next_close:
+                if next_open != -1 and next_open < next_close and s[next_open+2] in (' ', '>'):
                     depth += 1
                     i = next_open + 2
                 else:
                     depth -= 1
-                    i = next_close + 4
+                    i = next_close_end
             end_pos = i
             is_subset = False
             for (start, end), _ in matches:


sefaria/helper/normalization.py
```diff
@@ -149,73 +149,47 @@ def norm_to_unnorm_indices_with_mapping(normalized_indices, removal_map, subst_e
...
+            i = i_open_end
+            while depth > 0:
+                next_open = s.find('<i', i)
+                next_close = s.find('</i>', i)
+                if next_close == -1:
+                    # malformed HTML, bail out
+                    break
+                if next_open != -1 and next_open < next_close:
+                    depth += 1
+                    i = next_open + 2
```
akiva10b, line 185: `next_open + 2` and `next_open + 4` rely heavily on the HTML being correct, so the scenario of malformed itag like `</ i>` would fail in this scenario, but would be caught by bs which is being replaced.
I am not sure exactly what to recommend. So I will leave it to you:
1. You can run another find for the last ">" to bullet proof the code. (you can add a if checking that next_open + 4 = ">" and only run it when it isn't. You can also skip the find and run a if on place +4 and +5 which are the only legitimate ones)
2. you can ignore but leave a comment regarding the scenario

nsantacruz: I tried using regex to detect the start and end but the time went from 6ms per run to 44ms. So I took your suggestion which leads to code that's a bit harder to read but certainly more bullet proof! The speed difference from the original code is negligible as you'd expect.

PR #2851, opened 2025-11-20
fix-context-terms-from-full-node-path

nsantacruz:
## Description
When pulling term context from a context ref, we only considered the match_templates at the book level. We should add all terms from match_templates at any level.

E.g. if context ref is Arukh HaShulchan, Orach Chayim 23 and my citation is "Shulchan Arukh Seif 3", pull "Orach Chayim" as a term context even though it's lower in the node tree.

### DIFF
--- a/sefaria/model/linker/ref_resolver.py
+++ b/sefaria/model/linker/ref_resolver.py
@@ -591,7 +591,7 @@ class RefResolver:
             if len(match_templates) != 0:
                 # not clear which match_template to choose. shortest has advantage of adding minimum context to search
                 shortest_template = min(match_templates, key=lambda x: len(list(x.terms)))
-                term_contexts.extend([TermContext(term) for term in longest_template.terms])
+                term_contexts.extend([TermContext(term) for term in shortest_template.terms])
             if node.parent is None:
                 break
             node = node.parent


sefaria/model/linker/ref_resolver.py
```diff
@@ -584,11 +584,22 @@ def _get_all_term_contexts(node: schema.SchemaNode, include_root=False) -> List[
...
+            match_templates = list(node.get_match_templates())
+            if len(match_templates) != 0:
+                # not clear which match_template to choose. shortest has advantage of adding minimum context to search
+                longest_template = min(match_templates, key=lambda x: len(list(x.terms)))
+                term_contexts.extend([TermContext(term) for term in longest_template.terms])
+            if node.parent is None:
+                break
+            node = node.parent
+            recursion_depth += 1
+            if recursion_depth > 10:
```
yonadavGit, line 599: why fear an infinite loop?

sefaria/model/linker/ref_resolver.py
```diff
@@ -584,11 +584,22 @@ def _get_all_term_contexts(node: schema.SchemaNode, include_root=False) -> List[
...
-        # not clear which match_template to choose. shortest has advantage of adding minimum context to search
-        longest_template = min(match_templates, key=lambda x: len(list(x.terms)))
-        return [TermContext(term) for term in longest_template.terms]
+        term_contexts = []
+        recursion_depth = 0
+        while True:
+            match_templates = list(node.get_match_templates())
+            if len(match_templates) != 0:
+                # not clear which match_template to choose. shortest has advantage of adding minimum context to search
+                longest_template = min(match_templates, key=lambda x: len(list(x.terms)))
```
yonadavGit: is it longest or shortest?
what's the reason for picking only shortest (or only longest)?

nsantacruz:
```suggestion
                shortest_template = min(match_templates, key=lambda x: len(list(x.terms)))
```

sefaria/model/linker/ref_resolver.py
```diff
@@ -584,11 +584,22 @@ def _get_all_term_contexts(node: schema.SchemaNode, include_root=False) -> List[
...
+                term_contexts.extend([TermContext(term) for term in longest_template.terms])
```
nsantacruz:
```suggestion
                term_contexts.extend([TermContext(term) for term in shortest_template.terms])
```

PR #2853, opened 2025-11-23
feat: capture object memory snapshot

akiva10b:
This pull request adds a new admin endpoint to provide a summary of memory usage using the `pympler` library. This feature is intended to help staff members monitor memory consumption by object type in the application.

New admin memory summary endpoint:

* Added a new URL pattern `/admin/memory/summary` in `sefaria/urls.py` to route requests to the new memory summary view.
* Implemented the `memory_summary` view in `sefaria/views.py`, which returns a JSON summary of memory usage by object type. The endpoint supports an optional `limit` query parameter and handles errors such as missing dependencies or invalid parameters.

### DIFF
--- a/sefaria/views.py
+++ b/sefaria/views.py
@@ -870,6 +870,31 @@ def cache_dump(request):
 
 
 @staff_member_required
+def memory_summary(request):
+    try:
+        from pympler import muppy, summary
+    except ImportError:
+        return jsonResponse({"error": "pympler is not installed in this environment"}, status=500)
+
+    limit_param = request.GET.get("limit")
+    try:
+        limit = int(limit_param) if limit_param is not None else None
+    except ValueError:
+        return jsonResponse({"error": "limit must be an integer"}, status=400)
+
+    all_objects = muppy.get_objects()
+    summarized = summary.summarize(all_objects)
+    if limit is not None:
+        summarized = summarized[: limit]
+
+    data = [
+        {"type": type_name, "count": int(count), "size": int(size)}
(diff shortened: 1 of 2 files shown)


sefaria/views.py
```diff
@@ -869,6 +869,31 @@ def cache_dump(request):
     return jsonResponse(resp)
 
 
+@staff_member_required
+def memory_summary(request):
+    try:
+        from pympler import muppy, summary
```
nsantacruz, line 875: pympler is in requirements.txt. why would it not be installed (except on local, but it's fine for this to fail non-gracefully on local).

sefaria/views.py
```diff
@@ -869,6 +869,31 @@ def cache_dump(request):
...
+    limit_param = request.GET.get("limit")
+    try:
+        limit = int(limit_param) if limit_param is not None else None
+    except ValueError:
+        return jsonResponse({"error": "limit must be an integer"}, status=400)
+
+    all_objects = muppy.get_objects()
+    summarized = summary.summarize(all_objects)
+    if limit is not None:
+        summarized = summarized[: limit]
```
nsantacruz, line 888: remove extra space in array index

sefaria/views.py
```diff
@@ -869,6 +869,31 @@ def cache_dump(request):
     return jsonResponse(resp)
 
 
+@staff_member_required
+def memory_summary(request):
```
nsantacruz, line 873: write docs that explain the GET params and the return format.

PR #2859, opened 2025-11-24
Bug/sc 36893/configure google oauth to whitelist

stevekaplan123:
## Description

## Code Changes

## Notes


yitzhakc (approved):

YishaiGlasner:
> >    Interesting. .. Do you have a specific issue in mine where you think the cookie would cease to work? @YishaiGlasner
no, it's just that we'll have explicit domain rather than only path and i wonder if there will be side effects

PR #2864, opened 2025-11-25
feat: Updated the header CTA buttons with the expected padding

dcschreiber:
created a new variant for Button component using a new class
Moved the relevant css to the common-components

I also moved the colors vars duplicate from `s2` to `color-palette`

### DIFF
--- a/static/js/Header.jsx
+++ b/static/js/Header.jsx
@@ -719,7 +719,7 @@ const HelpButton = () => {
 
 const SignUpButton = () => {
   return (
-    <Button className="cta-button" href="/register" targetModule={Sefaria.LIBRARY_MODULE}>
+    <Button className="auto-width-button" href="/register" targetModule={Sefaria.LIBRARY_MODULE}>
       <InterfaceText>Sign Up</InterfaceText>
     </Button>
   )
@@ -728,7 +728,7 @@ const SignUpButton = () => {
 const CreateButton = () => {
 
   return (
-    <Button className="cta-button" href="/sheets/new" targetModule={Sefaria.VOICES_MODULE}>
+    <Button className="auto-width-button" href="/sheets/new" targetModule={Sefaria.VOICES_MODULE}>
       <InterfaceText text={{ 'en': 'Create', 'he': 'דף חדש' }} />
     </Button>
   );
--- a/static/css/common-components.css
+++ b/static/css/common-components.css
@@ -82,9 +82,9 @@ a.sefaria-common-button.fillwidth {
     height: 53px;
     width: 100%;
 }
-/** cta-button variant for the signup and create button in the header **/
-button.sefaria-common-button.cta-button,
-a.sefaria-common-button.cta-button {
+/** auto-width-button variant for the signup and create button in the header **/
+button.sefaria-common-button.auto-width-button,
+a.sefaria-common-button.auto-width-button {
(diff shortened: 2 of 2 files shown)


static/css/color-palette.css
```diff
@@ -7,12 +7,15 @@
...
   --beit-midrash-grey: #333333;
+  --darkest-grey: #333333;
+  --darker-grey: #575757;
   --dark-grey: #666666;
   --medium-grey: #6f6f6f;
   --medium-grey-legacy: #999999;
   --light-grey: #CCCCCC;
   --lighter-grey: #EDEDEC;
   --lightest-grey: #FBFBFA;
+  --black: #000000;
```
YishaiGlasner, line 18: there is already css `black` so we don't need var

static/css/color-palette.css
```diff
@@ -7,12 +7,15 @@
   --highlight-blue: #DDEEFF;
   --highlight-blue-light: #F0F7FF;
   --beit-midrash-grey: #333333;
+  --darkest-grey: #333333;
+  --darker-grey: #575757;
```
YishaiGlasner, line 11: please make sure with product we really want to add greys.

static/js/Header.jsx
```diff
@@ -698,7 +698,7 @@ const HelpButton = () => {
 
 const SignUpButton = () => {
   return (
-    <Button href="/register" targetModule={Sefaria.LIBRARY_MODULE}>
+    <Button className="header" href="/register" targetModule={Sefaria.LIBRARY_MODULE}>
```
YishaiGlasner: `header` is the classname of the most upper div of the header. here we need more specific classname

static/css/color-palette.css
```diff
@@ -33,6 +36,6 @@
   --reference-orange: #D4896C;
   --miscelaneous-green: #3B5849;
   --selected-option: #000000;
-  --select-blue: #0B71E7;
+  --select-blue: #0056b3;
```
saengel, line 39: Why was this value changed?

static/css/color-palette.css
dcschreiber: I just merged the two lists of colors that were almost perfect duplication so I'd rather not change what there is

static/css/color-palette.css
dcschreiber: I just merged the two lists of colors that were almost perfect duplication so I'd rather not change what there is

saengel (changes requested):
If I understand correctly, what we're trying to do here is adjust the background color of the common `<Button/>` component. 

If you read through the well documented [Button](https://github.com/Sefaria/Sefaria-Project/blob/modularization-main/static/js/common/Button.jsx) component, you'll see we were looking to avoid the need to add new classes unless necessary. 

Since the changes here seem to be ones we want across the board for a `--sheets-green` variant of the button, I'd recommend re-configuring the button component itself, versus layering on specific header CSS, as much as is possible (since the bulk of the changes aren't specific to the header as far as I can see). 

What do you think?

static/css/color-palette.css
dcschreiber: It wasn't. I just merged the two lists, and there was a conflict, so I took the value that is being used in practise.

saengel (approved):

saengel (approved):

PR #2873, opened 2025-11-26
Updated the keyboardto 18px size and updated search placeholder color

dcschreiber:
updated the keyboard CSS to 18px size
Checked all the uses of keyboard approved them with Penina.
Removed the opacity reduction for the dictionary search keyboard icon at Penina's request

Updated the search placeholder color to --dark-grey overriding Bootstrap's default
Approved with Penina that this should happen in all searches (other than the dictionary search)


static/css/header.scss
```diff
@@ -118,6 +118,14 @@ $interfaces: english hebrew;
   font-size: 18px;
 }
 
+.search::placeholder,
+.topic-landing-search-input::placeholder,
+.readerSearch::placeholder,
+.searchFiltersInput::placeholder,
+.topicFilterInput::placeholder {
```
YishaiGlasner: i think this shouldn't be here but in s2 (for now). if we have general styles for search box, here we should have only overridings

dcschreiber: done

PR #2874, opened 2025-11-26
Header refactor + breakpoint + scss

YishaiGlasner:
## Description
The task was changing the header to response to some breakpoints.
Since css vars are not working in `@media`, it was decided that we should use scss.
So this branch includes introducing of scss into our code, the fonts refactor which was the first scss POC, introducing of breakpoints and header refactor 

## Introducing SCSS
package.json & package-lock.json - adding sass, concurrently (for running the scss parsing) and husky (for pre commit hook).
.husky/pre-commit - pre commit hook to parse scss

## Fonts
fonts.scss - structured file to generaete fonts.css which is similar to the old one.

## Breakpoints
breakpoints.scss - mixin for getting `@media` blocks according to 3 modes - desktop, tablet and mobile. Also it put 2 vars into root for using them in js (so the breakpoints are defined in only one place).
sefaria.js - js function for getting the breakpoint

## Header
Header.jsx - 
1. use breakpoint for getting the mobile header.
2. remove logo img which is done by css.
3. refactor `textLinks` out of the big return statement and make it cleaner.
4. Wrap the header icons in one div.
5. fix `MobileNavMenu` to be `secondary`.
6. change ProfilePic to be 24px (Penina requested).
7. remove unused ProfilePicMenu component.
s2.css - remove all the logics of header. some of it redundant and some to header.scss.
header.scss - new file for all the css of header - but not the css for dropdown that comes from header (logged in/out menu, module switcher, search autocomplete, mobile menu).
samech-logo.svg - use one samech logo so the color will be determined by header.scss.

### DIFF
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -3592,7 +3592,14 @@ _media: {},
     DESKTOP: 'desktop',
   },
   getBreakpoint: () => {
-    // ReaderApp is re-rendered every resize, so static function will work.
+    /**
+     * Returns the current responsive breakpoint.
+     *
+     * This is a plain JS utility (not a React hook) because the top-level ReaderApp component already listens to
+     * window resize events and triggers re-renders on breakpoint changes.
+     * That means this function can reliably read the up-to-date breakpoint without needing its own hook or state.
+     */
+
     const rootStyles = getComputedStyle(document.documentElement);
     const tabletMin = parseInt(rootStyles.getPropertyValue('--bp-tablet-min'));
     const desktopMin = parseInt(rootStyles.getPropertyValue('--bp-desktop-min'));


static/js/Header.jsx
```diff
@@ -511,7 +485,8 @@ const MobileNavMenu = ({ onRefClick, showSearch, openTopic, openURL, close, visi
                   <InterfaceText>Saved & History</InterfaceText>
                 </Button>
                 <Button
-                  icon={unread ? "notifications-1_mdl" : "notifications_mdl"}
+                  variant="secondary"
+                  icon={Sefaria.notificationCount ? "notifications-1_mdl" : "notifications_mdl"}
```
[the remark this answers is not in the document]
YishaiGlasner: unread was ReferenceError. unread defined in another place in the file as `!! Sefaria.notificationCount` and someone forgot to pass it.

static/css/header.scss
```diff
@@ -0,0 +1,148 @@
...
+  ".dropdownLinks-menu",
+  ".autocomplete-dropdown",
+  ".languageToggle",
+);
+$not-selector: "";
+@each $sel in $exclude-flex {
+  $not-selector: "#{$not-selector}:not(#{$sel}):not(#{$sel} *)";
+}
+.headerInner,
+.headerInner *#{$not-selector} {
```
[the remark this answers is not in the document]
YishaiGlasner: nothing nested here. embarssing mistake of copilot, it's just chain of `not`s (which is not so readable but for that we use scss)

static/js/sefaria/sefaria.js
```diff
@@ -3586,6 +3586,21 @@ _media: {},
     const next = Sefaria.activeModule === Sefaria.VOICES_MODULE ? '' : 'texts';
     return `/logout?next=/${next}`;
   },
+  getBreakpoint: () => {
+    // ReaderApp is re-rendered every resize, so static function will work.
```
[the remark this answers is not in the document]
YishaiGlasner: the function is static, so engineers can think it shouldn't be used in jsx (and build a hook for that). the comment say it functions as a hook, for any resize cause re-render of the whole app

static/js/Header.jsx
```diff
@@ -301,30 +273,31 @@ const Header = (props) => {
...
+          }
+        </div>
       </div>
     </>
   );
 
   // Language toggle logic - show on mobile for specific menu pages
   const languageToggleMenus = ["navigation", "saved", "history", "notes"];
-  const hasLanguageToggle = mobile && Sefaria.interfaceLang !== "hebrew" && languageToggleMenus.includes(props?.firstPanel?.menuOpen);
+  const hasLanguageToggle = Sefaria.interfaceLang !== "hebrew" && languageToggleMenus.includes(props?.firstPanel?.menuOpen);
```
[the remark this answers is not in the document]
YishaiGlasner: `hasLanguageToggle` is used inly in `mobileHeaderContent`, so it won't appear in desktop. it was redundant from the first place.

static/img/samech-logo.svg
dcschreiber, line 1: Why use CSS to color the logo and not stick to the same logic as the other logo for consistency?

YishaiGlasner: i think that having one file rather than 2 is preferred. here it also means both have the same color var so the theme handles the color.
i'd it also for the big logo, but it contains more than one color.

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm run scss-compile
+exit 0
```
dcschreiber:
```suggestion
npm run scss-compile || exit 1
```

In case there is an error

static/js/sefaria/sefaria.js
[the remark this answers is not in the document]
yitzhakc: Would it be more clear if we call the function `getBreakpointhHook`?

static/js/sefaria/sefaria.js
```diff
@@ -3586,6 +3586,21 @@ _media: {},
...
+    const rootStyles = getComputedStyle(document.documentElement);
+    const tabletMin = parseInt(rootStyles.getPropertyValue('--bp-tablet-min'));
+    const desktopMin = parseInt(rootStyles.getPropertyValue('--bp-desktop-min'));
+    const width = window.innerWidth;
+
+    if (width < tabletMin) {
+      return 'mobile';
```
yitzhakc: Magic strings :-( 
Please use constants.

static/js/Header.jsx
```diff
@@ -195,7 +191,7 @@ const Header = (props) => {
     }
   }, []);
   
-  const mobile = !props?.multiPanel;
+  const mobile = Sefaria.getBreakpoint() === 'mobile';
```
yitzhakc: Use constant

package.json
```diff
@@ -23,7 +23,7 @@
     "build-sefaria": "webpack --config ./node/webpack.sefaria.js",
     "build-timeline": "webpack --config ./node/webpack.timeline.js ",
     "build-categorizeSheets": "webpack --config ./node/webpack.categorizeSheets.js ",
-    "watch": "webpack --config ./node/webpack.config.js --progress -w --watch-options-stdin",
+    "watch": "concurrently \"npm run scss\" \"webpack --config ./node/webpack.config.js --progress -w --watch-options-stdin\"",
```
yitzhakc: Would it work to use single quotes? Makes the line more readable
`"concurrently 'npm run scss' ...`

static/js/sefaria/sefaria.js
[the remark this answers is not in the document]
YishaiGlasner: i think it will be confusing for it isn't a hook and it's in pure js file

.husky/pre-commit
YishaiGlasner: done in another way

package.json
YishaiGlasner: i understand it won't work on windows.
gpt suggests:
```
"watch": "concurrently \"npm:scss\" \"npm:webpack\"",
"scss": "npm run scss",
"webpack": "webpack --config ./node/webpack.config.js --progress -w --watch-options-stdin"
```
seems reasonable to me. what do you think?

static/js/sefaria/sefaria.js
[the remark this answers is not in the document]
yitzhakc: Looking at it again, I'm thinking mentioning ReaderApp is confusing. Since it's not a hook, and not state-dependent, maybe something like this?

`// This is a pure function, not a React hook. It can be safely called in render methods or event handlers, as it always reads the latest window size.`

package.json
```diff
@@ -23,7 +23,8 @@
     "build-sefaria": "webpack --config ./node/webpack.sefaria.js",
     "build-timeline": "webpack --config ./node/webpack.timeline.js ",
     "build-categorizeSheets": "webpack --config ./node/webpack.categorizeSheets.js ",
-    "watch": "webpack --config ./node/webpack.config.js --progress -w --watch-options-stdin",
+    "webpack": "webpack --config ./node/webpack.config.js --progress -w --watch-options-stdin",
+    "watch": "concurrently \"npm:scss\" \"npm:webpack\"",
```
yitzhakc, line 27: It's definitely more readable now, but I'd really like you to see if you can use single quotes. If it doesn't work, that's fine. But if it does work, then it's more readable then escaped-double-quotes

PR #2887, opened 2025-12-01
fix(HeaderAutocomplete): Added a catch for ref searches in voices to …

dcschreiber:
Blocking searching for refs in the voices module
Code changes:
1. Moved `MODULE_AUTOCOMPLETE_TYPES` to be global and renamed to `MODULE_ALLOWED_SEARCH_TYPES`
2. Added a filter inside `getQueryObj` to only return for `is_ref` if `ref` is inside the active module allowed search list

### DIFF
--- a/static/js/HeaderAutocomplete.jsx
+++ b/static/js/HeaderAutocomplete.jsx
@@ -29,7 +29,7 @@ const type_title_map = {
 };
 
 const MODULE_ALLOWED_SEARCH_TYPES = {
-  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Collection', 'Term'],
+  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Term'],
   [Sefaria.VOICES_MODULE]: ['Topic', 'User', 'Collection']
 };
 


static/js/HeaderAutocomplete.jsx
```diff
@@ -111,7 +116,8 @@ const getQueryObj = (query) => {
         return getQueryObj(repairedQuery);
       }
 
-      if (d["is_ref"]) {
+      const allowedTypes = MODULE_ALLOWED_SEARCH_TYPES[Sefaria.activeModule];
```
stevekaplan123, line 119: Something seems off with this approach.  Why is 'ref' the only one in the list that we have to specifically check for? @dcschreiber

YishaiGlasner (approved):

static/js/HeaderAutocomplete.jsx
dcschreiber: I agree it seams a bit off.
I didn't want to do a big refactor and this seamed like the only change we needed.

It seems like the behaviour we need to block is unique to `ref`s (I may be wrong about this) and other types in this function are allowed on all modules. At first I just hardcoded an exception for voices but thought working with `allowedTypes` will avoid breaks on if we add modules or start allowing `ref` searches on voices.

The type name in the query result (`d`) and in our list is not the same so I had to do this explicit check.

static/js/HeaderAutocomplete.jsx
```diff
@@ -28,6 +28,11 @@ const type_title_map = {
   "User": "Users"
 };
 
+const MODULE_ALLOWED_SEARCH_TYPES = {
+  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Collection', 'Term'],
```
stevekaplan123: remove Collection

dcschreiber:
```suggestion
  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Term'],
```

stevekaplan123:
```suggestion
  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Term'],
```

stevekaplan123 (approved):

PR #2888, opened 2025-12-01
Feature/sc 38607/show message to returning users about changes

yodem:
## Description

Refactors Module Switcher popover styling for mobile responsiveness and accessibility. Adds 44x44px touch targets, fixes typography sizing, sets a fixed mobile width, and simplifies structure by removing wrapper divs.

### `static/css/popover.scss`
- Fixed mobile popover width to 264px
- Set desktop "Learn More" button font-size to 14px
- Added mobile font-size 16px to popover content
- Removed `popover-button-wrapper` styles
- Added `.accessible-touch-target` class with `::before` pseudo-element for 44x44px mobile touch targets
- Fixed mobile button height (removed problematic `min-height`)

### `static/js/ModuleSwitcherPopover.jsx`
- Removed `popover-button-wrapper` divs
- Added `learn-more` and `accessible-touch-target` classes directly to buttons
- Updated content text

### `static/js/common/Popover.jsx`
- Added `popover-button` class to `PopoverClose` component
- Updated className merging logic

## Notes

- Accessibility: 44x44px touch targets meet WCAG guidelines via invisible `::before` pseudo-elements
- Mobile UX: Fixed width prevents edge collisions; height grows naturally for text
- Code simplification: Removed wrapper divs for cleaner structure


static/css/popover.scss
```diff
@@ -14,6 +14,11 @@ $popover-text: white;
   box-shadow: 0px 1px 3px rgba(0, 0, 0, 0.25);
   z-index: 9999;
   max-width: 350px;
+
+  @include breakpoint(mobile) {
+    width: 264px;
+    max-width: 264px;
```
YishaiGlasner: isn't this redundant?

static/css/popover.scss
```diff
@@ -126,9 +161,10 @@ $popover-text: white;
   }
 
   @include breakpoint(mobile) {
-    min-height: 44px;
-    font-size: 16px;
-    padding: 8px 12px;
+    font-size: 16px; // English System Body / Hebrew System Body (16px)
+    padding: 8px 12px; // 8px top, 12px right, 8px bottom, 12px left
+    height: auto;
+    min-height: auto; // Height controlled by container
```
YishaiGlasner: this seems to do nothing

static/css/popover.scss
```diff
@@ -23,12 +28,15 @@ $popover-text: white;
...
   color: $popover-text;
   border-radius: 8px;
-  font-size: 14px;
+  font-size: 14px;  
   display: flex;
   flex-direction: column;
   align-items: flex-start;
   justify-content: flex-start;
 
+  @include breakpoint(mobile) {
```
YishaiGlasner: unite all the `breakpoint(mobile)`

static/css/popover.scss
```diff
@@ -76,30 +84,57 @@ $popover-text: white;
...
-    > a,
-    > button {
-      min-height: 44px;
-      height: auto;
-      font-size: 16px;
-      padding: 8px 12px;
+    align-items: center;
+
+    // Learn More link keeps its underline style but gets 16px font
+    > a.learn-more {
```
YishaiGlasner: why do we need the element and not only class here and in other places in this file?

static/css/popover.scss
```diff
@@ -126,9 +161,10 @@ $popover-text: white;
...
   @include breakpoint(mobile) {
-    min-height: 44px;
-    font-size: 16px;
-    padding: 8px 12px;
+    font-size: 16px; // English System Body / Hebrew System Body (16px)
+    padding: 8px 12px; // 8px top, 12px right, 8px bottom, 12px left
+    height: auto;
+    min-height: auto; // Height controlled by container
   }
 }
```
YishaiGlasner, line 109: it seems to me that much can be done for making this file great - removing too specific selectors, unite doubled selectors, nesting (scss feature)

static/js/ModuleSwitcherPopover.jsx
```diff
@@ -46,6 +46,7 @@ const ModuleSwitcherPopover = ({ children }) => {
       const isModuleSwitcherClick = 
         event.target.closest('.headerDropdownMenu') || 
         event.target.closest('.menuButton');
+      console.log('event.target', event.target, 'isModuleSwitcherClick', isModuleSwitcherClick);
```
YishaiGlasner: is this intentional or forgotten?

YishaiGlasner (changes requested):

static/css/popover.scss
yodem: its to override the property of line 16 -  max-width: 350px;

static/css/popover.scss
yodem: what do you mean? there is one @include breakpoint(mobile) per classname

static/js/ModuleSwitcherPopover.jsx
yodem: forgotten, thanks

static/css/popover.scss
YishaiGlasner:
```
@include breakpoint(mobile) {
  .class1 {
    <rules>
  }
  .class2 {
    <rules>
}
```
I think this works also in plain css. in Scss you can also nest classes.

static/css/popover.scss
YishaiGlasner: but the browser takes the lesser

static/css/popover.scss
YishaiGlasner: you can say you prefer a structure that unite different things, so the https://github.com/media will be in several places, but now it isn't structured at all

static/css/popover.scss
yodem: @YishaiGlasner I did it again. when there was 2 accourances of css i turn them into mixin, learned and used some scss features

PR #2894, opened 2025-12-03
fix: [sc-38660] [Voices][History] History is displayed in duplicate

dcschreiber:
## Problem

History/Saved items were displaying in duplicate when:
- First visiting the History tab (items appeared, then duplicated after a pause)
- Switching between History and Saved tabs (duplicates accumulated with each switch)

Root cause: `useScrollToLoad` tracked skip state separately from the actual item count in the store. When the component remounted (tab switch), the hook's skip reset to 0 while the store still had cached items. Guard logic tried to prevent duplicate fetches but didn't work for <10 cached items.

Additionally, using `itemsPreLoaded` in the `useCallback` closure caused stale values during rapid scrolling - the callback would fire with outdated skip values before React re-rendered.

## Solution

Simplified `useScrollToLoad` by:
1. Tracking fetched count in a ref (`fetchedCountRef`) instead of state
2. Reading/updating the ref synchronously, avoiding stale closure issues

## Changes

- Removed `skip` state and `setSkip`
- Removed `isFirstRender` ref and guard logic
- Added `fetchedCountRef` to track pagination position (updates synchronously)
- Added `loadingRef` for synchronous loading check (prevents concurrent fetches)
- General cleanup: JSDoc, modern syntax, clearer variable names

## Additional patch
I added a catch for empty string in sheet summaries and titles.
Yitzhak approved this patch.
It was causing problems that didn't allow me to debug

I also removed functions in `Hooks.jsx` that weren't being used

Story: https://app.shortcut.com/sefaria/story/38660

### DIFF
--- a/static/css/keyboard.css
+++ b/static/css/keyboard.css
@@ -265,7 +265,11 @@
 }
 
 .keyboardInputInitiator {
-  margin:0px 3px;
-  vertical-align:middle;
-  cursor:pointer;
+  margin: 0;
+  padding: 0;
+  vertical-align: middle;
+  cursor: pointer;
+  width: 18px;
+  height: 18px;
+  box-sizing: border-box;
 }
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -4747,6 +4747,14 @@ h1.topic-landing-header {
   margin-top: 2px;
 }
 
+.search::placeholder,
+.topic-landing-search-input::placeholder,
+.readerSearch::placeholder,
+.searchFiltersInput::placeholder,
+.topicFilterInput::placeholder {
+  color: var(--dark-grey);
+}
+
 .topic-landing-search-suggestion{
   display: flex;
   list-style-type: none;
@@ -9860,9 +9868,7 @@ div.aboutSheetMetadata span:not(:first-child)::before{
   float: right;
 }
 .searchBox .keyboardInputInitiator {
-    width: 22px;
-    height: 20px;
-    padding: 4px 7px 2px;
+    margin: 0 7px 0 0;
     position: initial;
     vertical-align: 0px;
 }
--- a/static/js/DictionarySearch.jsx
+++ b/static/js/DictionarySearch.jsx
@@ -166,7 +166,7 @@ class DictionarySearch extends Component {
         return; //this prevents the icon from flashing on every key stroke.
       }
       if(Sefaria.interfaceLang === 'english'){
-          const opacity = show ? 0.4 : 0;
+          const opacity = show ? 1 : 0;
           $(ReactDOM.findDOMNode(this)).find(".keyboardInputInitiator").css({"opacity": opacity});
       }
   }
--- a/static/js/HeaderAutocomplete.jsx
+++ b/static/js/HeaderAutocomplete.jsx
@@ -28,6 +28,11 @@ const type_title_map = {
   "User": "Users"
 };
 
+const MODULE_ALLOWED_SEARCH_TYPES = {
+  [Sefaria.LIBRARY_MODULE]: ['Topic', 'ref', 'TocCategory', 'Term'],
+  [Sefaria.VOICES_MODULE]: ['Topic', 'User', 'Collection']
+};
+
 function type_icon(itemType, itemPic) {
     if (itemType === "User" && itemPic !== "") {
       return itemPic;
@@ -111,7 +116,8 @@ const getQueryObj = (query) => {
         return getQueryObj(repairedQuery);
       }
 
-      if (d["is_ref"]) {
+      const allowedTypes = MODULE_ALLOWED_SEARCH_TYPES[Sefaria.activeModule];
+      if (d["is_ref"] && allowedTypes.includes('ref')) {
         return {'type': 'Ref', 'id': d["ref"], 'is_book': d['is_book']};
       } else if (!!d["topic_slug"]) {
         return {'type': 'Topic', 'id': d["topic_slug"], 'is_book': d['is_book']};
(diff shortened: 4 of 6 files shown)


static/js/Hooks.jsx
```diff
@@ -51,56 +51,67 @@ function useDebounce(value, delay) {
...
-    $scrollable.on("scroll", handleScroll);
-    return (() => {$scrollable.off("scroll", handleScroll);})
-  }, [scrollableRef.current, loadedToEnd, skip, loading]);
+  const loadMore = useCallback(() => {
+    if (loadedToEnd || loadingRef.current) return;
+
+    loadingRef.current = true;
+    const skip = fetchedCountRef.current;
+    const separator = url.includes('?') ? '&' : '?';
+    const nextUrl = `${url}${separator}skip=${skip}&limit=${pageSize}`;
```
YishaiGlasner: please check if we have function in sefaria.js for that.
anyway, use js functions for parsing urls.

static/js/Hooks.jsx
```diff
@@ -51,56 +51,67 @@ function useDebounce(value, delay) {
...
+ *
+ * @param {Object} options
+ * @param {React.RefObject} options.scrollableRef - Ref to the scrollable container element
+ * @param {string} options.url - API endpoint (must support `skip` and `limit` query params)
+ * @param {function} options.setter - Callback to handle fetched data (receives array of items)
+ * @param {number} [options.itemsPreLoaded=0] - Number of items already in cache; skips initial fetch if > 0
+ * @param {number} [options.pageSize=20] - Number of items to fetch per request
+ */
+function useScrollToLoad({scrollableRef, url, setter, itemsPreLoaded = 0, pageSize = 20}) {
   const [loadedToEnd, setLoadedToEnd] = useState(false);
```
YishaiGlasner: the same reason tu use `useRef` apllies also here, isn't it?

static/js/Hooks.jsx
```diff
@@ -51,56 +51,67 @@ function useDebounce(value, delay) {
...
       setter(data);
+      fetchedCountRef.current += data.length;
       if (data.length < pageSize) {
         setLoadedToEnd(true);
       }
-      setLoading(false);
+      loadingRef.current = false;
     });
-  }, [skip]);
+  }, [url, setter, pageSize, loadedToEnd]);
```
YishaiGlasner: i'd say:
```suggestion
  }, [scrollableRef.current, pageSize, loadedToEnd]);
```
because it seems that if `scrollableRef` stays the same `url` and `setter` wouldn't change, but i can imagine a theoretical situation where url and setter are the same but the ref was changed. i can be wrond, i don't understand this deeply. maybe all of them are needed?

static/js/Hooks.jsx
```diff
@@ -51,56 +51,67 @@ function useDebounce(value, delay) {
...
+    if (itemsPreLoaded === 0) {
+      loadMore();
     }
-  }, [])
-}
+  }, []);
+
+  // Scroll listener for infinite loading
+  useEffect(() => {
+    const $scrollable = $(scrollableRef.current);
```
YishaiGlasner: can you remove jquery from here?
you can use js `scrollTop` `clientHeight` `scrollHeight` and event listener.
GPT suggestion, i didn't check:
```
  const el = scrollableRef.current;
  if (!el) return;

  const scrollMargin = 600;

  const handleScroll = () => {
    const scrollPosition = el.scrollTop + el.clientHeight;
    const scrollThreshold = el.scrollHeight - scrollMargin;

    if (scrollPosition >= scrollThreshold) {
      loadMore();
    }
  };

  el.addEventListener('scroll', handleScroll);
  return () => el.removeEventListener('scroll', handleScroll);
```

YishaiGlasner (commented):
after debugging do you still need the patch?

static/js/Hooks.jsx
dcschreiber: Couldn't find something specific. Could you have a look at this?

static/js/Hooks.jsx
dcschreiber: Well `url, setter` are not meant to change I think and are there because they are used in the callback so it's best practice to put them there.
You can't use scrollableRef.current in the dependency array because React doesn't track changes to .current properties

static/js/Hooks.jsx
YishaiGlasner: os there is no. you can make an addParam or addParams func and do it with js URL rather than this

dcschreiber: I'm now using `URL`.
Do you think a `Sefaria.addParams()` is needed considering there is `urlObj.searchParams.set`?

dcschreiber: Current code for easy review
```
const urlObj = new URL(url, window.location.origin);
urlObj.searchParams.set('skip', skip);
urlObj.searchParams.set('limit', pageSize);
const nextUrl = urlObj.pathname + urlObj.search;
```

dcschreiber:
@YishaiGlasner note, I added a cleanup of not used code to this PR

static/js/Hooks.jsx
```diff
@@ -51,97 +51,71 @@ function useDebounce(value, delay) {
...
+  const loadMore = useCallback(() => {
+    if (loadedToEndRef.current || loadingRef.current) return;
+
+    loadingRef.current = true;
+    const skip = fetchedCountRef.current;
+
+    const urlObj = new URL(url, window.location.origin);
+    urlObj.searchParams.set('skip', skip);
+    urlObj.searchParams.set('limit', pageSize);
+    const nextUrl = urlObj.pathname + urlObj.search;
```
YishaiGlasner, line 81: why not
```suggestion
    const nextUrl = urlObj.toString();
```

static/js/Hooks.jsx
YishaiGlasner: i don't know. i'll approve it also without (just had a liitle comment)

static/js/Hooks.jsx
dcschreiber: Well that produces a different string and make break the logic

PR #2896, opened 2025-12-03
Notification Badge Bug - remove badge when user enter the page

yodem:
## Description
Fixed notification red dot not disappearing when viewing notifications by implementing proper React state management and fixing backend scope handling for notification count calculations.

### `static/js/ReaderApp.jsx`
- Added `notificationCount: Sefaria.notificationCount` to component state initialization
- Modified `setUnreadNotificationsCount()` method to use `setState({ notificationCount: n })` instead of `forceUpdate()`
- Added `notificationCount={this.state.notificationCount}` prop to Header component

### `static/js/Header.jsx`
- Updated notification count logic to read from `props.notificationCount` instead of global `Sefaria.notificationCount`
- Modified both desktop and mobile notification icon logic to use the prop value

### `reader/views.py`
- **notifications_api**: Changed default scope from `LIBRARY_MODULE` to `VOICES_MODULE`
- **notifications_read_api**: 
  - Added scope parameter extraction from POST request
  - Updated `NotificationSet().unread_for_user()` to include scope parameter
  - Fixed `UserProfile().unread_notification_count()` to use the correct scope parameter
  - Changed default scope from `LIBRARY_MODULE` to `VOICES_MODULE`
  - Added debug print statement for scope parameter

## Notes
- **Root Cause**: The backend was ignoring the `scope` parameter sent by the frontend, causing it to return the total unread count across all modules instead of the module-specific count
- **Breaking Change**: Default notification scope changed from `LIBRARY_MODULE` to `VOICES_MODULE` 
- **React Pattern**: Replaced `forceUpdate()` anti-pattern with proper state management to ensure Header component re-renders when notification count changes
- **Module Scoping**: Notifications are now properly scoped to the active module (Library vs Voices), ensuring accurate unread counts per module

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -117,7 +117,7 @@ class ReaderApp extends Component {
       showSignUpModal: false,
       translationLanguagePreference: props.translationLanguagePreference,
       editorSaveState: 'saved',
-      notificationCount: props.notificationCount,
+      notificationCount: props.notificationCount || 0,
     };
   }
   setEditorSaveState = (nextState) => {
--- a/static/js/sefaria/sefaria.js
+++ b/static/js/sefaria/sefaria.js
@@ -3729,7 +3729,6 @@ Sefaria.unpackBaseProps = function(props){
       "following",
       "blocking",
       "calendars",
-      "notificationCount",
       "notifications",
       "saved",
       "userHistory",


reader/views.py
```diff
@@ -3033,10 +3033,13 @@ def notifications_read_api(request):
     """
     if request.method == "POST":
         notifications = request.POST.get("notifications")
+        print("notifications", request.POST.get("scope", VOICES_MODULE))
```
yodem:
```suggestion
```

static/js/NotificationsPanel.jsx
```diff
@@ -84,7 +83,7 @@ class NotificationsPanel extends Component {
     const notifications = Sefaria.notifications.map(n => 
       <Notifications type={n.type} props={n} key={n._id} />
     );
-    
+    console.log(Sefaria.notifications, Sefaria.notificationCount);
```
yodem:
```suggestion
```

static/js/ReaderApp.jsx
```diff
@@ -1744,8 +1745,8 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     }
   }
   setUnreadNotificationsCount(n) {
-    Sefaria.notificationCount = n;
-    this.forceUpdate();
+    Sefaria.notificationCount = n;  // Keep global in sync
```
YishaiGlasner: if you make it Header prop, why do we need global?
but why to change the old implementation and add another prop? i think storing it in `Sefaria.notificationCount` is better.
maybe better option is to add this to the context, and get it from there in Header and NotificationsPanel.

yodem: I think I"ll remove it from global Sefaria and turn it into a state. its needed only in one place, there is no need for it to be global

static/js/ReaderApp.jsx
```diff
@@ -117,6 +117,7 @@ class ReaderApp extends Component {
       showSignUpModal: false,
       translationLanguagePreference: props.translationLanguagePreference,
       editorSaveState: 'saved',
+      notificationCount: props.notificationCount || 0,
```
yodem, line 120: @YishaiGlasner In general I like to have fallbacks like this one but we basically never do it. shall I remove it?

static/js/ReaderApp.jsx
```diff
@@ -1744,8 +1745,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     }
   }
   setUnreadNotificationsCount(n) {
-    Sefaria.notificationCount = n;
-    this.forceUpdate();
+    this.setState({ notificationCount: n });
```
yodem, line 1748: @YishaiGlasner I got rid of the global state

PR #2905, opened 2025-12-07
feat: add maintenance mode to remote config

akiva10b:
This pull request introduces a site-wide maintenance mode feature, allowing the site to display a custom maintenance page for non-admin users when enabled. The implementation uses remote configuration keys to control activation and messaging, and includes a new, modernized maintenance page design.

**Maintenance Mode Feature:**

* Added new remote configuration keys `ENABLE_SITE_MAINTENANCE_MODE` and `SITE_MAINTENANCE_MESSAGE` to `remote_config/keys.py` for controlling maintenance mode activation and custom message.
* Implemented `_maintanance_mode_check` in `reader/views.py` to check if maintenance mode is enabled and display the maintenance page to non-admin users, using the configured message.
* Updated `render_template` in `reader/views.py` to invoke the maintenance mode check and render the maintenance page if maintenance mode is active.

**Maintenance Page Redesign:**

* Completely redesigned the maintenance page in `templates/static/maintenance.html` with improved styles, accessibility, and messaging, including a badge, logo, custom message, and contact information.

**Code Cleanup:**

* Replaced unused import of `ENABLE_WEBPAGES` with the new maintenance mode keys in `reader/views.py`.


reader/views.py
```diff
@@ -89,13 +89,22 @@
...
+from remote_config import remoteConfigCache
 
 if USE_VARNISH:
     from sefaria.system.varnish.wrapper import invalidate_ref, invalidate_linked
 
 import structlog
 logger = structlog.get_logger(__name__)
 
+def _maintanance_mode_check(request):
+    # Maintenance Mode Check, do to the sensativity of this check - we check that the value is exactly 613
```
yitzhakc: typo `due`

reader/views.py
```diff
@@ -89,13 +89,22 @@
...
+    if remoteConfigCache.get(ENABLE_SITE_MAINTENANCE_MODE) == 613:
```
yitzhakc: Please don't use magic numbers. Make it a constant.

reader/views.py
```diff
@@ -89,13 +89,22 @@
...
+        # Check if current user is admin/staff and skip maintenance mode if so
+        if not request.user.is_authenticated or not request.user.is_staff:
```
yitzhakc: Why do you need the `not request.user.is_authenticated`? By defintion, a non-authenticated (`AnonymousUser`) cannot have `is_staff == True`

reader/views.py
```diff
@@ -89,13 +89,22 @@
...
+            template_name = "static/maintenance.html"
```
yitzhakc: This is not intutive to me. To me, a `check` function returns `True` or `False`.

Perhaps the template rendering logic should be in the view, and this function should only check if we bypass maintanence mode or not.

templates/static/maintenance.html
```diff
@@ -1,28 +1,130 @@
...
 <body>
-	<div id="maintenancePage">
-		{% include "elements/sefaria.logotype.svg" %}
-		<h2>
-			{% autoescape off %}
-				{{ message }}
-			{% endautoescape %}
-		</h2>
-	</div>
+  <main id="maintenancePage" role="main">
```
yitzhakc, line 112: These should be translated strings, using Django's translation system.

remote_config/keys.py
```diff
@@ -1,3 +1,5 @@
 CURRENT_LINKER_VERSION = "feature.linker.current_version"
 REF_CACHE_LIMIT_KEY = "feature.text.ref_cache_limit"
-ENABLE_WEBPAGES = "feature.webpages.enable"
\ No newline at end of file
+ENABLE_WEBPAGES = "feature.webpages.enable"
+ENABLE_SITE_MAINTENANCE_MODE = "feature.site.maintenance_mode"
+SITE_MAINTENANCE_MESSAGE = "feature.site.maintenance_message"
```
yitzhakc: If we're adding the maintenance message to the RemoteConfig, we should remove it from all the environment variables so as not to create confusion as to the source of the message.

reader/views.py
```diff
@@ -118,6 +127,8 @@ def render_template(request, template_name='base.html', app_props=None, template
     props.update(app_props)
     propsJSON = json.dumps(props, ensure_ascii=False)
     template_context["propsJSON"] = propsJSON
+    if remoteConfigCache.get(ENABLE_SITE_MAINTENANCE_MODE):
+        template_name, template_context = _maintanance_mode_check(request)
```
yitzhakc: My concern here is that if we render the maintenance template for any (or all) URLs, this could cause improper caching or SEO penalty. 
Perhaps we should instead force a redirect to the static maintenance page.

yitzhakc: I'd say that this logic is critical enough that it would warrant a unittest.

Please add a unit test that mocks or patches the relevant value in the `remoteConfigCache` and ensures that whatever result you desire (totally different template, a redirect to mainteance page etc.) happens.

reader/views.py
akiva10b: I am actually conflicted about what to do here. In theory - we don't need the check that it is 613 at all, as this is just a way to make sure that we don't accidentally screw up. Technically it should also be a password that is in the secrets or env vars. In this case, it isn't really a magic string, 613 is the check. I imagine we could declare a var VALIDATION_NUMBER and put 613 in it, but we would never use it anywhere else.

reader/views.py
akiva10b: There is a performance hack I did here that made the code less readable. 99.9999% of cases, remoteConfig.get(ENABLE_SITE_MAINTENANCE_MODE) returns false. I only check the value ENABLE_SITE_MAINTENANCE_MODE if the truthiness check passed. Because this endpoint runs millions of times, I went with the truthiness check which is slightly faster. However, this leads to a function which still has check logic in it. To fix this - I will revert back to my original "in code" solution which elongates the function, btu is clearer

reader/views.py
akiva10b: great point! I will just add a 503 status to cover that

remote_config/keys.py
akiva10b: the env var maintenance mode does something different - it works ever when the web pod is the last thing standing an all other systems have broken down

reader/views.py
yitzhakc: If you're adding a check, add it properly 🙃

reader/views.py
yitzhakc: Maybe my comment wasn't clear - I'm 100% fine with you checking the value of `ENABLE_SITE_MAINTENANCE_MODE` twice.

My comment:
> Perhaps the template rendering logic should be in the view, and this function should only check if we bypass maintanence mode or not.

Was referring to `_maintanance_mode_check`, that it should return a boolean, and not return the tuple `(template_name, template_context)`

reader/views.py
```diff
@@ -89,13 +89,22 @@
...
 logger = structlog.get_logger(__name__)
 
+def _maintanance_mode_check(request):
+    # Maintenance Mode Check, do to the sensativity of this check - we check that the value is exactly 613
+    if remoteConfigCache.get(ENABLE_SITE_MAINTENANCE_MODE) == 613:
+        # Check if current user is admin/staff and skip maintenance mode if so
+        if not request.user.is_authenticated or not request.user.is_staff:
+            template_name = "static/maintenance.html"
+            template_context = {"message": remoteConfigCache.get(SITE_MAINTENANCE_MESSAGE)}
+            return template_name, template_context
```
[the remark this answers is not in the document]
yitzhakc: Seemingly no longer relevant?

yitzhakc:
@akiva10b I assume we still want this, correct?

PR #2908, opened 2025-12-07
feat: Hide ToC language toggle on narrow screens

dcschreiber:
## Description
Hid the `.languageToggle` from the ToC page on narrow screens because it appears in the header

## Code Changes
* Added in S2 that not to display the component on narrow screens
* Removed the old JS filter on `multiPanel`

## Notes
1. I tried to implement this in the header SCSS using the existing breakpoint vars, but S2 loads after Header.css so it wasn't working
2. I thought making a new SCSS file just for this was overkill for a 2 point task and didn't have other ideas how to avoid hardcoding the breakpoint

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -1328,6 +1328,12 @@ a.interfaceLinks-option.active, a.interfaceLinks-option.inactive {
   align-items: center;
   justify-content: center;
 }
+/* Hide ToC language toggle on narrow screens where mobile header shows its own toggle */
+@media (max-width: 842px) {
+  .navTitle .languageToggle {
+    display: none;
+  }
+}
 .navTitleTab {
   margin-inline-end: 26px;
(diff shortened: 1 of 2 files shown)


YishaiGlasner:
@schreibersefaria @yitzhakc we have css mixins for breakoints.
we shouldn't use the breakpoints in css, which cause them to be duplicated, and enlarge our breakpoints jungle.
that's why i told Penina this should be a part of refactor that takes some of our css out to be scss. i think that this is not urgent (how many people use narrow screen on desktop?) and should be done as part of such a refactor.

if my opinion rejected, this pr still should remove the old rule of hiding.

dcschreiber:
@YishaiGlasner I did remove the old rule of hiding, it was in JS, not CSS

YishaiGlasner (approved):

PR #2913, opened 2025-12-09
fix(NavSidebar): Fixed the navbar footer donate link to also support …

dcschreiber:
Fixed the Navbar footer donate link to also support the correct heb link using `Sefaria._v`
I left the hardcoded links to match the task scope

Story details: https://app.shortcut.com/sefaria/story/39393

### DIFF
--- a/static/js/NavSidebar.jsx
+++ b/static/js/NavSidebar.jsx
@@ -294,7 +294,7 @@ const getSidebarFooterData = () => [{'he': 'אודות','en': 'About', 'url': `$
                                     {'he': 'יוטיוב','en':'YouTube', 'url':'https://www.youtube.com/user/SefariaProject'},
                                     {'he': 'חנות','en':'Shop', 'url': 'https://store.sefaria.org/'},
                                     {'he': 'אפשרויות תרומה','en':'Ways to Give', 'url': `${Sefaria.getModuleURL(Sefaria.LIBRARY_MODULE).origin}/ways-to-give`},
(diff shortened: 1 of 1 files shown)


static/js/NavSidebar.jsx
```diff
@@ -294,7 +294,7 @@ const getSidebarFooterData = () => [{'he': 'אודות','en': 'About', 'url': `$
                                     {'he': 'יוטיוב','en':'YouTube', 'url':'https://www.youtube.com/user/SefariaProject'},
                                     {'he': 'חנות','en':'Shop', 'url': 'https://store.sefaria.org/'},
                                     {'he': 'אפשרויות תרומה','en':'Ways to Give', 'url': `${Sefaria.getModuleURL(Sefaria.LIBRARY_MODULE).origin}/ways-to-give`},
-                                    {'he': 'תרומות','en':'Donate', 'url': 'https://donate.sefaria.org/give/451346/#!/donation/checkout?c_src=Footer'},
+                                    {'he': 'תרומות','en':'Donate', 'url': Sefaria._v({en: 'https://donate.sefaria.org/give/451346/#!/donation/checkout?c_src=Footer', he: 'https://donate.sefaria.org/give/468442/#!/donation/checkout?c_src=Footer'})},
```
YishaiGlasner, line 297: we have the same links in Misc.jsx (and maybe more places).
can we have them in only one place?

dcschreiber: where would be the correct place for this?
Being as this is intended as a small task I thought to just leave it as is. The search and replace is so easy these days.
But if there is a correct place already, I don't mind doing the change

yitzhakc: Let's not break scope of this story please :-)

PR #2916, opened 2025-12-09
feat(middleware): add SessionCookieDomainMiddleware for cross-subdoma…

yodem:
## Description
Adds `SessionCookieDomainMiddleware` to enable cross-subdomain cookie sharing between modules (library ↔ voices) within the same language domain. This allows users to remain logged in when navigating between different Sefaria modules. Also fixes a bug in `_find_longest_common_domain_suffix` that was failing for local testing domains.

## Code Changes

**`sefaria/system/middleware.py`**
- Added new `SessionCookieDomainMiddleware` class that:
  - Dynamically sets `SESSION_COOKIE_DOMAIN` and `CSRF_COOKIE_DOMAIN` based on `DOMAIN_MODULES` configuration
  - Builds an approved domains mapping at initialization using `get_cookie_domain()` from `domains_and_languages.py`
  - Sets cookie domain in `process_request` before `SessionMiddleware` runs
  - Includes `process_response` fallback to modify cookie domains for edge cases
- Added import for `_get_hostname_without_port` from `domains_and_languages`

**`sefaria/utils/domains_and_languages.py`**
- Fixed `_find_longest_common_domain_suffix` algorithm:
  - **Before**: Only matched suffixes that already started with a dot (failed for local domains like `localsefaria.xyz`)
  - **After**: Prepends a dot when checking if hostname ends with the suffix (`hostname.endswith(f'.{common_suffix}')`)
- Added docstring example for local testing domains
- Returns `f'.{common_suffix}'` to ensure the result always starts with a dot

**`sefaria/system/tests/test_middleware.py`** (new file)
- Added comprehensive test suite for `SessionCookieDomainMiddleware`:
  - `TestBuildApprovedDomains`: Tests hostname → cookie domain mapping for local, production, cauldron, and empty configs
  - `TestCookieDomainSetting`: Tests that cookies are set with correct domains for English/Hebrew hosts
  - `TestUnapprovedDomains`: Tests that unapproved domains and localhost don't get cookie domains modified
  - `TestEdgeCases`: Tests error handling for missing cookies, empty config, and production domains

## Notes
- The fix to `_find_longest_common_domain_suffix` is mainly needed for **local testing domains** (e.g., `localsefaria.xyz`) where the base domain doesn't contain a leading dot before the TLD. Production domains like `sefaria.org` already worked because `.sefaria.org` starts with a dot.
- Single hostname per language returns no cookie domain (need 2+ hostnames to find a common suffix)
- All test configs now include both `"en"` and `"he"` keys to match the expected `DOMAIN_MODULES` structure

### DIFF
--- a/sefaria/system/middleware.py
+++ b/sefaria/system/middleware.py
@@ -15,10 +15,10 @@ from sefaria.site.site_settings import SITE_SETTINGS
 from sefaria.model.user_profile import UserProfile
 from sefaria.utils.util import short_to_long_lang_code, get_lang_codes_for_territory
 from sefaria.utils.views_utils import add_query_param
-from sefaria.utils.domains_and_languages import current_domain_lang, get_redirect_domain_for_language, needs_domain_switch, get_cookie_domain, _get_hostname_without_port
+from sefaria.utils.domains_and_languages import current_domain_lang, get_redirect_domain_for_language, needs_domain_switch, get_cookie_domain, get_hostname_without_port
 from sefaria.system.cache import get_shared_cache_elem, set_shared_cache_elem
 from django.utils.deprecation import MiddlewareMixin
-from urllib.parse import quote, urlparse, urljoin
+from urllib.parse import quote, urljoin
 from sefaria.constants.model import LIBRARY_MODULE
 
 import structlog
@@ -195,8 +195,9 @@ class SessionCookieDomainMiddleware(MiddlewareMixin):
     Enables cross-subdomain cookie sharing within the same language domain,
     allowing users to remain logged in when navigating between modules (e.g., library ↔ voices).
     
-    Works by setting Django's SESSION_COOKIE_DOMAIN and CSRF_COOKIE_DOMAIN in process_request,
-    before SessionMiddleware creates the cookies.
+    Works by storing the cookie domain on the request object in process_request,
+    then applying it to cookies in process_response. This approach is thread-safe
+    as it avoids modifying global Django settings.
     
     Examples:
         - www.sefaria.org, voices.sefaria.org → cookie domain '.sefaria.org'
@@ -205,23 +206,22 @@ class SessionCookieDomainMiddleware(MiddlewareMixin):
     
     def __init__(self, get_response=None):
         super().__init__(get_response)
-        self.approved_domains = self._build_approved_domains()
-        # Store original settings
-        self.original_session_domain = getattr(settings, 'SESSION_COOKIE_DOMAIN', None)
-        self.original_csrf_domain = getattr(settings, 'CSRF_COOKIE_DOMAIN', None)
+        self._approved_domains = self._build_approved_domains()
+    
+    def _get_cookie_domain_for_request(self, request):
+        """
+        Get the cookie domain for a request based on its hostname.
+        
+        Returns:
+            str or None: The cookie domain (e.g., '.sefaria.org') if hostname is approved,
+                        None otherwise.
+        """
+        hostname = get_hostname_without_port(request)
+        return self._approved_domains.get(hostname)
     
     def process_request(self, request):
-        """Set cookie domain in Django settings before SessionMiddleware runs."""
-        hostname = _get_hostname_without_port(request)
-        cookie_domain = self.approved_domains.get(hostname)
-        
-        if cookie_domain:
-            settings.SESSION_COOKIE_DOMAIN = cookie_domain
-            settings.CSRF_COOKIE_DOMAIN = cookie_domain
-        else:
-            # Reset to original if host not in approved list
-            settings.SESSION_COOKIE_DOMAIN = self.original_session_domain
-            settings.CSRF_COOKIE_DOMAIN = self.original_csrf_domain
+        """Determine cookie domain for this request and store it on the request object."""
+        request._cookie_domain = self._get_cookie_domain_for_request(request)
     
     def _build_approved_domains(self):
         """
@@ -260,12 +260,11 @@ class SessionCookieDomainMiddleware(MiddlewareMixin):
     
     def process_response(self, request, response):
         """
-        Fallback: modify cookie domains in response if they weren't set correctly.
+        Apply the cookie domain to session and CSRF cookies in the response.
         
-        This handles edge cases where cookies are set outside the normal request flow.
+        Uses the cookie domain stored on the request object by process_request.
         """
-        hostname = _get_hostname_without_port(request)
-        cookie_domain = self.approved_domains.get(hostname)
+        cookie_domain = getattr(request, '_cookie_domain', None)
         
         if cookie_domain:
             # Update session cookie domain if present
--- a/sefaria/system/tests/test_middleware.py
+++ b/sefaria/system/tests/test_middleware.py
@@ -4,8 +4,6 @@ Tests for SessionCookieDomainMiddleware.
 Tests that session and CSRF cookie domains are dynamically set based on 
 an approved list derived from DOMAIN_MODULES.
 """
-
-import pytest
 from django.test import RequestFactory, override_settings
 from django.http import HttpResponse
 from sefaria.system.middleware import SessionCookieDomainMiddleware
@@ -95,7 +93,7 @@ class TestBuildApprovedDomains:
             'chiburim.localsefaria-il.xyz': '.localsefaria-il.xyz'
         }
         
-        assert middleware.approved_domains == expected
+        assert middleware._approved_domains == expected
     
     @override_settings(DOMAIN_MODULES=PRODUCTION_CONFIG)
     def test_builds_correct_mapping_for_production_config(self):
@@ -109,14 +107,14 @@ class TestBuildApprovedDomains:
             'chiburim.sefaria.org.il': '.sefaria.org.il'
         }
         
-        assert middleware.approved_domains == expected
+        assert middleware._approved_domains == expected
     
     @override_settings(DOMAIN_MODULES=EMPTY_CONFIG)
     def test_handles_empty_config(self):
         """Test that empty DOMAIN_MODULES results in empty approved list."""
         middleware = SessionCookieDomainMiddleware(get_response=lambda r: HttpResponse())
         
-        assert middleware.approved_domains == {}
+        assert middleware._approved_domains == {}
     
     def test_cauldron_domain_strips_only_first_subdomain(self):
         """Test that cauldron domain preserves second-level domain."""
@@ -128,7 +126,7 @@ class TestBuildApprovedDomains:
                 'voices.cauldron.sefaria.org': '.cauldron.sefaria.org'
             }
             
-            assert middleware.approved_domains == expected
+            assert middleware._approved_domains == expected
     
     def test_single_hostname_per_language_returns_no_cookie_domain(self):
         """Test that single hostname per language returns no cookie domain (need 2+ for common suffix)."""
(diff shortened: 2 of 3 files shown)


dcschreiber:
Note for next time, the description seems to cover the last changes you made, not the whole diff.
So it has things about moving the position of `SessionCookieDomainMiddleware` and removing the `localhost` fallback

sefaria/system/middleware.py
```diff
@@ -188,6 +188,40 @@ def process_request(self, request):
...
+    
+    Works by modifying cookies in response.cookies after Django's SessionMiddleware 
+    and CsrfViewMiddleware have set them.
+    
+    Examples:
+        - www.sefaria.org → '.sefaria.org'
+        - voices.sefaria.org → '.sefaria.org'
+        - www.sefaria.org.il → '.sefaria.org.il'
+    
+    Note: For localhost/dev environments, set SESSION_COOKIE_DOMAIN explicitly in local_settings.py
```
dcschreiber: This won't work on localhost. You can't share cookies on localhost

sefaria/system/middleware.py
```diff
@@ -188,6 +188,40 @@ def process_request(self, request):
...
 
 
+class SessionCookieDomainMiddleware(MiddlewareMixin):
+    """
+    Sets appropriate domain on session and CSRF cookies based on request host.
+    
+    This middleware enables cross-subdomain cookie sharing within the same language domain,
+    allowing users to remain logged in when navigating between modules (e.g., library ↔ voices).
+    
+    Works by modifying cookies in response.cookies after Django's SessionMiddleware
```
dcschreiber: Why is this needed if you changes SESSION_COOKIE_DOMAIN?

yitzhakc: I spoke with Yotam, we agreed that the current implementation doesn't solve the issue for the multiple top-level domains. He's fixing it and will re-request review.

sefaria/settings.py
```diff
@@ -113,9 +113,10 @@ def get_static_url():
 MIDDLEWARE = [
     'django_hosts.middleware.HostsRequestMiddleware',  # must be first
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'sefaria.system.middleware.SessionCookieDomainMiddleware',
+    'django.middleware.csrf.CsrfViewMiddleware',
```
dcschreiber, line 117: The description you wrote says the new middleware is `after SessionMiddleware and CsrfViewMiddleware`

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        
+        if cookie_domain:
+            settings.SESSION_COOKIE_DOMAIN = cookie_domain
+            settings.CSRF_COOKIE_DOMAIN = cookie_domain
+        else:
+            # Reset to original if host not in approved list
+            settings.SESSION_COOKIE_DOMAIN = self.original_session_domain
+            settings.CSRF_COOKIE_DOMAIN = self.original_csrf_domain
+    
+    def _build_approved_domains(self):
```
dcschreiber, line 226: We already have a `get_cookie_domain` and a `_find_longest_common_domain_suffix`. Did we check if we can use them?

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        # Find longest common suffix by progressively removing left parts
+        common = hostnames[0]
+        for hostname in hostnames[1:]:
+            while common and not (hostname == common or hostname.endswith('.' + common)):
+                parts = common.split('.', 1)
+                common = parts[1] if len(parts) > 1 else ''
+        
+        return '.' + common if common else None
+    
+    def process_response(self, request, response):
```
dcschreiber, line 261: Not sure what this is for. Can only see it used in tests. Could you also give an example for when it's needed?

sefaria/system/tests/test_middleware.py
```diff
@@ -0,0 +1,318 @@
...
+from django.test import RequestFactory, override_settings
+from django.http import HttpResponse
+from sefaria.system.middleware import SessionCookieDomainMiddleware
+
+
+# ============================================================================
+# TEST CONFIGURATIONS
+# ============================================================================
+
+LOCAL_CONFIG = {
```
dcschreiber, line 16: Shouldn't we also have a cauldron config?
Also, locally this won't work anyway so I'm not sure we need to test this

sefaria/system/tests/test_middleware.py
```diff
@@ -0,0 +1,318 @@
...
+        "library": "https://www.sefaria.org",
+        "voices": "https://voices.sefaria.org"
+    },
+    "he": {
+        "library": "https://www.sefaria.org.il",
+        "voices": "https://chiburim.sefaria.org.il"
+    }
+}
+
+EMPTY_CONFIG = {}
```
dcschreiber, line 49: Why do we want to have an empty config? don't we want to have some error in that case?

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        approved = {}
+        domain_modules = getattr(settings, 'DOMAIN_MODULES', {})
+        
+        for modules in domain_modules.values():
+            # Collect all hostnames for this language
+            hostnames = [
+                urlparse(url).hostname 
+                for url in modules.values() 
+                if urlparse(url).hostname
+            ]
```
dcschreiber, line 245:
```suggestion
            hostnames = [
                hostname
                for url in modules.values()
                if (hostname := urlparse(url).hostname)
            ]
```

sefaria/system/tests/test_middleware.py
yodem: It was mainly to see if the function that creates the approved domain doesnt crash and provide an empty object.

sefaria/system/tests/test_middleware.py
yodem: why wouldnt it work? if we set custom hosts it would work (worked for me)

sefaria/system/middleware.py
yodem: I did see _find_longest_common_domain_suffix, but it operates slightly different and return different results in some cases (4 tests fails).
since I didnt see any tests for _find_longest_common_domain_suffix I didnt want to modify it (although I think we need to).

sefaria/system/middleware.py
yodem: Its a django middleware thing - https://stackoverflow.com/questions/33407254/django-middleware-difference-between-process-request-and-process-view

sefaria/system/tests/test_middleware.py
yitzhakc: Definitely want to test this :-) 

Cauldrons specifically don't have this problem, since it's all the same (top-level) subdomain...

sefaria/system/middleware.py
yitzhakc: @dcschreiber can you and @yodem meet and see if we can combine them, or at least better understand why we can't/shouldn't?

sefaria/settings.py
```diff
@@ -113,9 +113,10 @@ def get_static_url():
 MIDDLEWARE = [
     'django_hosts.middleware.HostsRequestMiddleware',  # must be first
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'sefaria.system.middleware.SessionCookieDomainMiddleware',
```
yitzhakc, line 115: Doesn't our `SessionCookieDomainMiddleware` need to be _before_ the Django `SessionMiddleware`? Otherwise, I think the [`SessionMiddleware`](https://docs.djangoproject.com/en/1.11/_modules/django/contrib/sessions/middleware/) will override our manually set cookie (since responses are processed bottom-up)

yitzhakc (changes requested):

sefaria/system/middleware.py
yodem: @dcschreiber @yitzhakc  The key difference: the middleware checks if the hostname ends with dot + common, while _find_longest_common_domain_suffix only looks at substrings that start with a dot.
I think we can change _find_longest_common_domain_suffix Im just afraid it would break other things, and since we dont have tests we wouldnt know.
However its a minor change and the logic would stay the same so I believe we can make this change - take the _build_approved_domains logic and assign it to _find_longest_common_domain_suffix.
what do you think?

sefaria/settings.py
yodem: AFAIK, the SessionMiddleware generates the unique id of the sessions and then we modify it in SessionCookieDomainMiddleware.
Plus, it does work now, but I"ll see if something changes if I change the order

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+    
+    def _build_approved_domains(self):
+        """
+        Build mapping of hostname -> cookie domain from DOMAIN_MODULES.
+        
+        Groups hostnames by language and finds the common domain suffix for each group,
+        enabling cookie sharing across all modules within that language.
+        
+        Returns:
+            dict: {hostname: cookie_domain}, e.g., {'www.sefaria.org': '.sefaria.org'}
```
YishaiGlasner, line 234: why not to pass this function the hostname, and then return `cookie_domain`. it seems to save redundant calculatios that are done in any request

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        Examples:
+            ['www.sefaria.org', 'voices.sefaria.org'] → '.sefaria.org'
+            ['localsefaria.xyz', 'voices.localsefaria.xyz'] → '.localsefaria.xyz'
+            ['name.cauldron.sefaria.org'] → '.cauldron.sefaria.org'
+            ['name.cauldron.sefaria.org', 'voices.cauldron.sefaria.org'] → '.cauldron.sefaria.org'
+        """
+        if len(hostnames) == 1:
+            # Single hostname: strip first subdomain if multi-part
+            parts = hostnames[0].split('.')
+            return '.' + '.'.join(parts[1:]) if len(parts) > 1 else '.' + hostnames[0]
```
YishaiGlasner: maybe
```suggestion
            parts = hostnames[0].split('.')[1:]
            return f'.{".".join(parts) if parts else hostnames[0]}'
```

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        
+        # Find longest common suffix by progressively removing left parts
+        common = hostnames[0]
```
YishaiGlasner: moving this line upward will save other occurrences of hostnames[0]

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+        for hostname in hostnames[1:]:
+            while common and not (hostname == common or hostname.endswith('.' + common)):
```
YishaiGlasner:
```suggestion
            while common and hostname != common and not hostname.endswith(f'.{common)}':
```

sefaria/system/middleware.py
```diff
@@ -188,6 +188,125 @@ def process_request(self, request):
...
+                parts = common.split('.', 1)
+                common = parts[1] if len(parts) > 1 else ''
```
YishaiGlasner: maybe
```suggestion
                common = common.split('.', 1)[1] if '.' in common else ''
```
it seems better to me because it's 'telling' exactly what you want to di

sefaria/settings.py
yitzhakc: It seems to me that if - for whatever reason - the session is modified, the conditional on [line 46](https://github.com/django/django/blob/c669cf279ae7b3e02a61db4fb077030a4db80e4f/django/contrib/sessions/middleware.py#L46) will validate as True, and the block will be executed. Assuming the status code is not 500, the Session cookie will then be overridden (on line 65) with whatever the "default" `settings.SESSION_COOKIE_DOMAIN` is (`None`?). 
So, it's an edge case - it would cause a bug only if the session changes as part of a login/register flow on the subdomain, since then the session cookie would not be set properly on the right domain - but it is a possibility nontheless.

sefaria/system/middleware.py
dcschreiber: @yodem `get_cookie_domain` takes into account the `.` if I'm not mistaken. Want to huddle about it?

sefaria/system/tests/test_middleware.py
dcschreiber: It's a chrome thing. You can't share cookies across domains on localhost

yodem: you define custom hosts in etc/hosts

dcschreiber: yes. But unless I'm very mistaken, it just won't work locally. 
Did you test and see if it's working?

yodem: yes

sefaria/system/middleware.py
yodem: Not sure I understood but this function runs only once on startup

YishaiGlasner: you have twice
```
        hostname = _get_hostname_without_port(request)
        cookie_domain = self.approved_domains.get(hostname)
```
that means that in `__init__` you should define `self.cookie_domain` rather than `self.approved_domains`. the approved domains are only 'private' var in the process of getting the cookie.
so you should have a function that gives you it. now i don't think the host should be passed to the function, as i wrore previously. you can also use `_get_hostname_without_port` in this func.

dcschreiber (approved):

PR #2917, opened 2025-12-09
deprecate-non-cts-part

nsantacruz:
## Description

## Code Changes

## Notes

### DIFF
--- a/sefaria/model/linker/ref_resolver.py
+++ b/sefaria/model/linker/ref_resolver.py
@@ -366,33 +366,29 @@ class RefResolver:
         self._thoroughness = thoroughness
         if reset_ibids:
             self.reset_ibid_history()
-        resolved = []
-        for raw_ref in raw_refs:
-            temp_resolved = self._resolve_raw_ref_and_update_ibid_history(raw_ref, book_context_ref)
-            resolved += temp_resolved
-        return resolved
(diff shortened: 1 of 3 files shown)


sefaria/model/linker/ref_resolver.py
```diff
@@ -420,53 +423,34 @@ def split_non_cts_parts(self, raw_ref: RawRef) -> List[RawRef]:
                     pass
                 curr_parts = []
                 curr_part_start = ipart+1
-        return split_raw_refs
+        return split_raw_refs[0]
```
yonadavGit, line 426: theoretically we now that the LHS after the non-CTS part should be a valid ibid citation.
Maybe we can look into salvaging it

sefaria/model/linker/tests/linker_test_utils.py
```diff
@@ -155,7 +155,7 @@ def print_spans(raw_ref: RawRef):
         print(f'{i}) {part.text}')
         
         
-def get_matches_from_resolver_data(resolver_data) -> list[PossiblyAmbigResolvedRef]:
+def get_matches_from_resolver_data(resolver_data) -> PossiblyAmbigResolvedRef:
```
yonadavGit, line 158: nice, i can see why the deprecation makes things simpler

PR #2919, opened 2025-12-09
Bug/sc 39129/implement the fix and form to dropdown in misc

dcschreiber:
## Summary
- Fix focus of dropdown first item showing on mouse click (should only show when keyboard navigating)
- Fix keyboard selection in AddToSourceSheetBox (Enter/Space wasn't working)

## Changes

**static/css/s2.css**
- Add CSS rule for `.dropdownOption.focused` that only shows focus ring when user is tabbing and listbox has focus (so it no longer shows if the users moves to the `AddToSourceSheetBox` form)

**static/js/Misc.jsx**
- Remove inline focus style from `Dropdown` options (now handled by CSS)

**static/js/AddToSourceSheet.jsx**
- Fix `handleSheetListSelect`: was using non-existent `this.props.userSheets`, now uses `Sefaria.sheets.userSheets()`
- Remove inline focus style, add `focused` class to classNames
- Clean up variable declarations (`let` → `const`, reorder for clarity)
- Remove unnecessary `onKeyDown` and `tabIndex` on individual options (listbox handles keyboard)

Decided against migrating the `AddToSourceSheetBox` to `Dropdown` because I think it's not enough duplicate code to justify the extra complication to `Dropdown`

Story details: https://app.shortcut.com/sefaria/story/39129

### DIFF
--- a/static/js/AddToSourceSheet.jsx
+++ b/static/js/AddToSourceSheet.jsx
@@ -340,11 +340,10 @@ const isSelected = this.state.selectedSheet?.id === sheet.id;
       const isFocused = i === this.state.focusedSheetIndex;
       const classes = classNames({dropdownOption: 1, noselect: 1, selected: isSelected, focused: isFocused});
       const title = Sefaria.sheets.getSheetTitle(sheet?.title);
-      const selectSheet = this.selectSheet.bind(this, sheet);
       return (
         <div
           className={classes}
-          onClick={selectSheet}
+          onClick={() => this.selectSheet(sheet)}
           key={sheet.id}
           role="option"
           aria-selected={!!isSelected}


static/js/AddToSourceSheet.jsx
```diff
@@ -336,23 +336,21 @@ class AddToSourceSheetBox extends Component {
     const titleRef = this.makeTitleRef();
     const sheets     = Sefaria._uid ? Sefaria.sheets.userSheets(Sefaria._uid) : null;
     let sheetsList = Sefaria._uid && sheets ? sheets.map((sheet, i) => {
-      let classes     = classNames({dropdownOption: 1, noselect: 1, selected: this.state.selectedSheet && this.state.selectedSheet.id == sheet.id});
-      let title = Sefaria.sheets.getSheetTitle(sheet?.title);
-      let selectSheet = this.selectSheet.bind(this, sheet);
-      const isSelected = !!(this.state.selectedSheet && this.state.selectedSheet.id == sheet.id);
+      const isSelected = this.state.selectedSheet && this.state.selectedSheet.id == sheet.id;
```
stevekaplan123:
``` suggestion
const isSelected = this.state.selectedSheet?.id === sheet.id;
```

static/js/AddToSourceSheet.jsx
```diff
@@ -115,8 +115,8 @@ class AddToSourceSheetBox extends Component {
    * Selects the currently focused sheet from the list.
    */
   handleSheetListSelect = () => {
-    const sheets = this.props.userSheets;
-    if (sheets && sheets[this.state.focusedSheetIndex]) {
+    const sheets = Sefaria.sheets.userSheets(Sefaria._uid);
```
stevekaplan123, line 118: why aren't we using the userSheets prop?

static/js/AddToSourceSheet.jsx
```diff
@@ -336,23 +336,21 @@ class AddToSourceSheetBox extends Component {
...
     let sheetsList = Sefaria._uid && sheets ? sheets.map((sheet, i) => {
-      let classes     = classNames({dropdownOption: 1, noselect: 1, selected: this.state.selectedSheet && this.state.selectedSheet.id == sheet.id});
-      let title = Sefaria.sheets.getSheetTitle(sheet?.title);
-      let selectSheet = this.selectSheet.bind(this, sheet);
-      const isSelected = !!(this.state.selectedSheet && this.state.selectedSheet.id == sheet.id);
+      const isSelected = this.state.selectedSheet && this.state.selectedSheet.id == sheet.id;
       const isFocused = i === this.state.focusedSheetIndex;
+      const classes = classNames({dropdownOption: 1, noselect: 1, selected: isSelected, focused: isFocused});
+      const title = Sefaria.sheets.getSheetTitle(sheet?.title);
+      const selectSheet = this.selectSheet.bind(this, sheet);
```
stevekaplan123: 'bind' isn't used much these days.  I think
it's better to use an arrow function ```() => this.selectSheet(sheet)}```

static/js/AddToSourceSheet.jsx
```diff
@@ -336,23 +336,21 @@ class AddToSourceSheetBox extends Component {
...
       return (
         <div
           className={classes}
           onClick={selectSheet}
```
stevekaplan123:
```suggestion
onClick={() => this.selectSheet(sheet)}```

static/js/AddToSourceSheet.jsx
dcschreiber: Primarily because it doesn't exist.

stevekaplan123: What do you mean that it doesn't exist?  There is no prop 'userSheets' that gets passed down?  Are there any other mentions of it in this component?  Those should be removed as well

PR #2923, opened 2025-12-09
Bug/sc 39367/fixing page title for sheets

dcschreiber:
Now setting the title directly once it's available 

Story details: https://app.shortcut.com/sefaria/story/39367

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -430,9 +430,9 @@ class ReaderApp extends Component {
       }
 
       // Update history when sheet title becomes available (was showing placeholder)
-      if (next.mode === "Sheet" && prev.mode === "Sheet" && prev.sheetID === next.sheetID
-          && Sefaria.sheets.loadSheetByID(next.sheetID)?.title
-          && document.title === Sefaria.getPageTitle("", "sheet")) {
+      if (next.mode === "Sheet"
+          && Sefaria.sheets.loadSheetByID(next.sheetID)?.title // sheet data now in cache
+          && document.title === Sefaria.getPageTitle("", "sheet")) { // title is still placeholder
         return true;
       }
     }


static/js/sheets/Sheet.jsx
```diff
@@ -34,6 +34,9 @@ class Sheet extends Component {
     Sefaria.sheets.loadSheetByID(this.props.id, this.onDataLoad);
   }
   onDataLoad(data) {
+    // Set page title for client-side navigation (SSR handles this on direct navigation)
+    document.title = Sefaria.getPageTitle(data.title.stripHtml(), "sheet");
+    history.replaceState(history.state, document.title, window.location.href);
```
YishaiGlasner: isn't this `ReferenceError`? history was not defined.
this seems as sort of soemthing that should happen in ReaderApp

dcschreiber: "_history is a browser global object (window.history). It's part of the Web History API and is available in all browser environments without needing to import it._"

dcschreiber: Regarding putting it in ReaderApp, how would I go about that?
I could create a callback that will update the history from ReaderApp instead of Sheet. Do you think that would be more correct?

YishaiGlasner: my bad :man-facepalming: 
now, i don't know. we have really complicated router that controled by ReaderApp. the question is if this fits with it.
anyway, now it seems to me this is a patch. what's the uniqueness of this page that we need to change the title like that?

dcschreiber: When we are doing client-side navigation (not SSR) the title isn't cached on time and therefore isn't being set correctly in History.
I spent a bunch of time trying to find a more "correct" place for this and didn't find.
If you have a specific idea, I'd be happy to check it out

YishaiGlasner: client side navigation is controled by ReaderApp router

static/js/ReaderApp.jsx
```diff
@@ -428,6 +428,13 @@ class ReaderApp extends Component {
           return true; // both are set, compare arrays
         }
       }
+
+      // Update history when sheet title becomes available (was showing placeholder)
+      if (next.mode === "Sheet" && prev.mode === "Sheet" && prev.sheetID === next.sheetID
```
YishaiGlasner: can we remove `prev.mode === "Sheet"`? if next is sheet and prev has the sane sheetID, i guess it means also prev is sheet (but i don't know and maybe you also).
maybe there are more things that can be removed. this is very long condition. it can be broken intoe functions with meaningfull names.

dcschreiber: hows the new code?

PR #2926, opened 2025-12-10
Bug/sc 39280/ library notes preexisting pre mdl notes on a

stevekaplan123:
## Description
If there is a note on a sheet, link it to voices module.  The previous functionality was actually wrong.  I've now added a function `getNoteURL` because the behavior is quite different.  This comment explains:
    // This helper function returns the a tag linking to the given ref of the note  When the URL is to a ref in the library, 
    // returning the appropriate a tag is very simple, but when the ref is to a sheet, we need to modify the ref and set the data-target-module attribute to the voices module.
    // If the ref is to a sheet, the ref will be something like "Sheet 3.4" but it needs to link to "/sheets/3.4" in the voices module.

### DIFF
--- a/static/js/sefaria/util.js
+++ b/static/js/sefaria/util.js
@@ -294,11 +294,8 @@ class Util {
                 window.location.pathname + window.location.search;
     }
     static fullURL(relativePath, moduleTarget, params=null) {
-      if (!relativePath.startsWith("/")) {
-        return relativePath;
-      }
       const baseURL = Sefaria.getModuleURL(moduleTarget);
-      if (!baseURL) {
+      if (!relativePath.startsWith("/") || !baseURL) {
         return relativePath;
       }
       try {


static/js/NoteListing.jsx
```diff
@@ -34,7 +34,9 @@ class NoteListing extends Component {
   }
   render() {
     var data = this.props.data;
-    var url  = "/" + Sefaria.normRef(data.ref) + "?with=Notes";
+    const module = Sefaria.isSheetRef(data.ref) ? Sefaria.VOICES_MODULE : Sefaria.LIBRARY_MODULE;
+    const path = "/" + Sefaria.normRef(data.ref) + "?with=Notes";
+    const url = Sefaria.getModuleURL(module).origin + path;
```
YishaiGlasner: please use js fnctions for url parsing

stevekaplan123: Which function do you have in mind?  Ones in our util.js files or JS library?

YishaiGlasner: something like:
```
const url = new URL(Sefaria.normRef(data.ref), Sefaria.getModuleURL(module).origin);
url.searchParams.set("with", "Notes");
const somename = url.toString();
```

static/js/NoteListing.jsx
```diff
@@ -32,28 +32,45 @@ class NoteListing extends Component {
...
+    if (isSheet) {
+      const parsedRef = Sefaria.parseRef(ref);
+      path = `/sheets/${parsedRef.sections.join('.')}`;
+      noteURL = new URL(path, Sefaria.getModuleURL(module));
+    }
+    else {
+      path = `/${ref}`;
+      noteURL = new URL(path, Sefaria.getModuleURL(module));
+      noteURL.searchParams.set('with', 'Notes');  // side panel should only open to show notes in Library module
+    }
```
YishaiGlasner: you can insert the module into the if-else blocks to avoid checking isSheet twice.
it still will have `noteURL = new URL(path, Sefaria.getModuleURL(module));` twich which i really don't like, but i didn't find the right way to do it. it's worth to give it another think. maybe some of the logic here can be generalized to a separate helper function in Sefaria.

stevekaplan123: OK I just pushed again.  Let me know what you think of my re-factor.  I used a helper function we already had with some changes.

static/js/sefaria/util.js
```diff
@@ -293,10 +293,26 @@ class Util {
...
-        ? Sefaria.getModuleURL(moduleTarget).origin + relativePath
-        : relativePath;
+    static fullURL(relativePath, moduleTarget, params=null) {
+      if (!relativePath.startsWith("/")) {
+        return relativePath;
+      }
+      const baseURL = Sefaria.getModuleURL(moduleTarget);
+      if (!baseURL) {
+        return relativePath;
+      }
```
YishaiGlasner:
```suggestion
      const baseURL = Sefaria.getModuleURL(moduleTarget);
      if (!relativePath.startsWith("/") || !baseURL) {
        return relativePath;
      }
```

static/js/NoteListing.jsx
```diff
@@ -32,28 +32,44 @@ class NoteListing extends Component {
...
+      module = Sefaria.VOICES_MODULE;
+      const parsedRef = Sefaria.parseRef(ref);
+      path = `/sheets/${parsedRef.sections.join('.')}`;
+      noteURL = Sefaria.util.fullURL(path, module);
+    }
+    else {
+      module = Sefaria.LIBRARY_MODULE;
+      path = `/${ref}`;
+      noteURL = Sefaria.util.fullURL(path, module, {"with": "Notes"});  
+    }
```
YishaiGlasner:
```suggestion
    let path, noteURL, module, params;
    if (Sefaria.isSheetRef(ref)) {
      module = Sefaria.VOICES_MODULE;
      const parsedRef = Sefaria.parseRef(ref);
      path = `/sheets/${parsedRef.sections.join('.')}`;
    }
    else {
      module = Sefaria.LIBRARY_MODULE;
      path = `/${ref}`;
      params = {"with": "Notes"};  
    }
    noteURL = Sefaria.util.fullURL(path, module, params);
```

YishaiGlasner (approved):

PR #2931, opened 2025-12-15
Feature/sc 38589/change the shade of green used for voices

stevekaplan123:
## Description
Change midrash-green to sheets-green in voices and change sheets-green to be a different color.


YishaiGlasner (approved):
why do we need all the identical samech svgs?

PR #2933, opened 2025-12-16
Bug/sc 39568/ voices export to google on initial signup if do

stevekaplan123:
## Description
Error handling for exporting sheets to google docs at multiple levels.  A number of possible edge cases exist such as if the user's token has expired (called an 'invalid_grant'), if the user clicks "Cancel" midway through the process of setting up google auth ('access_denied'), or if the user doesn't click the checkbox when trying to set up google auth (which causes the obscure "scope_mismatch" originally discussed in the corresponding story).

## Code Changes
We've already been using "#" in the URL because it doesn't get lost when modifying history.  I decided to piggy back on this existing mechanism by adding a "gauth_error" parameter in the hash string.  These can be generated at several stages in the back-end process, and I defined three errors in GAUTH_ERROR_MESSAGES that can show up to the user in any of these cases.

### DIFF
--- a/sefaria/gauth/decorators.py
+++ b/sefaria/gauth/decorators.py
@@ -28,6 +28,7 @@ def gauth_required(scope, ajax=False):
             # Try grabbing credential from storage
             profile = UserProfile(user_obj=request.user)
             credentials_dict = profile.gauth_token
+            
             if credentials_dict is None or not set(scope).issubset(set(credentials_dict['scopes'])):
                 # If no credentials or scopes don't match, redirect to authentication
                 request.session['next_view'] = request.path


sefaria/gauth/decorators.py
```diff
@@ -30,13 +28,14 @@ def inner(request, *args, **kwargs):
             # Try grabbing credential from storage
             profile = UserProfile(user_obj=request.user)
             credentials_dict = profile.gauth_token
-
-            if credentials_dict is None or not set(scope).issubset(set(credentials_dict['scopes'])):
+            needs_auth = not (credentials_dict is not None and set(scope).issubset(set(credentials_dict.get('scopes', []))))
```
stevekaplan123: I thought this was clearer but no functionality is changed with this variable

yodem: maybe two lines would be better:
`existing_scopes = credentials_dict.get('scopes', []) if credentials_dict else []

needs_auth = not set(scope).issubset(existing_scopes)`

static/js/sheets/SheetModals.jsx
```diff
@@ -137,6 +137,22 @@ const SaveModal = ({historyObject, close}) => {
...
+    en: "You declined permission to connect with Google. Export requires Google Drive access.",
+    he: "סירבת לתת הרשאה להתחבר לגוגל. הייצוא דורש גישה ל-Google Drive."
+  },
+  invalid_grant: {
+    en: "The authorization expired or was already used. Please try again.",
+    he: "ההרשאה פגה או כבר נוצלה. אנא נסה שוב."
+  },
+  scope_mismatch: {
+    en: "There was a permission mismatch. Please try again.",
+    he: "הייתה אי התאמה בהרשאות. אנא נסה שוב."
```
yodem: maybe its worth letting mickey see these error messages just to make sure?

static/js/sheets/SheetModals.jsx
```diff
@@ -149,24 +165,46 @@ const GoogleDocExportModal = ({ sheetID, close }) => {
...
+      const hashParams = Sefaria.util.parseHash(window.location.hash);
+      const gauthError = hashParams.gauth_error;
+      
+      if (gauthError) {
+        // Found a gauth_error in the hash, get the appropriate error message
+        // Remove gauth_error from hash, keeping other hash params like afterLoading
+        delete hashParams.gauth_error;
+        const remainingHash = Object.entries(hashParams)
+          .map(([k, v]) => `${k}=${v}`)
+          .join('&');
```
yodem: Sefaria might do something custom with the params, but you can use
`const remainingHash = new URLSearchParams(hashParams).toString();`

static/js/sheets/SheetModals.jsx
```diff
@@ -149,24 +165,46 @@ const GoogleDocExportModal = ({ sheetID, close }) => {
...
+          } else {
+            // Export succeeded
+            setGoogleDriveLink(data.webViewLink);
+            setGoogleDriveText(googleDriveState.exportComplete);
+          }
+        } catch (error) {
+          setGoogleDriveText({
+            en: "A network error occurred. Please check your connection and try again.",
+            he: "אירעה שגיאת רשת. אנא בדוק את החיבור שלך ונסה שוב."
+          });
```
yodem: I dont think its should be here, maybe in strings or as a const above as you did with GAUTH_ERROR_MESSAGES

static/js/sheets/SheetModals.jsx
```diff
@@ -177,12 +215,16 @@ const GoogleDocExportModal = ({ sheetID, close }) => {
   const getExportMessage = () => {
     if (currentlyExporting()) {
       return <InterfaceText text={googleDriveText}/>;
-    } else {
+    } else if (googleDriveLink) {
+      // Success - show link
       return <>
         <InterfaceText text={googleDriveText}/>&nbsp;
```
yodem: &nbsp isnt best practice.
wrapping it with a div with flex, gap would be a better solution

static/js/sheets/SheetModals.jsx
```diff
@@ -177,12 +215,16 @@ const GoogleDocExportModal = ({ sheetID, close }) => {
   const getExportMessage = () => {
     if (currentlyExporting()) {
```
yodem: it looks like the logic is:
```
 if (!currentlyExporting() && googleDriveLink) {
 <>
        <InterfaceText text={googleDriveText}/>&nbsp;
        <a href={googleDriveLink} target="_blank" className="successMessage"><InterfaceText>View in Google
          Docs</InterfaceText></a>
      </>
} else {
      return <InterfaceText text={googleDriveText}/>;

}
```

static/js/sheets/SheetModals.jsx
stevekaplan123: thanks Ill let him know

sefaria/gauth/views.py
```diff
@@ -46,14 +57,19 @@ def index(request):
...
 
 @login_required
 def auth_return(request):
     """
     Step 2 of Google OAuth 2.0 flow.
     """
+    # Check for OAuth errors from Google (e.g., user denied access)
+    oauth_error = request.GET.get('error', None)
+    if oauth_error:
+        if oauth_error in GAUTH_ERROR_CODES:
```
yitzhakc: I don't quite understand the benefit of having the outer condition (`if oauth_error`).
Why not `if oauth_error is not None and oauth_error in GAUTH_ERROR_CODES`?
Or even without `is not None`...

static/js/sheets/SheetModals.jsx
yitzhakc: Is there a reason we can't store these translations along with all the other translated strings, and use the keys to lookup instead? I don't love having these translation strings dispersed across our codebase

sefaria/gauth/decorators.py
stevekaplan123: This seemed to change the functionality when I tried to split it up.  I decided to just revert to the original code in this case as I think it's just as clear

PR #2934, opened 2025-12-17
chore: use en instead of primaryTitle

stevekaplan123:
## Description
Collections for Nechama Leibowitz couldn't be navigated.

## Code Changes
Changed `topic.primaryTitle.en` to `topic.en`.  Normally our pattern is `topic.primaryTitle.en` because `topic.contents()` returns the structure `topic.primaryTitle.en`.  However, sheet topics are actually stored in our database in the sheets collection with the structure `topic.en` or `topic.he`, so we should use this instead.

### DIFF
--- a/static/js/CollectionPage.jsx
+++ b/static/js/CollectionPage.jsx
@@ -124,7 +124,7 @@ class CollectionPage extends Component {
     filter = n(filter);
     
     //title and each topic in he and en
-    let filterableData  = sheet.topics.map(topic => [n(topic.en), n(topic.he), n(topic.asTyped)]).flat();
+    const filterableData  = sheet.topics.map(topic => [n(topic.en), n(topic.he), n(topic.asTyped)]).flat();
     filterableData.push(n(sheet.title.stripHtml()));
     
     //this may be confusing- in the exact case, "includes" is an array func and returns if any of the above match filter exactly, 


static/js/CollectionPage.jsx
```diff
@@ -124,7 +124,7 @@ class CollectionPage extends Component {
     filter = n(filter);
     
     //title and each topic in he and en
-    let filterableData  = sheet.topics.map(topic => [n(topic.primaryTitle.en), n(topic.primaryTitle.he), n(topic.asTyped)]).flat();
+    let filterableData  = sheet.topics.map(topic => [n(topic.en), n(topic.he), n(topic.asTyped)]).flat();
```
yodem: I know its an old code, but maybe we should consider changing `let` to `const` wherever we can.
what do you think?

yodem: Also, what about adding Optional chaining (?.) ? 
Can it be that a sheet wont have topic? so maybe add:
`sheet?.topics.map`

dcschreiber: I'm answering because I want to get this PR through before I can do my own. 
The variable is edited so it can't be a const.

dcschreiber: If I understand from your comment, you actually mean `sheet.topics?.map`, right?
Also I'll add that some topics didn't have a `primaryTitle`. Are we sure they all have .en / .he. if not we may want to do `topic?.en || ""` (Although I do see that this branch solves the bug I'm working on, so maybe this isn't needed. )

yodem: @dcschreiber I think `sheet?.topics.map ` in case topics would be undefined.  if topics would return, empty or not, `.map` is still valid

yodem: its a mutation change not reassignment. it can be done with const

stevekaplan123: The topic should have an 'en' and 'he' in this case as it is stored in the sheets topic key.  I'll add the chaining for sheet?.topics though as there might not be a topics key

stevekaplan123: Actually, now I see there is no reason to as 'topics' is always set to a list at this point in the code even if there are no topics.

PR #2935, opened 2025-12-17
chore: comment out operations

stevekaplan123:
## Description
Commented out migration operations.

## Code Changes
Works with postgres2 cauldron.  All I had to do was add the modularization main helm chart to postgres2.yaml and specify that it has its own postgres db, in addition to the code changes in this PR.
The cauldron behaves normally and in the logs, you can see that the migrations ran fine:
``` 
  Apply all migrations: admin, auth, contenttypes, django_topics, guides, remote_config, sessions, sites
  Applying django_topics.0012_rename_sheets_pool_to_voices... OK
  Applying django_topics.0013_revert_voices_to_sheets... OK 
```


yitzhakc (approved):

django_topics/migrations/0012_rename_sheets_pool_to_voices.py
```diff
@@ -40,5 +40,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(rename_sheets_to_voices, reverse_voices_to_sheets),
+        # migrations.RunPython(rename_sheets_to_voices, reverse_voices_to_sheets),
```
yitzhakc: Please add a comment explaining why we need to comment this out

yitzhakc (changes requested):

django_topics/migrations/0012_rename_sheets_pool_to_voices.py
stevekaplan123: Added comments

check-sandbox.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
yitzhakc: I don't think you meant to commit this?

yitzhakc (changes requested):

PR #2940, opened 2025-12-23
feat(AiInfoTooltip): update tooltip to accept custom display text

yodem:
[sc-39626]
## Description

Makes `AiInfoTooltip` accept custom text while keeping a default. Sets the default to "Some of the text on this page has been AI generated." for TopicPage, and allows GuideBox to pass context-specific text about the Learning Guide.

### `static/js/Misc.jsx`
- Modified `AiInfoTooltip` to accept an optional `displayText` prop with default parameter
- Default text: "Some of the text on this page has been AI generated." (both English and Hebrew)
- Removed `&nbsp;` spacing between text and links
- Added optional chaining (`?.`) when accessing `displayText.en` and `displayText.he` properties
- Component now uses `displayText` prop if provided, otherwise falls back to default

### `static/js/GuideBox.jsx`
- Added `DISPLAY_TEXT` constant containing the Learning Guide-specific message
- Updated `AiInfoTooltip` usage to pass custom `displayText` prop with context-specific text: "The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know."

### `static/js/TopicPage.jsx`
- No changes required - continues to use `<AiInfoTooltip/>` without props, which now displays the new default text

## Notes

- The component maintains backward compatibility - existing usages without props will automatically use the new default text
- The default text is more generic and appropriate for TopicPage context
- GuideBox maintains its specific messaging about Learning Guides
- All text continues to use the `Sefaria._()` translation function for proper internationalization
- The component structure (InterfaceText, EnglishText, HebrewText, links) remains unchanged

### DIFF
--- a/static/js/Misc.jsx
+++ b/static/js/Misc.jsx
@@ -1420,7 +1420,7 @@ const AiFeedbackLink = ({lang}) => {
 }
 
 const AiInfoTooltip = ({ displayText = "Some of the text on this page has been AI generated." }) => {
-  const [showMessage, setShowMessage] = useState(false);  
+  const [showMessage, setShowMessage] = useState(false);
   const aiInfoIcon = (
       <img
           className="ai-info-icon"
@@ -1434,10 +1434,10 @@ const AiInfoTooltip = ({ displayText = "Some of the text on this page has been A
   const aiMessage = (
       <div className="ai-info-messages-box" onMouseEnter={() => setShowMessage(true)} onMouseLeave={() => setShowMessage(false)}>
           <div className="ai-info-first-message">
-          <InterfaceText context="AiInfoTooltip">
-              {displayText}
-          </InterfaceText>
-          <AiLearnMoreLink />
+            <InterfaceText context="AiInfoTooltip">
+                {displayText}
+            </InterfaceText>
+            <AiLearnMoreLink />
         </div>
         <hr className="ai-info-messages-hr" />
         <div className="ai-info-last-message">


static/js/sefaria/strings.js
```diff
@@ -701,6 +701,7 @@ const Strings = {
     },
     "AiInfoTooltip": {
       "The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know.": "השאלות והתשובות הכלולות במדריך הלמידה נכתבו בעזרת בינה מלאכותית ונבדקו על ידי צוות הלמידה של ספריא. מצאתם שגיאה? שלחו לנו הודעה.",
+      "Some of the text on this page has been AI generated.": "חלק מהטקסט בעמוד זה נוצר על ידי בינה מלאכותית.",
```
dcschreiber, line 704: Was this copy given to you?
I'm asking because `generated by AI` seems more correct to me.
Also, I think that so far all the AI content was human reviewed.

static/js/Misc.jsx
```diff
@@ -1435,10 +1438,12 @@ const AiInfoTooltip = () => {
       <div className="ai-info-messages-box" onMouseEnter={() => setShowMessage(true)} onMouseLeave={() => setShowMessage(false)}>
             <div className="ai-info-first-message">
             <InterfaceText>
```
dcschreiber: You can just send `InterfaceText` the EN text with the prop `context` and it will translate with `Sefaria._()`
This way you can simplify the whole flow

static/js/GuideBox.jsx
```diff
@@ -183,7 +184,10 @@ class GuideBox extends Component {
           </div>
           <div className="guideHeaderTags">
             <span className="experimentLabel">Experiment</span>
-            {this.state.promptState !== COMMENTARIES && <AiInfoTooltip/>}
+            {this.state.promptState !== COMMENTARIES && <AiInfoTooltip displayText={{
```
dcschreiber: If you decide against sending the EN text to `InterfaceText` (see comment in Misc) then you can call `Sefaria._()` here and then the translation is decided before `AiInfoTooltip`

static/js/sefaria/strings.js
yodem: Yes its from production (at least was in production).
you can see it here - https://www.modularization.cauldron.sefaria.org/topics/peace?tab=admin

static/js/Misc.jsx
```diff
@@ -1434,11 +1434,13 @@ const AiInfoTooltip = () => {
   const aiMessage = (
       <div className="ai-info-messages-box" onMouseEnter={() => setShowMessage(true)} onMouseLeave={() => setShowMessage(false)}>
             <div className="ai-info-first-message">
-            <InterfaceText>
-                <EnglishText>{Sefaria._("The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know.", "AiInfoTooltip")}
-                  &nbsp;<AiLearnMoreLink lang="english" />
+            <InterfaceText context="AiInfoTooltip">
+                <EnglishText>
```
dcschreiber: I don't think you need `EnglishText`. Check but I think it will work without it and then completely remove the `HebrewText`.
Also, don't you need to add the context `AiInfoTooltip`?
Did you test this, I'm not sure it will work like this.

static/js/Misc.jsx
```diff
@@ -1434,11 +1434,13 @@ const AiInfoTooltip = () => {
...
+                  {displayText}
+                  <AiLearnMoreLink lang="english" />
```
dcschreiber: Could you not hardcode the lang?
We have functions for this I think, probably in `Sefaria`

static/js/Misc.jsx
```diff
@@ -1434,11 +1434,13 @@ const AiInfoTooltip = () => {
...
                 </EnglishText>
-                <HebrewText>{Sefaria._("The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know.", "AiInfoTooltip")}&nbsp;
+                <HebrewText>
+                  {displayText}
```
dcschreiber: I'm not sure this will work, did you test it?
I think you have to either remove the `hebrew/englishText` and just put english or use `Sefaria._()`

dcschreiber (changes requested):
I have one main comment about removing `HebrewText`/`EnglishText` if possible and testing it works. but looking good

static/js/Misc.jsx
yodem: Hebrew and English text components does basically nothing:
`const HebrewText = ({children}) => (
    <>{children}</>
);
const EnglishText = ({children}) => (
    <>{children}</>
);`

It was there before and its being used in a lot of places in our code so I decided to leave it as is (dont mind changing it, but it looks like an accepted pattern here)

second, the context should be added to the InterfaceText because it is the only component that uses it.

And yes I did test it, you can also check the cauldrons here - https://app.shortcut.com/sefaria/story/39626/ai-marker-tooltip-change-for-pirkei-avot-also-impacted-the-marker-on-topics-pages#activity-39710

static/js/Misc.jsx
yodem: I did test it and it does work.
 If you look at InterfaceText at line 116 - `textResponse = Sefaria._(children, context);` it does it automatically

yodem:
@dcschreiber InterfaceText automatically uses Sefaria._() - `textResponse = Sefaria._(children, context);` if there is one children. so both in AiLearnMoreLink and aiMessage i dont even mention the lang, I just use InterfaceText with context

static/js/sefaria/strings.js
```diff
@@ -701,6 +701,8 @@ const Strings = {
...
+      "Learn More": "לפרטים נוספים",
```
dcschreiber, line 705: If you could be bothered, we already have 2 "Learn More" strings. Maybe we can consolidate.
But no pressure on this

static/js/Misc.jsx
```diff
@@ -1431,18 +1430,14 @@ const AiInfoTooltip = () => {
...
-                <EnglishText>{Sefaria._("The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know.", "AiInfoTooltip")}
-                  &nbsp;<AiLearnMoreLink lang="english" />
-                </EnglishText>
-                <HebrewText>{Sefaria._("The questions and answers in this Learning Guide have been written and curated by AI with human review. Something not right? Let us know.", "AiInfoTooltip")}&nbsp;
-                  <AiLearnMoreLink lang="hebrew" />
-                </HebrewText>
-            </InterfaceText>
-
+          <div className="ai-info-first-message">
+          <InterfaceText context="AiInfoTooltip">
```
dcschreiber: indentation is off

static/js/Misc.jsx
```diff
@@ -1420,8 +1419,8 @@ const AiFeedbackLink = ({lang}) => {
   );
 }
 
-const AiInfoTooltip = () => {
-  const [showMessage, setShowMessage] = useState(false);
+const AiInfoTooltip = ({ displayText = "Some of the text on this page has been AI generated." }) => {
+  const [showMessage, setShowMessage] = useState(false);
```
dcschreiber: If you are already making whitespace changes could you remove the extra space. It doesn't really matter this time, this is just a note because white space changes happen a lot with AI edits and they make the diff harder

static/js/sefaria/strings.js
yodem: we have 3 with the mdl...
its not the same context so its ok, but it is a product matter that need to be fixed

dcschreiber: Maybe you can just note it to someone and if they want to change it they can open a new ticket. again, no pressure

PR #2942, opened 2025-12-23
fix(CollectionPage): added optional chaining to avoid error when topi…

dcschreiber:
Problem:
The bug was that we had empty `topic.primaryTitle` and then we got an error when trying to access `topic.primaryTitle.en`
Solution:
Added safety catches for cases where `topic.primaryTitle` or `topic.asTyped` are empty

Story details: https://app.shortcut.com/sefaria/story/39659

### DIFF
--- a/static/js/CollectionPage.jsx
+++ b/static/js/CollectionPage.jsx
@@ -124,7 +124,7 @@ class CollectionPage extends Component {
     filter = n(filter);
     
     //title and each topic in he and en
-    let filterableData  = sheet.topics.map(topic => [n(topic.primaryTitle.en), n(topic.primaryTitle.he), n(topic.asTyped)]).flat();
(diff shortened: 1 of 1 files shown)


static/js/CollectionPage.jsx
```diff
@@ -124,7 +124,7 @@ class CollectionPage extends Component {
     filter = n(filter);
     
     //title and each topic in he and en
-    let filterableData  = sheet.topics.map(topic => [n(topic.primaryTitle.en), n(topic.primaryTitle.he), n(topic.asTyped)]).flat();
+    let filterableData  = sheet.topics.map(topic => [n(topic.primaryTitle?.en || ''), n(topic.primaryTitle?.he || ''), n(topic.asTyped || '')]).flat();
```
YishaiGlasner, line 127: i think it will be better to have `getPrimaryTitle(lang)` func

YishaiGlasner (changes requested):

PR #2984, opened 2026-01-08
feat(modtools): Backend foundation for bulk version editing (Part 1/3)

dcschreiber:
# Part 1 of 3: Backend Foundation

**Related PRs:**
- Part 2: UI Refactoring → #2985
- Part 3: BulkVersionEditor Feature → #2986

**Merge order:** PR1 → PR2 → PR3 (all target master, merge in sequence)

**!! Don't Merge into Master Until After MDL and Post MDL !!**

---

## Summary

This PR adds the backend foundation:
- PyMongo 4.x compatibility fixes (`update()` → `update_many()`)
- Activity feed logging for version metadata changes (new rev_type: `edit version_metadata`)
- New API endpoints for bulk version operations
- Enhanced Index API error handling
- Minor fixes for export and link helpers

---

## New API Endpoints (sefaria/views.py)

Three new endpoints were added to support bulk version editing:

**`/api/version-indices` (GET)**
- Returns a list of indices that have versions matching a specific versionTitle
- Used by BulkVersionEditor to show which indices a version appears in
- Example: `?versionTitle=Tanach with Nikkud&language=he` → `{indices: ["Genesis", "Exodus", ...], metadata: {...}}`
- Includes metadata with categories for each index (uses cached `library.get_index()` for efficiency)

**`/api/version-bulk-edit` (POST)**
- Bulk updates Version metadata across multiple indices
- Request: `{versionTitle, language, indices: [...], updates: {...}}`
- Response: `{status: "ok"|"partial"|"error", successes: [...], failures: [{index, error}, ...]}`
- **Security**: Whitelisted fields (`VERSION_BULK_EDIT_ALLOWED_FIELDS`) prevent arbitrary attribute injection
- **Partial success handling**: Processes all indices even if some fail, returning detailed success/failure arrays
- **Field clearing**: `null` values are treated as "delete this field" - backend uses `delattr` to remove from MongoDB
- **History logging**: Uses `tracker.update_version_metadata()` to log changes to activity feed

**`/api/check-index-dependencies/<title>` (GET)** - Disabled in frontend
- Checks dependencies before index title changes
- Used by NodeTitleEditor (currently disabled)
- Returns: dependent indices, version count, link count, has_dependencies flag

---

## Activity Feed / History Logging

Added audit logging for version metadata changes. Previously, metadata edits (license, status, priority, etc.) were not tracked anywhere in the system.

**New rev_type: `edit version_metadata`**

Changes are now logged to the MongoDB `history` collection and displayed on the `/activity` page.

**Implementation:**

| File | Addition |
|------|----------|
| `sefaria/model/history.py` | `log_version_metadata(user, old_dict, new_dict, **kwargs)` - Creates history entry with old/new values, title, version, language |
| `sefaria/model/__init__.py` | Added `log_version_metadata` to exports |
| `sefaria/tracker.py` | `update_version_metadata(user, version, updates, **kwargs)` - Centralized function: captures old state → applies updates → saves → logs to history |

**Endpoints updated to use tracker:**
- `version_bulk_edit_api` (sefaria/views.py) - The new bulk editor
- `flag_text_api` (reader/views.py) - Flag text feature
- `lock_text_api` (reader/views.py) - Lock/unlock text feature

**Activity Feed Display:**
- `templates/elements/activity_feed.html` - Added template block for version_metadata events
- `templates/activity.html` - Added "Version Metadata" filter option to dropdown
- `sefaria/history.py` - Added `version_metadata` case to `filter_type_to_query()`

---

## Enhanced reader/views.py

**Index API Error Handling:**
Enhanced the `index_post` function with categorized error responses:
- Supports both old format (`{update: {...}}`) and new format (direct index data)
- Dependency pre-checks when renaming indices
- Categorized errors: `dependency_error`, `validation_error`, `title_validation_error`, `general_error`
- **Backward-compatible**: Works with both request formats
- **Enhanced error display**: Changed from semicolon-separated strings to bullet-list format with newlines

**Version Metadata Logging:**
Updated `flag_text_api` and `lock_text_api` to use `tracker.update_version_metadata()` for history logging.

---

## PyMongo 4.x Migration

Updated deprecated MongoDB operations:
- `sefaria/model/history.py:210`: `update()` → `update_many()`
- `sefaria/helper/text.py:322`: `update()` → `update_many()`
- Both maintain identical functionality with updated API

---

## Minor Improvements
- `sefaria/export.py`: Fixed missing `user_id` parameter in `import_versions_from_file()`
- `sefaria/helper/link.py`: Added bytes decoding for CSV uploads (handles both bytes and str)

---

## Files Changed

**Backend (Python):**
- `sefaria/views.py` - New API endpoints, uses tracker for history logging
- `reader/views.py` - Enhanced Index API error handling, flag_text_api and lock_text_api now use tracker
- `sefaria/urls_shared.py` - URL routing for new endpoints
- `sefaria/tracker.py` - New `update_version_metadata()` function
- `sefaria/model/history.py` - PyMongo 4.x migration, new `log_version_metadata()` function
- `sefaria/model/__init__.py` - Added `log_version_metadata` export
- `sefaria/history.py` - Added `version_metadata` filter type
- `sefaria/helper/text.py` - PyMongo 4.x migration
- `sefaria/helper/link.py` - CSV bytes handling
- `sefaria/export.py` - user_id parameter fix

**Templates:**
- `templates/elements/activity_feed.html` - Added version_metadata display block
- `templates/activity.html` - Added "Version Metadata" filter option

---

---

## Next PR

➡️ **Part 2 is a draft:** #2985 - Remove from draft status when this PR is merged.


sefaria/tracker.py
```diff
@@ -140,6 +140,38 @@ def modify_node(jagged_array_node):
...
 
 
+def update_version_metadata(user: int, version: model.Version, updates: dict, **kwargs) -> model.Version:
+    """
+    Update version metadata fields and log the change to history.
+
+    Use this for metadata-only changes (license, status, priority, etc.) that should
+    be tracked separately from text content changes. Creates 'edit version_metadata'
+    history records.
+
```
stevekaplan123, line 157: I don't think we need a special function to track this.  I believe we have other functions for tracking that kick in automatically upon save.

sefaria/model/history.py
```diff
@@ -207,4 +236,4 @@ def process_version_title_change_in_history(ver, **kwargs):
         "version": kwargs["old"],
         "language": ver.language,
     }
-    db.history.update(query, {"$set": {"version": kwargs["new"]}}, upsert=False, multi=True)
\ No newline at end of file
+    db.history.update_many(query, {"$set": {"version": kwargs["new"]}})
```
stevekaplan123, line 218: You said that update is deprecrated.  Do we use update elsewhere in the codebase?  If so, we should create a story for dealing with its deprecation.

sefaria/helper/link.py
```diff
@@ -512,7 +512,10 @@ def create_link_cluster(refs, user, link_type="", attrs=None, exception_pairs=No
 
 def add_links_from_csv(file, linktype, generated_by, uid):
     csv.field_size_limit(sys.maxsize)
-    reader = csv.DictReader(StringIO(file.read().decode()))
+    content = file.read()
+    if isinstance(content, bytes):
+        content = content.decode('utf-8')
+    reader = csv.DictReader(StringIO(content))
     fieldnames = reader.fieldnames
```
stevekaplan123, line 519: Is this backwards-compatible?  As in, does this change work with other ways of using add_links_from_csv?

reader/views.py
```diff
@@ -2352,23 +2410,28 @@ def flag_text_api(request, title, lang, version):
...
+                    updates[flag] = None  # None signals deletion in tracker
                 else:
-                    setattr(vobj, flag, flags[flag])
-        vobj.save()
-        return jsonResponse({"status": "ok"})
+                    updates[flag] = flags[flag]
 
-    _attributes_to_save = Version.optional_attrs + ["versionSource", "direction", "isSource", "isPrimary"]
+        if updates:
+            tracker.update_version_metadata(user_id, vobj, updates)
```
stevekaplan123, line 2584: I mentioned we have other ways of doing this.  I may be wrong.  Look into a few other APIs where we save index changes, text changes, etc. to see how we handle tracker.  It seems strange that we have a special tracker function for version metadata.

reader/views.py
```diff
@@ -1742,9 +1742,67 @@ def determine_user_type_and_id(request):
                     return CONTENT_TYPE, apikey["uid"]
         return None, None
 
-    def index_post(request, uid, j, method, raw):
-        func = tracker.update if 'update' in j else tracker.add
```
stevekaplan123, line 1927: tracker.update and tracker.add are the typical ways we use tracker

templates/elements/activity_feed.html
```diff
@@ -110,6 +110,14 @@
                 <pre>{{ event.old|pprint }}</pre>
             </div>
 
+        {% elif "version_metadata" in event.rev_type %}
+            <span class="topline">{{ event.user|user_link }}
+                {% trans "edited version metadata for" %}
+                {% if he %}{{ event.title|he_ref_link }}{% else %}{{ event.title|ref_link }}{% endif %}.
+                {% if event.method == "API" %} (via API) {% endif %}
+            </span>
```
stevekaplan123, line 118: What are activity_feed and activity.html and why do we show version metadata information there?

reader/views.py
```diff
@@ -1742,9 +1742,67 @@ def determine_user_type_and_id(request):
...
-        return jsonResponse(func(uid, Index, j, raw=raw, method=method).contents(raw=raw))
+    def index_post(request, uid, json_data, method, raw):
+        # NOTE: The enhanced error handling and new format support below was added for
+        # NodeTitleEditor and BulkIndexEditor, which are currently disabled in ModeratorToolsPanel.
```
stevekaplan123: I'm pretty confused.  If NodeTitleEditor is disabled in ModeratorToolsPanel, then why are we writing code for it?

stevekaplan123 (changes requested):
Does this PR do anything on its own?  If someone were to use modtools or existing functionality if this were merged into master, would everything work the same?  Otherwise, it's a problem because it will mean we have introduced a new bug(s).  What we might want to do is to create a fourth branch that each of these merges into and then at the end we merge it all into master.

reader/views.py
stevekaplan123: I see that we do have a modify_version.  We should investigate why we need this and if we really do need it, then do we also need a tracker for version_metadata

reader/views.py
stevekaplan123: Also, this is a major function to be changing.  Have you tested various cases with this?  I suspect that these checks should be put in the Index class' _validate function.  See abstract.py's "save" function to see how we save things normally.   Virtually every save of every object runs through abstract.py's "save" (and in this case, tracker.add and tracker.update call "save").  And "save" calls "_validate" and so it makes sense that these checks should be there.  We already do some validation and normalization in the Index class itself I believe and so you could add the checks there.

templates/elements/activity_feed.html
dcschreiber: It's an internal tool Shmuel uses that tracks changes to the data. Of course, it's incomplete because we do changes directly to the database, but the original thought was that this would cover all changes. 
Shmuel wants to see the metadata version changes here.

sefaria/model/history.py
dcschreiber: Created a new story
https://app.shortcut.com/sefaria/story/40354/update-the-pymongo-update-deprecated-to-use-update-one-update-many

sefaria/helper/link.py
dcschreiber: Yes, it's backwards-compatible. The original code assumed file.read() always returns bytes and unconditionally called .decode(). The new code handles both bytes and strings - it only decodes if needed.

sefaria/tracker.py
dcschreiber: So I did try to see if we can combine this new `rev_type` into the existing functions, but I didn't manage. I think Claude's summary is probably better than mine:

_Good question! tracker.update wouldn't work here because:
  1. Version.history_noun = 'text' - would create "edit text" entries (same as content changes)
  2. tracker.update uses load_from_dict(attrs) requiring the entire Version dict including the huge chapter array
  3. We need targeted field updates and field deletion support

  The notification system in abstract.py is for cascading title changes, not history logging. We could add a rev_type override to log_update, but we'd still need custom logic for the targeted updates._

reader/views.py
dcschreiber: See comment above. If you still have an intuition that this could be done, I could dive in deeper or try to find someone who understands this code.

reader/views.py
dcschreiber: The comment above: they have different types of edits. Here is Claude's summary again for this specific question: 

_modify_version and update_version_metadata serve different purposes:

  - modify_version: For bulk text import - handles chapter content and creates "edit text"/"add text" history entries. Traverses the index schema to log individual text segment changes.
  - update_version_metadata: For metadata-only changes (license, status, priority) - creates "edit version_metadata" history entries. Doesn't touch chapter content at all.

  We need both because metadata changes shouldn't appear as text edits in the history feed. Using modify_version for metadata would either:
  1. Log nothing (if chapter unchanged)
  2. Create misleading "edit text" entries_

reader/views.py
dcschreiber: You are right, I don't want to get into this if I don't need to. We were trying to preserve work done for the disabled tools but let's leave that for future tasks.
I've reverted index_post to its original simple version. Added a TODO comment at the disabled tools location noting that index_post will need enhanced error handling when re-enabling them.

sefaria/tracker.py
stevekaplan123: It seems like the only reason we need this function is because of 'log_version_metadata'.  The 'log_version_metadata' function does make sense to me but it seems to me that this part could be done elsewhere as it looks really similar to other patterns we have:
```
    old_dict = version.contents()

    for field_name, field_value in updates.items():
        if field_value is None:
            # None is sentinel for "delete this field"
            if hasattr(version, field_name):
                delattr(version, field_name)
        else:
            setattr(version, field_name, field_value)

    version.save()
```

stevekaplan123: In tracker.update, kwargs could be old_dict and attrs could be updates.  Then we just need to override this "edit {}" so that maybe it can handle this specific case: 
```
def log_update(user, klass, old_dict, new_dict, **kwargs):
    kind = klass.history_noun
    rev_type = "edit {}".format(kind)
    return _log_general(user, kind, old_dict, new_dict, rev_type, **kwargs)
```

stevekaplan123: In the kwargs, maybe we can override history_noun

dcschreiber: Addressed in the conversation we had together.

stevekaplan123 (dismissed):

stevekaplan123 (dismissed):

reader/views.py
```diff
@@ -2474,11 +2474,11 @@ def lock_text_api(request, title, lang, version):
...
 
     if request.GET.get("action", None) == "unlock":
-        vobj.status = None
+        updates = {"status": None}  # None signals deletion in tracker
     else:
-        vobj.status = "locked"
+        updates = {"status": "locked"}
 
-    vobj.save()
+    tracker.update_version_metadata(request.user.id, vobj, updates)
```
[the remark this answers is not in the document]
stevekaplan123: @dcschreiber Do we have validation anywhere that the user must add existing versions?  I suspect we already have this validation somewhere in one of the PRs, so that CoPilot's comment here is irrelevant but I just wnat to confirm.

reader/views.py
```diff
@@ -2494,23 +2494,28 @@ def flag_text_api(request, title, lang, version):
...
+                    updates[flag] = None  # None signals deletion in tracker
                 else:
-                    setattr(vobj, flag, flags[flag])
-        vobj.save()
-        return jsonResponse({"status": "ok"})
+                    updates[flag] = flags[flag]
 
-    _attributes_to_save = Version.optional_attrs + ["versionSource", "direction", "isSource", "isPrimary"]
+        if updates:
+            tracker.update_version_metadata(user_id, vobj, updates)
```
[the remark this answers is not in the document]
stevekaplan123: Ditto

sefaria/views.py
```diff
@@ -1670,6 +1671,114 @@ def text_upload_api(request):
...
+    return jsonResponse({"indices": sorted_indices, "metadata": metadata})
+
+
+# Allowed fields for version bulk edit - prevents arbitrary attribute injection
+# Note: versionTitle is excluded - it's the search key used to find versions, not editable
+VERSION_BULK_EDIT_ALLOWED_FIELDS = {
+    "versionTitleInHebrew", "versionSource", "license", "status",
+    "priority", "digitizedBySefaria", "isPrimary", "isSource", "versionNotes",
+    "versionNotesInHebrew", "purchaseInformationURL", "purchaseInformationImage", "direction"
+}
```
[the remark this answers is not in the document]
stevekaplan123: This is a really good point.  Are None values currently possible in required fields when using the tool?  That is, is it possible to delete data that is actually required for versions to work properly?

stevekaplan123 (changes requested):
I agreed with some of CoPilot's comments and left comments of my own.

sefaria/views.py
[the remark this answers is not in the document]
dcschreiber: [Claude] Yes, the UI in PR3 allows clearing required fields. _validate() would reject it on save, but with a confusing error. Added early validation with a clear message.

reader/views.py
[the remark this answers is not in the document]
dcschreiber: [Claude] No, this validation didn't exist anywhere - Added if not vobj checks to both.

reader/views.py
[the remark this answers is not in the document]
dcschreiber: Fixed, see the other comment

yitzhakc (dismissed):

stevekaplan123 (dismissed):

PR #2988, opened 2026-01-11
implement Elasticsearch reindexing cronjob

yodem:
## Description
Implement a centralized logging system for Elasticsearch reindexing that toggles between `INFO` and `DEBUG` levels via a CLI flag, and refactor the reindexing cronjob for better reliability and visibility.

### `Sefaria-Project/sefaria/search.py`
* Added `setup_logging(debug=False)` utility to centralize `logging.basicConfig` configuration.
* Replaced `structlog` and `print` statements with standard `logging` calls.
* Moved verbose operational logs (index creation, mapping application, version priority mapping) to `DEBUG` level.
* Added detailed failure tracking to `TextIndexer` class for improved reporting.
* Improved error handling and null-safety during sheet and text indexing.

### `Sefaria-Project/scripts/scheduled/reindex_elasticsearch_cronjob.py`
* Added `argparse` support with a `-d/--debug` flag to trigger detailed logging.
* Implemented `ReindexingResult` class to track and report detailed successes, failures, and warnings.
* Refactored main execution flow into discrete steps (PageSheetRank, Full Index, Sheet Catch-up) with individual error handling.
* Moved non-essential progress logs to `DEBUG` level, keeping `INFO` output clean and milestone-focused.
* Added pre-flight checks for Elasticsearch connectivity and index states.

## Notes
* The new logging system uses `logging.basicConfig(force=True)` to ensure configuration consistency across the project when `search.py` is imported.
* Operational output is significantly quieter by default, focusing only on major milestones and summary reports.

### DIFF
--- a/scripts/scheduled/reindex_elasticsearch_cronjob.py
+++ b/scripts/scheduled/reindex_elasticsearch_cronjob.py
@@ -141,9 +141,9 @@ class ReindexingResult:
             return ""
         
         lines = [
-            "\n" + "=" * 80,
+            "\n" + SEPARATOR_LINE,
             "DETAILED FAILURE REPORT",
-            "=" * 80,
+            SEPARATOR_LINE,
         ]
         
         if self.failed_text_versions:
@@ -173,7 +173,7 @@ class ReindexingResult:
             if len(self.skipped_text_versions) > 20:
                 lines.append(f"... and {len(self.skipped_text_versions) - 20} more")
         
-        lines.append("=" * 80)
+        lines.append(SEPARATOR_LINE)
         return "\n".join(lines)
 
 
@@ -289,6 +289,30 @@ def run_index_all(result: ReindexingResult) -> bool:
         return False
 
 
+def should_retry_with_backoff(attempt: int, max_retries: int, context: str = "") -> bool:
+    """
+    Sleep with backoff and determine if operation should be retried.
+    
+    Args:
+        attempt: Current attempt number (1-indexed)
+        max_retries: Maximum number of retry attempts
+        context: Optional context string for logging
+        
+    Returns:
+        True if should retry (after sleeping), False if max retries exceeded
+    """
+    if attempt >= max_retries:
+        return False
+    
+    wait_time = attempt * 30  # Linear backoff (30s, 60s, 90s)
+    log_msg = f"Retrying in {wait_time} seconds..."
+    if context:
+        log_msg = f"{context} - {log_msg}"
+    logger.debug(log_msg)
+    time.sleep(wait_time)
+    return True
+
+
 def run_sheets_by_timestamp(timestamp: str, result: ReindexingResult, max_retries: int = 3) -> bool:
     """
     Call the sheets-by-timestamp API with retry logic.
@@ -315,10 +339,7 @@ def run_sheets_by_timestamp(timestamp: str, result: ReindexingResult, max_retrie
                 logger.warning(f"API returned non-200 status - attempt: {attempt}, "
                       f"status_code: {r.status_code}, response_preview: {r.text[:200]}")
                 
-                if attempt < max_retries:
-                    wait_time = attempt * 30  # Linear backoff (30s, 60s, 90s)
-                    logger.debug(f"Retrying in {wait_time} seconds...")
-                    time.sleep(wait_time)
+                if should_retry_with_backoff(attempt, max_retries, f"HTTP {r.status_code}"):
                     continue
                 else:
                     result.record_step_failure("sheets_by_timestamp", error_msg)
@@ -348,16 +369,14 @@ def run_sheets_by_timestamp(timestamp: str, result: ReindexingResult, max_retrie
             
         except requests.exceptions.Timeout:
             logger.warning(f"API request timed out - attempt: {attempt}")
-            if attempt < max_retries:
-                time.sleep(attempt * 30)
+            if should_retry_with_backoff(attempt, max_retries, "Timeout"):
                 continue
             result.record_step_failure("sheets_by_timestamp", "Request timed out after all retries")
             return False
             
         except requests.exceptions.RequestException as e:
             logger.warning(f"API request failed - attempt: {attempt}, error: {str(e)}")
-            if attempt < max_retries:
-                time.sleep(attempt * 30)
+            if should_retry_with_backoff(attempt, max_retries, "RequestException"):
                 continue
             result.record_step_failure("sheets_by_timestamp", f"Request failed: {str(e)}")
             return False
--- a/sefaria/search.py
+++ b/sefaria/search.py
@@ -913,12 +913,12 @@ def index_sheets_by_timestamp(timestamp):
     failed = []
     total = len(ids)
 
-    for i, id in enumerate(ids):
-        did_succeed = index_sheet(curr_index_name, id)
+    for i, _id in enumerate(ids):
+        did_succeed = index_sheet(curr_index_name, _id)
         if did_succeed:
-            succeeded.append(id)
+            succeeded.append(_id)
         else:
-            failed.append(id)
+            failed.append(_id)
         
     # Only log if there are failures
     if len(failed) > 0:
@@ -983,7 +983,7 @@ def clear_index(index_name):
         else:
             logger.debug(f"Index does not exist, nothing to delete - index_name: {index_name}")
     except NotFoundError:
-        # Index doesn't exist (race condition or double-check), which is fine
+        # Index doesn't exist - handle race condition where index is deleted between exists() check and delete() call
         logger.debug(f"Index not found when attempting to delete - index_name: {index_name}")
     except Exception as e:
         logger.error(f"Error deleting Elasticsearch index - index_name: {index_name} - error: {str(e)}")


scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Elasticsearch Reindexing Cronjob
+
+This script performs a full reindex of Elasticsearch indexes for text and sheet content.
+It includes:
+- Verbose logging for better debugging
```
YishaiGlasner: add explanation about why print is preferred on logging

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+        log_index_state('text', result)
+        log_index_state('sheet', result)
+        result.complete_step("preflight_index_check", "Index states logged")
+    except Exception as e:
+        result.fail_step("preflight_index_check", str(e))
+        print(result.summary())
+        sys.exit(1)
+    
+    # Step 1: Update PageSheetRank
+    print("=" * 40)
```
YishaiGlasner: it isn't really matter but previously you used 60
maybe it won't be an overkill to have a global =s-line and -s-line for getting a consistent print

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+        self.end_time = None
+        self.steps_completed = []
+        self.steps_failed = []
+        self.warnings = []
+        
+        # Detailed failure tracking
+        self.failed_text_versions = []
+        self.skipped_text_versions = []
+    
+    def complete_step(self, step_name: str, details: str = None):
```
YishaiGlasner: this sounds like this func completes the step. rename it to reflect its action

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+        """Mark a step as completed."""
+        entry = {"step": step_name, "time": datetime.now().isoformat()}
+        if details:
+            entry["details"] = details
+        self.steps_completed.append(entry)
+        print(f"Step completed: {step_name}" + (f" - {details}" if details else ""))
+    
+    def fail_step(self, step_name: str, error: str):
```
YishaiGlasner: same thing
also i think `message` is better than `error`

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+        entry = {"message": message, "time": datetime.now().isoformat()}
+        if details:
+            entry["details"] = details
+        self.warnings.append(entry)
+        print(f"WARNING: {message}" + (f" - {details}" if details else ""))
+    
+    def finalize(self):
+        """Finalize the result and calculate duration."""
+        self.end_time = datetime.now()
+        self.duration = self.end_time - self.start_time
```
YishaiGlasner: add duration to init

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+    
+    def is_success(self) -> bool:
+        """Return True if no critical failures occurred."""
+        return len(self.steps_failed) == 0
+    
+    def summary(self) -> str:
+        """Generate a human-readable summary."""
+        self.finalize()
```
YishaiGlasner: this is weird that funalize is side effect of summary. if the assumption is that summary will be called only in the end you can move finalize logic to here and have no finalize.

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+    def finalize(self):
+        """Finalize the result and calculate duration."""
+        self.end_time = datetime.now()
+        self.duration = self.end_time - self.start_time
+    
+    def is_success(self) -> bool:
+        """Return True if no critical failures occurred."""
+        return len(self.steps_failed) == 0
+    
+    def summary(self) -> str:
```
YishaiGlasner:
```suggestion
    def get_summary(self) -> str:
```

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+
+
+def run_index_all(result: ReindexingResult) -> bool:
+    """Run full index with error handling and failure capture."""
+    print("Starting full index rebuild")
+    try:
+        index_all()
+        
+        # Capture failures from TextIndexer class after indexing
+        from sefaria.search import TextIndexer
```
YishaiGlasner: why not on top?

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,432 @@
...
+            )
+            
+            # Check for HTTP errors
+            if r.status_code != 200:
+                error_msg = f"HTTP {r.status_code}: {r.text[:500]}"
+                print(f"WARNING: API returned non-200 status - attempt: {attempt}, "
+                      f"status_code: {r.status_code}, response_preview: {r.text[:200]}")
+                
+                if attempt < max_retries:
+                    wait_time = attempt * 30  # Exponential backoff
```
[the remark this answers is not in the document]
YishaiGlasner: that's true. the other option is to have it linear without the comment

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,440 @@
...
+                else:
+                    result.fail_step("sheets_by_timestamp", error_msg)
+                    return False
+            
+            # Check for error in response body
+            if "error" in r.text.lower():
+                # Try to parse as JSON first
+                try:
+                    response_json = r.json()
+                    if "error" in response_json:
```
YishaiGlasner: shouldn't you have `lower` also here?

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,432 @@
...
+                        error_msg = response_json["error"]
+                        result.fail_step("sheets_by_timestamp", f"API error: {error_msg}")
+                        return False
+                except:
```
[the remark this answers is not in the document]
YishaiGlasner: or at least `except Exception`

sefaria/search.py
```diff
@@ -4,19 +4,13 @@
 
 Writes to MongoDB Collection: index_queue
```
YishaiGlasner, line 5: this is not a part of the cron job, so changing logging to print here seems to be problematic

sefaria/search.py
```diff
@@ -117,41 +109,62 @@ def index_sheet(index_name, id):
...
+    # Validate all critical fields upfront
+    owner_id = sheet.get("owner")
+    sheet_title = sheet.get("title")
+    summary = sheet.get("summary")
+    datePublished = sheet.get("datePublished")
+    dateCreated = sheet.get("dateCreated")
+    dateModified = sheet.get("dateModified")
+    
+    if not all([owner_id, sheet_title, summary is not None, datePublished is not None, 
+                dateCreated is not None, dateModified is not None]):
```
YishaiGlasner:
```suggestion
    if any(x is None for x in [summary, datePublished, dateCreated, dateModified]) 
    or not (owner_id and sheet_title):
```

sefaria/search.py
```diff
@@ -213,10 +241,34 @@ def get_stemmed_english_analyzer():
...
+            doc_count = stats.get('_all', {}).get('primaries', {}).get('docs', {}).get('count', 0)
+            
+            if doc_count > 0 and not force:
+                error_msg = f"Index {index_name} already exists with {doc_count} documents. Use force=True to recreate."
+                print(f"ERROR: Refusing to recreate index with existing data - index_name: {index_name}, doc_count: {doc_count}")
+                raise ValueError(error_msg)
+            
+            print(f"WARNING: Index exists with documents, will be cleared - index_name: {index_name}, doc_count: {doc_count}")
+        except Exception as e:
+            if "ValueError" in str(type(e)):
```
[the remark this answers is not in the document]
YishaiGlasner: do 2 except statements one for ValueError and one for Exception

sefaria/search.py
```diff
@@ -394,13 +462,19 @@ def get_search_categories(oref, categories):
 
 
 class TextIndexer(object):
+    
+    # Class-level failure tracking
+    _failed_versions = []
+    _skipped_versions = []
```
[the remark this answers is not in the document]
YishaiGlasner: !!!

sefaria/search.py
```diff
@@ -483,62 +570,188 @@ def excluded_from_search(version):
...
             if len(vlist) == 0:
                 continue
-            cls.curr_index = vlist[0].get_index()
+            
+            try:
+                cls.curr_index = vlist[0].get_index()
+            except Exception as e:
+                failed += len(vlist)
+                for v in vlist:
+                    cls._failed_versions.append({
```
YishaiGlasner: it seems you have similar or identical many times. add function for appending failed

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,432 @@
...
+        return -1
+
+
+def log_index_state(index_type: str, result: ReindexingResult):
+    """Log the current state of indexes for a given type."""
+    try:
+        names = get_new_and_current_index_names(index_type, debug=False)
+        current_index = names['current']
+        new_index = names['new']
+        alias = names['alias']
```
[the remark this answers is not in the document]
akiva10b: would be a shame if the script broke because of logging - make sure all logging is in a try-catch

sefaria/search.py
```diff
@@ -666,35 +900,71 @@ def index_sheets_by_timestamp(timestamp):
...
-    for id in ids:
-        index_sheet(index_name, id)
+    total = len(ids)
+    print(f"Found public sheets to index - total: {total}, first_10_ids: {ids[:10] if ids else []}")
+    
+    succeeded = 0
+    failed = 0
+    failed_ids = []
+    
+    for i, id in enumerate(ids):
```
YishaiGlasner: id is saved. use `_id`

sefaria/search.py
```diff
@@ -710,11 +980,21 @@ def clear_index(index_name):
...
+    print(f"Attempting to delete Elasticsearch index - index_name: {index_name}")
     try:
-        index_client.delete(index=index_name)
+        # Check if index exists before trying to delete
+        if index_client.exists(index=index_name):
+            index_client.delete(index=index_name)
+            print(f"Successfully deleted Elasticsearch index - index_name: {index_name}")
+        else:
+            print(f"Index does not exist, nothing to delete - index_name: {index_name}")
+    except NotFoundError:
```
YishaiGlasner, line 985: isn't this checked by `if index_client.exists(index=index_name)`?

yodem: Cursor suggested that we should keep it, as another process might have been created after the last check. I don’t think that will ever happen, but I don’t mind keeping it since it is technically possible; the job could run twice simultaneously because several different things can trigger it.

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,485 @@
...
+        except Exception as e:
+            return f"ERROR: Failed to generate summary - {str(e)}"
+    
+    def detailed_failure_report(self) -> str:
+        """Generate detailed report of all failures."""
+        if not self.failed_text_versions and not self.skipped_text_versions:
+            return ""
+        
+        lines = [
+            "\n" + "=" * 80,
```
YishaiGlasner: now you have a var for this

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -1,27 +1,485 @@
...
+                except Exception as e:
+                    logger.warning(f"Error parsing response JSON - error: {str(e)}")
+            
+            # Success!
+            result.record_step_success("sheets_by_timestamp", f"Response: {r.text[:200]}")
+            return True
+            
+        except requests.exceptions.Timeout:
+            logger.warning(f"API request timed out - attempt: {attempt}")
+            if attempt < max_retries:
```
YishaiGlasner: move this to a sleep_before_attempt (or better name) func. you ha ve it thrice - one with logger and others w/o

sefaria/search.py
YishaiGlasner: ok, so add a comment for others not to remove it

sefaria/search.py
```diff
@@ -666,35 +896,69 @@ def index_sheets_by_timestamp(timestamp):
...
-        print(e)
+        logger.error(f"Error querying sheets by timestamp - timestamp: {timestamp}, error: {str(e)}", exc_info=True)
         return str(e)
 
     succeeded = []
     failed = []
+    total = len(ids)
 
-    for id in ids:
+    for i, id in enumerate(ids):
```
YishaiGlasner: `_id`

PR #3002, opened 2026-01-14
Bug/sc 39858/sandbox not getting ready for 50 minutes2

stevekaplan123:
## Description
This PR fixes git actions so that the sandbox health checks can pass and it also fixes the django_topics pytest problem so that the pytest suite can at least run.

## Code Changes
1. The solution to the django_topics issue is simply to mock get_pools so that get_pools no longer accesses the DB during pytests.
2. Of failing tests, I marked one as expected fail and wasn't sure if I should just remove the test.

### DIFF
--- a/build/ci/waitForSandbox.bash
+++ b/build/ci/waitForSandbox.bash
@@ -6,6 +6,9 @@ HEALTH_URL="https://sandbox-${GIT_COMMIT}.cauldron.sefaria.org/health-check"
 
 echo "Waiting for ${HEALTH_URL} to load for $waitDuration seconds"
 
-timeout $waitDuration bash -c 'while [[ "$(curl -s -k -o /dev/null -w ''%{http_code}'' -L '"${HEALTH_URL}"')" != "200" ]]; do sleep 5; done'
-
-echo "Reached server. Proceeding."
+if timeout $waitDuration bash -c 'while [[ "$(curl -s -k -o /dev/null -w ''%{http_code}'' -L '"${HEALTH_URL}"')" != "200" ]]; do sleep 5; done'; then
+  echo "Reached server. Proceeding."
+else
+  echo "Timed out waiting for ${HEALTH_URL}."
+  exit 1
+fi


build/ci/waitForSandbox.bash
```diff
@@ -1,22 +1,10 @@
...
-
-echo "Waiting for ${HEALTH_URL} to return 200 for up to ${waitDuration} seconds"
+echo "Waiting for https://sandbox-${GIT_COMMIT}.cauldron.sefaria.org/health-check to load for $waitDuration seconds"
 
-timeout $waitDuration bash -c "
-while [[ \"\$(curl -s -k -o /dev/null -w '%{http_code}' '${HEALTH_URL}')\" != \"200\" ]]; do
-    sleep 5
-done
-"
+timeout $waitDuration bash -c 'while [[ "$(curl -s -k -o /dev/null -w ''%{http_code}'' -L https://sandbox-${GIT_COMMIT}.cauldron.sefaria.org/health-check)" != "200" ]]; do sleep 5; done'
```
yodem: what did you change in this file? it looks like the logic/functionality is the same.
I like the HEALTH_URL var because you use this url twice and overall the previous structure was more readable

sefaria/model/tests/index_schema_test.py
```diff
@@ -310,6 +310,7 @@ def test_schema_load(self):
 
         i.delete()
 
+    @pytest.mark.xfail(reason="Likutei Moharan has a different structure")
```
yodem: since we know it will fail since the test doesnt fit the structure of the text, I dont think the test is neccesary.
I might didnt understand it so correct me if im wrong

build/ci/waitForSandbox.bash
stevekaplan123: I was actually using code from Yitzhak's version of an old version of the file.  The thing is that that version works better simply because of the -L flag and I wasn't sure how necessary some of these if statements were in the end of the file.  The -L tells curl to follow HTTP redirects instead of getting stuck.  I actually don't know why we need re-directs in the first place but since Yitzhak's code works, I used it.

PR #3033, opened 2026-01-20
feat: event to load webpages in memory

akiva10b:
> [!NOTE]
> **Medium Risk**
> Introduces new auth-like token generation (AES-GCM) and user gating logic that affects request context, profile writes, and script injection; misconfiguration could unintentionally enable/disable the chatbot or break navigation/script loading.
> 
> **Overview**
> Introduces a **chatbot integration** that is only enabled for users opted into an experiments cohort: server-side builds an encrypted `chatbot_user_token`, injects a configurable chatbot script URL/type into `base.html`, and passes chatbot props (API base URL, max input chars via remote config) through to `ReaderApp`, which renders an `lc-chatbot` element.
> 
> Adds a new `UserExperimentSettings` model + migration and extends Django admin and the account settings UI/API so staff can grant experiment eligibility and users can toggle `profile.experiments`; profile updates now strip `experiments` changes unless the user is eligible. Also adds a `sefaria:bootstrap-url` browser event to trigger in-app navigation, plus minor layout/copy-handling tweaks to accommodate the chatbot.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0b5352987301b65e1b5260fad651475db438a0d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


reader/migrations/0001_initial.py
```diff
@@ -0,0 +1,68 @@
...
+
+            constraint_name = "%s_pkey" % table
+            qn = connection.ops.quote_name
+            cursor.execute(
+                "ALTER TABLE %s ADD CONSTRAINT %s PRIMARY KEY (%s)"
+                % (qn(table), qn(constraint_name), qn(pk_column))
+            )
+
+    operations = [
+        migrations.RunPython(_ensure_user_pk, reverse_code=migrations.RunPython.noop),
```
yitzhakc: I would delete this entire action of `migrations.RunPython(_ensure_user_pk)`. Django migrations are very straightforward, this is unnecessary vibe coded stuff.

yitzhakc: You could even delete the whole file, and just run `python manage.py makemigrations`...

static/js/ReaderApp.jsx
```diff
@@ -1161,6 +1163,179 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       e.preventDefault();
     }
   }
+  handleBootstrapUrlEvent(event) {
```
yitzhakc, line 1166: What is all this code? It doesn't seem to be related at all to the chatbot?

helm-chart/sefaria/templates/configmap/local-settings-file.yaml
```diff
@@ -140,6 +140,8 @@ data:
     CSRF_COOKIE_SAMESITE = os.getenv("CSRF_COOKIE_SAMESITE", "Lax")
 
     SECRET_KEY = os.getenv("SECRET_KEY")
+    CHATBOT_USER_ID_SECRET = os.getenv("CHATBOT_USER_ID_SECRET", 'secret')
```
yitzhakc: Perhaps we shouldn't set a default here? I see that on other variables we don't set a default, probably to trigger exceptions to prevent un-secure environments.

yitzhakc (changes requested):

reader/migrations/0001_initial.py
akiva10b: I actually added this because in dev there wasn't the index, so I had to create it, before we push to prod we should verify that it has the index in place, otherwise the migration will fail

static/js/ReaderApp.jsx
akiva10b: it is the code that handles creating the panels for urls inside app without redirect. It takes a js event navigating the library, and bootstraps the library with that url without any navigation

akiva10b: it is the code that handles creating the panels for urls inside app without redirect. It takes a js event navigating the library, and bootstraps the library with that url without any navigation

sefaria/system/context_processors.py
```diff
@@ -107,3 +115,20 @@ def large_data(request):
...
     return {"EMBED": "embed" in request.GET}
+
+
+@user_only
+def chatbot_user_token(request):
+    if not request.user.is_authenticated:
+        return {"chatbot_user_token": None, "chatbot_enabled": False}
+    if not CHATBOT_USER_ID_SECRET:
+        return {"chatbot_user_token": None, "chatbot_enabled": False}
+    profile = UserProfile(user_obj=request.user)
```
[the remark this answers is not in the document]
akiva10b: this is incorrect, see 
```
if not user_has_experiments(request.user):
        return False
```
accounts for this

PR #3040, opened 2026-01-21
Links Disambiguator Task

yonadavGit:
## Description

## Code Changes

## Notes

### DIFF
--- a/sefaria/helper/linker/tasks.py
+++ b/sefaria/helper/linker/tasks.py
@@ -8,7 +8,7 @@ from celery import signature
 from celery.signals import worker_init
 from sefaria.settings import USE_VARNISH
 from sefaria import tracker
-from sefaria.model import library, Link, LinkSet, Version
+from sefaria.model import library, Link, LinkSet, Version, TermSet
 from sefaria.celery_setup.app import app
 from sefaria.model.marked_up_text_chunk import MarkedUpTextChunk, MUTCSpanType, LinkerOutput, MarkedUpTextChunkSet
 from sefaria.model import Ref
@@ -252,7 +252,7 @@ def _load_recent_non_segment_cases(linking_args: LinkingArgs) -> list[NonSegment
 
 
 def _is_non_segment_or_perek_ref(ref_str: str, oref: Optional[Ref] = None) -> bool:
-    """Return True for non-segment refs or Talmud perakim refs treated as non-segment."""
+    """Return True for non-segment refs or refs treated as non-segment (perakim/parshiot)."""
     if oref is None:
         try:
             oref = Ref(ref_str)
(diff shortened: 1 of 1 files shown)


scripts/dispatch_library_links_disambiguation_tasks.py
```diff
@@ -0,0 +1,262 @@
...
+from sefaria.model import Ref
+from sefaria.system.exceptions import InputError
+from sefaria.system.database import db
+from sefaria.settings import CELERY_QUEUES, CELERY_ENABLED
+from sefaria.celery_setup.app import app
+from dataclasses import asdict
+from sefaria.helper.linker.disambiguator import AmbiguousResolutionPayload, NonSegmentResolutionPayload
+
+# Global flag for debug mode
+DEBUG_MODE = True  # Set this to False for full analysis
```
nsantacruz: why would I set to false for full analysis? Presumably debug mode gives full analysis

sefaria/helper/linker/disambiguator.py
```diff
@@ -0,0 +1,1464 @@
...
+
+
+@dataclass(frozen=True)
+class NonSegmentResolutionResult:
+    resolved_ref: str
+    method: str
+
+# Configuration
+DICTA_URL = os.getenv("DICTA_PARALLELS_URL", "https://parallels-3-0a.loadbalancer.dicta.org.il/parallels/api/findincorpus")
+SEFARIA_SEARCH_URL = os.getenv("SEFARIA_SEARCH_URL", "https://www.sefaria.org/api/search/text/_search")
```
nsantacruz: SEFARIA_SEARCH_URL should be based on the local setting `SEARCH_URL`. On prod, this is set to an internal URL that accesses the elasticsearch pod directly. This is much faster than using a global URL (like here) that goes out of the cluster and then back in.

Still, when running on the cauldron we'll set SEARCH_URL to sefaria.org which effectively means all search on the cauldron will be routed to prod. But that's ok.

sefaria/helper/linker/tasks.py
nsantacruz, line 1: I didn't realize how much code there would be here! I think we should create a new folder in the linker folder called `disambiguator`. In that folder there are 2 files: disambiguator.py and tasks.py. Then in `celery_setup/app.py` add another entry in the `autodiscover_tasks` list. The new entry should be `sefaria.helper.linker.disambiguator`

sefaria/helper/linker/tasks.py
```diff
@@ -123,7 +136,320 @@ def link_segment_with_worker(linking_args_dict: dict) -> None:
...
+    """Load ambiguous cases from the latest LinkerOutput for this ref."""
+    linker_output = LinkerOutput().load({
+        "ref": linking_args.ref,
+        "versionTitle": linking_args.vtitle,
+        "language": linking_args.lang,
+    })
+    if not linker_output:
+        return []
+
+    spans = linker_output.contents().get("spans", [])
```
nsantacruz: why not `linker_output.spans`?

PR #3052, opened 2026-01-27
fix: voices ref search

akiva10b:
Redirect Voices search for valid refs to sheets-with-ref

Added a server-side redirect in voices /search when q is a valid ref, preserving other query params and sending users to /sheets-with-ref/<ref>. (views.py)
Added a client-side guard in voices search flow to redirect immediately on valid refs. (ReaderApp.jsx)


reader/views.py
```diff
@@ -941,6 +941,20 @@ def search(request):
     Search or Search Results page.
     """
     search_params = get_search_params(request.GET)
+    if request.active_module == VOICES_MODULE and search_params["query"]:
```
yitzhakc, line 944: Do we have any unit tests for this feature? It feels like we should have been able to catch this.. (Although I'm still not 100% sure what the problem is)

reader/views.py
```diff
@@ -941,6 +941,20 @@ def search(request):
     Search or Search Results page.
```
yitzhakc, line 941: Please add docstring to explain what is happening now, it is non-trivial!

PR #3053, opened 2026-01-27
feat(reader): update link handling to support dynamic subdomain URLs …

yodem:
## Description

Fixes SC-41141: CTRL-clicking reference links in sheets opened them on the wrong subdomain. Updates React components to render full URLs in `href` attributes for crawler compatibility.

### `static/js/ReaderApp.jsx`
- Updated `handleInAppClickWithModifiers` to set `href` for links with `data-target-module` on modifier key clicks (Ctrl/Cmd/Shift/Alt)

### `static/js/sheets/SheetContentSegments.jsx`
- Updated Hebrew and English reference links (lines 37, 46) to use `Sefaria.util.fullURL()` instead of relative paths

### `static/js/TopicPage.jsx`
- Updated 5 links with `data-target-module={Sefaria.LIBRARY_MODULE}` to use `Sefaria.util.fullURL()` (lines 428, 453, 1071, 1080, 1096)

## Notes

- Fixes CTRL-click bug: links now open on the correct domain
- Crawler-friendly: full URLs rendered in HTML
- Backward compatible: preserves `data-target-module` attributes
- No breaking changes: regular clicks continue to work

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -382,15 +382,6 @@ def catchall(request, tref, sheet=None):
     Handle any URL not explicitly covers in urls.py.
     Catches text refs for text content and text titles for text table of contents.
     """
-    def reader_redirect(uref):
-        # Redirect to standard URLs
-        url = "/" + uref
-
-        response = redirect(iri_to_uri(url), permanent=True)
-        params = request.GET.urlencode()
-        response['Location'] += "?%s" % params if params else ""
-        return response
-
     active_module = getattr(request, "active_module", LIBRARY_MODULE)
 
     for version in ['ven', 'vhe']:
@@ -411,7 +402,7 @@ def catchall(request, tref, sheet=None):
         # Normal processing for library module
         uref = oref.url(False)
         if uref and tref != uref:
-            return reader_redirect(uref)
+            return redirect_to_module(request, uref, target_module=None)
 
         return text_panels(request, ref=tref)
     else:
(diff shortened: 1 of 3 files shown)


static/js/ReaderApp.jsx
```diff
@@ -1099,6 +1099,14 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     if (linkTarget) { // We want the absolute target of the event to be a link tag, not the "currentTarget".
       // Dont trigger if user is attempting to open a link with a modifier key (new tab, new window)
       if (e.metaKey || e.shiftKey || e.ctrlKey || e.altKey) { //the ctrl/cmd, shift and alt/options keys in Windows and MacOS
+        // Update href for links with data-target-module to ensure correct subdomain
```
stevekaplan123, line 1102: I think we should have one function that is used by handleModuleLinkRightClick and this function as the code is identical or almost identical

reader/views.py
```diff
@@ -399,7 +399,8 @@ def reader_redirect(uref):
 
     if sheet is None:
         if active_module != LIBRARY_MODULE:
-            raise Http404
+            # Redirect to library module for text references
+            return redirect_to_module(request, f"/{tref}", LIBRARY_MODULE)
```
yitzhakc: I would prefer if we _first_ checked that this is a valid text ref _before_ redirecting, so we don't redirect invalud urls, resulting in a 301 to 404...

reader/views.py
```diff
@@ -4760,14 +4761,22 @@ def talmud_person_index_redirect(request):
     return redirect(iri_to_uri('/topics/category/talmudic-figures'), permanent=True)
 
 
-def redirect_to_voices(request, target_path):
+def redirect_to_module(request, target_path, target_module):
```
yitzhakc: Do we have tests for this function? If we do, please ensure they're updated.

yodem: there isnt

yodem: added

yitzhakc (approved):

PR #3062, opened 2026-01-29
Prevent duplicate submissions in the NewsletterSignUpForm component

relyks:
## Description
Once the NewsletterSignUpForm is populated with the submitter's information and the submission completes successfully, the submitter should be prevented from submitting again. Doing so will prevent duplicate submissions from being entered into Salesforce and ActiveCampaign

## Code Changes
* State variable added based on whether the user's submission has successfully gone through
* disabled property added to each of the text input fields
*  Conditionally render the submission icon based on that same state variable

### DIFF
--- a/static/js/NewsletterSignUpForm.jsx
+++ b/static/js/NewsletterSignUpForm.jsx
@@ -73,7 +73,7 @@ export function NewsletterSignUpForm({
             onChange={e => setEmail(e.target.value)}
             onKeyUp={(e) => Util.handleEnterKey(e, handleSubscribe)}/>
       </span>
-            {!showNameInputs && !isSubscribed ? <img src="/static/img/circled-arrow-right.svg" alt={Sefaria._("Submit")} onClick={handleSubscribe}/> : null}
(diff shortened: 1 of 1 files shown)


static/js/NewsletterSignUpForm.jsx
```diff
@@ -15,12 +15,17 @@ export function NewsletterSignUpForm({
...
+    const [isSubscribed, setIsSubscribed] = useState(false);
 
     function handleSubscribe() {
+        if (isSubscribed) {
+            return;
+        }
         if (showNameInputs === true) { // submit
             if (firstName.length > 0 && lastName.length > 0) {
                 setSubscribeMessage("Subscribing...");
                 subscribe(firstName, lastName, email, educatorCheck, additionalNewsletterMailingLists).then(res => {
```
yitzhakc, line 31: Since this is the call to the API, do you perhaps want to disable the button _prior_ to the call, as soon as the user has Clicked, and keep it disabled until the request returns - either as a 200 success, or a failure? This way, the button will be clickable until the request returns..

relyks: Yep, that would be better to handle that scenario too. I'll implement it. I forgot to consider that. Thanks

PR #3063, opened 2026-01-29
Bug/sc 40665/ voices search results display user id instead

yodem:
## Description
Add Django database health check to Elasticsearch reindexing cronjob and clean up excessive debug logging.

## Code Changes

**`scripts/scheduled/reindex_elasticsearch_cronjob.py`:**
- Added `check_django_database_connection()` preflight check to prevent sheet indexing with broken user lookups (would produce "User {id}" instead of real names)
- Removed truncation limits on failure/skip reporting (was capping at 50/20 items)
- Removed excessive debug logging throughout the script

**`sefaria/search.py`:**
- Removed excessive debug logging (countdown messages, index state checks, progress logs)
- Cleaned up redundant log statements in `clear_index()`, `index_all_of_type()`, and `index_all_of_type_by_index_name()`

## Notes
The Django database check is critical for Kubernetes deployments - if `local-settings-secret` is misconfigured, sheet indexing silently produces garbage data. The cronjob now fails fast with a clear error message instead.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -27,6 +27,7 @@ We're grateful to the following organizations for providing us with donated serv
 - [PyCharm](https://www.jetbrains.com/pycharm/)
 - [Slack](https://slack.com)
 
+
 ## Storybook
 
 We use [Storybook](https://storybook.js.org/) with the Vite builder to develop and review shared React components. To start Storybook locally:
--- a/docs/openAPI.json
+++ b/docs/openAPI.json
@@ -2969,7 +2969,7 @@
   },
     "/api/related/{tref}": {
       "summary": "The Related API",
-      "description": "The Related API returns all related links, sheets, notes, webpages, topics, manuscripts, and other connected media for a given `Ref`. ",
+      "description": "The Related API returns all related links, sheets, notes, topics, manuscripts, and other connected media for a given `Ref`. ",
       "get": {
         "operationId": "get-related",
         "tags": [
@@ -3490,7 +3490,6 @@
                         }
                       ],
                       "notes": [],
-                      "webpages": [],
                       "topics": [],
                       "manuscripts": [],
                       "media": []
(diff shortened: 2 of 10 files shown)


scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -190,11 +187,69 @@ def check_elasticsearch_connection() -> bool:
...
+        logger.debug(f"Django database connection verified - user_count: {user_count}")
+
+        # Extra sanity check: verify we can get a real user name
+        first_user = User.objects.first()
+        if first_user:
+            logger.debug(f"Sample user lookup successful - id: {first_user.id}, "
+                        f"name: {first_user.first_name} {first_user.last_name}")
+
+        return True
+    except Exception as e:
```
stevekaplan123, line 243: I wonder if instead of catching a general Exception, we should try...catch User.objects.count separately from User.objects.first.

scripts/scheduled/reindex_elasticsearch_cronjob.py
```diff
@@ -149,29 +152,23 @@ def detailed_failure_report(self) -> str:
         if self.failed_text_versions:
             lines.append(f"\nFailed Text Versions: {len(self.failed_text_versions)}")
             lines.append("-" * 40)
-            for i, failure in enumerate(self.failed_text_versions[:50], 1):
+            for i, failure in enumerate(self.failed_text_versions, 1):
```
stevekaplan123, line 155: I don't understand this change or what the original code intended.  Why were we stopping at 50 failed_text_versions before and why are we no longer doing this?  Same question about only looking at 20 skipped_text_versions below.

sefaria/search.py
```diff
@@ -974,17 +963,13 @@ def clear_index(index_name):
...
         # Check if index exists before trying to delete
         if index_client.exists(index=index_name):
             index_client.delete(index=index_name)
-            logger.debug(f"Successfully deleted Elasticsearch index - index_name: {index_name}")
-        else:
-            logger.debug(f"Index does not exist, nothing to delete - index_name: {index_name}")
     except NotFoundError:
         # Index doesn't exist - handle race condition where index is deleted between exists() check and delete() call
-        logger.debug(f"Index not found when attempting to delete - index_name: {index_name}")
+        pass
```
stevekaplan123, line 972: Curious why we are removing so many logger statements.  Couldn't they be useful?   Especially, here... it seems bad to have a "pass" in an exception

PR #3067, opened 2026-02-01
fix(tasks): make fallback for lookups in CELERY_QUEUES

nsantacruz:
## Description

## Code Changes

## Notes


sefaria/tracker.py
```diff
@@ -11,9 +11,10 @@
 import sefaria.model as model
 from sefaria.system.exceptions import InputError
 try:
-    from sefaria.settings import USE_VARNISH
+    from sefaria.settings import USE_VARNISH, CELERY_ENABLED
```
YishaiGlasner: if CELERY_ENABLED raises ImportError it will cause USE_VARNISH to be False. i don't think this is what you want.

PR #3070, opened 2026-02-02
show-all-topics-in-admin-view

nsantacruz:
This pull request updates the Django admin interface for topic pools, making it easier to filter, manage, and perform bulk actions on topics across different pools. The main changes include expanding filter and action options for pools, improving default behaviors, and simplifying the queryset logic.

**Admin filter and action improvements:**

* Expanded the `PoolFilter` lookup options to include `Library Pool`, `Sheets Pool`, and an `All Topics` option, allowing admins to filter topics by these pools or view all topics at once.
* Updated the default filter behavior so that if no pool is selected, the filter defaults to the `Library Pool`, improving usability for admins.
* Added bulk actions in `TopicAdmin` to add or remove topics from `Library Pool` and `Sheets Pool`, in addition to existing pools, enabling more flexible topic management.

**Codebase simplification:**

* Removed the custom `get_queryset` method in `TopicAdmin` that previously restricted topics to only those in the `Library Pool`, allowing the improved filter to handle pool selection instead.

### DIFF
--- a/django_topics/admin.py
+++ b/django_topics/admin.py
@@ -67,11 +67,9 @@ class PoolFilter(admin.SimpleListFilter):
 
 @admin.register(Topic)
 class TopicAdmin(admin.ModelAdmin):
-    list_display = []
+    list_display = ['slug', 'en_title', 'he_title', 'sefaria_link']
     for pool_type in PoolType:
         list_display.append(f'is_in_pool_{pool_type.value}')
-    list_display.append('slug', 'en_title', 'he_title', 'sefaria_link')
-    list_display = tuple(list_display)
     list_filter = (PoolFilter,)
     filter_horizontal = ('pools',)
     search_fields = ('slug', 'en_title', 'he_title')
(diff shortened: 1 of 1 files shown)


django_topics/admin.py
```diff
@@ -43,17 +43,28 @@ class PoolFilter(admin.SimpleListFilter):
...
-        return queryset
+        if pool_name == 'all':
+            return queryset
+        if pool_name is None:
+            pool_name = PoolType.LIBRARY.value  # Default to LIBRARY
+        pool = TopicPool.objects.get(name=pool_name)
+        return queryset.filter(pools=pool)
+
+    def value(self):
+        value = super().value()
```
stevekaplan123, line 62: It seems we are defaulting the pool_name to PoolType.LIBRARY.value in two places -- on line 66 and line 59.  Why do we do this?

nsantacruz: This is a good question and I'm not sure... I don't understand Django admin very well. Please feel free to research why this is.

stevekaplan123: It was duplicate code.  Then, I removed and also removed magic strings like 'general_en'.  Finally, I added columns for "sheets" and "library" as previously the view only displayed columns for general en, general he, and torah tab.  I think it's ready to merge.

PR #3093, opened 2026-02-12
Exclude parsha talmud perek from sidebar

nsantacruz:
This pull request prevents Talmud perek and Parsha links from being returned from the links API.

Along the way, it refactors the handling of Talmud perek and Parasha reference sets by moving their implementation from `sefaria/helper/linker/tasks.py` to `sefaria/helper/text.py`, and updating all relevant usages to import them from the new location. This centralizes the logic, improves maintainability, and eliminates code duplication.

Refactoring and code organization:

* Moved the implementations of `get_talmud_perek_ref_set` and `get_parasha_ref_set` from `sefaria/helper/linker/tasks.py` to `sefaria/helper/text.py`, and decorated them with `@lru_cache` for efficient caching.
* Updated imports in `sefaria/client/wrapper.py` and `sefaria/helper/linker/tasks.py` to reference these functions from `sefaria/helper/text.py` instead of their previous location. [[1]](diffhunk://#diff-51d001363d5bf6bf0af5d5af3593b97bbca2534ca63a8b6d8615acd458342146R5-R7) [[2]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR28-L33)

Bug fix and logic update:

* Modified the link filtering logic in `get_links` (in `sefaria/client/wrapper.py`) to use the new functions for excluding perek-level Talmud and Parasha references, ensuring consistent behavior.
* Refactored `_is_non_segment_or_perek_ref` in `sefaria/helper/linker/tasks.py` to use the new centralized functions, removing the old implementations and their caching.

### DIFF
--- a/sefaria/model/text.py
+++ b/sefaria/model/text.py
@@ -5068,6 +5068,9 @@ class Library(object):
         self.get_virtual_books(rebuild=rebuild)
         
         # functions backed by lru cache
+        if rebuild:
+            get_talmud_perek_ref_set.cache_clear()
+            get_parasha_ref_set.cache_clear()
         get_talmud_perek_ref_set()
         get_parasha_ref_set()
         


sefaria/model/text.py
```diff
@@ -5064,6 +5066,11 @@ def init_shared_cache(self, rebuild=False):
         self.get_simple_term_mapping(rebuild=rebuild)
         self.get_simple_term_mapping_json(rebuild=rebuild)
         self.get_virtual_books(rebuild=rebuild)
+        
+        # functions backed by lru cache
+        get_talmud_perek_ref_set()
```
yonadavGit, line 5074: i can see why 
there's also an option of "lazy" cache initialization upon first user request, but i can see why the current approach maeks sense.
what about clearing the cache? it's not as crucial since the data of this cache should hardly change but still maybe worth thinking about providing en easy way for reloading this cache

nsantacruz: Good point! Added a way to rebuild. There is already a URL admin/rebuild/shared-cache that runs this function.

PR #3095, opened 2026-02-15
fix(search): handle missing TimePeriod.start in ES indexing

yodem:
## Description

This PR does two things: it fixes Elasticsearch indexing for texts whose author has only a death year (no birth year), and it updates the reindex cronjob so the failure report lists all failed and skipped versions instead of truncating them.

## Code Changes

- **`sefaria/search.py`**
  - In `TextIndexer.make_text_index_document` (line 867), replaced direct access to `tp.start` with `getattr(tp, 'start', getattr(tp, 'end', 3000))`.
  - Handles `TimePeriod` objects that lack a `start` attribute (e.g. author with only death year).
  - Avoids `AttributeError` that prevented indexing for texts like "Tashbetz Katan".

- **`scripts/scheduled/reindex_elasticsearch_cronjob.py`**
  - In `detailed_failure_report()`, removed the cap on how many failures and skips are printed.
  - Previously: only the first 50 failed versions and 20 skipped versions were shown, with a “… and X more” line.
  - Now: all failed and skipped versions are listed in the report.

## Notes

- **Fixes:** SC-39971
- **Reindexing:** After deployment, run a full reindex so previously skipped texts (e.g. Tashbetz Katan) are indexed.
- **Report size:** With no truncation, the failure report can be large when there are many failures or skips.

### DIFF
--- a/sefaria/search.py
+++ b/sefaria/search.py
@@ -863,7 +863,7 @@ class TextIndexer(object):
         indexed_categories = get_search_categories(oref, categories)
 
         tp = cls.best_time_period
-        comp_start_date = int(getattr(tp, 'end', None) or getattr(tp, 'start', 3000)) # far in the future if no end date
+        comp_start_date = int(getattr(tp, 'end', None) or getattr(tp, 'start', 3000)) # If there is no end/start date, use 3000 which make it appear at the end of the results
         ref_data = RefData().load({"ref": tref})
         pagesheetrank = ref_data.pagesheetrank if ref_data is not None else RefData.DEFAULT_PAGESHEETRANK
 


sefaria/search.py
```diff
@@ -864,7 +864,7 @@ def make_text_index_document(cls, tref, heTref, version, lang, version_priority,
 
         tp = cls.best_time_period
         if tp is not None:
-            comp_start_date = int(tp.start)
+            comp_start_date = int(getattr(tp, 'start', getattr(tp, 'end', 3000)))
```
yodem: @YishaiGlasner Are we sure this is what we want?

YishaiGlasner: i think we want end before start, and for more readability i'd suggest:
```
int(getattr(tp, 'end', None) or getattr(tp, 'start', 3000))
```
in a second thought. remove all of the if-else and this line will do the job (but keep the comment about 3000)

PR #3098, opened 2026-02-17
feat(chatbot): add prop `is-moderator`.

YishaiGlasner:


static/js/ReaderApp.jsx
```diff
@@ -2453,6 +2453,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
                 <lc-chatbot
                   user-id={this.props.chatbot_user_token}
                   api-base-url={this.props.chatbot_api_base_url}
+                  is-moderator={this.props.is_moderator || undefined}
```
yodem, line 2456:
```suggestion
                  is-moderator={!!this.props.is_moderator}
```

YishaiGlasner: we're passing here a boolean to html, so it's converted to string. this will make it 'false' which is truthy. the right way is to pass undefined (which means not passing the param.

PR #3104, opened 2026-02-18
feat(linker): optimize text modification by processing only changed segments

nsantacruz:
This pull request refactors how text modifications are tracked and processed in `sefaria/tracker.py`. The main change is to only process and log segments of text that have actually changed, rather than re-processing the entire text every time. This should improve efficiency and accuracy in change tracking and downstream processing.

Key changes:

**Segment-level change tracking:**

* Added a new helper function `_post_modify_changed_segments` that recursively compares the old and new text, and only calls `post_modify_text` for segments that have changed. This ensures that only modified segments are processed and logged, rather than the entire text.

* Updated `modify_text` to use `_post_modify_changed_segments` instead of calling `post_modify_text` directly, and adjusted how the `count_after` parameter is handled to avoid redundant counting and indexing.

### DIFF
--- a/sefaria/tracker.py
+++ b/sefaria/tracker.py
@@ -36,15 +36,13 @@ def modify_text(user, oref, vtitle, lang, text, vsource=None, **kwargs):
     if vsource:
         chunk.versionSource = vsource  # todo: log this change
     if chunk.save():
-        kwargs['skip_links'] = kwargs.get('skip_links', False) or chunk.has_manually_wrapped_refs()
+        skip_links = kwargs.pop('skip_links', False) or chunk.has_manually_wrapped_refs()
+        count_after = kwargs.pop("count_after", 1)
         version_id = str(chunk.full_version._id)
 
-        orig_count_after = kwargs.get("count_after", 1)
-        kwargs['count_after'] = False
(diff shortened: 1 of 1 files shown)


sefaria/tracker.py
```diff
@@ -37,7 +37,14 @@ def modify_text(user, oref, vtitle, lang, text, vsource=None, **kwargs):
         chunk.versionSource = vsource  # todo: log this change
     if chunk.save():
         kwargs['skip_links'] = kwargs.get('skip_links', False) or chunk.has_manually_wrapped_refs()
-        post_modify_text(user, action, oref, lang, vtitle, old_text, chunk.text, str(chunk.full_version._id), **kwargs)
+        version_id = str(chunk.full_version._id)
+
+        orig_count_after = kwargs.get("count_after", 1)
+        kwargs['count_after'] = False
+
+        _post_modify_changed_segments(user, action, oref, lang, vtitle, old_text, text, version_id, **kwargs)
```
yonadavGit: the "kwargs all the way down" style here is a bit confusing, especially now that the counting logic has been modified. Maybe worthwhile, while were at it, taking out the the **counting-related vars** out of kwargs land so it's a bit clearer what's going on.. (concretely, trying to make these vars explicit as much as possible)

PR #3110, opened 2026-02-25
Feature/sc 41912/welcome experience s

stevekaplan123:
## Description
See Code Pilot's description: https://github.com/Sefaria/Sefaria-Project/pull/3110#pullrequestreview-3913949238

### DIFF
--- a/chatbot/models.py
+++ b/chatbot/models.py
@@ -29,6 +29,8 @@ def get_chatbot_welcome_messages():
     for any missing or empty value.
     """
     stored = remoteConfigCache.get(CHATBOT_WELCOME_MESSAGES, {})
+    if not isinstance(stored, dict):
+        stored = {}
     return {key: stored.get(key) or default for key, default in DEFAULTS.items()}
 
 


chatbot/models.py
```diff
@@ -0,0 +1,86 @@
+from django.db import models
+
+from sefaria.system.cache import django_cache
+
+
+DEFAULT_WELCOME_EN = "Welcome! Ask me anything about Jewish texts."
+DEFAULT_WELCOME_HE = "ברוך הבא! שאל אותי כל מה שתרצה על טקסטים יהודיים."
```
yitzhakc, line 7: This should probably be `DEFAULT_WELCOME_HE`?

reader/migrations/0002_add_chatbot_welcome_message.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
```
yitzhakc: Reading through, it seems like you decided to move the model to a separate django app.
I'd perform a bit more of a cleanup - you can just delete both of the migration files, and then delete the records from your local db and the cauldron db.

chatbot/migrations/0001_initial.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
```
yitzhakc: I feel like it's a bit overkill to create an entire Django model that will have a single record (`default`?).
In addition, it's very brittle - any new modifiable string they want would require a database migration.

Let's try and use the `RemoteConfig` json object. But, since we want to have a slightly better editing experience for these strings then just a large textbox with JSON objects, we can look for different solutions:
- Write code that uses a custom widget to make the RemoteConfig object more writeable
- Create a custom Django Admin form page that attaches to RemoteConfig only for this single record
- See if there's a package that supports something like this (django-jsonform)

Happy to discuss this offline.

chatbot/tests/test_chatbot_welcome_message.py
```diff
@@ -0,0 +1,74 @@
+"""Tests for ChatbotWelcomeMessage model and get_chatbot_welcome_messages()."""
```
yitzhakc: These tests don't provide any additional value over the base functionality of built-in Django ORM...

reader/views.py
```diff
@@ -356,6 +357,7 @@ def base_props(request):
         "chatbot_enabled": False,
         "chatbot_api_base_url": CHATBOT_API_BASE_URL,
         'chatbot_max_input_chars': remoteConfigCache.get(CHATBOT_MAX_INPUT_CHARS, default=500),
+        'chatbot_welcome_messages': json.dumps(get_chatbot_welcome_messages()),
```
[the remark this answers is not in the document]
yitzhakc: This is something I'd want a test for, to ensure that we're not doing this kind of double serialization.

static/js/ReaderApp.jsx
```diff
@@ -2435,7 +2435,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     var classes = classNames(classDict);
     const mobile = Sefaria.getBreakpoint() === Sefaria.breakpoints.MOBILE;
     const isLibraryModule = Sefaria.activeModule === Sefaria.LIBRARY_MODULE;
-    const displayChatbot = this.props.chatbot_enabled && this.props.chatbot_user_token && !mobile && isLibraryModule && this.props.interfaceLang === "english" && !(this.props.remoteConfig?.chatbot?.hide === 1);
```
yitzhakc, line 2438: I don't think we want to display the chatbot on the Hebrew site yet; please revert this modification

chatbot/models.py
```diff
@@ -0,0 +1,41 @@
...
+def get_chatbot_welcome_messages():
+    """
+    Returns dict with welcome_english, welcome_hebrew, restart_english,
+    restart_hebrew, new_session_english, new_session_hebrew for the chatbot.
+
+    Reads from the RemoteConfig JSON entry; falls back to hard-coded defaults
+    for any missing or empty value.
+    """
+    stored = remoteConfigCache.get(CHATBOT_WELCOME_MESSAGES, {})
+    return {key: stored.get(key) or default for key, default in DEFAULTS.items()}
```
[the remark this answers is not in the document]
yitzhakc: Adding a safety check is important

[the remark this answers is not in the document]
yitzhakc: On point 😉

chatbot/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ChatbotWelcomeMessageProxy(RemoteConfigEntry):
```
yitzhakc, line 37: This is good!

PR #3113, opened 2026-02-25
Add API endpoint to retrieve works by author

yonadavGit:
## Description

## Code Changes

## Notes

### DIFF
--- a/sefaria/helper/topic.py
+++ b/sefaria/helper/topic.py
@@ -372,6 +372,10 @@ def _get_author_indexes_from_topic(author_topic: AuthorTopic, include_aggregatio
 
 
 def get_author_indexes(slug: str, include_aggregations: bool = False, include_descriptions: bool = False) -> Optional[dict]:
+    """Return author-indexes payload for `slug`, or `None` if slug is missing/non-author.
+    Optional flags: `include_aggregations`, `include_descriptions`.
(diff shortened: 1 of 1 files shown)


sefaria/urls_shared.py
```diff
@@ -165,6 +165,7 @@
 
     url(r'^api/(?P<action>(block|unblock))/(?P<uid>\d+)$', reader_views.block_api),
 
+    url(r'^api/authors/(?P<author_slug>[^/]+)/works/?$', reader_views.author_works_api),
```
akiva10b: "works" isn't a db object, really we are returning "index" objects. We should stick to our terminology throughout the code. Even if the client knows it as "works". The reason is that in a restful philosophy, the API is supposed to articulate real objects, so authors/(?P<author_slug>[^/]+)/indexes/ tells a dev or AI what the object being returned is, and how they can map it.

While "Author" is also not a db object, we do have a data class for AuthorTopic which is more traversable. My recommendation would be to change the terminology of "work" to "index" throughout the code.

reader/views.py
```diff
@@ -3335,6 +3336,22 @@ def topics_list_api(request):
     return response
 
 
+@catch_error_as_json
+def author_works_api(request, author_slug):
+    if request.method != "GET":
+        return jsonResponse({"error": "This API only accepts GET requests."})
```
akiva10b: correct error code here is 405

sefaria/helper/topic.py
```diff
@@ -324,6 +324,56 @@ def get_bulk_topics(topic_list: list) -> TopicSet:
...
+        he_desc = getattr(index, "heShortDesc", None) or getattr(index, "heDesc", None)
+        if en_desc:
+            description["en"] = en_desc
+        if he_desc:
+            description["he"] = he_desc
+        response["description"] = description
+    return response
+
+
+def _get_author_works_from_topic(author_topic: AuthorTopic, include_aggregations: bool = False, include_descriptions: bool = False) -> dict:
```
akiva10b: this looks good overall, add a doctoring with return examples for what include_aggregations and include_descriptions will do

PR #3120, opened 2026-02-27
NEW Newsletter Sign-Up Page

relyks:
### Description

This branch adds a new dedicated newsletter sign-up page at `/newsletter`, replacing the old footer-style sign-up widget with a full-page experience. The page supports both logged-out (new subscriber) and logged-in (existing subscriber) flows within a single React form component.

**What a user can do:**

- Pick from newsletter lists, each with its own icon and display name. The newsletter lists are fetched live from ActiveCampaign, Sefaria's email campaign management provider.
- Enter their name + email (logged-out) or just review and edit their existing selections (logged-in).
- Optionally provide a "learning level" (1–5) survey component, so future content can be tailored and so the organization can get a better understanding of the userbase.
- Logged-in users can also opt out of all marketing email via a toggle, which is stored on their `UserProfile` in MongoDB.
- See preview iframes of two real example emails (Weekly Parashah Study Companion, Timeless Topics) on the page.

---

### Key Design Decisions

- **Why a dedicated `NewsletterSignUpPage/` directory instead of files in `static/js/`?**

  The page is large enough (form container, two views, three reusable components (`SelectableOption`, `MarketingEmailToggle`, `LearningLevelSurvey`), an API adapter, a state symbols file (`stateSymbols.js`), a custom init hook (`useNewsletterFormInit.js`), and a bilingual strings file) that scattering it across `static/js/` would be hard to navigate. Nothing outside this folder needs to know about this feature except `StaticPages.jsx`, which renders it.

- **State-machine pattern in `NewsletterSignUpPageForm.jsx`**

  The form has three sequential stages: newsletter selection → confirmation → success. Rather than using URL routes or modal toggles, the component holds a `currentStage` symbol (`STAGE.NEWSLETTER_SELECTION | CONFIRMATION | SUCCESS`) and renders one view at a time. Separate URL routes are unnecessary.

- **ActiveCampaign personalization variables used as a lightweight CMS**

  ActiveCampaign offers several ways to attach metadata to a list, and they all have tradeoffs:

  | Option in AC                                     | Why it didn't fit                                                                                                                                                                 |
  | ------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
  | Hard-code list metadata in a static js file      | Sefaria team can't edit emoji/icon/display-name without a deploy.                                                                                                                 |
  | AC custom fields on the list object              | AC doesn't support custom fields on lists, only on contacts.                                                                                                                      |
  | AC tags                                          | Tags are flat strings — no place to store a JSON payload like icon + language.                                                                                                    |
  | Separate metadata service / DB table like Strapi | New database table, new sync logic — a heavyweight solution for what's essentially "label this list." Creates a stronger dependency on another CMS.                               |
  | Personalization variables ✅                      | Already a key/value store in AC. Each variable has a tag (the key) and content (free-form text). Editable in the AC UI by non-engineers. No extra infrastructure or dependencies. |

  So the convention this branch establishes is:

  - For every newsletter list we want to expose on the sign-up page, create a personalization variable in AC.
  - The variable's tag follows the pattern `list_{N}_meta`, where N is the AC numeric list ID. AC normalizes tags to lowercase, so this pattern is enforced via `re.match(r'list_(\d+)_meta', tag)` in `extract_list_id_from_tag`.
  - The variable's name is the human-readable newsletter title (e.g., "Sefaria News & Resources" or its Hebrew equivalent).
  - The variable's content is a small JSON blob: `{"icon": "news-and-resources.svg", "language": "english"}`.

  `get_newsletter_list()` in `newsletter_service.py` then:

  1. Fetches all lists from AC.
  2. Fetches all personalization variables from AC.
  3. Filters variables to those whose tag matches `list_{N}_meta`.
  4. For each, parses the JSON content and matches it to a list by ID.
  5. Returns only lists that have both a real AC list and a matching metadata variable.

  That last step is intentional: a list without a `list_{N}_meta` variable is invisible to the sign-up page, even if it exists in AC. This gives MarCom a clean way to expose only lists that users can self-manage, while keeping other lists internal. MarCom has full control over the visible options without touching the codebase.

- **Shared form for logged-in and logged-out users with different semantics**

  Same component, different behavior, controlled by `formStatus.isLoggedIn`:

  | Flow       | Endpoint                           | Semantics                                                                                                        | Why                                                                                                                                                                                                                   |
  | ---------- | ---------------------------------- | ---------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
  | Logged-out | `POST /api/newsletter/subscribe`   | UNION — adds new selections to whatever the contact already has in AC. Never removes anything.                   | An unauthenticated user could spoof someone else's email and allow them to manage that someone's subscriptions potentially. UNION ensures the worst case is harmless re-subscription, not destructive unsubscription. |
  | Logged-in  | `POST /api/newsletter/preferences` | REPLACE — the user's submitted selections become their full subscription list. Unchecked lists are unsubscribed. | The user is explicitly managing their own preferences with session auth. REPLACE is what a "preferences" screen should do.                                                                                            |

  This is also a security boundary. There is no logged-out unsubscribe path at all on this page. If a logged-out user wants to unsubscribe, they have to either (a) log in, or (b) click the unsubscribe link in an actual AC email (which AC handles directly, outside our code). This prevents anyone from POSTing `{email: "victim@example.com", newsletters: {}}` and clearing someone else's subscriptions.

- **UNION subscribe mechanics (`subscribe_with_union`)**

  For a logged-out POST to `/api/newsletter/subscribe`, the backend:

  1. Validates `firstName`, `lastName`, and `email` are present, and at least one newsletter is selected.
  2. Maps selected `stringids` to AC numeric list IDs via `map_stringids_to_list_ids`, which simultaneously validates that every stringid exists in the cached newsletter list (raises InputError → 400 for unknown IDs).
  3. Calls `find_or_create_contact(email, first_name, last_name)` — either looks up the existing AC contact or creates a new one.
  4. Fetches the contact's current *active* list memberships via `get_contact_list_memberships(contact_id, active_only=True)`.
  5. Deduplicates: computes `to_add = new_list_ids - existing_active_list_ids` so lists the user is already subscribed to are skipped.
  6. Calls `subscribe_contact_to_lists(contact_id, to_add)` — a single batched call that POSTs one `{"contactList": {"contact": ..., "list": ..., "status": 1}}` per list in `to_add`. `status: 1` means "subscribed" (AC requires this field — omitting it returns 422).
  7. Returns the union of (pre-existing active subscriptions + newly added ones) so the frontend can show the full picture in the confirmation screen.

- **REPLACE preferences mechanics (`update_user_preferences_impl`)**

  For a logged-in POST to `/api/newsletter/preferences`, the backend:

  1. Reads the user from `request.user` — never from the request body, so the email can't be spoofed.
  2. Fetches the user's current AC list memberships.
  3. Diffs requested list-membership against current list-membership to compute `to_add` and `to_remove`.
  4. Calls `add_contact_to_list` (status 1) for each list in `to_add`.
  5. Calls `remove_contact_from_list` (status 2) for each list in `to_remove`. AC unsubscribe is just another POST to `/contactLists` with `status: 2`.
  6. If the request includes `marketingOptOut: true`, also unsubscribes the user from every AC list returned by `get_all_ac_list_ids()` — including unmanaged lists that aren't on the sign-up page. This is how the "opt out of marketing emails" toggle becomes a global opt-out.
  7. Persists `wants_marketing_emails` to MongoDB on the `UserProfile`.

- **`Promise.all` gate before rendering the form**

  On mount, the form fires off `getNewsletterLists()` and (if logged in) `fetchUserSubscriptions()` in parallel via `useNewsletterFormInit`. The form isn't rendered until both are resolved.

- **`wants_marketing_emails` lives in MongoDB, not ActiveCampaign**

  Added two fields to `UserProfile`:
  - `learning_level` (1–5, nullable)
  - `wants_marketing_emails` (bool, defaults to `True` — opted-in)

  The opt-out flag is MongoDB-only. When a user opts out, the backend unsubscribes them from every AC list. This flag isn't mirrored to an AC custom field because AC's source-of-truth is "is this contact on this list, yes/no" — the flag is a Sefaria-side intent, not an AC-managed value.

- **Stale opt-out detection (frontend-only)**

  If MongoDB says the user opted out, but AC shows they're actually subscribed to lists (because they re-subscribed via a different channel like the footer signup or an external Unbounce form), the form trusts AC and shows the toggle as ON. MongoDB is not corrected on read — it's corrected only when the user actually submits the form. This is a deliberate "fix on write" choice that keeps GET requests cheap and side-effect-free.

- **Module-level `requests.Session` singleton in `newsletter_service.py`**

  One signup hits AC 5–10 times back-to-back (lookup contact, create if missing, fetch lists, update subscriptions, fetch fields, write custom field, etc.). A fresh HTTPS handshake costs hundreds of ms each, so one `Session` is kept open per worker process and the `Session` reuses the TCP/TLS connection. The thread-safety contract — headers set once in `__init__`, never mutated.

- **`requires_ac_client` decorator + `is_newsletter_service_configured()` flag**

  The whole module gracefully degrades when `ACTIVECAMPAIGN_API_TOKEN` is `None` (typical for local dev or CI). The views return HTTP 503 `newsletter_service_not_configured` instead of crashing, and the frontend shows a "not available right now" message.

- **Bilingual text in one file**

  All English/Hebrew strings live in `BILINGUAL_TEXT` (`bilingualUtils.js`) as `{en, he}` objects, which is the shape InterfaceText already understands. A few strings ("Last Name (Optional)", "Confirm Email Address", etc.) were added to `sefaria/strings.js` because they came from shared components like InterfaceText's standard translation table.

---

### Code Changes

#### Backend (`api/`)

**`newsletter_service.py`**

The entire ActiveCampaign integration layer. Public functions used by views:

- `is_newsletter_service_configured()` — Returns True if AC credentials are set. Used as a pre-check by views to return 503 early.
- `get_newsletter_list()` — The CMS-style metadata join described above. Returns the list of `NewsletterInfo` dicts shown on the sign-up page.
- `subscribe_with_union(email, first_name, last_name, newsletter_stringids, valid_newsletters=None)` — UNION subscribe flow for logged-out users.
- `fetch_user_subscriptions_impl(email, valid_newsletters, user)` — Returns the user's currently-subscribed lists plus their `learning_level` and `wants_marketing_emails` from `UserProfile`.
- `update_user_preferences_impl(email, first_name, last_name, ...)` — REPLACE preferences flow for logged-in users, including the global opt-out branch.
- `update_learning_level_impl(email, learning_level)` — Writes learning level to AC (as a custom field on the contact) and to MongoDB. Both writes happen so ActiveCampaign can store the change and Sefaria can read it server-side without an AC round-trip.
- `get_cached_newsletter_list()` — Wrapper around `get_newsletter_list()` that reads from Django's cache (NEWSLETTER_LIST_CACHE_KEY, 1-hour TTL). Used by views for subscription validation to avoid hitting AC twice per request.

Internal helpers worth knowing:

- `NewsletterClient` — Thin wrapper around `requests.Session`. One method: `make_request(endpoint, method='GET', data=None)`. Translates Timeout, ConnectionError, HTTPError, and unknown exceptions into a single `ActiveCampaignError`.
- `_client` — Module-level `NewsletterClient` singleton, instantiated at import time if credentials are present.
- `requires_ac_client` — Decorator that raises `ActiveCampaignError` if `_client` is None.
- `get_ac_field_id_by_perstag(perstag)` — Resolves a perstag (personalization tag) like `LEARNING_LEVEL` to AC's numeric field ID. Caches the full `perstag`→`id` map on first call (`_field_id_cache`, module-level dict).
- `extract_list_id_from_tag(tag)`, `parse_metadata_from_variable(variable)`, `_parse_variable_entry(v)` — The personalization-variable parsing pipeline used by `get_newsletter_list()`.
- `find_or_create_contact`, `add_contact_to_list`, `remove_contact_from_list`, `subscribe_contact_to_lists` — AC operations.
- `get_all_ac_list_ids()` — Returns every list ID in the account, not just managed ones. Used by the marketing opt-out path to ensure unmanaged lists are also unsubscribed.

**`newsletter_views.py`**

Thin Django view adapters. One function per endpoint:

- `get_newsletter_lists(request)` — GET `/api/newsletter/lists`. Calls `get_cached_newsletter_list()` (defined in `newsletter_service.py`). Returns 503 if AC isn't configured. No CSRF needed (GET).
- `subscribe_newsletter(request)` — POST `/api/newsletter/subscribe`. `@csrf_protect` — requires a CSRF token (sent by the frontend via the `X-CSRFToken` header). Validates body, calls `subscribe_with_union`.
- `get_user_subscriptions(request)` — GET `/api/newsletter/subscriptions`. Requires auth (returns 401 otherwise). Reads `request.user.email` directly — never trusts an email from the request body.
- `update_user_preferences(request)` — POST `/api/newsletter/preferences`. `@csrf_protect`. Requires auth. Calls `update_user_preferences_impl`.
- `update_learning_level(request)` — POST `/api/newsletter/learning-level`. `@csrf_protect`. Works for both authed and un-authed users; if authed, ignores any email in the body and uses `request.user.email`.

**`conftest.py`**

pytest fixtures shared across the newsletter test suite. Notably:

- A fixture that clears `_field_id_cache` between tests to prevent module-level cache leakage.
- A fixture that mocks the `_client` singleton so tests don't need real AC credentials.

**`test_newsletter.py`**

Unit tests with fully mocked AC. Covers every public function in `newsletter_service.py`: regex parsing, JSON parsing, list/variable merging, contact lookup/create, UNION subscribe, REPLACE preferences, learning level writes, error propagation, and the not-configured branch.

**`test_newsletter_api_integration.py`**

View-level integration tests using `pytest-django`. Exercises the HTTP surface: authentication, status codes, response shapes, CSRF behavior, and validation errors.

**`test_newsletter_integration.py`**

End-to-end backend tests that wire the views to the service layer with mocked AC at the HTTP boundary only. Catches integration bugs that pure unit tests miss.

---

#### Frontend (`static/js/NewsletterSignUpPage/`)

**`NewsletterSignUpPageForm.jsx`**

The orchestrator. Owns two state objects and one reducer:

- `formData` — what the user typed and selected.
- `formStatus` — managed via `useReducer(formStatusReducer, ...)` with `dispatchFormStatus`. State-machine fields (`currentStage`, `status`, `isLoggedIn`, `userEmail`, etc.). All transitions are centralized in `formStatusReducer` (a module-level function) using `FORM_STATUS_ACTION` symbols from `stateSymbols.js`.
- `validationState` — `fieldErrors` map and `hasAttemptedSubmit` flag (so errors only appear after the first submit attempt, not on first keystroke).

The loading lifecycle (fetching newsletters + subscriptions on mount) is delegated to the `useNewsletterFormInit` hook, which returns `authState`, `newsletters`, `newslettersLoading`, `serviceUnavailable`, `initialSubscriptionData`, and `baselineRef`.

Refs (intentionally not state, because they don't drive rendering):

- `formTitleRef` — used to scroll to the form title when validation fails, so the user sees the error summary at the top.
- `containerRef` — used to scroll the form into view on stage transitions.
- `baselineRef` (in `useNewsletterFormInit`) — write-once snapshot of the user's subscriptions + marketing preference at page load, with shape `{ subscriptions: {}, wantsMarketing: true }`. Used to compute the diff at submit time so we can show "Added X, Removed Y" copy on the confirmation screen and skip the API call if nothing changed.

Notable functions:

- `buildFieldValidators(formData, formStatus)` — Returns an object of field-name → validator function. Each validator returns either an error message or null.
- `validateFormData()` — Runs every validator and returns the errors object.
- `handleFieldBlur(fieldName)` — On-blur (out-of-focus) revalidation, but only if the user has already attempted submit (to avoid premature errors).
- `getSubscriptionDiffs()` — Compares the initial snapshot against the current state to return `{added: [...], removed: [...]}`.
- `handleSubscribeSubmit` and `handleSaveLearningLevel` — The two submit handlers; both branch on `isLoggedIn` and handle the loading/error states. There is also `handleSkipLearningLevel` for the "No thanks" path (redirects to `/`).

**`NewsletterFormView.jsx`**

Stage 1 view. Renders all the inputs: name, email, confirm email, newsletter checkboxes, marketing-email toggle (logged-in only), submit button, and the error summary block. Receives all data and handlers as props from the container; has no state of its own. The container/view split keeps the form testable without needing to mount the API-talking parent.

**`NewsletterConfirmationView.jsx`**

Stage 2 view. Shows the user what they just subscribed to / changed, prompts them for an optional learning level, and has "Save" / "No thanks" buttons. Receives `subscriptionDiffs` from the parent to render different copy depending on whether the user added or removed lists.

**`SuccessView.jsx`**

Stage 3 view. Thank-you screen.

**`SelectableOption.jsx`**

Reusable card-style option used for newsletter checkboxes. Has a `type` prop that switches between `'checkbox'` and `'radio'` behavior.

**`MarketingEmailToggle.jsx`**

Reusable toggle for the "opt out of marketing emails" switch shown to logged-in users.

**`LearningLevelSurvey.jsx`**

Reusable component that renders the learning-level radio options in the confirmation view.

**`newsletterApi.js`**

Plain fetch-based adapter. Five named exports (`subscribeNewsletter`, `updatePreferences`, `updateLearningLevel`, `fetchUserSubscriptions`, `getNewsletterLists`) and two private helpers: `parseJsonResponse` (checks Content-Type and parses JSON) and `apiRequest` (the shared dispatcher that attaches headers, optionally adds the CSRF token, calls `fetch`, and normalizes errors).

**`stateSymbols.js`**

Exports `FORM_STATUS`, `STAGE`, and `FORM_STATUS_ACTION` (all frozen Symbol enums). These are the action types consumed by `formStatusReducer` in `NewsletterSignUpPageForm.jsx`.

**`useNewsletterFormInit.js`**

Custom hook that owns the form's loading lifecycle: detects auth state synchronously at first render, fires `getNewsletterLists()` and (if logged in) `fetchUserSubscriptions()` in parallel via `Promise.all`, and records a `baselineRef` snapshot for later diff computation. Returns `{ authState, newsletters, newslettersLoading, serviceUnavailable, initialSubscriptionData, baselineRef }`.

Note: `LEARNING_LEVELS` is exported from `bilingualUtils.js` (alongside `BILINGUAL_TEXT`), not from a separate constants file.

**`__tests__/`**

Jest tests for every component plus the API adapter:

- `NewsletterSignUpPageForm.test.js` — The container; uses a mock-and-capture-props strategy, mocks the child views to assert what props the container passes them.
- `NewsletterFormView.test.js`, `NewsletterConfirmationView.test.js`, `SuccessView.test.js` — Per-view tests.
- `SelectableOption.test.js`, `MarketingEmailToggle.test.js` — Small reusable component tests.
- `NewsletterPage.test.js` — Top-level page (the wrapper from `StaticPages.jsx`).
- `newsletterApi.test.js` — API adapter with `global.fetch` mocked.

---

#### Modified files

**`sefaria/model/user_profile.py`**

Two new fields and matching serializer lines:

- `self.learning_level = None` and `self.wants_marketing_emails = True` in `__init__`.
- Both added to the `to_mongo_dict()` return dict.
- `learning_level` also added to `to_api_dict()` (under `other_info`), so the frontend can read it without a separate request. `wants_marketing_emails` is intentionally not in `to_api_dict()` — the newsletter form has its own endpoint that returns the effective value (with stale opt-out reconciliation applied).

**`sefaria/urls_shared.py`**

Imports `api.newsletter_views` as `newsletter_views` and registers five new routes:

| Endpoint                         | Method | Auth Required |
| -------------------------------- | ------ | ------------- |
| `/api/newsletter/lists`          | GET    | No            |
| `/api/newsletter/subscribe`      | POST   | No            |
| `/api/newsletter/subscriptions`  | GET    | Yes           |
| `/api/newsletter/preferences`    | POST   | Yes           |
| `/api/newsletter/learning-level` | POST   | No            |

**`static/js/StaticPages.jsx`**

The `NewsletterPage` component was rewritten. It now:

1. Wraps `NewsletterSignUpPageForm` and passes it an `onStageChange` callback.
2. Tracks the form's current stage in local state, and uses that to conditionally hide the testimonials/example-emails block once the user advances past stage 1.

**`playwright.config.ts`**

Added two new projects:

- `setup` — runs `auth.setup.js`, which logs in a real user and saves the session to `e2e-tests/.auth/user.json`.
- `authenticated` — runs `*-real.spec.js` test files using that saved session via Playwright's `storageState` mechanism.

The `chrome-newsletter` project now has a `testIgnore` for `auth.setup.js` and `*-real.spec.js`, so unauthenticated tests still work without env vars.

---

### Testing Infrastructure

This branch introduces two new testing patterns that are worth understanding before reviewing the test code, because they don't exist elsewhere in the codebase yet.

#### Pattern 1: Playwright HAR fixtures for deterministic CI replay (NEW CONVENTION)

The problem: Playwright tests need to hit `/api/newsletter/*`, which in turn hits the real ActiveCampaign API. Doing that in CI is slow, fragile, and would burn through AC rate limits. Mocking with `page.route()` per test is tedious and ends up obsolete whenever an endpoint shape changes.

The solution: HAR-replay fixtures. A HAR file is a standard JSON format that captures HTTP requests and responses. Playwright has a built-in `context.routeFromHAR()` that replays them. This branch adds a small helper, `e2e-tests/support/har-fixture.js`, that wraps the built-in API with a record/replay dual mode:

```js
await routeWithHarFixture(context, 'newsletter-loggedout');
```

How it works:

| Mode             | Triggered by                                                 | Behavior                                                                                                              |
| ---------------- | ------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------- |
| Replay (default) | `e2e-tests/fixtures/newsletter-loggedout.har` exists on disk | Intercepts `/api/newsletter/` and serves responses from the HAR. Unmatched requests fall through to the real network. |
| Record           | `RECORD_HAR=1` env var                                       | Passes everything through to the real network AND updates the HAR file with the actual responses.                     |
| Fall-through     | No HAR and no `RECORD_HAR`                                   | Returns early; tests hit the real network.                                                                            |

This makes the test workflow:

1. Developer runs the local Django server with real AC credentials.
2. Developer records a HAR by running tests with `RECORD_HAR=1`. Two HAR files are committed: `newsletter-loggedout.har` and `newsletter-loggedin.har`.
3. CI runs the same tests in replay mode — no AC credentials needed, deterministic.
4. When a backend response shape changes, the developer re-records the HAR and commits the new file.

Why this matters as a convention: other features in the codebase that hit external APIs from Playwright tests could adopt this same pattern. The conventions are "use `/api/<feature>/` as the URL filter" and "commit the HAR alongside the test."

The fixtures themselves live in `e2e-tests/fixtures/*.har` — they're stable, deterministic JSON, but they're not small (the logged-in HAR captures a real subscription flow with all 5–10 AC calls).

#### Pattern 2: Real-authentication Playwright project (NEW CONVENTION)

Most Playwright tests in this repo run unauthenticated. To test the logged-in flow against real Django sessions (not just a `_uid` stub), this branch adds `e2e-tests/auth.setup.js`, which runs once at the start of the `authenticated` project, before any spec files. It:

1. GETs `/login` to receive Django's CSRF cookie and extract the CSRF token from the form's hidden input.
2. POSTs to `/login/` with `email`, `password`, and `csrfmiddlewaretoken` — all via `page.request.post`, which shares the browser's cookies so Django's `sessionid` cookie is captured automatically.
3. Verifies login by navigating to `/newsletter` and asserting the user's email appears on the page (the form shows it for logged-in users).
4. Calls `page.context().storageState({ path: '.auth/user.json' })` to save the full cookie + localStorage state to disk.

Playwright's `playwright.config.ts` wires `auth.setup.js` to the `authenticated` project, which has `storageState: 'e2e-tests/.auth/user.json'` and `dependencies: ['setup']`. Every test in that project starts with the user already logged in.

The `chrome-newsletter` project ignores both `auth.setup.js` and `*-real.spec.js` (via `testIgnore`), so contributors who don't have a test account can still run the unauthenticated suite without env vars.

`.auth/user.json` is git-ignored (the file would contain a real session cookie). Generated fresh on each run.

---

### Why the tests are split the way they are

Both the backend and frontend test suites are split by purpose and seam, not by feature. This is intentional — when a test fails, the file it lives in tells you which layer to investigate first.

#### Backend test layout

| File                                     | Style                | Scope                                                                           |
| ---------------------------------------- | -------------------- | ------------------------------------------------------------------------------- |
| `api/test_newsletter.py`                 | pytest, fully mocked | Unit tests for service-layer functions. Patches `_client.make_request`.         |
| `api/test_newsletter_api_integration.py` | pytest + django_db   | View-level HTTP tests. Calls Django's test client.                              |
| `api/test_newsletter_integration.py`     | pytest               | Wires real views to mocked AC at the HTTP boundary; exercises end-to-end logic. |

The file split is deliberate — bugs caught only at the integration level (like the original missing `data` parameter on POST/PUT in `_make_ac_request`) wouldn't be visible in pure unit tests because the mock would always match the expected call shape.

The newsletter service has three meaningful seams: service functions ↔ AC HTTP, views ↔ services, and HTTP ↔ views. Each test file targets exactly one:

**`api/test_newsletter.py`**
- **Targets:** Service ↔ AC HTTP
- **What it does:** Patches `_client.make_request` directly. Asserts that service functions issue the right calls with the right arguments.
- **What it can catch:** Logic bugs: regex parsing, JSON parsing, list/variable merging, REPLACE vs UNION semantics, contact lookup-or-create branching.
- **What it can't catch:** Anything view-layer — auth, request parsing, CSRF, status codes.

**`api/test_newsletter_api_integration.py`**
- **Targets:** View ↔ Service
- **What it does:** Uses Django's test client with `pytest-django`. Stubs the service-layer functions called by views. Asserts on HTTP behavior.
- **What it can catch:** Auth handling (401 when logged out), HTTP method validation (405 on wrong method), request body parsing, response shapes, 503 fall-through when AC isn't configured.
- **What it can't catch:** Whether the service layer is actually correct — those are stubbed out.

**`api/test_newsletter_integration.py`**
- **Targets:** HTTP ↔ Service ↔ AC (mocked at AC boundary)
- **What it does:** Wires real views to real service functions, then mocks only the AC HTTP boundary. End-to-end exercise of one request.
- **What it can catch:** Integration bugs between layers — for example, a view passing arguments in the wrong order to a service function.
- **What it can't catch:** Production-only behaviors (real network errors, real AC rate limits).

**`api/conftest.py`**
- **Targets:** Shared infrastructure
- **What it does:** pytest fixtures that all three pytest files use — chiefly clearing `_field_id_cache` between tests and providing a mocked `_client` singleton.

The split exists because the bug-finding power of each layer is genuinely different. Running all four gives confidence that (a) the service-layer logic is correct in isolation, (b) the view layer makes the right calls, and (c) the two layers actually compose. Without the integration file, the unit and view tests can each pass while the system silently breaks at the join.

#### Frontend test layout

Frontend tests follow Jest's file-per-source-file convention (Foo.jsx → `__tests__/Foo.test.js`), which gives every component a dedicated test file. The container vs view split mirrors the source-code split:

- `NewsletterSignUpPageForm.test.js` tests only the orchestrator. The three views are mocked with stub components that capture the props they receive. Container tests assert "did the container pass the right props for this state?" without rendering DOM that belongs to the views.
- `NewsletterFormView.test.js`, `NewsletterConfirmationView.test.js`, `SuccessView.test.js` each render their respective view in isolation with synthetic props.
- `SelectableOption.test.js`, `MarketingEmailToggle.test.js` test the small reusable inputs directly.
- `NewsletterPage.test.js` tests the page wrapper that lives in `StaticPages.jsx`, including its conditional supplemental-content rendering.
- `newsletterApi.test.js` tests the API adapter against a mocked `global.fetch`.

A failing test in `NewsletterFormView.test.js` is unambiguously a rendering/interaction problem in that view; a failing test in `NewsletterSignUpPageForm.test.js` is unambiguously a state-management or orchestration problem in the container.

#### Playwright spec layout

The Playwright specs are split by concern rather than by user flow. One concern per file means a failure points to one cause.

| File                                      | Concern                                                                                             | Notes                                                                                                                                           |
| ----------------------------------------- | --------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
| `newsletter-signup-loggedout.spec.js`     | Happy-path logged-out flow                                                                          | Uses `newsletter-loggedout.har` fixture. The main end-to-end coverage.                                                                          |
| `newsletter-signup-loggedin-real.spec.js` | Logged-in flow against real Django auth                                                             | Only runs under the `authenticated` project. Suffix `-real` is what `playwright.config.ts`'s `testMatch: /.*-real.spec.js/` looks for.          |
| `newsletter-signup-validation.spec.js`    | Field-level validation (empty fields, invalid email, email mismatch, no list selected)              | Isolated so validation failures don't get drowned out by happy-path failures.                                                                   |
| `newsletter-signup-accessibility.spec.js` | Keyboard navigation, ARIA roles, focus management, error-summary announcement, skip-to-main-content | Self-contained because a11y violations are reviewed separately.                                                                                 |
| `newsletter-signup-responsive.spec.js`    | Mobile/tablet/desktop viewport behavior                                                             | Calls `page.setViewportSize` for each breakpoint and re-asserts layout.                                                                         |
| `newsletter-signup-api-states.spec.js`    | Error states: AC 503, missing newsletters, malformed responses                                      | HAR fixtures vary per test to simulate each backend condition.                                                                                  |

Each spec runs an expensive `beforeEach` (navigate to the page, set up a HAR fixture, dismiss the cookies banner), so amortizing that cost across many assertions within a single concern is also a practical reason to bundle them this way.

---

### Operational gotchas

- AC must be configured externally for newsletters to appear. Each list needs a `list_{N}_meta` personalization variable in AC with JSON content `{"icon": "...", "language": "..."}`. If the variable is missing or malformed, that list is silently filtered out of the response.
- Newsletter list is cached for 1 hour (`NEWSLETTER_LIST_CACHE_TTL = 60 * 60` in `newsletter_service.py`). Adding a new list in AC won't appear until the cache expires or is manually flushed.
- `_field_id_cache` in `newsletter_service.py` is a module-level dict. In tests, it can leak state between cases if not patched — there's a fixture for this in `api/conftest.py`.
- The three POST endpoints (`subscribe_newsletter`, `update_user_preferences`, `update_learning_level`) are `@csrf_protect` decorated. The frontend sends the CSRF token via the `X-CSRFToken` header (read from the `csrftoken` cookie by `js-cookie`). The frontend also uses `parseJsonResponse` to detect non-JSON responses (e.g., a CSRF-rejection HTML page) and surface a generic error instead of a parse exception.

---

### Security-relevant behaviors

- There is no logged-out unsubscribe path on this page. Logged-out users can only subscribe (UNION) — they cannot remove any list they're already on. This is by design: it prevents an unauthenticated attacker from POSTing `{email: "victim@example.com", newsletters: {}}` to clear someone else's subscriptions. Removal goes through either (a) the logged-in preferences endpoint with session auth, or (b) AC's own unsubscribe links in delivered email.
- `get_user_subscriptions` and `update_user_preferences` read email from `request.user`, never from the request body. Same reasoning — prevents email spoofing under an authenticated session.
- The subscribe endpoint validates `stringids` against the cached newsletter list. If a client tries to subscribe to a `stringid` that doesn't correspond to an AC list with a `list_{N}_meta` variable, the request is rejected. This prevents the endpoint from being used to enumerate or subscribe to unmanaged lists.

---

### Behavioral surprises

- Newsletter selection is required only for logged-out users. A logged-in user can submit with everything unchecked — that's a valid "unsubscribe me from all managed lists" action, and the marketing-opt-out toggle handles the broader intent (unmanaged lists too).
- Logged-in users skip the API call entirely if nothing changed. `getSubscriptionDiffs()` compares against the initial snapshot, and if no lists were added or removed and the opt-out didn't change, the form jumps to the confirmation stage with no network call.
- The opt-out toggle hits even unmanaged lists. When a logged-in user opts out, the backend calls `get_all_ac_list_ids()` (every list in the account, not just managed ones) and unsubscribes them from all of them. This is intentional — "marketing emails" means all marketing emails — but it's a meaningfully larger blast radius than "the lists shown on this page."
- Stale opt-out is read-fixed but not write-fixed. If MongoDB says opted-out but AC shows active subscriptions, the form trusts AC and shows the toggle ON. MongoDB is not corrected until the user submits.
- Newsletter IDs (not display labels) are used in analytics.
- Email examples are English-only. The Hebrew variant of `NewsletterPage` hides the example-emails block entirely. This is intentional because those emails don't have corresponding Hebrew versions.


api/newsletter_service.py
```diff
@@ -0,0 +1,1501 @@
...
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+from django.core.cache import cache
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email as django_validate_email
+
+try:
+    from sefaria.local_settings import (
```
yitzhakc, line 39: This should be imported via `from django.conf import settings`

api/newsletter_service.py
```diff
@@ -0,0 +1,1501 @@
+"""
+Newsletter Service Module
```
yitzhakc, line 2: This module is very heavy, with a lot of entrypoints, views and functions. I'd suggest a structural refactor into a separate module, multiple files, with a single file exposing the public services. I'd also recommend more distinctive naming for both the internal and external functions, in a way that the code will be easier to navigate by both human and agent.

api/newsletter_views.py
```diff
@@ -0,0 +1,516 @@
...
+    update_user_preferences_impl,
+    update_learning_level_impl,
+)
+
+logger: logging.Logger = logging.getLogger(__name__)
+P = ParamSpec("P")
+
+
+def handle_newsletter_view_errors(
+    action: str,
```
yitzhakc, line 37: I suggest that the action be an enum or constant, and not use magic strings

build/ci/createJobFromRollout.sh
```diff
@@ -24,7 +24,7 @@ EOF
 kubectl get rollout $DEPLOY_ENV-web -o yaml | yq '.spec.template.spec' > spec.yaml
 yq -i '.spec.template.spec += load("spec.yaml")' job.yaml
 yq -i '.spec.template.spec.restartPolicy = "Never"' job.yaml
-yq -i '.spec.template.spec.containers[0].args = ["-c", "pip3 install pytest-django; pytest -v -m \"not deep and not failing\" ./sefaria; echo $? > /dev/stdout; exit 0;"]' job.yaml
+yq -i '.spec.template.spec.containers[0].args = ["-c", "pip3 install pytest-django; pytest -v -m \"not deep and not failing\" ./sefaria ./api; echo $? > /dev/stdout; exit 0;"]' job.yaml
```
yitzhakc, line 27: While I'm 100% in favor of adding the api unit tests to our CI, I think we have some other tests there that might be failing, so adding the entire directory might be out of scope. Perhaps as part of the structural refactor you'll move the tests too, and then add them explicitly here.

static/js/NewsletterSignUpPage/bilingualUtils.js
```diff
@@ -0,0 +1,260 @@
...
+ */
+
+/**
+ * Bilingual text constants for common form labels
+ * Keeps translations in one place for easy maintenance
+ */
+export const BILINGUAL_TEXT = {
+  // Hero section (page header above the form)
+  HERO_TITLE: {
+    en: "Stay Connected",
```
yitzhakc, line 14: I think these strings might belong in weblate?

yitzhakc (changes requested):
Left some comments. I recommend merging in from `master` and reconciling any conflicts too.

api/newsletter_service.py
stevekaplan123, line 39: **The newsletter test suite only passes when a real ActiveCampaign token is present.**

These credentials are imported straight from `sefaria.local_settings` at module load instead of through `django.conf.settings`, so the CI/example fallbacks in `sefaria/local_settings_ci.py:205-206` and `local_settings_example.py:150-151` are never consulted. When the names are absent, both values become `None` and the module-level `_client` (lines 287-291) is frozen to `None` for the life of the process.

Every test that does `@mock.patch("api.newsletter_service._client.make_request")` (36 sites in `api/test_newsletter.py`, 12 in `api/test_newsletter_integration.py`) then errors at setup, and the two `_get_client()` calls at `api/test_newsletter.py:142,158` hit the `assert _client is not None`.

Reproduced on a checkout whose `local_settings.py` has no `ACTIVECAMPAIGN_*` names:

```
python -m pytest -c sefaria/pytest.ini api/test_newsletter.py -k TestGetAllLists -x -q
FAILED ...TestGetAllLists::test_get_all_lists_success - AttributeError: None does not have the attribute 'make_request'
```

Because this PR also adds `./api` to `build/ci/createJobFromRollout.sh:27`, the pytest job now passes only if the sandbox pod happens to have `ACTIVECAMPAIGN_API_TOKEN` injected from the optional `local-settings-secrets` (helm `local-settings-file.yaml:350-351` reads it via `os.getenv`). Nothing in `api/conftest.py` installs a fake client.

Suggested fix: read the values lazily via `django.conf.settings` inside `_get_client()` (as @yitzhakc also suggested above), and have `api/conftest.py` install a `NewsletterClient("test", "test")` (or patch `_get_client`) so the suite never depends on the environment.

api/newsletter_service.py
```diff
@@ -0,0 +1,1501 @@
...
+    if a caller mutates and saves that object, a missing Mongo profile document
+    will be created then. Returns None only when no Django user can be resolved
+    for `email`.
+
+    Why this exists as a separate helper: the MongoDB lookup happens exactly
+    once per call to the parent service function. Explicit loading at the
+    call site makes the cost visible.
+    """
+    try:
+        profile = UserProfile(email=email, user_registration=True)
```
stevekaplan123, line 1310: **A Django user with no Mongo profile document ends up with a permanent empty slug.**

`UserProfile(email=..., user_registration=True)` skips the constructor's auto-create branch at `sefaria/model/user_profile.py:413-419` (`elif self.exists() and not user_registration:` -> `assign_slug()` + `save()`). It still returns an object with `id` set, `_id = None` and `slug = ""` (default at line 347), so this helper returns it as a valid profile.

The callers then mutate and save it: `_update_wants_marketing_emails` (line 1336) and `update_learning_level_impl` (line 1474). With `_id = None`, `save()` takes the `insert_one` branch at `user_profile.py:509-511` and writes a document with `slug: ""`. On every later `UserProfile(id=N)` load the document is found at line 407, so the `assign_slug` branch never runs again. Consequences: `profileUrl` becomes `/profile/` (line 844), and a second such user's next profile edit fails `errors()` because slug `""` is already taken (lines 536-538).

The precondition (Postgres user, no `db.profiles` doc) is the case the constructor comment at lines 414-416 describes, and the spam-cleanup path at `sefaria/views.py:1337-1341` produces it by deleting the profile doc while keeping the Django user. Because `/api/newsletter/learning-level` accepts an arbitrary email when logged out, this is reachable anonymously.

Suggested fix: either drop `user_registration=True` here so the standard create path runs, or, if the intent is "don't create profiles from this endpoint", return `None` when `profile._id is None` so the callers skip the save.

static/js/StaticPages.jsx
```diff
@@ -547,8 +550,8 @@ const EducatorsPage = () => (
 
     <ButtonRow white={true} enTitle="" heTitle="">
       { [
-          ["Get Sefaria for Educators", "מדריך למשתמש בספריא", "https://newsletter.sefaria.org/f/40", "https://chiburim.sefaria.org.il/sheets/361600?lang=he"],
-          ["Past Educator Newsletter", "ספריא לתלמידים", "/collections/educator-newsletters?tab=sheets&utm_source=sefaria&utm_medium=landingpage&utm_campaign=educators", "/collections/educator-newsletters?tab=sheets&utm_source=sefaria&utm_medium=landingpage&utm_campaign=educators"],
+          ["Get Sefaria for Educators", "מדריך למשתמש בספריא", "/newsletter", "/newsletter"],
```
stevekaplan123, line 553: **The Hebrew "user guide" button now opens the newsletter form, and the utm params on the next row were dropped.**

Before this PR (origin/master line 550) this row was:

```
["Get Sefaria for Educators", "מדריך למשתמש בספריא", "https://newsletter.sefaria.org/f/40", "https://chiburim.sefaria.org.il/sheets/361600?lang=he"]
```

Repointing the English href to `/newsletter` looks intentional, but the Hebrew href (index 3) was also replaced, while its label still reads "מדריך למשתמש בספריא" (Sefaria user guide). `SimpleButton` renders the `int-he` anchor with `href={he_href || href}` (lines 2287-2296), so a Hebrew-interface visitor who reads the paragraph above (line 545, describing a step-by-step guide) and clicks the button lands on the English sign-up page instead of the guide sheet.

The row below (line 554) also lost `&utm_source=sefaria&utm_medium=landingpage&utm_campaign=educators` from both hrefs, so GA campaign attribution for `/collections/educator-newsletters` from this page drops to zero. The other `ButtonRow`s on the same page (lines 585-586, 616-617) keep their utm params and distinct Hebrew links.

Suggested fix: restore the Hebrew href to the chiburim guide sheet (or, if the Hebrew button is meant to change too, update its label), and restore the utm params on line 554.

stevekaplan123 (commented):
Three items from an automated review pass that Copilot's review did not raise (details inline).

static/js/NewsletterSignUpPage/bilingualUtils.js
stevekaplan123: Yes, though this branch hasn't merged in master.  Once you merge in master, you'll find the files Sefaria-Project/static/js/sefaria/i18n/interface/en.json and he.json that you can put these strings.  See the new `_` in sefaria.js (`_: function(inputStr)`) function so you can see how to properly handle en and he strings. @relyks

PR #3191, opened 2026-03-25
Ref api

YishaiGlasner:
## Summary

- Add a new **`/api/ref/<tref>`** endpoint that returns structured metadata for any Sefaria `Ref`
- Provide consistent navigation and structure metadata across node types (JaggedArrayNode, SchemaNode, Dictionary, Sheets, etc.)
- Avoids redundant MongoDB queries by ensuring a single vstate fetch per request for standard text refs (i.e. not virtual nodes)
- Fix `prev_segment_ref` / `next_segment_ref` to correctly handle virtual nodes (e.g. Siddur)
- Add optional `state_ja` parameter to avoid redundant DB calls when state is already available
- Add a pymongo `QueryCounter` listener for asserting query counts in tests
- Add comprehensive test coverage and OpenAPI documentation

## API Details

**New endpoint: `GET /api/ref/<tref>`**

Returns a JSON object with:
- `is_ref` — whether the input resolves to a valid ref (returns `{is_ref: false}` for invalid input)
- `normalized`, `hebrew`, `url_ref` — normalized representations
- `index_title`, `node_type` — index and node metadata
- `depth`, `address_types`, `section_names` — structure info (for JaggedArrayNode / DictionaryEntryNode)
- `start_indexes`, `start_labels`, `end_indexes`, `end_labels` — section position
- `navigation_refs` — contextual navigation:
  - `lineage_refs_top_down` — ancestor refs from root to parent
  - `first_available_section_ref` — first section with content
  - `first_subref` / `last_subref` — child navigation (non-segment, non-range)
  - `prev_section_ref` / `next_section_ref` — section-level navigation
  - `prev_segment_ref` / `next_segment_ref` — segment-level navigation
- `children` — child node titles (for SchemaNode)
- `default_child_node` — default child metadata when applicable
- `sheet_id`, `lexicon_name`, `headword` — type-specific fields

### Considerations
- **Navigation scope**  
  `prev_*` and `next_*` are only defined for section-level and segment-level refs.  
  Navigation at higher levels is intentionally not exposed to avoid ambiguity. Consumers can traverse upward (via `lineage_refs_top_down`) and derive such relationships if needed.
- **Field presence**  
  Fields that are not applicable to a given ref type are omitted.  
  Fields that are applicable but have no value (e.g. no previous or next ref exists) are returned as `null`.

## Changes in Ref
- Fix `prev_segment_ref` and `next_segment_ref` to support DictionaryEntryNode
- Add optional `state_ja` parameter to selected methods (already supported in others) to improve performance
- Add function `get_subrefs_count` and refactor `all_subrefs`, `prev_segment_ref` and `next_segment_ref` to use it.

## pymongo listener
Adds `QueryCounter`, a pymongo `CommandListener` used in tests to:
- Count MongoDB queries per request
- Record query tracebacks for debugging
Tests reset the counter before each API call and assert on `QueryCounter.count`. On failure, full query tracebacks are printed to help identify unnecessary database hits.
The listener is only registered in test environments (`sys._called_from_test`), so there is zero production overhead.

## Note on tests
`api/tests.py` is currently not part of the CI suite (historical decision).  
All new tests were added there and can be run locally.


api/views.py
```diff
@@ -67,3 +67,89 @@ def get(self, request, *args, **kwargs):
...
+            'url_ref': oref.url(),
+            'index_title': index.title,
+            'node_type': type(index_node).__name__,
+            'navigation_refs': {
+                'lineage_refs_top_down': [r.normal() for r in oref.all_context_refs(False, True)]
+            }
+        }
+
+        if return_object['node_type'] != 'SheetNode':
+            return_object['navigation_refs']['first_available_section_ref'] = oref.first_available_section_ref(vstate=vstate).normal()
```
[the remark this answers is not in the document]
YishaiGlasner: it can't get ComplexBookLevelRefError - it will go to padded_ref only if the node is ja, but this error will raise only if the node has no depth (i.e. not ja).
fixing the none issue

sefaria/model/text.py
```diff
@@ -3625,34 +3626,48 @@ def prev_segment_ref(self):
             d["sections"] = d["toSections"] = r.sections[:-1] + [r.sections[-1] - 1]
             return Ref(_obj=d)
         else:
-            r = r.prev_section_ref()
+            r = r.prev_section_ref(vstate=vstate)
             if not r:
                 return None
+            if self.index_node.is_virtual:
+                return r.all_subrefs()[0]
```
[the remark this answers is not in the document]
YishaiGlasner: changing to -1
the assumption that previous section has segments is also when not vurtual.

sefaria/model/text.py
```diff
@@ -3625,34 +3626,48 @@ def prev_segment_ref(self):
...
         r = self.ending_ref()
         if not r.is_segment_level():
             return r
+        if self.index_node.is_virtual:
+            section_ref = self.context_ref()
+            siblings = section_ref.all_subrefs()
+            curr_index = siblings.index(self)
+            if len(siblings) == curr_index + 1:
+                next_section = section_ref.next_section_ref(vstate=vstate)
+                return next_section.all_subrefs()[0] if next_section else None
```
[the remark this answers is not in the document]
YishaiGlasner: i can't imagone the scenario of having LexiconEntry with headword but withno text. it technically can be, for the text is defined specifically for any lexicon, but that not something i think we should guard

docs/openAPI.json
```diff
@@ -13904,6 +14042,206 @@
...
+            "description": "The type of the index node. Determines which additional fields are present in the response.",
+            "type": "string",
+            "enum": ["JaggedArrayNode", "SchemaNode", "DictionaryNode", "DictionaryEntryNode", "SheetNode"]
+          },
+          "navigation_refs": {
+            "description": "An object containing related references for navigating within the text.",
+            "type": "object",
+            "properties": {
+              "lineage_refs_top_down": {
+                "description": "An array of references representing the hierarchical path from the book level down to the immediate parent. Empty array if at book level.",
```
stevekaplan123, line 14081: I'm not crazy about the name 'lineage_refs_top_down'.  What about 'ancestor_refs'?   When do you foresee it being used by API users?

sefaria/model/text.py
```diff
@@ -3648,34 +3649,48 @@ def prev_segment_ref(self):
...
 
         If this ref is not segment level, will return ``self```
 
+        :param vstate: optional pre-fetched VersionState to avoid DB calls
         :return: :class:`Ref`
         """
         r = self.ending_ref()
         if not r.is_segment_level():
             return r
+        if self.index_node.is_virtual:
```
stevekaplan123, line 3675: Instead of special casing for VirtualNodes, I'm wondering if you could take advantage of the fact that VirtualNodes and JaggedArrayNodes inherit from TitledTreeNodes and create a function in TitledTreeNodes that handles the base case.  Then you can override that function in VirtualNode.  That way, here you can just call this function on self.index_node rather than having a special case both here and in prev_segment_ref.

stevekaplan123 (changes requested):

docs/openAPI.json
stevekaplan123: It's essentially the shortest path to the root so that fact could also factor in the name: 'shortest_path_to_root', but 'ancestor_refs' might be sufficient

YishaiGlasner: i neither crazy, but i didn't want to use 'ancestor', for we nodes where we use the word ancestor, like Ramban on genesis is the father of Ramban on Genesis, Introuduction; so i'm not sure we want to call Genesis 1 the father of Genesis 1:5. what do you think?

sefaria/model/text.py
YishaiGlasner: i don't think the node is the right place for function that finds the next *segment*. the place to find it is the vstate.

stevekaplan123: That may be true, and we special case VirtualNodes a lot it seems.  And in this function half of it is dealing with the special case of VirtualNodes.  What do you mean the place to find it is the vstate?  This is the Ref class.

YishaiGlasner: i mean that we're finding the next segment by checking the vstate, but the node is just the jagged array. it can tell you its depth. maybe you can expect it to tell you the next jagged array node. but it doesn't know what populates it (in some cases we do have a limitation on the top level, like 50 chapters in Genesis, but no regarding the verses in any chapter). so i don't think the node should have a method for the next segment.

YishaiGlasner:
Spec ing ref API and implementing [sc-40554]

docs/openAPI.json
stevekaplan123: I see what you mean.  We have a node function `is_ancestor_of`.  That means to me that `shortest_path_to_root` might be a good idea as it avoids 'ancestor' or 'lineage' entirely.

sefaria/model/text.py
stevekaplan123: I talked it over with my friend Cursor.  What do you think of something like this?  It allows us to be agnostic of node type in next_segment_ref and prev_segment_ref because of a new function _section_segment_count.
``` 
def _section_segment_count(self, section_ref, vstate=None):
        """Number of segments in the given section_ref."""
        if self.index_node.is_virtual:
            return len(section_ref.all_subrefs())
        ja = self.get_state_ja(vstate=vstate)
        return ja.sub_array_length([i - 1 for i in section_ref.sections])

    def prev_segment_ref(self, vstate=None):
        """
        Returns a :class:`Ref` to the next previous populated segment.

        If this ref is not segment level, will return ``self```

        :param vstate: optional pre-fetched VersionState to avoid DB calls
        :return: :class:`Ref`
        """
        r = self.starting_ref()
        if not r.is_segment_level():
            return r
        if r.sections[-1] > 1:
            d = r._core_dict()
            d["sections"] = d["toSections"] = r.sections[:-1] + [r.sections[-1] - 1]
            return Ref(_obj=d)
        else:
            r = r.prev_section_ref(vstate=vstate)
            if not r:
                return None
            last_index = self._section_segment_count(r, vstate=vstate)
            d = r._core_dict()
            d["sections"] = d["toSections"] = r.sections + [last_index]
            return Ref(_obj=d)

    def next_segment_ref(self, vstate=None):
        """
        Returns a :class:`Ref` to the next populated segment.

        If this ref is not segment level, will return ``self```

        :param vstate: optional pre-fetched VersionState to avoid DB calls
        :return: :class:`Ref`
        """
        r = self.ending_ref()
        if not r.is_segment_level():
            return r
        section_ref = r.section_ref()
        section_length = self._section_segment_count(section_ref, vstate=vstate)
        if r.sections[-1] < section_length:
            d = r._core_dict()
            d["sections"] = d["toSections"] = r.sections[:-1] + [r.sections[-1] + 1]
            return Ref(_obj=d)
        else:
            try:
                return r.next_section_ref(vstate=vstate).subref(1)
            except AttributeError:
                return None
```

stevekaplan123 (commented):
I suggested a possible approach for a re-factor that allows us to avoid checking node type in next_segment_ref or prev_segment_ref.  Curious what you think.

docs/openAPI.json
YishaiGlasner: done

sefaria/model/text.py
YishaiGlasner: great idea and i want to make it greater

stevekaplan123 (dismissed):

stevekaplan123 (dismissed):

docs/openAPI.json
```diff
@@ -19222,6 +19226,140 @@
...
+          "name": "tref",
+          "description": "A text reference string to validate and analyze. Can be any Sefaria reference, including book-level, section-level, segment-level, or ranged references.",
+          "schema": {
+            "type": "string"
+          },
+          "in": "path",
+          "required": true,
+          "examples": {
+            "Segment Level": {
+              "value": "Genesis 1:1"
```
saengel, line 19354: Can you add `"default": true` so this value is auto-populated in the field on the ReadMe GUI side?

saengel (changes requested):
Overall looks great!

Two small changes:
1. Making the first example a "default" value so it auto-populates in the docs and avoids misleading 404s (a GUI issue). Every other endpoint should have this (minus sheets) so you can look for examples. 
2. The `Ref` tag is coming after `MISC` and is making it look funny in the sidebar, `MISC` should be last. Can you rearrange so the MISC category appears last? Attaching a screenshot from the test env:

Image

yitzhakc (approved):

saengel (approved):

PR #3207, opened 2026-04-14
chore: guard texts_api against whole-book commentary+pad=0 requests

EliezerIsrael:
Reject /api/texts/<book>?commentary=1&pad=0 for book-level refs, which would otherwise expand links across the entire book.


akiva10b (commented):
looks great - I think you should add one negative test making sure you cannot get that api

akiva10b (approved):

PR #3215, opened 2026-04-16
Bug/sc 40416/ research copy and paste from ios creates odd2

stevekaplan123:
## Description
Generated image previews for social media included stray HTML entities such as `&nbsp;`.   This cleans this up as well as cleaning up any stray HTML tags.

## Code Changes
We should have one way of handling this across our apps so I intentionally copied the code in the front-end that handles copying text: `Sefaria.util.htmlToText`, and added tests.

## Notes
This is part of [sc-43402]


sefaria/image_generator.py
```diff
@@ -59,11 +60,37 @@ def calc_letters_per_line(text, font, img_width):
...
+    html = html.replace("</td>", "\t")
+    html = html.replace("</table>", "\n")
+    html = html.replace("</tr>", "\n")
+    html = html.replace("</p>", "\n")
+    html = html.replace("</div>", "\n")
+    html = html.replace("<br>", "\n")
+    html = re.sub(r"<br( )*/>", "\n", html)
+
+    # Parse HTML and extract text, mirroring the JS DOMParser+textContent path.
+    text = BeautifulSoup(html, "html.parser").get_text()
```
yitzhakc, line 84: I believe `BeautifulSoup` is a bit of a "heavy" dependency if we're just parsing a relatively small html chunk and extracting the text. Did you try any other options?

At the very least, you should be using the `lxml` parser (and maybe use `cchardet`) per [BS4's docs on improving performance](https://beautiful-soup-4.readthedocs.io/en/latest/#improving-performance)

yitzhakc (changes requested):

saengel (commented):
A little puzzled, if we need "one way of handling this across apps" and there is existing logic for this somewhere, can we just have one centralized function and call that? Or is this an effort to build that centralized function (in utils)?

sefaria/image_generator.py
stevekaplan123: I swapped out the parser from html to lxml to speed things up.  I kept BeautifulSoup though because I wanted to achieve parity as much as possible with the copy functionality in util.js `htmlToText`.  My thought was to implement the same bug fix in as identical a way as possible as we handle copying text in the front-end.

stevekaplan123:
> A little puzzled, if we need "one way of handling this across apps" and there is existing logic for this somewhere, can we just have one centralized function and call that? Or is this an effort to build that centralized function (in utils)?

The way to do that would be to change the texts API itself.  However, the problem from my perspective is that we don't want to filter out all HTML since some of it is useful for display purposes.  Therefore, we need to get the HTML in the API and then filter it out later downstream.

yitzhakc (approved):

PR #3246, opened 2026-04-26
Refactor CSV export to streamline functionality

yonadavGit:


sefaria/export.py
```diff
@@ -616,11 +616,22 @@ def export_all():
 # Row 4: Version Source
 
 
+def _sorted_segment_refs(refs):
+    """Sort walked segment refs into canonical textual order for stable CSV output."""
+    return sorted(refs, key=lambda tref: Ref(tref).order_id())
+
+
+def _walk_version_segments(version, action, schema, he_tref):
```
nsantacruz: I'm not a big fan of one line functions, the functions at the top here are an example of those.

nsantacruz (approved):

PR #3264, opened 2026-05-06
chore: pass short lang because svelte-i18n uses two letter language keys

stevekaplan123:
## Description
We should be passing 'en' or 'he' rather than 'english' or 'hebrew' because svelte-i18n uses two letter language keys


YishaiGlasner (commented):
if we submit the full language, i guess it means svelte uses it, so that seems as a breaking chage. did you check we don't assume somewhere we have the whole string?

stevekaplan123:
> if we submit the full language, i guess it means svelte uses it, so that seems as a breaking chage. did you check we don't assume somewhere we have the whole string?

That's a fair point, but I've made the relevant changes in the ai-chatbot in a separate PR.

YishaiGlasner (approved):

PR #3273, opened 2026-05-06
fix: remove punctuation when matching dibur hamatchil

nsantacruz:
This pull request enhances Hebrew text processing in the `sefaria` codebase, focusing on improving abbreviation handling and test coverage for ambiguous references. The main updates include adding punctuation-stripping to Hebrew string comparisons and expanding test cases for ambiguity resolution.

Improvements to Hebrew string comparison:

* Added a new helper function `_strip_hebrew_punctuation` in `sefaria/utils/hebrew.py` to remove punctuation from Hebrew strings before processing. This ensures more accurate comparisons, especially for abbreviation expansion and matching.
* Updated the `hebrew_starts_with` function to use `_strip_hebrew_punctuation` on both input strings, improving its robustness against punctuation differences.

Test coverage enhancements:

* Added a new test case to `test_multiple_ambiguities` in `sefaria/model/linker/tests/linker_test.py` to cover ambiguity resolution involving Tosafot on Berakhot 13a.

### DIFF
--- a/sefaria/model/linker/tests/linker_test.py
+++ b/sefaria/model/linker/tests/linker_test.py
@@ -95,6 +95,7 @@ def test_multiple_ambiguities():
 
 
 @pytest.mark.parametrize(('resolver_data', 'expected_trefs'), [
+    [crrd(["@תוספות", "@ברכות", '#י"ג ע"א', '''*ד"ה 'עד כאן\'''']), ("Tosafot on Berakhot 13a:36:1",)],
     # Numbered JAs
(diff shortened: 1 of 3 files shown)


YishaiGlasner:
does `hebrew_starts_with` run only on dibur hamatchil strings?
anyway,what about ד"ה אע"פ? would it find it as אעפ?
and what about hyphens and makaf? if they exist in dibur hamatchil you'll get one word rather than 2.

nsantacruz:
@YishaiGlasner good points... hebrew_starts_with is meant to check if one string starts with another but it takes into account abbreviations. It's currently only used by the linker but it should be somewhat general. Removing punctuation is a general thing that is good.

However, you're right that this breaks abbreviation parsing. I actually have tests for that that are now failing 😅 . I'll fix that.

YishaiGlasner (approved):

PR #3283, opened 2026-05-11
fix(sc-44205): fix password reset confirm URL pattern for Django 6

yodem:
## Summary
- The `password_reset_confirm` URL pattern used a restrictive regex (`{1,13}-{1,20}`) for the token that doesn't match Django 6's SHA-256-based token format (32+ char hash segments)
- This caused a `NoReverseMatch` error when generating password reset confirmation emails, completely breaking the password reset flow
- Switched from `re_path` with hardcoded regex to `path()` with simple converters, matching Django's own URL configuration
- Added regression tests to prevent future token format incompatibilities

## Root Cause
Django 6 uses SHA-256 for password reset tokens (vs SHA-1 in older versions), producing longer hash segments. The old regex `[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20}` rejected tokens like `cbbab4-e586b88a5c84910507df7a562596bf` where the hash is 32 chars.

## Test plan
- [ ] CI passes `reader/tests/password_reset_urls_test.py` (8 tests: SHA-256 token, legacy token, set-password internal token, resolve, parametrized hash lengths 20/32/40/64)
- [ ] Trigger password reset flow on staging — email should contain a valid confirmation link
- [ ] Click the confirmation link — should render the password reset form (not a 404/500)
- [ ] Complete the password reset — new password should work

Subtask of sc-40101 (Upgrade to Django 6)


stevekaplan123 (approved):

YishaiGlasner (commented):
are we changing here the allowed chars for password? and this is only on reset but not in first registration?

PR #3296, opened 2026-05-13
Feature/sc 43550/add ability to change version title to bulk

stevekaplan123:
## Description
Adds a “Rename Version Title” action to ModTools’ Bulk Version Editor by introducing a new per-index backend endpoint (/api/version-rename) and wiring a frontend workflow that iterates selected indices and aggregates queue/success/failure results.  The dependencies for versionTitle changes take a very long time as we modify the History collection and the ElasticSearch database.  In order to handle the sheer volume of dependency tasks, in production and cauldron environments, the version rename API takes advantage of a Celery queue and when its finally done, a Slack notification is sent to #engineering-signal for each of the versions renamed.


sefaria/helper/texts/tasks.py
```diff
@@ -90,6 +91,124 @@ def save_version(raw_version_change: dict):
...
+    try:
+        load_query = {"title": index_title, "versionTitle": version_title}
+        if language:
+            load_query["language"] = language
+        version = Version().load(load_query)
+        if not version:
+            return {"status": "error", "index": index_title, "error": f'No Version "{version_title}" found'}, 404
+
+        collision_query = {"title": index_title, "versionTitle": new_version_title}
+        if language:
```
YishaiGlasner: 1. i see no reason to prevent 2 versions with the same "langugae" (i.e. ltr/rtl) and versionTitle, but only to prevent them when the motherGamiliyLanguage is the same (but i i may be wrong)
2. why not taking the language from `version`?

sefaria/views.py
```diff
@@ -1815,6 +1815,76 @@ def version_bulk_edit_api(request):
...
+      POST {"versionTitle": "...", "newVersionTitle": "...", "index": "...", "language": "he" (optional)}
+
+    Response:
+      Celery enabled: 202 {"task_id": "..."}
+      Celery disabled: 200 {"status": "ok"} or non-200 {"error": "..."}
+
+    Callers that need to rename a versionTitle across many indices should call
+    this endpoint once per index and aggregate per-index results.
+    """
+    if request.method != "POST":
```
YishaiGlasner, line 1834: why not PATCH?

sefaria/views.py
```diff
@@ -1815,6 +1815,76 @@ def version_bulk_edit_api(request):
...
+    rename_payload = {
+        "user_id": request.user.id,
+        "versionTitle": version_title,
+        "newVersionTitle": new_version_title,
+        "index": index_title,
+        "language": language,
+    }
+
+    if CELERY_ENABLED:
+        from sefaria.helper.texts.tasks import rename_version_title
```
YishaiGlasner: any reason no to import at file beginning?

static/js/modtools/components/BulkVersionEditor.jsx
```diff
@@ -446,6 +455,86 @@ const BulkVersionEditor = () => {
...
+        setMsg({
+          type: MESSAGE_TYPES.WARNING,
+          message: `Queued ${successCount}/${total} renames.\n\nQueueing failed:\n${failureList}\n\n${slackReminder}`
+        });
+      } else {
+        const failureList = failures.map(f => `• ${f.index}: ${f.error}`).join("\n");
+        console.error("Rename queueing failed", { url, failureCount, total, failures });
+        setMsg({ type: MESSAGE_TYPES.ERROR, message: `All ${failureCount} renames failed to queue:\n${failureList}` });
+      }
+    } finally {
```
YishaiGlasner, line 538: no catch?

sefaria/helper/texts/tasks.py
stevekaplan123: Great point.  Changing...

sefaria/views.py
stevekaplan123: I tried PATCH due to your suggestion.  Turns out Varnish doesn't allow it

YishaiGlasner: this is weird - we have patch in reader/views.:1812

static/js/modtools/components/BulkVersionEditor.jsx
stevekaplan123: added

sefaria/views.py
stevekaplan123: added to beginning

YishaiGlasner (approved):

PR #3298, opened 2026-05-13
Small fix to reorder tags

saengel:
A small fix to the OpenAPI JSON file to fix the tag ordering, see screenshot:
Screenshot 2026-05-13 at 10 04 08


stevekaplan123 (approved):
Assuming that the difference is solely due to formatting

PR #3329, opened 2026-05-24
Create a a section sampling script from db to google cloud bucket

yonadavGit:
## Description

## Code Changes

## Notes

### DIFF
--- a/scripts/sample_library_sections.py
+++ b/scripts/sample_library_sections.py
@@ -3,12 +3,14 @@ Sample Hebrew section-level refs from the library and upload the resulting JSON 
 
 The sampler:
 - uniformly samples eligible sections via reservoir sampling
+- can collect all eligible sections with `--all`
 - excludes sheet/dictionary/reference works
 - keeps only sections whose sampled Hebrew version has non-empty text for every included segment
 - writes the JSON locally and uploads it to `custom_embeddings/<output filename>` in GCS
 
 Examples:
     python scripts/sample_library_sections.py 10
+    python scripts/sample_library_sections.py --all
     python scripts/sample_library_sections.py 100 --seed 613
     python scripts/sample_library_sections.py 100 --output scripts/output/my_sample.json
     python scripts/sample_library_sections.py 100 --bucket development-research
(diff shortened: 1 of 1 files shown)


scripts/sample_library_sections.py
```diff
@@ -0,0 +1,214 @@
...
+from google.cloud.exceptions import GoogleCloudError
+from sefaria.google_storage_manager import GoogleStorageManager
+from sefaria.model import Ref, VersionSet, library
+from tqdm import tqdm
+
+
+DEFAULT_BUCKET = "development-research"
+DEFAULT_OUTPUT = Path("scripts/output/sampled_library_sections.json")
+EXCLUDED_CATEGORIES = {"Sheets", "Dictionary", "Reference"}
+BUCKET_PREFIX = "custom_embeddings"
```
nsantacruz: Should be `BLOB_PREFIX`. It's not the bucket

scripts/sample_library_sections.py
```diff
@@ -0,0 +1,214 @@
...
+def build_sefaria_url(oref):
+    return f"https://www.sefaria.org/{oref.url()}"
+
+
+def is_supported_index(index):
+    categories = set(index.categories or [])
+    return categories.isdisjoint(EXCLUDED_CATEGORIES)
+
+
+def parent_section_ref(oref):
```
nsantacruz: this function is to optimize speed? Because oref.section_ref() exists (although it is slow)

scripts/sample_library_sections.py
```diff
@@ -0,0 +1,214 @@
...
+def build_hebrew_version_map(indexes):
+    titles = [index.title for index in indexes]
+    versions_by_title = defaultdict(list)
+    version_set = VersionSet({"title": {"$in": titles}, "language": "he"})
+    for version in tqdm(version_set, desc="Loading Hebrew versions"):
+        versions_by_title[version.title].append(version)
+    return versions_by_title
+
+
+def iter_section_payloads_for_index(index, versions_by_title):
```
nsantacruz: seems like a waste to have versions_by_title be a parameter. You only care about the versions for it. You also only use `index.title` and never need the index itself.

nsantacruz: actually you only use the first version even

scripts/sample_library_sections.py
```diff
@@ -0,0 +1,214 @@
...
+            section_key,
+            {
+                "ref_obj": section_ref,
+                "segments": OrderedDict(),
+            },
+        )
+
+        normalized_text = segment_text.strip()
+        if not normalized_text:
+            empty_segment_refs.add(section_key)
```
nsantacruz: this is a confusing variable name. should be something like `paritally_empty_sections`

PR #3331, opened 2026-05-25
Different social images for modules

relyks:
## Description
  - Adds modularization for social image generation.
  - Lets `/api/img-gen/` return the right kind of image for text refs, static pages, unknown paths, and module pages.
  - Adds Library and Voices branding for generated social images and fallback images.

  ## Code changes
  - Added page classification before image generation so static pages and module pages do not get treated like text refs.
  - Added fallback social images for empty paths, unsupported paths, and pages without custom image support.
  - Added Library and Voices logo assets in English and Hebrew (regular and white versions).
  - Updated generated text-ref images to use the correct module and language logo.
  - Updated `/api/img-gen/` so missing, invalid, or bi language values fall back to the request host language.
  - Added module detection for `/api/img-gen/` requests.
  - Updated OpenAPI docs for the expanded social image endpoint behavior.

  ## Testing
  - Added API tests for text refs, static pages, module fallback pages, empty paths, host language defaults, and active module handling.
  - Added image generator tests for module-specific logos, fallback logo behavior, static fallback images, and cache headers.
  - Added middleware tests to confirm social image requests use the host module.
  - Expanded Playwright coverage to check fallback images, no-path image requests, Twitter-sized images, and Hebrew image generation.

## Notes
  - Used Django URL resolution to classify pages because it reuses the same routing rules as the backend.
  - Avoids guessing from path strings, which is risky because paths like static pages, and module-specific pages can look similar but need different image types.
  - Other approaches, such as putting more logic into templates, would have required more page-specific wiring and more repeated logic across templates.
  - Keeping the classification in the image endpoint makes the behavior more centralized and easier to test.
  - `/api/img-gen/` is an API route, but unlike normal APIs it needs host/module context so it can generate
  the correct social image branding. 
  - Since `ModuleMiddleware` normally skips all `/api/` paths, these requests would otherwise default to the Library module even on Voices domains. The explicit exception lets image-generation requests set `request.active_module` from the host, while keeping the normal API-skip behavior for all other `/api/` routes.
  - Keeping this endpoint under `/api/img-gen/` preserves backwards compatibility with existing share URLs, metadata tags, and crawler requests that already use that path (or other consumers)


reader/tests/social_image_api_test.py
```diff
@@ -34,67 +48,198 @@ def contents(self):
...
-    return RequestFactory().get(path)
+def _request(path, active_module=None, host=None):
+    request = RequestFactory().get(path, HTTP_HOST=host or "localsefaria.xyz:8000")
+    if active_module:
+        request.active_module = active_module
+    return request
 
 
-def _capture_social_image_call(monkeypatch, path):
+def _capture_social_image_call(monkeypatch, path, active_module=None, host=None, tref="Genesis.1.1"):
```
yitzhakc: Why do you set a default for the `tref` of the utility function? Don't we want it to be specified (or `None`?)

reader/tests/social_image_api_test.py
```diff
@@ -34,67 +48,198 @@ def contents(self):
         }
 
 
-def _request(path):
-    return RequestFactory().get(path)
+def _request(path, active_module=None, host=None):
+    request = RequestFactory().get(path, HTTP_HOST=host or "localsefaria.xyz:8000")
```
yitzhakc: Can you add a docstring here? 
If we want host to default to `"localsefaria.xyz:8000"` when it's `None`, I would do it in the default at the function decleration (ie. line 41 `host="localsefaria.xyz:8000"`)

reader/views.py
```diff
@@ -670,6 +681,56 @@ def _extract_version_title_param(request, key):
     return params
 
 
+def _social_image_urlconf_for_module(module: str) -> str:
```
yitzhakc: Do you want to see if there's a builtin function as part of `django-hosts`, instead of hardcoding these values?
See https://django-hosts.readthedocs.io/en/latest/reference.html#python-helpers

yitzhakc (changes requested):

sefaria/image_generator.py
```diff
@@ -5,6 +5,45 @@
 from django.http import HttpResponse
 import io
```
yitzhakc, line 6: Could you please add docstrings to the functions?

yitzhakc (changes requested):

reader/views.py
relyks: Yeah, I'll look into it. I didn't realize we use `django-hosts` for the hostname resolution of the modules to the URL configs.

reader/tests/social_image_api_test.py
relyks: Yeah, you're right. That should only be defined as part of each test for consistency

reader/views.py
relyks: In order to take advantage of using `django_hosts` to resolve the host (and get the url conf for the module), I need to change the key for the `VOICES_MODULE` from 'sheets' to 'voices'. I'm pretty sure this won't be an issue, because `get_host` is called as a method on the request objects in the other places in the codebase (from what I found). The `name` parameter passed for the host definition in `sefaria/hosts.py` only acts as an internal lookup key for the host definitions from what I currently understand like an enum value.

reader/views.py
```diff
@@ -670,6 +681,56 @@ def _extract_version_title_param(request, key):
...
+
+    if match.func in {serve_static, serve_static_by_lang}:
+        # Static pages are shared between modules and should use the simple
+        # Sefaria fallback image, not Library or Voices module branding.
+        return SocialImagePageType.STATIC
+
+    if module == LIBRARY_MODULE and match.func == catchall:
+        # Only the Library module should generate text-ref images. A ref-like
+        # string on Voices should stay module-branded because Voices pages are
+        # not currently supported by the text image generator.
```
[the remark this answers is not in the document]
relyks: I will potentially address this in the next PR related to the social image generation. Not sure if this is actually needed

relyks:
@yitzhakc I made your requested changes. I also made some other ones:
* I changed `hosts.py` to make it more consistent and to make the key used `voices`
* I fixed a few tests related to the middleware
* I added the `raqm` library as a dependency for production in the Dockerfile. The fallback `python-bidi` still doesn't do bidirectional text layout properly with `pillow` and I changed the configuration process to get it working. Adding `raqm` as a dependency will also help with nikkud if there's ever a desire for text in the images to have that. Example still not working in production: https://www.sefaria.org/api/img-gen/Exodus.3.15?ven=english|The_Contemporary_Torah,_Jewish_Publication_Society,_2006&lang=bi. Once this PR is merged, I'm pretty sure I need to create a new release tag, and then someone or I need to update the tag reference in the infrastructure repo

reader/views.py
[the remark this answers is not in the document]
yitzhakc: So I'll leave it unresolved so you can track it

yitzhakc (approved):

PR #3332, opened 2026-05-25
Social image generation for TOCs (category and individual texts)

relyks:
## Description
- Adds social image support for Library table-of-contents pages.
- Book-level pages like `/Genesis` and category pages like `/texts/Tanakh` can now get generated social preview images instead of falling back to a generic image.
- Keeps the earlier behavior for text refs, static pages, and unsupported module pages.

## Code changes
- Added a new TOC social image type for pages that represent a book or Library category.
- Modeled the URL-classification result as a **tagged union** `SocialImageTarget = PassageTarget | TocTarget | StaticTarget | ModuleFallbackTarget`, so each case carries only its valid fields:
    - `TocTarget` — English and Hebrew title, English and Hebrew subtitle, category color, category path
    - `PassageTarget` — the parsed `Ref` for a text quote
    - `StaticTarget` / `ModuleFallbackTarget` — no payload
  The old `page_type` enum is gone; the class itself is the tag, so illegal field combinations are unconstructable.
- Dispatch in `/api/img-gen/` uses `match`/`case` over the union (with `assert_never` for exhaustiveness); a `resolve_passage()` helper isolates the lang+version text fetch for the quote case.
- Updated `/api/img-gen/` routing logic so it can tell the difference between:
    - regular text refs like `/Genesis.1.1`
    - book pages like `/Genesis`
    - category TOC pages like `/texts/Talmud`
    - nested category pages like `/texts/Tanakh/Targum`
    - static pages
    - module fallback pages
- Resolve the module's URLConf via django-hosts `get_host(module).urlconf` instead of a hardcoded module→URLConf map, keeping that mapping in one place (`sefaria/hosts.py`).
- Added `generate_toc_image()` and `make_toc_img_http_response()` for TOC-style images.
- Refactored shared image header/logo drawing into a helper so text-ref images and TOC images use the same header behavior.
- Added logic to choose the best category color from a nested category path.
- Passage/quote image failures (bad ref or version, render errors) are caught and logged, returning the module-fallback image so the crawler-facing endpoint never 500s or emits a blank image.
- Recovered fuller bidi handling in `prepare_text_for_drawing` / `get_text_direction` (`base_dir='L'/'R'` and an `ltr` direction hint) so English text with embedded Hebrew words renders with correct neutral-character placement.
- Restored per-function docstrings on `sefaria/image_generator.py` (kept existing inline c
- Moved Library and Voices logo assets into module-specific folders:
    - `static/img/library/`
    - `static/img/voices/`
- Increased module fallback logo size so fallback images have stronger branding.
- Added social image meta tags to the annual report page.

## Testing
- Added API tests for book-level refs becoming TOC images.
- Added API tests for English and Hebrew TOC titles/subtitles.
- Added API tests for category and nested category TOC paths.
- Added a test for URL-decoded category paths with spaces.
- Added image generator tests for TOC image rendering, category colors, headers, and logo behavior.
- Added tests for module fallback logo sizing and static fallback logo sizing.
- Added unit tests for the `make_social_image_response` dispatcher: each `match` arm returns the expected response, and a passage-resolution failure falls back to the module image.
- Added bidi tests locking in the `ltr` behavior for English-with-embedded-Hebrew text.
- Updated existing classification tests from `.page_type ==` assertions to variant `isinstance` checks.
- Updated existing logo path tests to match the new module-specific logo asset directories.


reader/views.py
```diff
@@ -746,42 +833,80 @@ def _classify_social_image_path(tref: str, module: str) -> SocialImagePageType:
...
+            return make_static_img_http_response(platform)
+        case ModuleFallbackTarget():
+            return make_module_fallback_img_http_response(lang, platform, module)
+        case TocTarget() as t:
+            title = t.title_he if lang == "he" else t.title_en
+            subtitle = t.subtitle_he if lang == "he" else t.subtitle_en
+            desc = t.desc_he if lang == "he" else t.desc_en
+            return make_toc_img_http_response(title, subtitle, t.category, lang,
+                                              platform, module, category_path=t.category_path,
+                                              desc=desc, is_primary_category=t.is_primary_category)
```
relyks, line 897: shouldn't this be different? len is checked before

PR #3337, opened 2026-05-26
Arch Doc for Search Improvements

saengel:
## Description
The deliverable for the POC phase of search improvements - an arch doc. 

One change is that I created an `arch/` subdirectory to store these documents moving forward. Please let me know if something else would be better.


saengel:
Creating a new branch for this PR to avoid cauldron issues

PR #3338, opened 2026-05-26
chore(sc-44454): startup timing patch — parallel branch to drive cauldron image build

yodem:
Parallel branch carrying the timing+ensure_indices observability patch on top of feature/sc-44454/community-book-upload-voices. The main feature PR (#3325) is in CONFLICTING state so its CI doesn't run; this draft PR exists solely to trigger image builds against the sefaria-web-sc-44454 registry that the community-books cauldron's imagepolicy tracks. Not for merge.

Cherry-picked from chore/startup-timing-logs (PR #3336).


yodem:
Superseded by PR #3339 — the feature branch this was based on is in CONFLICTING state, which suppresses CI. PR #3339 is a clean branch off master with the same patch and the same sc-44454 path segment, so it builds images to the same registry the community-books cauldron tracks.

PR #3344, opened 2026-05-26
Arch docs

saengel:
## Description
- A new sub-directory `docs/arch_docs/` for storing sdd arch docs
- Moving other arch docs into this directory
- Moving the agent docs into docs - `docs/agent_docs/`
- The arch doc for search improvements, the deliverable after the POC phase - `/docs/arch_docs/search_improvements_arch.md`

(see: sc-44124 and sc-45372)


docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+            ▼                               ▼
+       Sources tab             Topics / Authors / Books tabs
+```
+
+1. **Source search** — `Sefaria.search` function, uses our elastic search endpoint for texts aside from Tanakh. For Tanakh, it uses the Dicta search for fuzzy matching.
+2. **Entity search** — A new endpoint using Elastic Search, `/api/entity-search?q=<query>&type=<topic|author|book>`, is fired once per tab. This endpoint queries the dedicated `topic` and `book` Elasticsearch indices built for the POC; see [Backend Architecture](#backend-architecture) below.
+* Note - this API was called search-poc in the POC. It will be renamed for clarity to `entity-search`. 
+
+> **Note on history.** The POC originally fed the entity tabs from the autocomplete **Name API** (`/api/name/<query>?limit=50`), reusing the dropdown's existing result handling - this was not the right decision, since while the Name API excels at returning results sorted by type, **it is not searching!** The Name API is _**autocompleting**_, which is an entirely different functionality. (Additionally, the Name API `ref` type conflated books and section-level refs)
```
stevekaplan123, line 40: Is it necessary to include this bit of history if the goal is to generate code based on this arch doc?  It seems like it could be confusing to include this.

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+- **Prefix match (`phrase_prefix`, titles only)** — handles mid-typing. Elasticsearch treats each word as an indivisible token, so "Mos" doesn't match "Moses" in an exact search — it's not a recognized token. `phrase_prefix` solves this by treating the last word in the query as a prefix, so "Mos" matches "Moses", "Moshe", etc. Applied to title fields only (not descriptions) to avoid noise.
+- **Popularity boost (`function_score` on `numSources`)** — without this, "Mos" prefix-matches Moses, Mosquitoes, and Moser with nearly identical text scores. `numSources` breaks the tie using source count: Moses has 7,074 references, Mosquitoes has 3. The score is log-scaled so the multiplier stays reasonable (≈8.9× vs ≈1.4×), and is applied as a multiplier — not an additive offset — so it consistently separates results regardless of their base text score.
+
+**Routing:** `topic` and `author` queries both search the `topic` index, filtered by `subtype`. Book queries additionally boost `author_names` so a search for "Rambam" surfaces his works even when his name isn't in the book title.
+
+The prefix match is the dominant query cost and the main performance risk to validate at scale.
```
stevekaplan123: Should this be here or in limitations?

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Author-aware book results
+
+When the query resolves to an author, the endpoint returns that author's works aggregated by category rather than a flat list. The dozens of Mishneh Torah volumes, for example, collapse into a single "Mishneh Torah" entry. This reuses the model's existing `AuthorTopic` author-works aggregation. Category aggregations sort to the top; individual books below. When the query does not resolve to an author, the endpoint falls back to a flat full-text search over the `book` index.
+
+> **Bug fixed.** An earlier version triggered the author-works view whenever an author's name appeared anywhere in matched text — including book descriptions. This caused queries like "Genesis" to return all of Rashi's books because his name appeared in a description. Fixed: the aggregated-works view now only activates when the query directly matches an author entity in the `topic` index.
```
stevekaplan123: Again, since this arch doc is going to be separate from the code, it seems strange to word this as a history of the project.  Instead of "bug fixed", we could just say "Note: A query like "Genesis" should not return all of Rashi's books just because his name appeared in a description.  Instead, the aggregated-works view should only activate when the query directly matches an author entity in the `topic` index."

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+
+**Routing:** `topic` and `author` queries both search the `topic` index, filtered by `subtype`. Book queries additionally boost `author_names` so a search for "Rambam" surfaces his works even when his name isn't in the book title.
+
+The prefix match is the dominant query cost and the main performance risk to validate at scale.
+
+#### Author-aware book results
+
+When the query resolves to an author, the endpoint returns that author's works aggregated by category rather than a flat list. The dozens of Mishneh Torah volumes, for example, collapse into a single "Mishneh Torah" entry. This reuses the model's existing `AuthorTopic` author-works aggregation. Category aggregations sort to the top; individual books below. When the query does not resolve to an author, the endpoint falls back to a flat full-text search over the `book` index.
```
stevekaplan123: I may be misunderstanding but it's confusing that we say "the model's existing `AuthorTopic`..."  It's an existing function Sefaria has for AuthorTopic pages.

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+
+> **Bug fixed.** An earlier version triggered the author-works view whenever an author's name appeared anywhere in matched text — including book descriptions. This caused queries like "Genesis" to return all of Rashi's books because his name appeared in a description. Fixed: the aggregated-works view now only activates when the query directly matches an author entity in the `topic` index.
+
+To support useful labels in the aggregated view, the author-works aggregation was extended to report, per entry, whether it is a category aggregation and a localized category label. The Name API (`/api/name`) gained an opt-in `get_author_books` flag that returns the same aggregated structure, so the autocomplete path and the POC endpoint share this data.
```
stevekaplan123: I don't understand what this comment is talking about, but it may be from previous code since it's referencing the Name API.

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+      "titleVariants": ["Rambam", "Moses Maimonides", "Moses ben Maimon"],
+      "description_en": "Rabbi Moshe ben Maimon (1138–1204), prolific halakhic authority and philosopher.",
+      "numSources": 7074,
+      "era": "RI",
+      "birthYear": 1138,
+      "deathYear": 1204
+    }
+  ],
+  "total": 1
+}
```
stevekaplan123, line 175: This is great

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+## Elastic Search Indexing Operations
+
+### Scheduled reindex
+
+The scheduled cron job (`scripts/scheduled/reindex_elasticsearch_cronjob.py`) rebuilds the `topic` and `book` indices on the same schedule as `text`/`sheet`. Each index is rebuilt using a blue-green strategy: a fresh index is built in the background under a temporary name, and only swapped in as the live index once it's complete. This means search stays available on the old data during the rebuild, with zero downtime. A failure rebuilding one index type is recorded but does not block the other from completing.
+
+We need both sample scripts to minimally populate a dev environment, as well as a script for a **full reindex** which runs the reindex for `topic` and/or `book` on demand.
```
stevekaplan123: This arch doc shouldn't reference the sample scripts.  It could reference how the full reindex works though.

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Local development setup
+
+Running any indexing script from a fresh local checkout crashes with `ImportError: cannot import name 'SEARCH_INDEX_NAME_TOPIC' from sefaria.settings` unless those constants exist in `sefaria/local_settings.py`. The production `settings.py` defines them; the local dev override does not by default. Add to `local_settings.py` to unblock local indexing:
```
stevekaplan123: This seems a pretty indirect way of saying that we need to have these constants defined in local settings files.

docs/arch_docs/search_improvements_arch.md
```diff
@@ -0,0 +1,264 @@
...
+
+```python
+SEARCH_INDEX_NAME_TOPIC = 'topic'
+SEARCH_INDEX_NAME_BOOK = 'book'
+```
+
+## Limitations
```
stevekaplan123, line 255: I personally wonder if this entire limitations / questions / enrichment section should be its own markdown file since don't we want to be able to re-build the code from this document?

stevekaplan123 (changes requested):
Left comments.  Mostly this is great and it will be exciting to see Claude bring the code back to life from this arch doc.

docs/arch_docs/search_improvements_arch.md
saengel: Hmm, maybe useful so we don't repeat the same mistake? The difference between the Name API and Search is not super clear at a brief glance.

docs/arch_docs/search_improvements_arch.md
saengel: I'll move to limitations, good catch.

docs/arch_docs/search_improvements_arch.md
saengel: I think it just means that it passed along the aggregation to the Name API as well, i.e. extending this code there. I can delete since it is irrelevant to this work.

docs/arch_docs/search_improvements_arch.md
saengel: I think keeping the limitations close is important, helps the document provide a broad perspective on everything going on.

docs/arch_docs/search_improvements_arch.md
saengel: Fixed

sefaria-weblate (approved):

stevekaplan123 (approved):

PR #3350, opened 2026-05-27
fix: bootstrapUrl crash on non-ref URLs (topics, profiles, etc.)

yodem:
## Problem

Clicking a topic link in the chatbot's "Discover sources" appetizer has been **silently broken on sefaria.org** since the appetizer shipped.

### Root cause

`handleAppetizerClick` in the chatbot already dispatches `sefaria:bootstrap-url` with a topic path (e.g. `/topics/shabbat`) when on sefaria.org. `handleBootstrapUrlEvent` receives it and calls `bootstrapUrl`. Inside `bootstrapUrl`:

```js
const {path, ref} = this._getPathAndRefFromUrl(href);  // returns {path, ref} — url never returned
if (Sefaria.isRef(ref)) { ... return true; }           // false for topics — falls through
return this.openURL(path + url.search, true);           // 💥 ReferenceError: url is not defined
```

`url` is a local variable inside `_getPathAndRefFromUrl` — it was **never returned and never in scope** in `bootstrapUrl`. Since `Sefaria.isRef` returns `true` for text refs (early return), the crash only happens for non-ref routes: topics, profiles, calendars, etc. The bug was invisible for refs, silently swallowed for everything else.

There was also no null guard on the `false` return from `_getPathAndRefFromUrl` (invalid / cross-origin URLs), which would cause a destructuring crash on those paths.

## Fix

1. Return `url` (the `URL` object) from `_getPathAndRefFromUrl` alongside `path` and `ref`
2. Add `if (!result) { return false; }` null guard before destructuring in `bootstrapUrl`

`bootstrapUrl` can now correctly compute `path + url.search` and pass it to `openURL`.

## What this enables

Topic pages (and other non-ref routes) now open in-panel when the chatbot dispatches `sefaria:bootstrap-url` — the same in-page navigation that already works for text refs. Previously, the event was received, `bootstrapUrl` was called, and then it crashed before reaching `openURL`.

## Testing

- Dispatched `sefaria:bootstrap-url` with `{ url: '/topics/shabbat' }` locally → Shabbat topic page opens in-panel ✓
- Dispatched with `{ url: '/Genesis.1' }` → Genesis 1 still opens correctly (regression) ✓
- Dispatched with `{ url: '/topics/shabbat?tab=sources' }` → query param preserved via `url.search` ✓

### DIFF
--- a/static/js/ReaderApp.jsx
+++ b/static/js/ReaderApp.jsx
@@ -1199,7 +1199,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     }
     const path = decodeURI(url.pathname);
     const ref = path.slice(1).replace(/%3F/g, '?');
-    return {path, ref, url};
+    return {path, ref, search: url.search};
   }
   bootstrapUrl(href, options) {
     if (this.shouldAlertBeforeCloseEditor()) {
@@ -1210,13 +1210,13 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     const opts = options || {};
     const result = this._getPathAndRefFromUrl(href);
     if (!result) { return false; }
-    const {path, ref, url} = result;
+    const {path, ref, search} = result;
     if (Sefaria.isRef(ref)) {
       // Route bot refs through the same path as Header search ref navigation.
       this.handleNavigationClick(Sefaria.humanRef(ref), null, {replaceHistory: opts.replaceHistory});
       return true;
     }
-    return this.openURL(path + url.search, true);
+    return this.openURL(path + search, true);
   }
 
   updateModuleLinkHref(link) {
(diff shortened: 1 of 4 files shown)


static/js/ReaderApp.jsx
```diff
@@ -1199,7 +1199,7 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
     }
     const path = decodeURI(url.pathname);
     const ref = path.slice(1).replace(/%3F/g, '?');
-    return {path, ref};
+    return {path, ref, url};
```
YishaiGlasner: it isn't reasonable to return the url object. it's a private variable for parsing its patrs. also you son't need it outside. you just need the search attribute, so return it rather than the url object. that will make mush sense.

YishaiGlasner (changes requested):

YishaiGlasner (approved):

PR #3368, opened 2026-06-02
fix(search): restore default 'path' field for text filters (sc-44603)

yodem:
## Problem

[sc-44603](https://app.shortcut.com/sefaria/story/44603) — Selecting a path filter on a **text** search returns no results. Repro: search \`מעל\`, filter to תהילים (Psalms). The facet shows results in Psalms, but applying the filter shows nothing and \`/api/search-wrapper/es8\` returns **HTTP 500**.

Problematic URL: \`https://www.sefaria.org/search?q=%D7%9E%D7%A2%D7%9C&tab=text&tpathFilters=Tanakh%2FWritings%2FPsalms&tvar=1&tsort=relevance\`

### Backend: wrong filter type resolved for text path filters

Two separate flows land in \`make_filter\` with \`agg_type=None\`:

1. **URL-bootstrapped filters** — \`reader/views.py:get_search_params\` parses text filters from query-string params and sends \`agg_types = [None for _ in filters]\`. This was previously fine because \`make_filter\` treated all text filters as path filters by default.

2. **Empty \`filter_fields\` default** — \`get_filter_obj\` fills an empty \`filter_fields\` list with \`[None] * len(filters)\` as a documented "use default" sentinel (line 129).

Commit \`125265597\` ("keyword-search-in-connected-refs") tightened the text branch to only match \`agg_type == "path"\` (so that a new \`linked_refs\` agg type could use a \`Term\` filter), which accidentally excluded the \`None\` case:

\`\`\`diff
-    if type == "text":
+    if type == "text" and agg_type == "path":
         return Regexp(path=reg)
-    elif type == "sheet":
+    else:
         return Term(**{agg_type: agg_key})
\`\`\`

\`agg_type=None\` then fell through to \`Term(**{None: agg_key})\` → \`TypeError: keywords must be strings\` → 500.

### Frontend: applied filter shows count (0)

When \`buildFilterTree\` seeded applied filters with \`docCount: 0\` and then merged them with the aggregation buckets, the placeholder for e.g. "Psalms" inserted first would shadow the real bucket entry (which had the correct count). Result: a selected filter always rendered as "(0)" while siblings showed real counts.

### Backend — two changes

**1. Fix the root cause: \`reader/views.py\`** — now passes \`"path"\` explicitly for text filters from URLs instead of \`None\`, making the intent clear:

\`\`\`python
agg_types = ["path" for _ in filters]  # text search always filters on the "path" field
\`\`\`

**2. Keep the \`None\` guard in \`make_filter\` as a documented defensive fallback** (covers the \`filter_fields=[]\` → \`[None]\` case from \`get_filter_obj\` line 129 and any future caller):

\`\`\`python
if type == "text" and agg_type in (None, "path"):
\`\`\`

This preserves the \`Term\` path for \`linked_refs\` and all sheet filters.

### Frontend — pre-index bucket counts before building the tree

\`\`\`js
const bucketDocCounts = {};
aggregation_buckets.forEach(f => { bucketDocCounts[f["key"]] = f["doc_count"]; });
// seed applied filters with their real count from the aggregation
docCount: bucketDocCounts[fkey] || 0,
\`\`\`

Applied filters are always included in the tree so they appear even when the aggregation has no bucket for them. Now they're seeded with the real aggregation count (falling back to 0 when genuinely absent), so a selected filter shows its true count rather than always "(0)".

## Tests

- Added \`test_text_filter_default_agg_type\` (\`sefaria/helper/tests/search_test.py\`): a text filter with \`filter_fields=[None]\` and \`[]\` must produce the same \`path\` Regexp as \`["path"]\` and must not raise.
- Added e2e regression test asserting an applied path filter shows a non-zero count.


YishaiGlasner (commented):
I didn't understand when and why we have agg_type None (and if this fix is a patch or the right way to do it - it seems like a patch).
Also, there is no explanation to the client side change (bucketDocCounts) in the description.

nsantacruz (approved):

PR #3369, opened 2026-06-02
chore(helm): make local-settings-secrets the only source for SLACK_URL; retire slack-webhook secret [sc-44609]

yodem:
## Story
[sc-44609](https://app.shortcut.com/sefaria/story/44609)

## Problem
`SLACK_URL` was injected into pods from **two** sources:
1. An explicit `env:` entry from the dedicated `slack-webhook` secret (`sefaria.secrets.slackWebhook` helper).
2. The `SLACK_URL` key inside `local-settings-secrets` (loaded via `envFrom`).

In Kubernetes an explicit `env` entry **always wins** over an `envFrom` key of the same name, so `SLACK_URL` overrides set in an environment's / cauldron's `local-settings-secrets` were silently ignored — pods kept serving a stale webhook (`invalid_token`).

## Change — `local-settings-secrets` is the only source, `slack-webhook` retired
- **Removed the redundant explicit `SLACK_URL` env** from pods that already mount `local-settings-secrets` via `envFrom`: `rollout/web.yaml`, `rollout/task.yaml`, `cronjob/reindex-elasticsearch.yaml`.
- **`cronjob/postgres-packup.yaml`**: the `pgdump-uploader` container has no `envFrom` and its `upload-dumps.sh` posts to `${SLACK_URL}` under `set -e`, so it keeps an explicit `SLACK_URL` — now sourced from `{{ .Values.secrets.localSettings.ref }}` (key `SLACK_URL`) instead of `slack-webhook`.
- **`cronjob/mongo-backup.yaml`**: its `upload-dumps.sh` does not use `SLACK_URL`; the old env there was dead and is simply removed.
- **`analysistemplate/rollout-complete.yaml`** (blue-green deploy notifier): `SLACK_URL` now comes from `{{ .Values.secrets.localSettings.ref }}` (`optional`, and the curl is `|| /bin/true`). **No channel override is sent** — the Slack incoming webhook posts to its own configured channel — so no `SLACK_CHANNEL` is needed.
- **Deleted** the `slack-webhook` secret template, the `sefaria.secrets.slackWebhook` helper, and the `secrets.slackWebhook` values stanza. No `slack-webhook` references remain in the chart.

## Companion infra PR
**Sefaria/infrastructure#630** removes the now-orphaned `slack-webhook-helm` (dev) and `slack-webhook-production` (prod) secrets. `SLACK_URL` already exists in `local-settings-secrets`, so no secret edits are required.

## ⚠️ Deploy ordering
The running prod chart injects `SLACK_URL` via a **non-optional** `secretKeyRef` to `slack-webhook-production`. **Sequence:** merge + release this PR → deploy chart to **production** → only then merge/reconcile infra #630. Otherwise web/task pods crashloop.

## Verification
- `helm lint` → 0 failed.
- `helm template` → `SLACK_URL` resolves from `local-settings-secrets-*` on web/task/reindex (envFrom), the pg-backup uploader, and the analysistemplate. No `slack-webhook` refs remain.


yodem:
Thanks @copilot — two of the three were real and are fixed in e7a6d79:

**✅ `postgres-packup.yaml` (valid, high severity).** Correct: the `pgdump-uploader` container runs `upload-dumps.sh` → `curl ... ${SLACK_URL}` under `set -e`, and it has no `envFrom` (only the `postgres-dumper` initContainer does). Dropping the explicit env left `SLACK_URL` unset, which would fail the backup job after the dump succeeded. Restored `SLACK_URL` via `secretKeyRef` from the configured `localSettings` ref.

**✅ `rollout-complete.yaml` (valid).** Switched both `SLACK_URL` and `SLACK_CHANNEL` from the hardcoded `local-settings-secrets-{{ deployEnv }}` to `{{ .Values.secrets.localSettings.ref }}` — which prod/preprod set to `local-settings-secrets-production`, the secret that actually holds `SLACK_URL` (and where the companion infra PR adds `SLACK_CHANNEL`).

**◑ `mongo-backup.yaml` (comment was wrong, no functional break).** You're right the comment was inaccurate (that container has no `envFrom`). But `upload-dumps.sh` for **mongo** never references `SLACK_URL` — the explicit env that used to be there was dead. So I kept it removed and corrected the comment rather than adding an unused env.

Verified: `helm lint` clean; `helm template` shows `SLACK_URL` resolving from `local-settings-secrets-production` on the pg uploader, web/task/reindex (via `envFrom`), and the analysistemplate. No `slack-webhook` references remain.

BrendanGalloway (approved):

PR #3045, opened 2026-01-25
Produce an LLMs.txt for Sefaria.org to enhance discoverability

saengel:
# Summary
Add LLMS.txt file to provide structured guidance for AI systems accessing Sefaria's content
Optimized for cross-engine compatibility (OpenAI/GPT, Anthropic/Claude, Google/Gemini)
Includes API quick reference, reference format guide, best practices, and data access options

LLMS.txt is an emerging convention (similar to robots.txt for search engines) that tells AI systems how to interact with a domain. Major AI providers are beginning to recognize and respect these files. [See the docs here](https://llmstxt.org/).

OpenAI, Anthropic, and Google are all actively developing systems that read site-level instruction files
Early adopters of LLMS.txt will have their guidance incorporated as these systems mature
The cost is minimal (one static file); the upside is significant

This file:

- Positions Sefaria as the canonical source for Jewish textual content
- Provides actionable API instructions so models fetch verified text instead of hallucinating
- Establishes guardrails: directs users to rabbis for halakhic questions, presents texts within Jewish interpretive tradition
- Surfaces our infrastructure: Sefaria-Export for training, REST API for live queries, MCP server for Claude

## Strategic Value
**Discoverability**: AI developers building Jewish-focused applications will find our API documentation
**Accuracy**: Every AI response that cites Sefaria instead of generating from training data is a win for textual integrity
**Mission alignment**: "Making Jewish texts accessible" now includes making them accessible through the AI interfaces people increasingly use
**Attribution**: The file explicitly requests citation ("via Sefaria.org"), driving awareness back to us

PR #3370, opened 2026-06-02
feat(sc-44545): make registration discoverable on login page

yodem:
## Story
[SC-44545 — Update login page to handle anon LA candidates who don't have an account](https://app.shortcut.com/sefaria/story/44545) (epic: Library Assistant System and UX)

As part of simplifying the LA promo (logged-out users get a single "Log in to Try" entry point), the **login page** must make the path to registration hard to miss.

## Changes
- **New partial** `templates/registration/_cross_flow_nav.html` — reusable bilingual "prompt + inline link" line (EN/HE), preserves `?next=`.
- **`login.html`** — cross-flow link moved directly under the "Log in to Sefaria" title (was an easy-to-miss link at the bottom); old "Create a new account" link removed. Uses Django 6 `{% url … as %}` + `{% include … with %}`.
- **`static/css/static.css`** — scoped `.registration-prompt` styling from the Figma design (node `35:430`): prompt `--darker-grey`, link `--sefaria-blue` SemiBold underlined, 4px RTL-safe gap. General CSS vars, no hardcoded hex.
- **`auth-flow.spec.ts`** — HAM-A003 updated to select the new "Sign Up" link (scoped to `#login`).

Copy (source of truth = story): EN `Don't have an account? Sign Up` · HE `אין לכם חשבון? להרשמה`.

## Design docs
- Spec: `docs/superpowers/specs/2026-06-02-login-cross-flow-nav-design.md`
- Plan: `docs/superpowers/plans/2026-06-02-login-cross-flow-nav.md`

Figma (source of truth for layout/CSS): https://www.figma.com/design/2WflG98PDhWQ7OKrDkaCPb/Login---Registration-Wireframes?node-id=35-420

## Scope
Login page only. SC-44544 (backend auto-enroll into the experiment on auth) is a separate story; this design leaves the partial as a clean seam for it.

## Test plan
Preview cauldron (feature mode): **https://sc-44545-login.cauldron.sefaria.org**
- [ ] EN: prompt under title, "Sign Up" blue/SemiBold/underlined, old bottom link gone
- [ ] HE (RTL): "אין לכם חשבון? להרשמה" renders correctly
- [ ] `/login?next=/texts` → link href is `/register?next=/texts`
- [ ] Mobile width
- [ ] `register.html` links visually unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3371, opened 2026-06-02
fix: restore re_path in sites/sefaria/urls.py (url() breaks Django 6)

yodem:
## Problem
`sites/sefaria/urls.py` calls `url(...)` for the `metrics → healthz-rollout` route block (lines 66–78), but **`django.conf.urls.url` was removed in Django 4.0** and the file only imports `re_path`. On Django 6.0.4 (this repo's version) the URLConf module raises **`NameError: name 'url' is not defined`** at import → Django fails to load URLs → **every request returns HTTP 500**, including the `/healthz-rollout` startup probe.

### Impact
Because the rollout health probe 500s, web pods never pass startup and CrashLoop — **all new deploys and cauldrons built from `master` fail to come up** (observed on a fresh cauldron: web pod `0/1`, nginx serving 503 with `NameError: name 'url' is not defined`).

### Cause
Introduced by #3045 (LLMs.txt), which converted that route block from `re_path()` to `url()`.

## Fix
Revert the 13 affected routes from `url(` back to `re_path(` (anchored to line-leading calls; commented examples and `STATIC_URL` untouched). `python -m py_compile sites/sefaria/urls.py` passes.

## Test plan
- [x] `py_compile` clean
- [ ] URLConf loads without NameError (CI / app boot)
- [ ] `/healthz-rollout` returns 200 once `library.is_initialized()`
- [ ] New cauldron/deploy web pod reaches Ready

### DIFF
--- a/sites/sefaria/urls.py
+++ b/sites/sefaria/urls.py
@@ -63,19 +63,19 @@ static_pages_by_lang = [
 
 # Static and Semi Static Content
 site_urlpatterns = [
-    url(r'^metrics/?$', reader_views.metrics),
-    url(r'^digitized-by-sefaria/?$', reader_views.digitized_by_sefaria),
-    url(r'^(favicon\.ico|apple-touch-icon\.png|favicon\.svg)/?$', reader_views.module_favicon),
-    url(r'^(?P<filename>site\.webmanifest|manifest\.json)/?$', reader_views.dynamic_manifest),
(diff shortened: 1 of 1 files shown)


yodem:
Closing as duplicate — another PR already covers the url()→re_path() Django 6 fix in sites/sefaria/urls.py. Folding the same fix directly into the cauldron PR #3370 instead.

PR #3376, opened 2026-06-03
Feature/sc 44531/update la promo and dismissal experience

stevekaplan123:
## Description

## Code Changes

## Notes


static/js/sefaria/strings.js
```diff
@@ -37,12 +37,14 @@ const Strings = {
...
-    "Discover & explore texts in the Sefaria Library with our new AI-powered assistant.": "התחילו ללמוד ולגלות מקורות בספריית ספריא בעזרת העוזר החדש שלנו המופעל על ידי בינה מלאכותית",
+    "Enhance Your Learning Experience": "העמיקו את הלמידה שלכם",
     "Log in to Join": "התחברות כדי להצטרף", 
+    "Maybe later": "אולי מאוחר יותר",
+
     "Create an Account": "יצירת חשבון",
-    "Join the Experiment": "הצטרפו לניסוי",
-    "Joining...": "מצרפים אותך...",
+    "Try It": "נסו עכשיו",
+    "Log in to Try": "התחברות כדי להתנסות",
```
yodem, line 46: there are some inconsistencies with the capital letters at the beginning

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  const storagePrefix = `promo_backoff_${cookieName}`;
+  return {
+    state: `${storagePrefix}_state`,
+    sessionCounter: `${storagePrefix}_session_counter`,
+    lastSessionAtSec: `${storagePrefix}_last_session_at_sec`,
+  };
+};
+
+const getPromoSessionLengthSeconds = (promoSessionLengthSeconds) => {
+  const configuredSessionLengthSeconds = Number(promoSessionLengthSeconds);
```
yodem, line 24: Can it be something that isnt number? because the parsing might fail. can you try pass a number before the function? like parse it before the function call

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  return Number.isFinite(configuredSessionLengthSeconds) && configuredSessionLengthSeconds > 0
+    ? configuredSessionLengthSeconds
+    : DEFAULT_PROMO_SESSION_LENGTH_SECONDS;
+};
+
+const updatePromoSessionCounter = ({ storageKeys, sessionLengthSeconds }) => {
+  const nowSec = Math.floor(Date.now() / 1000);
+  const lastSessionAtSec = Number(localStorage.getItem(storageKeys.lastSessionAtSec));
```
yodem, line 32: do you want to have a default || 0 for this const as well? otherwise it might be undefiend and break other things?

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  const currentSessionCounter = Number(localStorage.getItem(storageKeys.sessionCounter)) || 0;
+  const isNewSession = !Number.isFinite(lastSessionAtSec) || (nowSec - lastSessionAtSec) >= sessionLengthSeconds;
+
+  if (!isNewSession) {
+    return currentSessionCounter;
+  }
+  const nextSessionCounter = currentSessionCounter + 1;
+  localStorage.setItem(storageKeys.sessionCounter, String(nextSessionCounter));
+  localStorage.setItem(storageKeys.lastSessionAtSec, String(nowSec));
```
yodem: are you sure you need to parse it to string?

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  if (localStorage.getItem(storageKeys.state)) {
+    return;
+  }
+  // A legacy dismissal is recorded only as `${cookieName}=1` with no timestamp, so we
+  // can't recover when it was set. But any existing legacy cookie necessarily predates
+  // this deploy, so we treat it as a single "Maybe later" from long ago: backdate the
+  // time gate (lastMaybeLaterAtSec: 0) so only the session gate remains before re-showing.
+  if (!document.cookie.includes(cookieName)) {
+    return;
+  }
```
yodem, line 58: can you do "||"(or) operator between them instead of 2 different if statements?

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  const migratedState = {
+    maybeLaterCount: 1,
+    lastMaybeLaterAtSec: 0,
+    sessionCountAtLastMaybeLater: 0,
```
yodem: im not sure that maybe is a good work in here, do you agree?

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  // User opted out for good (explicitly, or by hitting the click cap) — hide permanently.
+  if (state.dismissedForever || state.maybeLaterCount >= MAX_MAYBE_LATER_CLICKS) {
+    return true;
+  }
+
+  // No nudge rule for this dismissal count means there's nothing left to wait on — show it.
+  const nudgeRule = nudgeSchedule[state.maybeLaterCount];
+  if (!nudgeRule) {
+    return false;
+  }
```
yodem, line 83: is there a way to combine them?

static/js/SiteWideBanner.jsx
```diff
@@ -86,18 +229,18 @@ SiteWideBanner.propTypes = {
...
 
-const CHATBOT_BANNER_MAIN_TEXT = Sefaria._("Try Sefaria's new Library Assistant [Experimental]");
-const CHATBOT_BANNER_SECONDARY_TEXT = Sefaria._("Discover & explore texts in the Sefaria Library with our new AI-powered assistant.");
-const CHATBOT_BANNER_LEARN_MORE_URLS = {
-  en: "https://help.sefaria.org/hc/en-us/articles/26006423836828",
-  he: "https://help.sefaria.org/hc/he/articles/26006423836828-%D7%9B%D7%99%D7%A6%D7%93-%D7%9C%D7%94%D7%A9%D7%AA%D7%9E%D7%A9-%D7%91%D7%A2%D7%95%D7%96%D7%A8-%D7%94%D7%A1%D7%A4%D7%A8%D7%99%D7%99%D7%94-%D7%A9%D7%9C-%D7%A1%D7%A4%D7%A8%D7%99%D7%90",
-};
+const CHATBOT_BANNER_MAIN_TEXT = Sefaria._("Enhance Your Learning Experience");
+const CHATBOT_BANNER_SECONDARY_TEXT_HE = <div>נסו את <a href="https://help.sefaria.org/hc/he/articles/26006423836828-How-to-Use-the-Sefaria-Library-Assistant">עוזר הספרייה</a> שלנו, המופעל על ידי בינה מלאכותית, על מנת להעמיק את הבנתכם ולגלות מקורות חדשים.</div>;
+const CHATBOT_BANNER_SECONDARY_TEXT = <div>Try our AI-powered <a href="https://help.sefaria.org/hc/en-us/articles/26006423836828-How-to-Use-the-Sefaria-Library-Assistant">Library Assistant</a> to deepen your understanding and discover new texts.</div>;
```
yodem, line 240: dont we keep strings in other place and not hard code them?

yodem:
overall nice code, I believe the namings can be more indicative and also you shouldnt name a const anything that starts with "use" as it is a terminology of react hooks ("useState","useEffect") - it wont break anything but it is better. if you want to name a boolean its better to add "is" before the conts name

static/js/SiteWideBanner.jsx
stevekaplan123: Number() never throws errors — it will just return NaN or undefined — and the Number.isFinite(...) && > 0 guard catches that and falls back to `DEFAULT_PROMO_SESSION_LENGTH_SECONDS`

static/js/SiteWideBanner.jsx
stevekaplan123: Again, this is fine because Number doesn't error out -- it returns a number, NaN, or undefined.  The next line relies on !Number.isFinite(lastSessionAtSec) to treat NaN as a fresh session.

static/js/SiteWideBanner.jsx
stevekaplan123: You're right it's unnecessary

static/js/SiteWideBanner.jsx
stevekaplan123: These guard clauses are semantically distinct (already migrated vs. no legacy cookie to migrate) and each carries its own multi-line explanatory comment. Merging into one || would force both rationales into a single comment and lose that clarity.

static/js/SiteWideBanner.jsx
stevekaplan123: The reason is because it tracks clicks of a button called "Maybe later".  Do you see this as problematic?

static/js/SiteWideBanner.jsx
stevekaplan123: Right, but I did it this way because we have an a tag inside the string so it's not just a string.  `Sefaria._v` is a function that takes an object with 'en' and 'he' keys and chooses the right one based on the current interface language.

static/js/SiteWideBanner.jsx
stevekaplan123: I think the logic of this promo is fairly complicated and I liked this approach for readability's sake.  If we combine the two "return false" scenarios into one, we'll have to check if state is empty and then if not check nudgeSchedule[state.maybeLaterCount] and then later on define it as the variable nudgeRule again.  It seems awkward in either case.

static/js/SiteWideBanner.jsx
yodem: Lol its quite confusing, but I really dont have better suggestions

stevekaplan123: You're right.  I'm going to change these names

static/js/SiteWideBanner.jsx
```diff
@@ -3,6 +3,96 @@ import PropTypes from "prop-types";
...
+  if (!isNewSession) {
+    return currentSessionCounter;
+  }
+  const nextSessionCounter = currentSessionCounter + 1;
+  localStorage.setItem(storageKeys.sessionCounter, nextSessionCounter);
+  localStorage.setItem(storageKeys.lastSessionAtSec, nowSec);
+  return nextSessionCounter;
+};
+
+const migrateLegacyCookieToBackoffState = ({ cookieName, storageKeys }) => {
```
saengel, line 45: From copilot - 

Side effects during render (React anti-pattern):
- migrateLegacyCookieToBackoffState(...) and updatePromoSessionCounter(...) run in component body, not useEffect.
- This writes to storage during render and can run repeatedly/unpredictably on re-renders.
- Fix: move migration/session-counter update into useEffect (or lazy initializer + effect), then store counter in state/ref.

saengel (changes requested):
One note from copilot, otherwise looks good to me.

static/js/SiteWideBanner.jsx
stevekaplan123: Claude responds to CoPilot:
"Can run repeatedly/unpredictably on re-renders" — both functions are actually guarded against that:

Migration is idempotent: line 49 bails if state already exists.
The counter has a 30-min time gate (line 34) and persists lastSessionAtSec immediately, so a normal re-render (or a StrictMode double-invoke) sees isNewSession === false and returns the existing value. It won't increment on every re-render.
So it's not "unpredictable" in practice today — the guards make it mostly safe. The strongest real argument is the discarded-concurrent-render case, not runaway re-render increments.  And to be fully honest about severity: on plain, non-concurrent React this is close to a non-bug today. The guards make it well-behaved.   

I'll add since both Tzirel and I tested this, I think Claude is right that this is really a "non-bug"

saengel (approved):

PR #3391, opened 2026-06-04
Redirect about-sidebar pages from voices to sefaria.org

YishaiGlasner:
## Summary

Static "about" pages (About, Team, Terms, Privacy Policy, etc.) belong only on `sefaria.org` and have no context on the `voices` subdomain. Previously these pages had no cross-module behavior — they'd either render incorrectly on voices or be left out of voice-side navigation links entirely.

This PR:
- Introduces a single canonical list (`ABOUT_SIDEBAR_PAGES` in `site_settings.py`) as the source of truth for all 12 about-section pages, including their English and Hebrew display labels.
- Redirects direct navigation to these paths on voices (e.g. `voices.sefaria.org/about`) to `sefaria.org/about`.
- Opens these pages in a new tab when a user clicks a link to them from within the voices experience — handled centrally in `ReaderApp.openURL` so every link in the app is covered without per-component changes.
- Cleans up `SidebarFooter` (domain-prefix removal for about/terms/privacy-policy) and `_sidebar.html` (hardcoded `<li>` list replaced by a template loop driven by the canonical list).
- Consolidates `serve_static` and `serve_static_by_lang` into a single view; the `by_lang` distinction is now passed as a URL-pattern kwarg.

## Changes

| File | What changed |
|---|---|
| `sites/sefaria/site_settings.py` | Added `ABOUT_SIDEBAR_PAGES` — the single source of truth |
| `sites/sefaria/urls.py` | Replaced `serve_static_by_lang` reference with `serve_static, {"by_lang": True}` |
| `reader/views.py` | Merged `serve_static` / `serve_static_by_lang` into one view; added voices redirect |
| `templates/_sidebar.html` | Replaced 12 hardcoded `<li>` elements with a `{% for %}` loop over `SITE_SETTINGS.ABOUT_SIDEBAR_PAGES` |
| `templates/static/privacy-policy.html` | Fixed `whichPage='privacy'` → `'privacy-policy'` to match new path-based element IDs |
| `templates/static/link-to-annual-report.html` | Fixed `whichPage='annualreport'` → `'link-to-annual-report'` (same reason) |
| `static/js/ReaderApp.jsx` | `openURL` now opens about-sidebar paths in a new tab when on voices |
| `static/js/NavSidebar.jsx` | `getSidebarFooterData`: removed library-origin prefix from about/terms/privacy-policy |
| `reader/tests/test_catchall_redirects.py` | Added 3 tests for `serve_static` voices redirect behavior |

## Tests added

In `reader/tests/static_pages_test.py`:
- `test_voices_about_sidebar_page_redirects_to_library` — core: `/about` on voices → 301 to library
- `test_voices_about_sidebar_page_preserves_query_params` — redirect preserves query strings
- `test_library_about_sidebar_page_no_redirect` — regression: library doesn't redirect sidebar pages

Note: client-side `openURL` logic has no unit test — there is no existing pattern for testing `ReaderApp` methods in this codebase. Covered by manual verification below.

## Test plan

- [ ] On voices, navigate directly to `/about`, `/terms`, `/team` → should redirect (same tab) to `sefaria.org/<path>`
- [ ] On voices, click "About" or "Terms" in the sidebar footer → should open `sefaria.org/about` in a new tab
- [ ] On voices, click any other link that points to an about-sidebar path → opens in new tab
- [ ] On library, click "About" or "Terms" in the sidebar footer → navigates in the same tab
- [ ] On any about-sidebar page (about, team, terms, etc.), verify the sidebar nav renders correctly and highlights the active item


static/js/ReaderApp.jsx
```diff
@@ -1260,6 +1260,15 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
...
     const path = decodeURI(url.pathname);
+    if (Sefaria.activeModule === Sefaria.VOICES_MODULE) {
+      const sidebarPaths = new Set(
+        (Sefaria._siteSettings.ABOUT_SIDEBAR_PAGES || []).map(p => '/' + p.path)
+      );
+      if (sidebarPaths.has(path)) {
+        window.open(Sefaria.util.fullURL(path, Sefaria.LIBRARY_MODULE), '_blank');
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
YishaiGlasner: there are no search params for those pages. i don't think we should now write complicated code for case we'll change it.

reader/views.py
```diff
@@ -4660,19 +4660,15 @@ def search_path_filter(request, book_title):
...
 
 @ensure_csrf_cookie
-def serve_static_by_lang(request, page):
-    """
-    Serve a static page whose template matches the URL
-    """
-    return render_template(request,'static/{}/{}.html'.format(request.LANGUAGE_CODE, page), {"headerMode": True}, {"renderStatic": True})
+def serve_static(request, page, by_lang=False):
+    if request.active_module == VOICES_MODULE and page in _ABOUT_SIDEBAR_PATHS:
+        return redirect_to_module(request, f"/{page}", LIBRARY_MODULE)
```
yodem, line 4812: issue (claude): The voices→library redirect lives inside `serve_static`, which sits *below* URL routing — so it only fires for paths that fall through to the `serve_static` catch-all (`urls.py:73`). `/metrics` is in `ABOUT_SIDEBAR_PAGES` but is routed to `reader_views.metrics` at `urls.py:66`, which precedes the catch-all and has no module logic. Django matches the first pattern, so `serve_static` never runs for `/metrics` and it will **not** redirect on voices — unlike the other 11 paths. metrics isn't a one-off; it's a sign the redirect is at the wrong layer. Consider a centralized rule keyed on the path set (middleware or a routing-level redirect) that fires regardless of which view handles the request, or add a voices redirect to the `metrics` view and document the exception.

static/js/ReaderApp.jsx
```diff
@@ -1260,6 +1260,15 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       return true;
     }
     const path = decodeURI(url.pathname);
+    if (Sefaria.activeModule === Sefaria.VOICES_MODULE) {
+      const sidebarPaths = new Set(
+        (Sefaria._siteSettings.ABOUT_SIDEBAR_PAGES || []).map(p => '/' + p.path)
+      );
+      if (sidebarPaths.has(path)) {
+        window.open(Sefaria.util.fullURL(path, Sefaria.LIBRARY_MODULE), '_blank');
```
yodem: suggestion (claude): Two things on this `window.open`:

1. **No `'noopener,noreferrer'`** — the opened library tab keeps a live `window.opener` back to the voices page (reverse-tabnabbing surface). Low practical risk since the target is first-party, but this is newly-added code; pass `'noopener,noreferrer'` as the third arg.
2. **Query string is dropped** — `path` here is `decodeURI(url.pathname)` (pathname only), and `fullURL` doesn't forward `url.searchParams`. The *server-side* redirect preserves query params (you even test that). So `/about?lang=he&utm_source=x` clicked in-app on voices loses `?lang=he&utm_source=x`, diverging from the full-page-load behavior — breaks lang overrides and analytics attribution. Append `url.search` to the opened URL.

static/js/ReaderApp.jsx
```diff
@@ -1260,6 +1260,15 @@ toggleSignUpModal(modalContentKind = SignUpModalKind.Default) {
       return true;
     }
     const path = decodeURI(url.pathname);
+    if (Sefaria.activeModule === Sefaria.VOICES_MODULE) {
+      const sidebarPaths = new Set(
```
yodem: nitpick (claude): `sidebarPaths` is rebuilt (new array + new Set) on every `openURL` call while on voices, including the common case where the path doesn't match. `ABOUT_SIDEBAR_PAGES` is static config — hoist this to a module-level constant (mirroring the server's `_ABOUT_SIDEBAR_PATHS`) so it's computed once.

reader/tests/static_pages_test.py
```diff
@@ -0,0 +1,48 @@
...
+def _make_request(factory, path, active_module, params=None):
+    request = factory.get(path, params or {})
+    request.active_module = active_module
+    request.interfaceLang = "english"
+    request.LANGUAGE_CODE = "en"
+    request.user = AnonymousUser()
+    return request
+
+
+def test_voices_sidebar_page_redirects_to_library(factory):
```
yodem, line 27: suggestion (claude): These tests call `serve_static(...)` directly and never exercise Django URL dispatch (no `client.get(...)`), so they validate the function's branching but can't observe routing precedence — which is exactly why the `/metrics` gap (routed before the `serve_static` catch-all) is invisible to the suite. A request-level test against the real URLconf asserting `/metrics` on voices redirects would have caught it. Also minor: `assert library_domain in response["Location"]` is vacuously true if `DOMAIN_MODULES` lacks the library entry (`library_domain == ""`); assert against a fixed expected domain with `startswith` instead.

(Separately: the PR description says tests also live in `reader/tests/test_catchall_redirects.py`, but the diff doesn't touch that file — looks like a description error.)

yodem (commented):
**Multi-agent review (claude)** — ran a 5-agent pass (security, code-quality, tests, architecture, frontend) over the diff. Security came back clean: the redirect target and all template values are fixed server-side settings, not user input, so no open-redirect/XSS. Solid, focused PR. One substantive gap and a few cleanups posted inline below.

**Headline:** `/metrics` is listed in `ABOUT_SIDEBAR_PAGES` but is routed to `reader_views.metrics` (`urls.py:66`) *before* the `serve_static` catch-all (`urls.py:73`), so the new voices→library redirect silently never fires for it — it's the one about-page of twelve that won't redirect on direct navigation. Two independent agents flagged this as a layering issue: the redirect lives inside one view, so any about-page with its own view bypasses it.

_Verdict: comment (not blocking) — the metrics gap is worth a fix or an explicit accepted-exception note before merge._

reader/views.py
YishaiGlasner: nice catch!
there won't be weird behaviour for there is no link to metrics within voices.
that's true that all the about pages should have been treated as one package, but this PR is already quite bug refactor for small bug.

static/js/ReaderApp.jsx
YishaiGlasner: 1. this isn't low risk but no risk but it's considered best practice so i'll do it.
2. i already answered copilot on this (in his first review)

templates/_sidebar.html
```diff
@@ -6,42 +6,11 @@ <h1 class="int-en">{{title}}</h1>
...
-                    <span class="int-en">Terms of Use<i class="fa fa-chevron-right"></i></span>
-                    <span class="int-he">תנאי שימוש<i class="fa fa-chevron-left"></i></span></a></li>
-            <li><a id="privacy" href="{{ domain_modules.library }}/privacy-policy">
-                    <span class="int-en">Privacy Policy<i class="fa fa-chevron-right"></i></span>
-                    <span class="int-he">מדיניות פרטיות<i class="fa fa-chevron-left"></i></span></a></li>
+            {% for page in SITE_SETTINGS.ABOUT_SIDEBAR_PAGES %}
+            <li><a id="{{ page.path }}" href="/{{ page.path }}">
+                    <span class="int-en">{{ page.en }}<i class="fa fa-chevron-right"></i></span>
+                    <span class="int-he">{{ page.he }}<i class="fa fa-chevron-left"></i></span></a></li>
+            {% endfor %}
```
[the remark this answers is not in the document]
YishaiGlasner: fixed

reader/tests/static_pages_test.py
YishaiGlasner: answered on the main comment

yodem (dismissed):

yodem:
### CI is stuck because the URLconf fails to import — dangling `serve_static_by_lang` references

The `Continuous Testing: Sandbox Deploy` → `sandbox-ready` step hangs and eventually times out (50 min), which skips PyTest and fails `continuous-branch-protection`. This is a real code bug, not CI flakiness — re-running won't help.

**Root cause:** this PR merged `serve_static_by_lang` into `serve_static(request, page, by_lang=False)` and **deleted** `serve_static_by_lang`, but two references to the old name remain:

- `sites/sefaria/urls.py:75` — `re_path(..., reader_views.serve_static_by_lang)` (evaluated at **URLconf import time**)
- `reader/views.py:720` — `if match.func in {serve_static, serve_static_by_lang}:`

Because `sites/sefaria/urls.py` can't import, **every request 500s** during middleware URL resolution — including the `/healthz-rollout` readiness probe. So the sandbox web pod never becomes Ready and `sandbox-ready` polls until timeout.

Verified live in the sandbox pod (`sandbox-ca149a9-web`, image `sefaria-web-sc-39253:sha-ca149a9`): the app boots and `library` initializes fine, but hitting `/healthz-rollout` raises during `LocaleMiddleware` → `import_module('sites.sefaria.urls')` at the `serve_static_by_lang` line. (The `'str' object has no attribute 'GET'` you may see in the pod logs is a secondary crash in the `social_image_url` template tag while Django renders the error page — not the actual cause.)

**Fix (two lines):**

```python
# sites/sefaria/urls.py:75
re_path(r'^(%s)/?$' % "|".join(static_pages_by_lang), reader_views.serve_static, {'by_lang': True}),

# reader/views.py:720
if match.func is serve_static:
```

Worth grepping for any other `serve_static_by_lang` references before pushing, but those two are the only non-test occurrences in the tree.

yodem:
Pushed the fix as `c44406814` — both references updated (the URLconf route in `sites/sefaria/urls.py` and the img-gen page-type check in `reader/views.py`), routed through the unified `serve_static(by_lang=True)`. This unblocks the sandbox health-check; a fresh CI run should now go green. Feel free to squash/adjust the commit as you see fit.

yodem (dismissed):

yodem (approved):

PR #3406, opened 2026-06-11
chore: csrf first pass

stevekaplan123:
## Description
This fixes dual-csrftoken-cookie 403s on cauldrons.  When someone has visited both sefaria.org and a cauldron, their browser ends up holding two cookies both named `csrftoken`: one scoped to the parent .sefaria.org (from production) and one scoped to the cauldron host. Both get sent on every request. The bug is that our JavaScript and Django disagree on which one to use: js-cookie/our getCookie return the first match (the cauldron token), while Django's cookie parser keeps the last (the production token). So the X-CSRFToken header we send never matches the token Django validates against, producing a 403.

## Code Changes

This PR stops reading the token from the cookie and instead reads it from a value Django stamps into the page itself: a new <meta name="csrf-token" content="{{ csrf_token }}"> tag in base.html. That value is derived from the exact token Django will validate against, so the two can never disagree no matter how many duplicate cookies exist. A single new helper, getCsrfToken() (static/js/sefaria/csrf.js), reads the meta tag with a cookie fallback, and all ~8 previous direct cookie reads now route through it (sefaria.js, BookPage, TopicEditor, StaticPages, and the four modtools components). 

## Code Changes
- **`static/js/sefaria/csrf.js`** (new) — single `getCsrfToken()` helper that reads
  the meta tag. The ~8 previous scattered cookie reads (sefaria.js, BookPage,
  TopicEditor, StaticPages, and the four modtools components) now route through it.
- **`static/js/lib/django-csrf.js`** — the bundled jQuery `ajaxSend` hook. Kept (it
  auto-attaches the header to legacy jQuery POSTs) but no longer duplicates logic —
  it imports `getCsrfToken` from the shared helper.
- **`templates/js/django-csrf.js`** — the inline hook for server-rendered pages that
  don't load the webpack bundle. Because this file is itself a Django template
  (pulled in via `{% include %}`), it now inlines `{{ csrf_token }}` directly instead
  of carrying a hand-maintained copy of `getCsrfToken()`. This removes the last
  duplicate, leaving a single source of truth.
- **`templates/base.html`** / **`templates/edit_text.html`** — render the meta tag
  (edit_text.html is standalone and needs its own copy for the React bundle).
- **`templates/edit_profile.html`** — switched from `<script src="{% static
  'js/django-csrf.js' %}">` to `{% include %}`. This also fixes a pre-existing 404:
  the `static` path no longer exists (the file was moved to `lib/`), so the hook
  hadn't been loading on that page.
- **`static/js/sefaria/tests/csrf.test.js`** (new) — covers the meta-tag read and the
  no-cookie-fallback guarantee.

## Notes
This was only tested on two APIs, but in both cases, this PR works perfectly: the presence of the meta tag that holds the `csrf_token` was necessary for the APIs to work.

### DIFF
--- a/static/js/lib/django-csrf.js
+++ b/static/js/lib/django-csrf.js
@@ -33,8 +33,18 @@ function init() {
             return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
         }
 
+        function getCsrfToken() {
+            // Prefer the server-rendered token over the cookie. A cauldron host
+            // inherits the parent-domain (.sefaria.org) csrftoken cookie alongside
+            // its own, and getCookie returns the first match while Django validates
+            // against the last — the meta tag always matches what Django validates.
+            var meta = document.querySelector('meta[name="csrf-token"]');
+            if (meta && meta.content) { return meta.content; }
+            return getCookie('csrftoken');
+        }
+
         if (!safeMethod(settings.type) && sameOrigin(settings.url)) {
-            xhr.setRequestHeader("X-CSRFToken", getCookie('csrftoken'));
+            xhr.setRequestHeader("X-CSRFToken", getCsrfToken());
         }
     });
 }
--- a/static/js/lib/django-csrf.js
+++ b/static/js/lib/django-csrf.js
@@ -1,22 +1,10 @@
 var $  = require('jquery');
+// Canonical CSRF token reader (meta tag first, cookie fallback). Shared with the
+// fetch()-based call sites so the dual-cookie fix lives in exactly one place.
+var getCsrfToken = require('../sefaria/csrf').getCsrfToken;
 
 function init() {
     $(document).ajaxSend(function(event, xhr, settings) {
-        function getCookie(name) {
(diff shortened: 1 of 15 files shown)


static/js/lib/django-csrf.js
```diff
@@ -1,22 +1,8 @@
 var $  = require('jquery');
+var getCsrfToken = require('../sefaria/csrf').getCsrfToken;
```
YishaiGlasner: let

templates/js/django-csrf.js
```diff
@@ -1,21 +1,13 @@
 // --- Django CSRF support for AJAX -----------
+//
+// NOTE: getCsrfToken() below is intentionally a standalone copy of
```
YishaiGlasner: maybe we can inject this in ssr as Sefaria.getCsrfToken and have one source of truth?

stevekaplan123: great point!

YishaiGlasner (approved):

PR #3407, opened 2026-06-11
feat: standalone name (autocomplete) service

EliezerIsrael:
## Summary

Extracts the autocompleters out of every web/worker process into an opt-in standalone "name service", behind a single helm flag (`nameService.enabled`, default **off**).

**What it does when enabled:**
- A `name` Rollout (web image, `NAME_SERVICE=true`) owns the full/lexicon/cross-lexicon autocompleters and serves only `/api/name`, `/api/words/completion`, `/api/opensearch-suggestions`, and `/search-autocomplete-redirecter` (byte-identical URL patterns and views — `sefaria/urls_name.py`, selected via django-hosts in `sefaria/hosts.py`).
- **varnish routes** those paths to a revision-pinned `name_service` backend in the VCL; everything else flows to web as today. Traffic path (nginx → varnish → backend) and the endpoints' never-cached semantics are unchanged.
- Web and Celery task pods run with `DISABLE_AUTOCOMPLETER=true`: they skip the builds (today every web worker *and all task pods* build the tries at boot) and their readiness stops gating on completer construction → faster rollouts, lighter pods.
- Rebuilds propagate over the existing multiserver Redis channel; topic-CRUD views now publish the rebuild event when completions are served externally (local-only semantics preserved otherwise).
- An nginx `prePromotionAnalysis` gate (`rollout-ready-name`) prevents the blue/green traffic flip until the new revision's name pods finish building.
- On KEDA environments, a `keda.nameService` scaledobject scales the name pod with web/redis so the whole stack sleeps and wakes together (after an idle wake, completions return ~minutes behind the rest of the site while the tries rebuild).

## ⚠️ Behavior change (deliberate, underwritten by Lev)

**`DISABLE_AUTOCOMPLETER` is now strict in all cases**: a process with the flag set neither builds completers nor serves them — the completion endpoints fail fast with a clear error instead of triggering the legacy multi-minute lazy in-request build (which could wedge single-worker cauldrons). Environments that set the flag and want autocomplete should deploy the name service (`create-cauldron.sh -N` on cauldrons) or unset the flag. The transitional `NAME_SERVICE_DEPLOYED` flag was removed before merge.

**Other cleanups:**
- Unknown lexicon names raise `InputError` instead of triggering an unauthenticated full in-request rebuild.
- Vestigial `_topic_auto_completer_is_ready` flag and dead `include_people` kwarg removed.

## Test plan

- [x] `helm template` renders cleanly with the flag on and off
- [x] Tests in `sefaria/model/tests/autocompleter_disabled_test.py`: strict no-op builds, fail-fast accessors, lexicon InputError, URLconf routing
- [x] hosts.py routing verified in both modes
- [x] Live cauldron verification (`autocompleterefactor.cauldron.sefaria.org`): completions served by the name pod through varnish; web pods ready in under a minute; KEDA sleep/wake cycle verified end to end


helm-chart/sefaria/templates/configmap/local-settings-file.yaml
```diff
@@ -215,6 +215,10 @@ data:
     # Turns off search autocomplete suggestions, which are reinitialized on every server reload
     # which can be annoying for local development.
     DISABLE_AUTOCOMPLETER = os.getenv("DISABLE_AUTOCOMPLETER", "false").lower() == "true"
+    # When true this process serves only the completion endpoints (sefaria/urls_name.py)
+    NAME_SERVICE = os.getenv("NAME_SERVICE", "false").lower() == "true"
+    # When true a name service serves completion traffic for this deployment
+    NAME_SERVICE_DEPLOYED = os.getenv("NAME_SERVICE_DEPLOYED", "false").lower() == "true"
```
BrendanGalloway: functionally, what is the difference between name service and name service deployed?

helm-chart/sefaria/templates/configmap/varnish-config.yaml
BrendanGalloway, line 1: Do we want to do this routing in varnish or in nginx? The rest of the routing currently sits in nginx, although a lot of what nginx does currently could in theory move to the gateway controller.

helm-chart/sefaria/templates/rollout/name.yaml
```diff
@@ -0,0 +1,229 @@
...
+kind: Rollout
+metadata:
+  name: {{ .Values.deployEnv }}-name
+  labels:
+    deployEnv: "{{ .Values.deployEnv }}"
+    app: name-{{ .Values.deployEnv }}
+  annotations:
+    rollout.argoproj.io/revision: "{{ .Release.Revision }}"
+spec:
+  strategy:
```
BrendanGalloway, line 18: You created an analysis template for this rollout, but have not made use of it here.

You probably also need to run the end-to-end analysis so that this pod rolls forward/backward with the version pinning the same as the other services.  I assume this service has the same hard version compatibility requirements to node/web that the other services currently have?

helm-chart/sefaria/templates/rollout/name.yaml
```diff
@@ -0,0 +1,229 @@
...
+        imagePullPolicy: Always
+        args: [ "gunicorn sefaria.wsgi --access-logfile - --error-logfile - --timeout 420 --threads {{ .Values.nameService.resources.gunicornThreadCount }} --worker-tmp-dir /dev/shm -b 0.0.0.0:80" ]
+        env:
+          # WEB_CONCURRENCY is used for determining the number of server workers
+          - name: WEB_CONCURRENCY
+            value: "{{ .Values.nameService.resources.gunicornWorkerCount }}"
+          - name: randomstringtoforceredeployments
+            value: {{ randAlphaNum 8 }}
+          - name: GOOGLE_APPLICATION_CREDENTIALS
+            value: /app/logging-secret.json
```
BrendanGalloway, line 95: Does name service need this?

helm-chart/sefaria/templates/rollout/name.yaml
```diff
@@ -0,0 +1,229 @@
...
+          periodSeconds: 60
+          timeoutSeconds: 60
+          failureThreshold: 10
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 80
+          periodSeconds: 60
+          timeoutSeconds: 60
+          failureThreshold: 10
```
BrendanGalloway: If this is the case healthz should be smarter.  Readiness should be false during rebuild, so that kube does not direct traffic to a name service that is not able to respond, but liveness probe should still complete to indicate that the pod is up and will return to service on its own.  This approach hides valid failure modes and potentially allows routing to a pod that is not able to respond.

helm-chart/sefaria/templates/rollout/nginx.yaml
```diff
@@ -19,6 +19,10 @@ spec:
         templates:
           - templateName: rollout-ready-{{ .Values.deployEnv }}
           - templateName: rollout-priming-{{ .Values.deployEnv }}
+          {{- if .Values.nameService.enabled }}
+          # don't flip traffic before the name service has built its completers
+          - templateName: rollout-ready-name-{{ .Values.deployEnv }}
+          {{- end }}
```
BrendanGalloway, line 25: why is this here?

helm-chart/sefaria/templates/configmap/local-settings-file.yaml
EliezerIsrael: So - the semantics had been:
* NAME_SERVICE_DEPLOYED is whether there is a name service in this deployment. 
* NAME_SERVICE indicates if this pod itself operates the name service. 

As I looked more into it, the existing DISABLE_AUTOCOMPLETER can do the work of NAME_SERVICE_DEPLOYED.  The agent created a new one so as to preserve a lazy-load behavior when DISABLE_AUTOCOMPLETER was true.  I judged that we don't need that lazy-load behavior.  We're jettisoning NAME_SERVICE_DEPLOYED.

So - target state is:
* DISABLE_AUTOCOMPLETER: if true, this pod doesn't run an autocompleter, ever.  Used both for local deployments and for web servers on cauldron/prod.  
* NAME_SERVICE : if true, this pod runs the name service.

helm-chart/sefaria/templates/configmap/varnish-config.yaml
EliezerIsrael: I put this behind Varnish, because these requests are good caching candidates.  They aren't being cached right now, either before or after this change, but I think that should change soon. 

I could imagine giving this service a separate Varnish server and keeping the routing higher up in nginx or gateway.  It would have the side-effect of making the VCL simpler and the purge logic on rebuild more straightforward.

helm-chart/sefaria/templates/rollout/name.yaml
EliezerIsrael: I think we should assume it has the same version compatibility requirements, as default.

But in practice - the sensitivity is much less.  

Regarding node and web - If node renders a DOM and the client code delivered my Django renders a different DOM, we have client side behavior inconsistency.  Any template change could cause this.  

For the name server, there is a very small class of changes that would bring it out-of-sync with web, and nothing that would bring it into conflict with node.

EliezerIsrael: Good catch — fixed in 8ba72ced7. The name rollout now runs the same end-to-end gate as web/nodejs (`rollout-ready-{env}` against `nginx-{env}-{rev}`), so it promotes and aborts in lockstep with the rest of the stack — a held deploy can no longer reap the old revision's name pods while that revision's varnish still pins them.

The `rollout-ready-name` template stays in nginx's prePromotionAnalysis for the inverse direction (don't flip traffic before the completers finish building). No deadlock between the two gates: both poll revision-pinned services, which route on pod readiness independent of promotion state — name pods go Ready → nginx's gate clears; the new nginx→varnish→web path goes healthy → name's gate clears.

EliezerIsrael: (That last one was Claude answering as me.  I expected it to answer with its own identity.  Such a weird world.)

helm-chart/sefaria/templates/rollout/name.yaml
EliezerIsrael: (Claude sez:) Agreed — implemented in 40203f648. The probes are now split: readiness fails after ~30s of unresponsiveness (15s period / 10s timeout / threshold 2) so kube de-routes a rebuild-blocked pod and re-admits it as soon as it answers again; liveness keeps the generous 10×60s budget so the pod isn't killed mid-rebuild and returns to service on its own. A fully-blocked sync worker fails the probe by timeout naturally, so no app-level smarts needed — the old config was suppressing exactly the de-routing you describe.

helm-chart/sefaria/templates/rollout/name.yaml
EliezerIsrael: Not sure.  I just lifted all the config from web and brought it over.  Deserves a review generally.   If we need it for logging, we need it everywhere.  If we don't need it here, we probably don't need it there either.

helm-chart/sefaria/templates/rollout/nginx.yaml
EliezerIsrael: web pods no longer build completers, so the generic gate passes minutes before the name pod is ready; without this line, every deploy has a 502 window on completion paths.

helm-chart/sefaria/templates/rollout/name.yaml
BrendanGalloway: We removed the need for some json secrets in the past, but I don't think the exercise to remove them completely was ever completed.  Logging as a whole can be handled by stdout in the pod feeding to stackdriver in google, I'm not sure what this secret was doing in web

helm-chart/sefaria/templates/rollout/name.yaml
BrendanGalloway: I meant the actual endpoints, not just the tuning of the parameters.  Although this is an improvement.  It would be much better that some separate thread is able to respond authoritatively saying the pod is up and able or not able to serve traffic than to guess at the cause of non-responsiveness.  This config means any other genuine failed state needs to wait 10 minutes to be reset, just in case it's actually a rebuild.

helm-chart/sefaria/templates/configmap/local-settings-file.yaml
nsantacruz: can we rename DISABLE_AUTOCOMPLETER to ENABLE_AUTOCOMPLETER? The current name is extremely confusing.

EliezerIsrael:
Closing this and moving to #3671

PR #3416, opened 2026-06-17
fix: [sc-41882] Floating bot (assistant) obscures header dropdowns (User menu, Module picker)

stevekaplan123:
## Description
When chatbot is in floating mode and user clicks one of the drop down menus in the header, the chatbot is covered.  Since the menus have 9999 z-index, it's necessary to make the z-index 10000 for the header drop down menus.  I've defined several z-index variables (one for 9999 and one for 10000) so we can keep track of z-indices.


saengel (dismissed):

YishaiGlasner (changes requested):
this is a z-index war - it's an anti patterm. you should find a better way to handle it

relyks:
> this is a z-index war - it's an anti patterm. you should find a better way to handle it

Too late. I started it 3 months ago lmao: [Let the z-index wars between Reader App and Library Assistant commence!](https://github.com/Sefaria/Sefaria-Project/pull/3167)

YishaiGlasner (approved):

PR #3417, opened 2026-06-17
Fix English header on Sheets-with-Ref view in Chiburim

saengel:
# Overview

## Description
This PR fixes a big where uses in the library (Hebrew interface) select a text, then click on "דפי מקורות" in the sidebar, and when they arrive at chiburim.sefaria.org.il, the title for the page is in English (see the visuals below). 

There were two issues:
1. No translation in Strings.s for `Sheets with`
2. Only passing the English ref, versus passing the Hebrew ref as well

## Code Changes
1. `static/js/ReaderPanel.jsx` - Passes the entire ref object (with Hebrew and English) as props, instead of just English. 
2. `static/js/sefaria/strings.js` - Add Hebrew copy
3. `static/js/sheets/SheetsWithRefPage.jsx` - Parsing out the ref prop passed to separate out the Hebrew and English, and pass them through as needed. 

# Visually

## Before
Screenshot 2026-06-17 at 10 36 52

## After
Screenshot 2026-06-17 at 10 36 11

### DIFF
--- a/static/js/sheets/SheetsWithRefPage.jsx
+++ b/static/js/sheets/SheetsWithRefPage.jsx
@@ -16,7 +16,9 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // we can return the searchState's available filters to the original list.  Once origAvailableFilters are loaded,
     // they are never changed.
 
-    const [refs, setRefs] = useState(srefs);
+    const [refs, setRefs] = useState(srefs.en);
+    const shortLang = Sefaria._getShortInterfaceLang();
+    const displayRef = srefs[shortLang] || srefs.en;
     const sortTypeArray = SearchState.metadataByType['sheet'].sortTypeArray;
 
     const cloneFilters = (availableFilters, resetDocCounts = true) => {
@@ -195,7 +197,7 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
           sortTypeArray={sortTypeArray}
           searchTopMsg="Sheets With"
           hits={sortedSheets}
-          query={refs}
+          query={displayRef}
           type={'sheet'}
           totalResults={new SearchTotal({value: sortedSheets.length})}
           compare={false}
--- a/static/js/sheets/SheetsWithRefPage.jsx
+++ b/static/js/sheets/SheetsWithRefPage.jsx
@@ -16,7 +16,7 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // we can return the searchState's available filters to the original list.  Once origAvailableFilters are loaded,
     // they are never changed.
 
-    const [refs, setRefs] = useState(srefs.en);
+    const enRefs = srefs.en;
     const shortLang = Sefaria._getShortInterfaceLang();
     const displayRef = srefs[shortLang] || srefs.en;
     const sortTypeArray = SearchState.metadataByType['sheet'].sortTypeArray;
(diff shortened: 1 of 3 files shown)


static/js/sheets/SheetsWithRefPage.jsx
```diff
@@ -16,7 +16,9 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // we can return the searchState's available filters to the original list.  Once origAvailableFilters are loaded,
     // they are never changed.
 
-    const [refs, setRefs] = useState(srefs);
+    const [refs, setRefs] = useState(srefs.en);
```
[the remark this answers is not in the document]
yodem: I would also say - if we dont use refs as a state - why do we need to do useState? cant we just do a const or use srefs directly?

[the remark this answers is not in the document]
YishaiGlasner: I would also say - if we dont use state, why do we need this var at all? is it used? also its name doesn't tell what it's - ut shoiuld be called `enRefs`, but how this is better than usin `refs.en`

stevekaplan123 (changes requested):
In sefaria.js, we have `_r` if the ref is in cache.  I don't know if it already is but it might be worth checking.  The sefaria.js function:
```
_r: function (inputRef) {
const oref = Sefaria.getRefFromCache(inputRef);
if (!oref) { return inputRef; }
return Sefaria.interfaceLang != "english" ? oref.heRef : oref.ref;
},```

File: static/js/sheets/SheetsWithRefPage.jsx [hunk 2]
```diff
@@ -16,7 +16,7 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // we can return the searchState's available filters to the original list.  Once origAvailableFilters are loaded,
     // they are never changed.
 
-    const [refs, setRefs] = useState(srefs.en);
+    const enRefs = srefs.en;
     const shortLang = Sefaria._getShortInterfaceLang();
     const displayRef = srefs[shortLang] || srefs.en;
```
YishaiGlasner:
```suggestion
    const displayRef = srefs[shortLang] || enRefs;
```

saengel:
> In sefaria.js, we have `_r` if the ref is in cache. I don't know if it already is but it might be worth checking. The sefaria.js function:
> 
> ```
> _r: function (inputRef) {
> const oref = Sefaria.getRefFromCache(inputRef);
> if (!oref) { return inputRef; }
> return Sefaria.interfaceLang != "english" ? oref.heRef : oref.ref;
> },```
> ```

What benefit would this give in this case?

static/js/sheets/SheetsWithRefPage.jsx
```diff
@@ -16,7 +16,9 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
     // we can return the searchState's available filters to the original list.  Once origAvailableFilters are loaded,
     // they are never changed.
 
-    const [refs, setRefs] = useState(srefs);
+    const enRefs = srefs.en;
```
stevekaplan123, line 19: I originally suggested using _r.  You should actually use _v:
```
  _v: function(langOptions){
    /* Takes an object {en: "something", he: "משהו"}
     * and returns the correct one according to interface language
     * Convenience method for when there are two data variables in an object one wishes to return
     * according to interface, in places where HTML is not allowed (inside <options> tag for ex.
    */
    const lang = Sefaria._getShortInterfaceLang();
    return langOptions[lang] ? langOptions[lang] : "";
  },
```

stevekaplan123 (changes requested):
I think you should use Sefaria._v

static/js/sheets/SheetsWithRefPage.jsx
saengel: Ah okay, got it, this makes sense. Thanks for pointing out the function.

stevekaplan123 (approved):

PR #3418, opened 2026-06-17
chore: remove divine name code

stevekaplan123:
## Description
Removes divine name feature.  This was already inaccessible via the UI but this code actually completely turns off the feature regardless of the sheet's settings.


sefaria/stats.py
```diff
@@ -206,23 +206,23 @@ def save_top_sources_sheet(self):
 		sheet = {
 			"title": "Top Sources in All Source Sheets - %s" % datetime.now().strftime("%B %Y"),
 			"sources": [{"ref": ref[0]} for ref in self.sorted_refs[:self.show_count]],
-			"options": {"numbered": 1, "divineNames": "noSub"}
+			"options": {"numbered": 1}
```
YishaiGlasner, line 209: it seems we can reduce duplication here, but i'm not sure about that nor it's the right spec

YishaiGlasner (approved):
i thought it's going to take me time before really checking...
please see and consider my comment on the code before merging.

PR #3420, opened 2026-06-17
fix: serialize and coalesce autocompleter rebuilds to stop OOM

yodem:
## Problem

Production web pods OOM-kill in a sawtooth pattern (RSS climbs to 12–15 GiB, killed, restart, repeat). A profiled incident showed a single pod going **4.5 → 11.9 GiB in 10 minutes**, with `Library.build_full_auto_completer` → `AutoCompleter.__init__` dominating allocations.

**Root cause:** the three autocompleter build methods in `Library` (`sefaria/model/text.py`) have no concurrency control. Under the gevent gunicorn worker (`worker_connections=2000`), a single rebuild trigger — a multiserver event replayed from a staff topic edit (`reader/views.py`), or the admin `rebuild_auto_completer` endpoint (`sefaria/views.py`, which fires 3 events) — fans out into **many concurrent ~1 GiB builds inside one worker**: each build yields on Mongo I/O, letting more greenlets enter the build before any finishes. The author's own `# I worry that these could pile up.` comments flagged exactly this.

## Fix (pure Python, `text.py` only)

Each of the three build methods (full / lexicon / cross-lexicon) now:

1. **Serializes** behind a gevent-cooperative `BoundedSemaphore(1)` — at most one build of a given completer runs at a time, collapsing the greenlet herd to one.
2. **Coalesces** redundant requests via a `requested`/`completed` counter — a rebuild request that arrives while a build is already in flight is satisfied by that build instead of re-running. An event storm of N triggers collapses to **1** rebuild, not N sequential ones.
3. **Atomically swaps** the new completer in (build into a local, then a single assignment) — readers keep serving the previous completer until the new one is fully ready, bounding transient memory to ~2× one completer instead of N×.

The lock import is guarded (`try: from gevent.lock import BoundedSemaphore / except ImportError: from threading import BoundedSemaphore`) because gevent is only present in the gunicorn runtime; dev/test fall back to a thread lock. gevent's semaphore acquires the hub lazily, so it is safe to construct in the master before fork/monkey-patching (the build path there is uncontended).

The getters keep their existing `KeyError`/`None` fallback as a cold-cache safety net — they just now call the safe, coalesced build. The three `# I worry…` comments are removed.

## Scope / non-goals

This is the OOM-stopping core. Deliberately **out of scope** (follow-on hardening): cross-pod Redis build lock, moving multiserver rebuilds to a background greenlet, `post_fork` connection reset, and MongoDB field projections in `AutoCompleter.__init__`. Note one residual cost: with `preload_app=True`, each worker still rebuilds independently — but now as a single, serialized, atomic build per worker rather than a concurrent herd.

## Verification

- Concurrency design independently reviewed and cross-checked against gevent docs: the atomic swap is safe (greenlets only yield at I/O/lock points, never mid-assignment); the lock-free readiness read is safe under gevent; no reentrancy/deadlock (`AutoCompleter.__init__` does not re-enter the completer getters); `BoundedSemaphore` chosen over `RLock` precisely to keep reentrancy a crash, not a silent mask.
- `python -m py_compile sefaria/model/text.py` passes.
- Plan: validate on the `production-web-dd-test` canary pod — compare allocation rate against baseline before full rollout.

### DIFF
--- a/sefaria/model/tests/autocompleter_coalesce_test.py
+++ b/sefaria/model/tests/autocompleter_coalesce_test.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8 -*-
+"""
+Concurrency tests for Library.build_full_auto_completer's serialize/coalesce logic.
+
+These do NOT build a real auto completer (that needs Mongo/Postgres, which is why
+autospell_test.py is disabled). Instead they stub out AutoCompleter and drive the real
+method over real OS threads, asserting the coalescing-counter behaviour directly.
+
+The build lock on the fake library is a threading.BoundedSemaphore so the test is identical
+on macOS dev (no gevent -> threading fallback) and on the gevent-on-Linux CI: the method only
+uses the lock through the BoundedSemaphore context-manager API, which both implementations
+share. What is under test is the requested/completed counter logic, not the lock flavour.
+
+AutoCompleter is supplied as a stub module in sys.modules so the test pulls in no real
+auto completer code (and no datrie/Mongo) -- build_full_auto_completer imports it lazily via
+`from .autospell import AutoCompleter`.
+"""
+import sys
+import time
+import types
+import threading
+
+import sefaria.model.text as text
+
+
+def _install_stub_autocompleter(monkeypatch, stub_cls):
+    """Make `from .autospell import AutoCompleter` (inside the build method) resolve to stub_cls."""
+    stub_module = types.ModuleType("sefaria.model.autospell")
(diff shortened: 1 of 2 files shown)


EliezerIsrael:
@yodem I'd like to see this go through a test that mirrors productions - at least two pods, each with multiple workers - either in the cauldron or pre-prod environment.   Let's aim to get a topology like that running the unpatched code to show the OOM behavior, and show this code to fix the OOM behavior.

PR #3421, opened 2026-06-18
chore: update_many instead of versionset

stevekaplan123:
### Description

Renaming a book fails in the case when the first version record for it in mongo is not primary.  The reason for this is unnecessary validation.

### Root cause

A title change fires `process_index_title_change_in_versions`, which called `VersionSet({"title": old}).update({"title": new})`, which in turn calls `AbstractMongoSet.update()`.  This `update` function doesn't do a bulk write — it iterates over all versions in the set and calls `.save()` on each version, triggering `Version._validate()`.

During a rename the `Index` title is updated **first**, so when each version is saved under the new title, `_validate()` runs a check that "there must be at least one primary version" corresponding to the Index:

```
        if not self.isPrimary and not is_any_other_primary:  # if all are False, return true
            raise InputError("There must be at least one version that is primary.")
```

The problem is if the first version checked isn't primary, this is guaranteed to fail because the primary version's title hasn't been changed yet.  This is order-dependent: failure occurs only when the first version iterated is non-primary.  Then, the Celery `update_index_title` task dies mid-cascade, and Texts/Links are never migrated.

### Why Now?!
The "isPrimary" check is only a year old.  In addition, it is order-dependent — it fails only when the first version iterated is non-primary.

### Code Changes

Replace the per-document `.save()` loop with a direct `db.texts.update_many()`. Renaming the `title` field on all versions is a pure data cascade, not a user edit, so it shouldn't run record-level validation.  Moreover the only title check in the version validation is to see that it's an existing Index record which we already know to be true.   Finally, no `Version` dependencies are registered on the `title` field (only on `versionTitle`), so no cascades are lost — and it's a single MongoDB call instead of N.


sefaria/model/tests/text_test.py
```diff
@@ -531,6 +531,58 @@ def construct_query(attribute, queries):
     return ret
 
 
+@pytest.mark.deep
+def test_index_name_change_with_nonprimary_version_first():
```
YishaiGlasner: i think that today we're doing tests by mocking rather than really manipulate our db

sefaria/model/tests/text_test.py
```diff
@@ -531,6 +531,53 @@ def construct_query(attribute, queries):
     return ret
 
 
+def test_index_rename_migrates_versions():
```
stevekaplan123, line 534: @YishaiGlasner I took out the previous test and instead put in a full integration test instead of the previous test I had.  Since it tests the entirety of saving and cascading, it's necessary to do a real DB test rather than mock test.

YishaiGlasner (approved):

PR #3439, opened 2026-06-24
Powered-by-Sefaria Data Wiring

saengel:
## Description
This PR wires up a basic backend setup for storing our projects powered by Sefaria in Postgres, as well as a basic endpoint that returns the raw data from the DB. 

## Code Changes
1. `powered_by/` - A new Django App for PBS (powered by Sefaria), most important of the subfiles is `powered_by/models.py` where the model lives. 
2. `sefaria/settings.py` - Configuration in settings, addition of `django.contrib.postgres` (According to Claude, `ArrayField` lives in `django.contrib.postgres.fields`, and Django refuses to use it unless `django.contrib.postgres` is in `INSTALLED_APPS`.)
3.  `sefaria/urls_shared.py` - URL added for the new API Endpoint


powered_by/models.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __str__(self):
+        return self.project_name
+
+    def contents(self):
+        """Return a JSON-safe dict of all fields (datetimes as ISO strings)."""
+        def iso(dt):
+            return dt.isoformat() if dt else None
+
+        return {
```
YishaiGlasner: i don't like this huge repeatitive code.
maybe you can use `from rest_framework import serializers`
rather something like
```
    data = model_to_dict(self)

    for k, v in data.items():
        if hasattr(v, "isoformat"):
            data[k] = v.isoformat()

    return data
```

powered_by/views.py
```diff
@@ -0,0 +1,15 @@
...
+@catch_error_as_json
+def powered_by_api(request):
+    """
+    Basic GET endpoint returning all Powered by Sefaria projects.
+
+    No filtering or pagination. Returns:
+        {"projects": [ { ...all project fields... }, ... ]}
+    """
+    projects = [project.contents() for project in Project.objects.all()]
+    return jsonResponse({"projects": projects})
```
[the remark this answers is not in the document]
YishaiGlasner: django must have something for priavte and public attrs. probably using it will be the best

YishaiGlasner (changes requested):
add tests

powered_by/models.py
saengel: From Claude (after adding protections for PII): model_to_dict/serializer approach is a blacklist (dump all fields, then strip private ones) — the wrong default for PII, since any future sensitive field leaks unless someone remembers to exclude it. Our explicit contents() is a whitelist that fails closed. Also, model_to_dict silently drops id, created_at, and updated_at (non-editable/auto fields), and a DRF serializer would be the only one of its kind in a codebase that standardizes on the contents() pattern

powered_by/models.py
```diff
@@ -0,0 +1,198 @@
...
+            models.Index(fields=["featured"]),
+            models.Index(fields=["status"]),
+        ]
+
+    def __str__(self):
+        return self.project_name
+
+    # PII / internal staff metadata. Only serialized for authenticated staff;
+    # see contents() and powered_by.views.powered_by_api.
+    PRIVATE_FIELDS = (
```
YishaiGlasner: i'd take this to the top of the class and add `PUBLIC_FIELDS`
then, in `contents`:
```
contents = {field: getattr(self, field) for field in self.PUBLIC_FIELDS}

for field in ("submission_date", "created_at", "updated_at", "last_checked"):
    contents[field] = iso(contents[field])
```

YishaiGlasner (approved):

PR #3441, opened 2026-06-25
Fix 404 bug on sheets

saengel:
## Description
We have a bug when when viewing a source sheet in Voices, the texts API was being called and 404'ing because Sheet IDs (i.e. `123456`) were being passed in as refs. 

## Code Changes
`static/js/sefaria/sefaria.js` - Adds a check in the `saveUserHistory` function if a ref is a sheet ID before calling the texts API, returns the sheet ID without the call for the sake of preserving history.


dcschreiber (approved):
The comment isn't so clear to me. Although not a bloker

PR #3442, opened 2026-06-25
Bug/sc 45009/title rename leaves index key versionstate

stevekaplan123:
## Description
The server failed to boot with KeyError: 'Rashi on Genesis 2' in all_index_records(). Root cause: _index_map is keyed by i.title while _index_title_maps is keyed by nodes.key; these diverge when an Index's title ≠ its root node key, and the lookup raised. This is one instance of a pervasive anti-pattern on the startup path: per-record loops over DB records with no per-item error handling, so any single malformed record (Index, Topic, Term, Collection, Lexicon entry, user profile, vstate doc, link) aborts the whole build and prevents server boot.

## Code Changes
Per-record try → log → skip guards across the boot path so one bad record is logged and skipped.  In `skip_tracking.py` we store the logs and report them all together in one Slack message.

Example of logs:
```
{"event": "[pathway:rebuild_toc,startup] TocTree vstate record: skipping ObjectId('54b0265afbfba21cd644530f'): 'title'", "timestamp": "2026-06-30T07:47:56.590355Z", "logger": "sefaria.model.category", "severity": "warning"}
{"event": "[pathway:rebuild_toc,startup] all_index_records key (title/nodes.key mismatch): skipping 'Rashi on Genesis 2': 'Rashi on Genesis 2'", "timestamp": "2026-06-30T07:47:56.670243Z", "logger": "sefaria.model.text", "severity": "error"}
```

Example of Slack message:
(In the most up-to-date code, instead of showing the confusing "init_library_cache", it shows "startup".

image


sefaria/model/text.py
```diff
@@ -5027,7 +5028,15 @@ def _build_index_maps(self):
...
-        self._index_map = {i.title: i for i in IndexSet() if i.nodes}
+        # Build the map record-by-record so one corrupt Index (e.g. accessing .title/.nodes
+        # raises) logs and is skipped rather than aborting the whole rebuild.
+        self._index_map = {}
+        for i in IndexSet():
+            try:
+                if i.nodes:
+                    self._index_map[i.title] = i
+            except BAD_RECORD_EXCEPTIONS as e:
+                logger.error("_build_index_maps: skipping malformed index record {}: {}".format(getattr(i, "_id", "<unknown>"), e))
```
stevekaplan123: add startup tag to this and others

sefaria/model/text.py
```diff
@@ -5242,9 +5255,14 @@ def get_topic_toc_json_recursive(self, topic=None, explored=None, with_descripti
             if child_topic is None:
                 logger.warning("While building topic TOC, encountered non-existant topic slug: {}".format(child))
                 continue
-            topic_json['children'] += [self.get_topic_toc_json_recursive(child_topic, explored, with_descriptions)]
+            # One corrupt child topic (e.g. missing title_group) must not abort the whole TOC build.
+            try:
```
stevekaplan123: context manager "with"

stevekaplan123: maybe we can put some of these checks in AbstractMongoRecord

stevekaplan123: maybe don't allow > X to fail so we don't get bombarded with engineering-signal msgs

stevekaplan123: 1. something more Pythonic to make the try...except nicer
2. def add engineering-signal and add [start-up] tag and dont worry about the "dont allow > X" for now

sefaria/helper/skip_tracking.py
```diff
@@ -0,0 +1,192 @@
...
+    propagate and abort loudly; pass `exceptions=` to narrow further (e.g. KeyError).
+
+    Usage:
+        skip_bad_record = bad_record_guard(logger)   # once, at module top
+        for rec in SomeSet():
+            with skip_bad_record("startup", "build_virtual_books", record=rec_id):
+                ...build using rec...
+    """
+    @contextmanager
+    def skip_bad_record(pathway, what, record=None, level="warning", exceptions=BAD_RECORD_EXCEPTIONS):
```
stevekaplan123: Rename 'what'

yitzhakc (commented):
Looks ok

sefaria/model/linker/category_resolver.py
```diff
@@ -45,10 +50,16 @@ class CategoryMatcher:
...
-                        self._title_to_cat[title] += [cat]
+            # One category whose match_template references a nonexistent/corrupt term
+            # (term.get_terms() yields None, or term.get_titles() raises) must not abort startup.
+            with skip_bad_record("startup", "CategoryMatcher category", record="/".join(getattr(cat, "path", []) or [])):
+                for match_template in cat.get_match_templates():
+                    for term in match_template.get_terms():
+                        if term is None:
+                            log_skip(logger, "startup", "CategoryMatcher category match_template", "category '{}' references a nonexistent term slug; skipping.".format("/".join(getattr(cat, "path", []) or [])))
+                            continue
+                        for title in term.get_titles(lang):
```
yodem, line 61: Is it a 4 layers nested for loop? even without this line there is 3 layers? Im not sure if there is something to do about it but it doesnt look good :(

stevekaplan123:
Skip log is mutable global state without a lock

stevekaplan123:
@yitzhakc
I updated the lock tests:

 Test 1 — test_concurrent_skips_and_summaries_lose_nothing. 8 threads record 200 skips each while the main thread repeatedly posts summaries; asserts all 1,600 show up across those summaries. (Note the sys.setswitchinterval(1e-6) fixture: Python switches threads only every 5ms by default, which is longer than the whole risky window, so without it the threads never interleave and the test passes even with the lock deleted). The fixture restores the original value in a finally — it's process-wide and would slow the rest of the suite.

Test 2 — test_lock_is_not_held_while_posting_to_slack. Guards the fact that the lock is released before the Slack call. Wrapping the whole function in one with _lock: would deadlock startup on any re-entrant path (threading.Lock isn't reentrant) and would reset away skips recorded during delivery. The test's fake Slack call checks _lock.acquire(blocking=False) succeeds and records a skip, then asserts that skip survives. Non-blocking on purpose: a blocking acquire would hang the suite instead of failing.

yitzhakc (approved):

PR #3446, opened 2026-06-25
fix(ci): correctly extract rollback app/chart versions in promotion PRs

yodem:
## Problem

The **Manual Promotion** workflow's auto-generated PRs always show the Rollback Reference as `app=`unknown`, chart=`unknown``. See [#3445](https://github.com/Sefaria/Sefaria-Project/pull/3445), where the table reads:

> | **Current on `preprod`** | app=`unknown`, chart=`unknown` |

This defeats the whole point of the section — there's no recorded version to roll back to.

## Root cause

`build/ci/generate-promotion-pr.sh` extracted the versions by piping the helmrelease YAML through `python3 -c "<regex>"` where the Python (and the regex inside it) lived in a **double-quoted bash string**. Bash mangled the regex's backslashes and quotes (`\s`, `[\"\'']`, etc.) before Python ever saw it, so the `re.search` never matched and both values fell back to `unknown`.

I reproduced this locally: the regexes match fine when handed to Python directly, but fail when routed through the exact `python3 -c "..."` bash-string path the script used.

## Fix

- Write the parser Python via a **single-quoted heredoc** so bash performs no substitution on the regex.
- Pass the YAML through an **environment variable** instead of escaping it on the command line.
- Extract both the app image tag and chart version in a single pass.
- The heredoc writes to a temp file (kept out of `$(...)`) so it stays portable to bash 3.2 as well as the bash 5.x on CI runners.

## Verification

Ran the real script end-to-end against `origin/preprod` with `gh` stubbed:

```
prev_app=v6.102.1-preprod.2
prev_chart=0.85.13-preprod.2
```

instead of `unknown`. `bash -n` passes on both bash 3.2 (macOS) and the CI bash. No behavior change to changelog/AI-summary/PR-creation logic.


yodem:
Superseded by #3382, which carries the same root-cause fix (the version extraction now uses a single-quoted heredoc instead of the bash-mangled `python3 -c "..."`, so the rollback reference no longer renders `app=unknown, chart=unknown`). Verified against the prod helmrelease that was live when #3447 was generated. Closing as redundant.

PR #3451, opened 2026-06-29
feat(pgvector): auto-update library_chunks on model dependency changes

nsantacruz:
## Summary

- Hooks into the existing `dependencies.py` subscriber system to keep the pgvector `library_chunks` table in sync when Sefaria model objects change (Index, Version, Topic, Category, Link, RefTopicLink, RefData)
- Dispatches targeted async Celery tasks so no model saves are blocked; silently skips if `vector_db` not in `settings.DATABASES`
- All field categories covered: index/version metadata, title renames, topic slug propagation, associated topics, linked refs, pagerank (3% change threshold), and text+embedding re-embedding

**New files:**
- `semantic_search/pgvector_dependencies.py` — thin subscriber callbacks, guard + `.delay()` only
- `semantic_search/tasks.py` — 13 Celery tasks with full business logic
- `sefaria/helper/vector/context.py` — shared `get_index_context`, `get_version_context`, `get_chunk_context` helpers (extracted from embed script)

**Modified files:**
- `semantic_search/semantic_text_chunk.py` — 11 new targeted update methods (bulk UPDATE, SQL array_replace for slugs, JSONB containment queries for `source_segment_refs`, raw SQL for title-prefix replacement in `ref`/`url`/`chunked_from_ref`)
- `sefaria/tracker.py` — hook `count_and_index` to dispatch `sync_text_and_embedding` on text save (same hook point as ES `SEARCH_INDEX_ON_SAVE`)
- `sefaria/model/topic.py` — fix `Topic.pkeys` to include `"slug"` so `attributeChange("slug")` fires correctly (latent bug: existing `marked_up_text_chunk` subscriber for slug changes was dead code)
- `sefaria/helper/vector/embed_library_to_pgvector.py` and `debug_single_ref.py` — refactored to import context helpers from `context.py`
- `sefaria/model/dependencies.py` — 14 new subscriber registrations under `if _PGVECTOR_DEPS:` guard

## Test plan

- [ ] `pytest sefaria/tests/ reader/tests/ -m "not deep and not failing"` passes (7 pre-existing failures in `TestVersionRenameAPI` excluded — confirmed present on unmodified `master`)
- [ ] Save an Index with a changed author in admin → verify `author_names`/`author_slugs` updated in `library_chunks` within seconds
- [ ] Change a Version title → verify `version_title` updated on all matching chunks
- [ ] Rename an Index → verify `index_title`, `ref`, `url`, `chunked_from_ref` all updated with new title prefix
- [ ] Change a Topic slug → verify `associated_topic_slugs` and `author_slugs` arrays updated in place
- [ ] Edit segment text via API → verify chunk `text` + `embedding` updated (Gemini re-called, no re-chunking)
- [ ] Add a ref-topic link → verify `associated_topic_names`/`associated_topic_slugs` updated on chunks containing that ref
- [ ] Confirm no Celery tasks fire when `vector_db` is absent from `settings.DATABASES`


sefaria/model/topic.py
```diff
@@ -132,7 +132,7 @@ class Topic(abst.SluggedAbstractMongoRecord, AbstractTitledObject):
         'person': 'PersonTopic',
         'author': 'AuthorTopic',
     }
-    pkeys = ["description"]
+    pkeys = ["description", "slug"]
```
yonadavGit, line 135: i take it it's an important though unrelated fix?

semantic_search/semantic_text_chunk.py
```diff
@@ -0,0 +1,133 @@
...
+    def filter(self, **kwargs) -> list[DjangoSemanticTextChunk]:
+        return list(DjangoSemanticTextChunk.objects.filter(**kwargs))
+
+    # --- Targeted field-level update methods ---
+
+    def update_index_metadata(self, index_title: str, fields: dict) -> int:
+        """Bulk UPDATE metadata fields for all chunks of an index. Returns row count."""
+        return DjangoSemanticTextChunk.objects.filter(index_title=index_title).update(**fields)
+
+    def update_index_title(self, old_title: str, new_title: str) -> int:
```
yonadavGit: im not sure that the hard coded query and string replacement is the ideal way to go.. 
also i think there's a hudden assumption here that the URL and the Title are the same.
is there a way to go about it without in-query string slicing?

semantic_search/semantic_text_chunk.py
```diff
@@ -0,0 +1,133 @@
...
+        ))
+
+    def get_chunks_containing_ref(self, index_title: str, ref_str: str) -> list[DjangoSemanticTextChunk]:
+        """Find chunks whose source_segment_refs (in chunker_metadata JSONB) contain the given ref."""
+        return list(DjangoSemanticTextChunk.objects.filter(
+            index_title=index_title,
+            chunker_metadata__source_segment_refs__contains=[ref_str],
+        ))
+
+    def replace_author_slug(self, old_slug: str, new_slug: str) -> int:
```
yonadavGit: i think it's more straightforward then the title-change query, but can we de-querify the logic?

semantic_search/semantic_text_chunk.py
```diff
@@ -0,0 +1,133 @@
...
+        """Replace old_slug with new_slug in the author_slugs array for all matching chunks."""
+        with connections['vector_db'].cursor() as cur:
+            cur.execute(
+                "UPDATE library_chunks SET author_slugs = array_replace(author_slugs, %s, %s) WHERE %s = ANY(author_slugs)",
+                [old_slug, new_slug, old_slug],
+            )
+            return cur.rowcount
+
+    def replace_associated_topic_slug(self, old_slug: str, new_slug: str) -> int:
```
yonadavGit: same as ibid.

semantic_search/tasks.py
```diff
@@ -0,0 +1,415 @@
+"""
```
yonadavGit, line 1: how would you describe in words the different logic implemented here v.s in semantic_search/semantic_text_chunk.py?

sefaria/model/topic.py
nsantacruz: Claude says: process_topic_slug_change_in_pgvector                                           
  (semantic_search/pgvector_dependencies.py:79) listens for the attributeChange event with          
  attr="slug" and dispatches update_topic_slug.delay(old_slug, new_slug) to fix up the              
  library_chunks Postgres table. Without "slug" in pkeys, Topic.save() would never emit that event, 
  and slug renames (e.g. via Topic.set_slug) would silently desync the pgvector store from Mongo.

Noah says: I agree

sefaria/model/dependencies.py
```diff
@@ -4,6 +4,13 @@
 
 from . import abstract, link, note, history, schema, text, layer, version_state, timeperiod, garden, notification, collection, library, category, ref_data, user_profile, manuscript, topic, place, marked_up_text_chunk
 
+_pgv = None
+try:
+    from semantic_search import pgvector_dependencies as _pgv
+except ImportError:
+    pass
+_PGVECTOR_DEPS = _pgv is not None
```
nsantacruz: unnecessary line

semantic_search/models.py
```diff
@@ -1,10 +1,127 @@
...
+            )
+            return cur.rowcount
+
+    def update_version_fields(self, index_title: str, version_title: str, fields: dict) -> int:
+        """Bulk UPDATE fields for all chunks of a specific index + version."""
+        return self.filter(
+            index_title=index_title, version_title=version_title
+        ).update(**fields)
+
+    def update_version_title(self, index_title: str, old_vtitle: str, new_vtitle: str) -> int:
```
nsantacruz, line 104: why is this function needed alongside update_version_fields? it seems like a special case.

semantic_search/pgvector_dependencies.py
```diff
@@ -0,0 +1,174 @@
...
+
+Each callback is thin: guard-check + dispatch a Celery task.
+All business logic lives in semantic_search.tasks.
+"""
+import structlog
+
+logger = structlog.get_logger(__name__)
+
+
+def _enabled() -> bool:
```
nsantacruz: the gate in dependencies.py should be this and not checking if you can import pgvector_dependencies.py. then remove all the if not _enabled() in this file since the gate in dependencies.py guards these functions.

semantic_search/pgvector_dependencies.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Index
+# ---------------------------------------------------------------------------
+
+def process_index_save_in_pgvector(index, **kwargs) -> None:
+    if not _enabled():
+        return
+    from semantic_search.tasks import update_index_metadata
```
nsantacruz: why are all these imports inside functions, move them to the top of the file.

semantic_search/tasks.py
nsantacruz: I actually removed semantic_text_chunk.py and put it in semantic_search/models.py (although it's pretty much the same, just a refactor)

according to claude:
models.py = "how to read/write chunks," content-agnostic; tasks.py = "what happened in  
  Sefaria and which chunk fields that implies need refreshing," domain-aware and stateful about the 
  surrounding Sefaria object graph. 

This sounds right to me

semantic_search/tasks.py
```diff
@@ -0,0 +1,412 @@
...
+from semantic_search.embedder import GeminiEmbedder
+from semantic_search.models import SemanticTextChunk
+
+logger = structlog.get_logger(__name__)
+
+# ---------------------------------------------------------------------------
+# Index-level tasks
+# ---------------------------------------------------------------------------
+
+@app.task(name="pgvector.update_index_metadata")
```
nsantacruz: change all task names from `pgvector` to `semantic_search` (e.g. semantic_search.update_index_metadata)

semantic_search/tasks.py
```diff
@@ -0,0 +1,412 @@
...
+@app.task(name="pgvector.update_version_attributes")
+def update_version_attributes(index_title: str, vtitle: str) -> None:
+    ver = Version().load({"title": index_title, "versionTitle": vtitle})
+    if not ver:
+        logger.warning(
+            "pgvector.update_version_attributes: version not found",
+            index_title=index_title, vtitle=vtitle,
+        )
+        return
+    fields = {
```
nsantacruz: shouldn't these fields be extracted in `sefaria.helper.vector.context`?

sefaria/model/topic.py
yonadavGit: guess i'll have no choice but to agree as well, even though i guess it's not trivial that "**event with
attr="slug"**" is emitted only for **pkeys field changes**.

sefaria/pagesheetrank.py
```diff
@@ -301,7 +301,19 @@ def calculate_pagerank():
     return pagerank_dict
 
 
+# update_pagesheetrank() overwrites pagesheetrank on essentially every ref_data record via a raw
```
yonadavGit, line 304: this is a significant bulk operation. not always a significant bulk write but ncecessarily a massive bulk read if i understand it correctly. has that been teseted? how long does it take to read the entire metadata of the vector db?

sefaria/pagesheetrank.py
```diff
@@ -301,7 +301,19 @@ def calculate_pagerank():
...
+# bulk_write, bypassing RefData.save() and the "save" notification it would otherwise emit (see
+# dependencies.py). Downstream listeners (e.g. pgvector's process_ref_data_save_in_pgvector) rely on
+# that notification to keep chunk pagerank in sync. Rather than skip the sync, we diff against the
+# ref_data values already in mongo and manually emit "save" notifications, but only for refs whose
+# pagesheetrank moved enough to matter -- otherwise we'd enqueue a sync task for nearly every segment
+# in the library on every run.
+PAGESHEETRANK_CHANGE_THRESHOLD = 0.03
```
yonadavGit, line 311: reading the code below seems like this threshold is a ratio, perhaps better to have the word "ratio" in the param name

semantic_search/pgvector_dependencies.py
```diff
@@ -0,0 +1,161 @@
+"""
```
yonadavGit, line 1: what  if text is added in a way that changes ref numbering / offsets? i fear we don't fully address this case. we'll have to somehow change the Ref column per affected row or re-embed a whole section.

semantic_search/pgvector_dependencies.py
```diff
@@ -0,0 +1,161 @@
...
+Subscriber callbacks for the sefaria.model.dependencies event system.
```
yonadavGit, line 2: If I understand this correctly, the file makes an important distinction between ordinary metadata fields and metadata columns that effectively form part of a row’s lookup key in Postgres.

When one of these de facto key columns changes, we need to preserve its previous value so that we can still locate the relevant row and update it.

Assuming that interpretation is correct, perhaps we could make this distinction more explicit by defining the de facto key columns in one place and handling them separately. This may also help deduplicate some of the logic, but the main benefit would be making the intent and behavior here much clearer.

semantic_search/pgvector_dependencies.py
```diff
@@ -0,0 +1,161 @@
...
+
```
yonadavGit, line 3: Just flagging that some of the logic here seems tightly coupled to the current Sefaria/Mongo data model. It may not be obvious to someone changing that model or its behavior in the future that this logic would likely need to be updated as well.

Not necessarily requesting a change here, just flagging

PR #3464, opened 2026-06-30
ES reindex: fix connection-hang root causes, parallelize into scatter-gather, integrate entity search [sc-45337]

yodem:
## Summary

The weekly full Elasticsearch reindex CronJob took ~26h and intermittently died with a "connection error," abandoning the entire run. This PR fixes the **four root causes** (not masks them), re-architects the job into an isolated, parallel **scatter-gather** so no section can stall another, and integrates cleanly with the Multi-Entity Search Results work now on master.

Story: **sc-45337** (Team Platform · Devops).

> **Consolidation note:** this PR now carries everything. #3654 was a temporary integration branch cut from this one so this PR stayed intact as a fallback while the master merge was validated. That merge and its follow-up fix are folded in here and #3654 is closed — this is the single PR for this work.

## Investigation (production logs + code)

Decoded the last completed run (~26h): ~14.5h real work + **~11.3h of silent multi-hour stalls**. Log hours 07/08/10–14 were entirely absent; the 6800→6900 stall was 2h17m ≈ Linux TCP keepalive (7200s); silent gaps had **zero ES traffic** → the process was blocked in Mongo reads. Bottleneck is Mongo/Python-side, not ES ingest.

## Root causes fixed

1. **🔴 Mongo dead-socket hangs (primary killer)** — `MongoClient` set no `socketTimeoutMS` (pymongo default = infinite), so a dead Mongo TCP socket blocked reads ~2h until OS keepalive. Now `socketTimeoutMS=300s` / `connectTimeoutMS=20s` / `serverSelectionTimeoutMS=60s` (global client; a generous ceiling, safe for the online path).
2. **🔴 Un-hardened ES client** — the alias swap and all 67k sheet writes used a client with no retries. Hardened (`request_timeout`/`retry_on_timeout`/`max_retries`/`http_compress`); bulk flush now retries `429`s.
3. **🟠 Sheet regression** — commit `12e9a845f` made optional sheet fields hard-required, silently dropping **~48,053 of 67,389 public sheets (71%)** from search. Restored graceful fallbacks (only `owner` required) → sheets searchable again.
4. **🟠 Single 26h failure domain** — split into a scatter-gather: `reindex_init` (pagesheetrank + create indexes + bulk-load settings) → K8s **Indexed Job** of N=8 `--mode shard` pods (deterministic size-aware sharding, per-shard `backoffLimitPerIndex` retry) → `reindex_finalize` (restore settings, **doc-count sanity gate**, atomic alias swap). A stall in one shard no longer holds the others hostage; the alias only swaps onto a verified-complete index. The orchestrator pod holds no ES/Mongo connection during the long phase.

## Integration with Multi-Entity Search Results

Master's `45bf5d34c` added `/api/entity-search` and three new ES index types — `topic`, `book`, `category` — reindexed via `index_entities()`. Merged in here; three files conflicted and were resolved as union merges keeping both sides:

| File | Resolution |
|---|---|
| `sefaria/search.py` | Master's entity-search machinery (mapping builders, `index_entities`, topic/author/book helpers) alongside this branch's phased `reindex_init`/`reindex_index_shard`/`reindex_finalize` split, bulk-load tuning, and hardened `index_sheet` fallbacks |
| `helm-chart/sefaria/values.yaml` | This branch's `ISOLATE_SEARCH_INDEXES` + master's `SEARCH_INDEX_NAME_TOPIC/BOOK/CATEGORY` |
| `helm-chart/.../configmap/local-settings.yaml` | Per-cauldron index-isolation templating for text/sheet + master's entity emission lines |

### The seam between the two refactors — found and fixed

The union merge was correct as a merge, but the two refactors had never met. `reindex_index_shard()` branched only on `'text'` and `'sheet'`, and `index_entities()` routes all three entity types through `index_all_of_type()` → `reindex_index_shard()`. Both the weekly production reindex (`index_all`) and `scripts/reindex_entity_search.py` reach it.

**Failure mode:** `reindex_init` *succeeds* for entity types, creating an empty index with bulk-load settings; then `reindex_index_shard` raises `ValueError`. `reindex_finalize` never runs, so:

1. Entity indexes go **stale, not empty** — the alias keeps serving the old index and `/api/entity-search` degrades quietly.
2. Each weekly run leaks an orphaned empty `*-a`/`*-b` index still carrying bulk-load settings.
3. **The job reports success.** No alert fires.

**Why CI did not catch it:** `reindex_pipeline_test.py` and `reindex_cronjob_test.py` `monkeypatch` `sefaria.search.reindex_index_shard` — the tests stub out the exact function that raises. The suite was green with the bug live.

**Fix** (`6b2de5392`): `reindex_index_shard` dispatches entity types to `index_topics`/`index_books`/`index_categories`, mirroring the dispatch that already exists and is correct in `index_all_of_type_by_index_name`. Entity corpora are small relative to text, so they are indexed single-shot and never fanned out — under a sharded invocation only shard 0 does the work, so N pods don't each rebuild the corpus. The `else: raise ValueError` stays, so a genuinely unknown type still fails loudly. Entity indexes also *gain* the dual-index a/b cutover gate they did not have on master.

`index_entities` keeps its per-type isolation but now collects failures and raises a summary once every type has been attempted.

> ⚠️ **Deliberate change to cronjob exit semantics:** an entity-type failure used to exit 0; it now exits **non-zero**, after text and sheet have already swapped their aliases. That is the point — the silence was the bug. The durable text/sheet cutover is committed by then, so a non-zero exit costs a re-run of entity indexing, not of the long text reindex. Expect a weekly job that can now go red where it previously went green quietly.

## Also included
- Bulk-load tuning (`refresh_interval:-1` / `replicas:0` during load, restore + `_refresh` in finalize).
- Sanity gate **fails closed** on an unreadable current-index count.
- Helm: orchestrator RBAC + cronjob wiring that propagates ES/Mongo env, secrets, and the `local_settings.py` volume to shard pods.

## Verification

- **112 passed** across `sefaria/tests/search_test.py`, `search_dependencies_test.py` and the four reindex suites (106 before this fix, plus 6 new tests).
- **51 passed** across `sefaria/helper/tests/search_test.py` + `reader/tests/entity_search_api_test.py`, unchanged.
- The six new tests were confirmed to **fail before** the fix. The one that matters is in `reindex_pipeline_test.py`: it exercises `index_all_of_type` for an entity type **without** stubbing `reindex_index_shard`, and failed with `ValueError: Unknown index type: topic`. Without that test this class of break can recur unnoticed.
- Test-count tripwire on the auto-merged test files: no duplicate test names, neither side's tests silently dropped.
- `helm template` with `cronJobs.reindexElasticSearch.enabled=true` renders both sides: orchestrator ServiceAccount/Role/RoleBinding, `SHARD_COUNT`, `reindex_orchestrator.py`, alongside master's entity index settings.
- Client separation holds: `entity_search()` uses the plain ES client; `_indexer_es_client` appears only in the two bulk-indexing call sites.

Note: master moved `sefaria/pytest.ini` → root `pytest.ini`; resolved to root-only. Running pytest against the wrong rootdir **exits 0** on a usage error, so a mis-rooted run can look like a pass. `requirements.txt` adds `kubernetes==31.*` — needs installing in any local venv running the reindex suites.

## Safety / rollout
- **CronJob ships disabled** (`reindexElasticSearch.enabled: false`) — merging changes nothing in prod.
- GKE cluster is v1.35.5 (Indexed-Job per-index retry fields are GA).

## ⚠️ Before enabling in prod (must do)
- [ ] Staging dry-run on `-debug` indexes end-to-end: `--mode init` → `--mode shard` for `shard-index 0..N-1` → `--mode finalize`; confirm shards index **disjoint** groups and `text-debug` doc count ≈ a monolith run.
- [ ] Validate `_index_size_map` resolves live `VersionState` counts; if it excepts it degrades to count-balanced (safe, just less head-balancing). Log a sample size-map.
- [ ] Confirm shard pods connect (ES secret + mounted `local_settings.py`) and that the orchestrator **exits non-zero and leaves the alias untouched** on an injected shard failure.
- [ ] End-to-end entity check on a cauldron with `ISOLATE_SEARCH_INDEXES`: run `scripts/reindex_entity_search.py --type all`, confirm the `topic`/`book`/`category` aliases each point at a freshly built index with a non-zero doc count and no orphaned empty `*-a`/`*-b` index is left behind.
- [ ] Known limitation (gate-mitigated): cross-pod shard determinism assumes stable Mongo reads during the run; sub-10% group loss could slip past the 90% gate — ticket as a follow-up.


yodem:
### Heads-up on CI: PyTest was never actually run on the master merge

Worth flagging before review. The `Continuous Testing: PyTest` job on this branch's last two pushes:

| commit | PyTest |
|---|---|
| `f431492d8` (Jul, pre-merge) | success — last real green |
| `360bf42a5` (old #3464 head) | **skipped** |
| `29616a8e1` (master merge, on #3654) | **skipped** — draft PRs skip PyTest |
| `6b2de5392` (this push) | failure |

So #3654's "CI green" never included Python tests. The first time PyTest actually ran against the merged code is this push, and it reports **4 failed, 4 errors, 1057 passed**: `category_test.py::Test_Category_Editor` (2), `search_dependencies_test.py` `TestBookHooks`/`TestCategoryHooks` (2 + 2 errors), `summaries_test.py::Test_Toc` (2 errors).

**These are not from the entity-type fix.** Evidence:

1. None of the failing tests reference `index_entities`, `reindex_index_shard`, or `index_all_of_type` — the commit's diff cannot reach them.
2. Run locally at `6b2de5392` and at its parent `29616a8e1` (i.e. without the fix): **identical results**, 28 passed / 2 failed both times.
3. The 2 tests that do fail locally (`Test_Toc::test_index_title_change`, `test_new_index_title_change`) fail the same way on plain `origin/master`, so they're pre-existing environmental.
4. All six new tests pass in CI.

What I can't yet say is whether the CI-specific failures come from the master merge itself or from CI infra/state — the log shows a cascade signature (`Could not get TOC object for Category Test ES Cascade Category` → TOC `IndexError` → book/category hook failures) that looks like leaked state from one root failure. A `--failed` rerun was inconclusive (its sandbox had been torn down, so PyTest produced no test output); a full clean rerun is in flight.

Flagging rather than sitting on it: **this PR should not merge on the assumption that CI was ever green on the merged code.**

yodem:
### Resolved — clean CI rerun is green

Correcting my previous comment. A full clean rerun of the same commit (`6b2de5392`) passes:

```
1063 passed, 27 skipped, 66 deselected, 9 xfailed in 29:37
```

**1063 vs the failing run's 1057 is exactly the six new tests.** `Continuous Testing: PyTest`, `Jest`, and `continuous-branch-protection` all green; PR is `MERGEABLE` with 14 success / 5 skipped / 1 neutral and nothing failing.

So the earlier `4 failed, 4 errors` was CI state flakiness, not a real regression — consistent with the cascade signature in that log (one root failure leaking `Test ES Cascade Category` into the TOC, then taking the book/category hook tests and `Test_Toc` down with it). The first comment's evidence that the failures weren't caused by the entity-type fix still stands; what I got wrong was leaving open the possibility that the master merge caused them. It didn't.

**One finding from that comment does still stand and is worth keeping:** PyTest was genuinely *skipped* on `360bf42a5` and on #3654's `29616a8e1` (draft PRs skip it), so this push is the first time Python tests actually ran against the merged code. They now pass. Worth knowing that a draft PR's green checks on this repo do not include PyTest.

yodem:
## Why the weekly reindex has not completed in 221 days — investigation, decisions, and rejected options

Three defects, found by tracing a real 12-hour failure rather than reasoning from the symptom. All three are **independent of this PR's own changes** — they reproduce on `sefariastaging`, which tracks master — but each blocked validating it.

### The failure chain

```
File "/app/sefaria/pagesheetrank.py", line 314, in update_pagesheetrank
    result = db.ref_data.bulk_write([
pymongo.errors.NetworkTimeout: mongo.default:27017: timed out
    (configured timeouts: socketTimeoutMS: 300000.0ms)
```

| # | Defect | Evidence | Fix |
|---|---|---|---|
| 1 | reindex cronjob hardcoded a **prod-only** `SEARCH_HOST` | `elasticsearch-1b-es-http` does not resolve in dev (`getent` exit 2, `curl` exit 6); the correct service answers from the same pod | `db14722b3` |
| 2 | **`ref_data.ref` is unindexed** | `explain()`: `COLLSCAN`, **1,273,527 docsExamined, 434 ms for one lookup** | `908335278` |
| 3 | reindex init is slow enough to span a 12h window | `best_time_period()` called **10,090,470×** for **6,601** distinct answers | `3c95b84a2` |

**Defect 2 is the one that made completion impossible.** `update_pagesheetrank()` ends with one `bulk_write` holding an `UpdateOne({"ref": tref}, upsert=True)` per tref. `ref_data` has 1,273,527 documents and only the default `_id_` index — the collection is **absent from `ensure_indices()` entirely**, so this index is missing in *every* environment, production included. At 434 ms per upsert that is ~1.6 trillion document examinations (~6.4 days); a single pymongo batch needs ~12.1h against a 300 s socket timeout, so `NetworkTimeout` was guaranteed. The observed pod ran **12h00m29s** and exited 1.

Why it hid so long: the job dies in `init`, **before** creating any ES index or spawning shards. Investigations starting from "search results are stale" looked at Elasticsearch and correctly found it healthy. Confirmed on the failed run — no `esreindexv2*` indices existed and zero shard jobs spawned.

### Two claims the investigation overturned

Recording these because both were plausible and both were wrong:

- **Offset pagination on links is not a bottleneck.** Measured: `skip(3,000,000)` costs 1.17 s, so the entire 51-page link walk wastes ~50 s. Streaming the documents costs more than skipping them.
- **The pure-Python PageRank iteration is not why it failed.** The log shows it *completing* in 22 iterations (L1 norm 1.359 → 0.0000485). Slow, not fatal.

### Options considered for the performance work

**Chosen — optimize in-process.** Memoize the era lookup; replace the quadratic `create_web` with flat COO arrays; vectorize the power iteration with a `bincount` matvec. One file, no infrastructure change, directly verifiable.

**Rejected — run pagerank and sheetrank concurrently.** They are genuinely independent, so it is safe in principle. But the payoff is capped at the shorter phase while requiring both the link graph and sheet dict in memory at once, and the orchestrator pod has a **6 Gi limit and no CPU request**. Trades a real OOM risk for a small constant factor. Revisit only if profiling shows sheetrank dominates.

**Rejected — shard init across the existing Indexed Job infrastructure.** This looked attractive since the scatter-gather already exists. It would **produce wrong answers**: the existing scheme partitions **by title**, valid for ES indexing because titles never overlap, but a ref's PageRank depends on in-links from refs in other shards' titles — so a by-title split yields *incorrect ranks, not merely slower-but-correct ones*. The infra also has **no result-return channel** (shards write straight to ES), so sharding the sheet walk needs entirely new reduce plumbing. The only drop-in piece is the write phase, which the index already made fast.

### Deliberate decision: preserve two quirks exactly

The adjacency build repeats each edge `int(round(weight))` times — weight 0.4 **disappears**, 0.6 counts once — while `number_out_links` accumulates the **raw float**. The transition matrix is `round(w(k→j)) / out_total(k)`: rounded numerator, unrounded denominator. This looks accidental, but search ranking is calibrated on it, so changing it is a product decision, not a refactor. The vectorized code reproduces it exactly, and there are tests pinning it.

`class web` / `create_web` / `step` are retained — off the production path — as the reference implementation the equivalence tests compare against. numpy only: scipy is not in the image and the job already does a fragile runtime `pip install`.

### Verification

| Graph | Result |
|---|---|
| simple / quantized / dangling / chain | **max\|diff\| = 0.00e+00** — bit-identical to the old implementation |
| randomized, 1,200 nodes / 3,518 edges / 77 iterations | max\|diff\| = 1.73e-17, ranking order preserved |
| speed on that graph | 2.32 s → 0.006 s (**413×**), and the gap widens with graph size |

Bit-identical output across *runs* was never a property to protect — `init_pagerank_graph` breaks same-era ties with an **unseeded `random.choice`**, so runs already differ.

### ⚠️ Deploy note

`ensure_indices()` only covers fresh databases. Existing databases — **including production** — need the index created explicitly, or the first run after deploy still scans:

```js
db.ref_data.create_index("ref")
```

A `unique` index is **not** possible today: `ref_data` currently holds **1,385 duplicated refs** (2,602 extra docs, up to 9 copies of one ref).

### Not addressed here — each worth its own ticket

- Unseeded `random.choice` makes pagerank non-reproducible run to run.
- Bare `except: pass` in the link loop and sheet-source counting silently drops malformed records with no count or log.
- The 1,385 duplicate `ref_data` refs: `UpdateOne` touches only the first match, so duplicates keep stale values forever — and they block the unique index a safely parallelized writer would need.
- `update_pagesheetrank()` is fail-loud in the orchestrator but warn-and-continue in the legacy monolith path (*"PageSheetRank update failed, continuing anyway"*), which can ship a reindex built on stale ranks.
- No test coverage of the `update_pagesheetrank()` call site in `reindex_orchestrator.main()`.
- Unguarded `math.log()` on the final merge path would crash after all the expensive work if any pagerank value were ≤ 0.

Full write-up with measurements is in the engineering wiki.

BrendanGalloway (dismissed):

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
     return v / numpy.sum(v)
 
 
+def _build_pagerank_arrays(g):
+    """Flatten the graph into numpy arrays for a vectorized power iteration.
```
nsantacruz, line 100: add comment explaining g format which is [(ref1, {ref2: weight1, ref3: weight2}), (ref4, {ref5: weight3})]. In this example, ref1 is connected to ref2 with weight weight1 and connected to ref3 with weight2 (connections are directional). ref4 is connected to ref5 with weight3.

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+    """
+    keys = [x[0] for x in g]
+    n = len(keys)
+    node2index = {r: i for i, r in enumerate(keys)}
+
+    out = numpy.zeros(n, dtype=numpy.float64)
+    dangling = numpy.ones(n, dtype=bool)
+    rows, cols, data = [], [], []
+
+    for r, links in g:
```
nsantacruz: change r to ref1

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+        r_ind = node2index[r]
+        for r_temp, count in links.items():
```
nsantacruz: change r_temp to ref2 and count to weight

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+            j = node2index[r_temp]
+            # legacy pops j from dangling_pages the first time it emits an edge,
```
nsantacruz: remove comment that doesn't seem to apply anymore

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+            # before adding the weight -- even when the weight rounds away
+            dangling[j] = False
+            out[j] += count
+            repeats = int(round(count))
```
nsantacruz: this variable is confusing. rename to rounded_weight

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+            if repeats:
```
nsantacruz: change to rounded_weight > 0 for more clarity

sefaria/pagesheetrank.py
```diff
@@ -96,21 +96,90 @@ def step(w, p, s=0.85):
...
+    Reproduces create_web()'s semantics exactly, including two quirks that are
+    load-bearing for search ranking and must NOT be "fixed" here:
+
+      * an edge is counted int(round(weight)) times, so weight 0.4 disappears
+        and 0.6 counts once -- edges rounding to 0 are simply not stored;
+      * number_out_links accumulates the RAW float weight, so the transition
+        matrix is round(w(k->j)) / out_total(k): rounded numerator over an
+        unrounded denominator.
+
+    Returns (n, rows, cols, data, out, dangling, keys) where rows/cols/data are
```
nsantacruz: change data to weights

sefaria/pagesheetrank.py
yodem: Added — with one correction, because the direction in your example is the reverse of what the code does.

In `g[ref1] = {ref2: weight}`, **ref2 links *into* ref1** and confers rank on it. Three things say so:

- `create_web` assigns those inner refs to `w.in_links[r_ind]` — they were always called *in-links* — and accrues out-degree on the inner one (`w.number_out_links[j] += count`).
- `init_pagerank_graph` builds it as `graph[older_ref][newer_ref]`, so the later commentary votes for the earlier text it quotes.
- Which is the direction we want for search: Rashi confers rank on Genesis, not the other way round.

The docstring now spells that out with a worked Genesis/Rashi example and names the consequences (out-degree accrues on the inner ref; an outer ref with an empty dict is dangling). I'm confident about the direction — less so about my prose, so worth a read.

sefaria/pagesheetrank.py
yodem: Done — `r` → `ref1`, and `r_ind` → `ref1_index` so the pair reads together.

sefaria/pagesheetrank.py
yodem: Done — `r_temp` → `ref2`, `count` → `weight`, and `j` → `ref2_index` for the same reason.

sefaria/pagesheetrank.py
yodem: Reworded rather than removed — the phrasing was bad but the comment is load-bearing.

It's there to explain why `dangling[ref2_index] = False` sits *outside* the `rounded_weight > 0` guard: a node stops being dangling the moment it appears as a source, even when its weight rounds to 0 and no edge gets stored. Move that line inside the guard and the ranks change. `test_dangling_detection_matches_legacy` is what catches it — the `DANGLING` fixture has `c: {"a": 0.4}`, where the edge rounds away but `a` still isn't dangling.

You were right about what was actually wrong with it: it narrated `create_web`'s `dangling_pages` dict, which doesn't exist in this function. It now states the invariant a future editor could break, instead of the history.

sefaria/pagesheetrank.py
yodem: Done — `repeats` → `rounded_weight`.

sefaria/pagesheetrank.py
yodem: Done — and it turns out to be more than cosmetic.

`if repeats:` also passed for a *negative* rounded weight, and legacy `create_web` stores nothing in that case (`[x] * -2 == []`). So `> 0` actually matches `create_web` where the truthiness check didn't. It can't arise today — weights only accumulate via `graph[str1][str2] += weight` with `weight` defaulting to 1.0 — but the form you asked for is the faithful one, not just the clearer one.

sefaria/pagesheetrank.py
yodem: Done — `data` → `rounded_weights` (docstring, the local, and the `_pagerank_step` parameter).

Went with `rounded_weights` over plain `weights` to match the `rounded_weight` rename two lines down, and because `rounded_weights / out[cols]` sitting side by side is exactly the trap the docstring warns about: a rounded numerator over an unrounded denominator. Happy to shorten it if you'd rather.

yodem:
Pushed the review responses, plus two commits that aren't answers to a comment — flagging them so they don't look like drive-by changes:

**`f8e4de27d` — `pagerank()` crashed on a single-node graph.** `numpy.squeeze` on the final 1×1 column vector returns a 0-d array, which `list()` can't iterate (`TypeError: iteration over a 0-d array`). Swapped to `reshape(-1)`, which is bit-identical for every n ≥ 2 and correct at n = 1; verified across n = 2…11, 50, 400, 3000 comparing `float.hex()`. `pagerank()` already had `if n == 0: return {}`, so this was the one degenerate size left uncovered. Not a regression from the vectorized rewrite — the legacy `create_web`/`step` path fails the same way. Test added.

**`3d26fe122` — preprod needed `allowSharedIndex: true`.** `envs/preprod/helmrelease.yaml` had `false`, but preprod uses the plain shared index names exactly like prod (it sets neither `ISOLATE_SEARCH_INDEXES` nor any `SEARCH_INDEX_NAME_*` override), so the guard would reject it. This PR's own test docstring already says "prod/preprod legitimately swap the shared aliases via `allowSharedIndex: true`" — the config contradicted it.

It's latent rather than live: preprod has `reindexElasticSearch.enabled: false` and the CronJob is gated on that, so nothing renders there today. It would fail the first time anyone turns it on. Prod and staging were already correct.

While checking that: `build/ci/preprod-values.yaml` also carries this flag, but it has no references in any current workflow and its only consumer (`build/ci/production-helm-deploy.sh`) is never invoked. It happens to hold the right value. Worth deleting separately so it stops looking authoritative.

Test status on `pagesheetrank`: 19 passed (the existing 18, which compare the vectorized path against `create_web`/`step` end-to-end, plus the new single-node case).

yitzhakc (approved):

Issue #3, opened 2011-10-19
Edit Text: line by line


blockspeiser:
After clicking "edit text" let each line be edited by clicking.
- Track changes (and display in header, e.g., "3 verses changed") 
- send changes as a batch on "Save" click
- account for texts of mixed source

PR #3470, opened 2026-06-30
ci: remove 20-min job deadline that killed healthy PyTest runs [sc-45322]

yodem:
## Problem

The **Continuous Testing: PyTest** job had been failing on every PR, for multiple compounding reasons. Example: [run 28442916591](https://github.com/Sefaria/Sefaria-Project/actions/runs/28442916591/job/84286668519) — **every test passed**, the suite reached **86%**, then the K8s Job was killed at exactly **1200s** with `DeadlineExceeded` ("Job was active longer than specified deadline").

## Root cause

PR #3455 (sc-45322) correctly added per-test hang protection (`pytest-timeout`) **and** a job-level `activeDeadlineSeconds: 1200` (20 min) as a "hard backstop". But the healthy full suite runs **~23–26 min single-process** against a networked Mongo sandbox — *longer than the 20-min backstop*. So the job deadline sat **below normal runtime** and guillotined healthy runs at ~86%.

Removing that deadline surfaced further issues, fixed one at a time as each was uncovered: a broken cluster-auth mode, pgvector/vector_db test setup errors, a stale-Postgres-DB error, a job that always reported success even on real failure, and — after those were fixed and the suite could finally run long enough to hit it — a ~10-minute-per-test hang in version-mutation tests caused by a stale cached `USE_VARNISH=True` from an import-time settings read.

## Fix

- **`createJobFromRollout.sh`**: drop `activeDeadlineSeconds=1200` (the mis-sized job-level deadline). Keep per-test `pytest-timeout` (now `--timeout=600`, tuned up from 120 after real non-hang tests needed more headroom) as the real hang bound. Add `--reuse-db` to avoid `DuplicateDatabase` errors against the shared sandbox's stale `test_sefaria_auth`. `backoffLimit` went `2 → 1 → 0`: `1` was tried first (one retry for infra-level pod loss, e.g. preemptible node eviction) but caused active harm — `waitForCIJob.bash` selects pods by label, not name, so a retry pod under the same label made `kubectl logs -l` return combined/duplicate output from both pods (duplicate PASSED/FAILED/ERROR lines for the same tests) and eventually hit the retry pod mid-`ContainerCreating` (see [run 28873987240](https://github.com/Sefaria/Sefaria-Project/actions/runs/28873987240/job/85797858085?pr=3470)). The wait script also isn't retry-aware — it reports the job failed on the first pod's `status.failed`, before any retry finishes — so the retry wasn't buying a real budget, just corrupted logs. Settled on `0`: no Job-level retry at all.
- **`continuous.yaml`**: the *Wait For Job To Finish* step `timeout-minutes` goes **60 → 45** (not 30→45 as originally stated here — corrected after review). This is the outer wall-clock backstop for the whole job (build + full suite, ~23–26 min today), replacing the broken `activeDeadlineSeconds`. It's a different layer from the per-test timeout: per-test timeout kills one hung test after 10 min with a traceback; this step-level timeout is what catches a systemic hang (like the Varnish issue below) so the whole job doesn't run forever. Also upgrades `get-gke-credentials@v2 → v3` on all 3 cluster-auth steps and drops the `use_auth_provider: true` input — v2's auth-provider mode writes a legacy kubeconfig entry that modern Helm/kubectl reject (upstream closed the issue as "not planned"); v3 doesn't need the flag.
- **`waitForCIJob.bash`**: rewritten to actually detect job success/failure (`status.succeeded`/`status.failed`) instead of always exiting 0, and to stream only new log lines instead of re-printing the whole log every 30s. Also caps the `kubectl unreachable` retry branch at 10 consecutive attempts (5 min) instead of retrying forever — a genuinely down cluster now fails loud with a clear message instead of silently burning the whole 45-min step timeout.
- **`build/ci/cleanup_test_data.py`** (new): removes stale `auto_linking` dummy fixtures ("Many to One on Genesis", "One to One on Genesis" — verified these are the only two the test suite creates) from the shared sandbox Mongo before each run, so a run killed mid-test doesn't leave dirty state for the next one.
- **`sefaria/conftest.py`**: pops the unreachable `vector_db` (pgvector) alias from `DATABASES` so pytest-django doesn't try to create a test DB for it; `embed_library_to_pgvector_test.py` is skip-marked until pgvector is provisioned in the CI sandbox. Also force-sets `USE_VARNISH = False` on `sefaria.settings` and re-patches any already-imported module that cached the old value at import time — several modules (`tracker.py`, `model/text.py`, `model/version_state.py`) read `USE_VARNISH` at import time before the CI settings override could apply, so `invalidate_linked()` was shelling out to a nonexistent `varnishadm` and blocking each version-mutation test for ~10 min until pytest-timeout killed it.
- **`sefaria/settings.py`**: no longer touched by this PR — an earlier commit here carried a duplicate `django.contrib.postgres` fix that became a stray merge artifact once `origin/master` fixed its own copy independently (#3479); reverted to match master exactly.

Net layering: **per-test hang kill (600s) → outer wall-clock bound (45m)** — two purposeful layers. A hung test still fails within ~10 min with a traceback; healthy runs now complete instead of being killed at 86%.

**Results:** before, 554 passed / 154 errors, killed at 86% or skipped entirely. After (final green run): **653 passed, 27 skipped, 0 errors**, runs to completion in ~25 min.

## Skipped tests (tracked separately, not fixed in this PR)

Two distinct, unrelated flakiness sources surfaced once the pipeline could finally run to completion — both skip-marked here with a follow-up card rather than blocking this PR further:

- **4 tests in `sefaria/helper/tests/auto_linking_test.py`** (`test_refresh_links_with_text_save`, `_many_to_one_default_node`, `_one_to_one_default_node`, `_complex`) — off-by-a-small-number link-count assertion mismatches (`assert 2019 == 2018`, etc.), believed to be link-count drift in the shared sandbox Mongo across concurrent/prior runs.
- **3 tests** (`links_test.py::test_get_links_without_exclusion_would_include_perek_refs`, `summaries_test.py::test_index_add_delete`, `summaries_test.py::test_index_attr_change`) — failed specifically in the `backoffLimit: 1` retry-corruption run above; not yet confirmed whether they're genuinely flaky on their own or only failed because of that bug. Worth re-verifying without the skip now that `backoffLimit: 0` is in place.

Both groups are tracked in follow-up card sc-45571 (assigned to @nsantacruz).

## Follow-up (not in this PR)

CI is much slower than local mainly because the job inherits the web rollout's pod anti-affinity (runs on a **different node from Mongo** → every query is a cross-node hop) and runs single-process. Low-risk speedup = co-locate the pytest pod with Mongo; larger speedup = `pytest-xdist` after adding per-worker DB isolation. Tracked separately. Also still open: `sandbox-ready` health-check flakiness tied to an unrelated Postgres migration (blocks all PRs, separate issue).


.github/workflows/continuous.yaml
```diff
@@ -333,12 +329,13 @@ jobs:
           DEPLOY_ENV: sandbox-${{ steps.get-sha.outputs.sha_short }}
       - name: Wait For Job To Finish
         run: ./build/ci/waitForCIJob.bash
-        timeout-minutes: 60
```
nsantacruz, line 336: PR description says you raised this from 30->45 but it actually being decreased. Is this timeout even needed if we have timeout per test?

.github/workflows/continuous.yaml
```diff
@@ -306,12 +303,11 @@ jobs:
           project_id: ${{ secrets.DEV_PROJECT }}
           install_components: 'gke-gcloud-auth-plugin'
       - name: Authenticate GHA Runner To Target Cluster
-        uses: google-github-actions/get-gke-credentials@v2
+        uses: google-github-actions/get-gke-credentials@v3
         with:
           cluster_name: ${{secrets.DEV_GKE_CLUSTER}}
           location: ${{secrets.DEV_GKE_REGION}}
           project_id: ${{secrets.DEV_GCP_PROJECT}}
-          use_auth_provider: true
```
nsantacruz, line 314: why remove this?

build/ci/cleanup_test_data.py
```diff
@@ -0,0 +1,17 @@
+# Generated by OrchestKit Claude Plugin
+# Created: 2026-07-02
+
+"""Remove stale test fixtures from prior pytest runs in the shared sandbox Mongo."""
+from sefaria.system.database import db
+
+STALE_TITLES = ["Many to One on Genesis", "One to One on Genesis"]
+title_filter = {"title": {"$in": STALE_TITLES}}
+ref_regex = "Many to One on Genesis|One to One on Genesis"
```
nsantacruz: this can be generated from STALE_TITLES. Also, are these really the only stale titles? Seems hard to imagine.

build/ci/createJobFromRollout.sh
```diff
@@ -12,7 +12,7 @@ metadata:
     test-name: pytest
   name: $DEPLOY_ENV-pytest-sandbox-$GITHUB_RUN_ID
 spec:
-  backoffLimit: 2   # in waitForCIJob, we look for 2 fails before declaring failure.  This could be made a variable.
+  backoffLimit: 1
```
nsantacruz: we is this changed? I didn't see this in the PR desc

.github/workflows/continuous.yaml
yodem: (claude) Good catch — the description is wrong here. This isn't 30→45, it's **60→45** (checked against `origin/master`, the pre-PR value was `timeout-minutes: 60`). I'll fix the description.

On "is this needed given per-test timeout": yes, they cover different things. `--timeout=600` (pytest-timeout) kills one hung *test* after 10 min with a traceback. This step's 45-min timeout is the outer wall-clock backstop for the *whole job* (build wait + full suite run, currently ~23–26 min end to end) — it's what replaces the old (broken) `activeDeadlineSeconds: 1200` job-level deadline that this PR removes. Without it, a systemic problem (e.g. the Varnish-hang issue fixed in the last commit here) could run indefinitely and never surface as a normal CI failure.

.github/workflows/continuous.yaml
yodem: (claude) This is a carry-over from the earlier commit in this same branch/story (`507eff91b`, chapter 1 of the sc-45322 fix), not new in this diff. `get-gke-credentials@v2`'s `use_auth_provider: true` writes a legacy (non-exec) kubeconfig entry that modern Helm/kubectl reject — the upstream GitHub Action issue for this was closed by the maintainers as "not planned," so `use_auth_provider` isn't fixable on v2. Upgrading to `@v3` writes the modern exec-based kubeconfig entry directly, so the flag is no longer needed (and doesn't exist as a valid input on v3 in the same way). This was fix #3 of the 6 compounding CI failures this story chain addressed — happy to add a code comment here if that'd help future readers.

build/ci/cleanup_test_data.py
yodem: (claude) Fixed — `ref_regex` is now built from `STALE_TITLES` (`"|".join(re.escape(t) for t in STALE_TITLES)`) so there's one list to maintain, not two (pushed in `e03e7a83e`).

On "are these really the only stale titles" — yes, verified: `grep -rn "Many to One on Genesis\|One to One on Genesis"` across the repo shows the only place these titles (or any other dummy index title) get created is `sefaria/helper/tests/auto_linking_test.py`, and it only ever creates these two named indexes. No other test module persists a named Index/Version/Text into the shared sandbox Mongo, so there's nothing else to add to the cleanup list today. If a future test adds another named fixture, it'll need its own entry here — this script only knows about what exists now.

build/ci/createJobFromRollout.sh
yodem: (claude) You're right, this should've been called out in the description explicitly — will fix.

Rationale: `backoffLimit: 1` allows one retry (e.g. for a preemptible-node loss / infra-level pod kill), not for retrying genuinely failed tests — a real test failure should surface as a real CI failure, not be silently retried. The old comment ("we look for 2 fails before declaring failure") is stale now anyway: `waitForCIJob.bash` was rewritten in this same PR to fail on the *first* observed `status.failed >= 1`, so it no longer waits for 2 pod failures before reporting red — that behavior predates this specific backoffLimit change. I'll restore an inline comment explaining the current rationale so it doesn't read as an unexplained magic-number change again.

yodem:
@nsantacruz can you please figure out why the 4 remaining tests are failing? I dont mind skip them as well

nsantacruz:
> @nsantacruz can you please figure out why the 4 remaining tests are failing? I dont mind skip them as well

@yodem I took a quick look at this and it's not very simple to fix. I would skip them for now and make a card to fix them (you can assign it to me)

yodem:
Skipped the 4 flaky tests in `60e0e22e1` and filed a follow-up card: [sc-45571](https://app.shortcut.com/sefaria/story/45571) — assigned to @nsantacruz per the discussion above.

yodem:
Root-caused the last failure (run [28873987240](https://github.com/Sefaria/Sefaria-Project/actions/runs/28873987240/job/85797858085?pr=3470)): it wasn't the known auto_linking flakiness (those stayed skipped, as expected) — it was `backoffLimit: 1` actively hurting us. `waitForCIJob.bash` selects pods by label, not name, so once k8s spawned a retry pod under the same label, `kubectl logs -l` started returning combined/duplicate output from both pods (duplicate PASSED/FAILED/ERROR lines for the same tests near the tail of that run) and eventually hit the retry pod mid-`ContainerCreating`. The wait script also isn't retry-aware — it declares failure on the first pod's `status.failed`, before any retry finishes, so the retry wasn't buying a real budget, just corrupted logs.

Fixed in `4383f8724`: `backoffLimit: 0` (no Job-level retry), and skipped 3 more tests that failed as a side effect of the retry rerunning the suite against state a prior, interrupted pod left behind.

build/ci/waitForCIJob.bash
```diff
@@ -1,18 +1,43 @@
...
+    if [[ "$FAILED" -ge "1" ]]; then
+        echo ""
+        echo "=========================================="
+        echo "Job failed (tests failed or pod error)"
+        echo "=========================================="
+        kubectl logs -l "$LABEL" --tail=-1 || true
+        exit 1
+    fi
+
+    TOTAL=$(kubectl logs -l "$LABEL" --tail=-1 2>/dev/null | wc -l)
```
YishaiGlasner, line 44: can that code stuck in the loop? if so maybe you should define max retries

sefaria/helper/tests/auto_linking_test.py
```diff
@@ -405,6 +407,7 @@ def test_refresh_commentary_links_many_to_one_default_node(self):
         link_count = LinkSet({"generated_by": linker._generated_by_string, "refs": {"$regex": regex}}).count()
         assert desired_link_count == link_count
 
+    @pytest.mark.skip(reason="flaky link-count assertion in shared CI sandbox")
```
YishaiGlasner, line 410: what about fixing or removing the tests?

sefaria/conftest.py
```diff
@@ -17,6 +17,28 @@ def pytest_configure(config):
...
+    _dj_settings.DATABASES.pop("vector_db", None)
+    # Bust the ConnectionHandler's cached settings so the popped alias is really gone.
+    _dj_connections.__dict__.pop("settings", None)
+
+    # Disable Varnish cache invalidation for the entire test session. With
+    # USE_VARNISH on (as in the CI sandbox), invalidate_linked() on a large
+    # index iterates every linked ref and spawns varnishadm subprocesses,
+    # hanging mutation tests for hours. Tests must not purge shared caches.
+    from sefaria import settings as sefaria_settings
+    sefaria_settings.USE_VARNISH = False
```
YishaiGlasner, line 37: do we have tests for when we do use varnish?

sefaria/settings.py
```diff
@@ -158,6 +158,7 @@ def get_static_url():
     'reader',
     'django.contrib.staticfiles',
     'django.contrib.humanize',
+    'django.contrib.postgres',
```
YishaiGlasner: this seems to be caused by a previous accident. you can remove the changes in this file.

build/ci/waitForCIJob.bash
yodem: Good catch — fixed in `ced04e743`. Capped it at 10 consecutive retries (5 min) before exiting loudly with a clear diagnostic, instead of silently spinning until the outer 45-min GitHub step timeout kills it with a generic message. Counter resets on any successful kubectl call so a transient hiccup doesn't count toward the limit.

sefaria/helper/tests/auto_linking_test.py
yodem: Agreed with @nsantacruz on the PR thread above — he took a look and said it's not a quick fix, so we're skipping for now with a follow-up card (assigned to him) to root-cause and fix properly rather than leaving them silently red or deleting the coverage.

sefaria/conftest.py
yodem: Answer is no, and this PR doesn't change that — `sefaria/system/tests/test_varnish.py` is the only Varnish-related test module, and its whole test class body is gated behind `if USE_VARNISH:` at import time. `USE_VARNISH` has been `False` in `local_settings_ci.py` since before this PR, so that class was already never collected in CI — 0 tests, not skipped, just never defined. This PR's conftest.py change doesn't remove any existing coverage; it just makes sure the `False` override actually takes effect everywhere (a few modules were caching a stale `True` from import-time reads before the settings override applied). There's a genuine, pre-existing gap here — no test exercises the actual `invalidate_linked()`/`varnishadm` behavior with `USE_VARNISH=True` (understandably, since there's no reachable Varnish instance in the CI sandbox) — but that predates this PR and is out of scope here.

sefaria/settings.py
yodem: You're right — fixed in `ced04e743`, reverted to match `origin/master` exactly (no diff on this file now). It was a merge artifact: this branch's own `django.contrib.postgres` entry (from re-merging master, which independently re-added it via #3479) survived at its original position while master's copy got dropped, a functional no-op re-shuffle that just added noise.

YishaiGlasner (approved):

nsantacruz (approved):

PR #3471, opened 2026-06-30
feat(sc-45064): Hirsch dedication page at /dedication/hirsch

dcschreiber:
Adds the **Rav Hirsch on Torah (English)** dedication page — in memory of **Joel L. Fleishman** — live at `/dedication/hirsch`, in EN and HE.

### Approach
Built as a **standalone static page** (no model/admin), per product direction to ship the Hirsch page first and retro-fit it onto the reusable template (sc-45220) later. The page lives at the same URL the template will reuse, so the retro-fit needs no redirect.

Follows the existing dedication-page pattern (`adin-even-israel-steinsaltz`, `nash-bravmann-collection`): a `templates/static/…` page registered in `static_pages` and served by `serve_static`, with EN/HE via `int-en`/`int-he`.

### Changes (3 files)
- `templates/static/dedication/hirsch.html` — the page
- `sites/sefaria/urls.py` — register `dedication/hirsch`
- `static/img/joel-fleishman.png` — portrait

### Notes
- All copy is **verbatim from the project's source-of-truth Google Doc** 
- Verified rendering in both EN and HE locally.

Story: https://app.shortcut.com/sefaria/story/45064


saengel (dismissed):
Looks good

YishaiGlasner (changes requested):
i think the picture should be in our google storage (try to investigate its structure to find the right place)

YishaiGlasner (dismissed):

saengel (approved):

PR #3481, opened 2026-07-05
chore: dedication template

stevekaplan123:
## Description
This PR creates a dedications Django app that will be used in the next PR to support the William Davdison dedication and the Hirsch dedication.  Each entry in the Django app dedications has a slug (which corresponds to its url), english and hebrew titles, and english and hebrew content


yodem (commented):
Review posted inline. (claude)

dedications/views.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Mirrors reader.views.serve_static: passes headerMode as app_props so the
+    site header renders in Node, and the dedication record as template context.
+    """
+    dedication = get_object_or_404(Dedication, slug=slug)
+    return render_template(
+        request,
+        "static/dedication/dedication.html",
+        {"headerMode": True},
+        {"dedication": dedication},
```
yodem: `serve_static` (which this view says it mirrors) also passes `{"renderStatic": True}` in the template context. Here it's missing. In `base.html`, the `{% block content %}` override (i.e. the dedication title/body markup) only renders inside `{% if renderStatic %}` — without this flag, the content block never renders and the page will show the header only. Suggest: `{"dedication": dedication, "renderStatic": True}`. (claude)

sites/sefaria/urls.py
```diff
@@ -74,6 +75,7 @@
     re_path(r'^llms\.txt/?$', reader_views.serve_llms_txt),
     re_path(r'^(%s)/?$' % "|".join(static_pages), reader_views.serve_static),
     re_path(r'^(%s)/?$' % "|".join(static_pages_by_lang), reader_views.serve_static_by_lang),
+    re_path(r'^dedication/(?P<slug>[\w-]+)/?$', dedication_views.dedication),
```
yodem: `static_pages` (line 76, pre-existing) already contains the literal entry `"dedication/hirsch"`, matched earlier in this urlpatterns list. Since Django resolves patterns in order, `/dedication/hirsch` will keep hitting `serve_static` (the old hardcoded template) and will never reach this new dynamic view — even after a `Dedication` row with slug `hirsch` is created. Per the PR description, the very next PR wires up the Hirsch dedication through this app, so `"dedication/hirsch"` should be removed from `static_pages` (or this route moved above it) before that lands. (claude)

sites/sefaria/urls.py
```diff
@@ -3,6 +3,7 @@
 from django.urls import re_path
 from django.http import HttpResponseRedirect
 import reader.views as reader_views
+import dedications.views as dedication_views
```
yodem: Nit: `dedication_views` (singular) for the `dedications` module — every other import in this file aliases to match the module name exactly (`reader_views` for `reader.views`). Consider `dedications_views` for consistency. (claude)

dedications/models.py
```diff
@@ -0,0 +1,46 @@
...
+        validators=[RegexValidator(
+            r"^[\w-]+$",
+            "Slug may only contain letters, numbers, underscores, and hyphens.",
+        )],
+        help_text="URL identifier. The page is served at /dedication/<slug> "
+                  "(e.g. 'hirsch' -> /dedication/hirsch).",
+    )
+    en_title = models.CharField(max_length=255, help_text="Page title (English).")
+    he_title = models.CharField(max_length=255, help_text="Page title (Hebrew).")
+    en_content = models.TextField(
```
yodem, line 34: `en_content`/`he_content` are stored raw HTML rendered with `|safe` in the template. The closest existing analog (`guides.InfoCard`) stores markdown instead, which is safer by default. Since only staff have admin access this isn't externally exploitable, but it's a deliberate deviation from the established pattern — worth a short note on why raw HTML was chosen here. (claude)

dedications/views.py
stevekaplan123: Good catch — fixed in 13a9d7d1ff, the view now passes `"renderStatic": True` alongside the dedication.

sites/sefaria/urls.py
stevekaplan123: Deliberately leaving `"dedication/hirsch"` in `static_pages` for this PR: no `Dedication` row exists yet, so removing it now would turn the live /dedication/hirsch page into a 404 (`get_object_or_404`). The follow-up PR that creates the Hirsch dedication record will remove this entry and delete `templates/static/dedication/hirsch.html` in the same change, so the cutover is atomic.

sites/sefaria/urls.py
stevekaplan123: Renamed to `dedications_views` in 13a9d7d1ff.

dedications/models.py
stevekaplan123: Raw HTML is intentional here: dedication pages are being migrated from existing static HTML templates (e.g. `templates/static/dedication/hirsch.html`), so storing HTML lets that markup be pasted into the admin verbatim instead of hand-converting to markdown. Only staff can edit these records, so the `|safe` rendering isn't exposed to untrusted input. Added a note explaining this to the model docstring in 13a9d7d1ff.

yodem (approved):

PR #3484, opened 2026-07-06
fix(sc-38063): stop deleted texts from crashing topic pages

dcschreiber:
Fixes [sc-38063](https://app.shortcut.com/sefaria/story/38063): deleting a text from the library left orphaned `RefTopicLink`s behind, and one orphaned curated source white-screens the topic page for moderators (`/topics/memory` → Admin tab → scroll → blank).

## Root cause

`TopicStoryDescBlock` (`Story.jsx`) calls `Sefaria.index(...).primary_category` on each curated source. For a deleted text, `Sefaria.index()` returns `undefined` → TypeError → React unmounts the whole tree. Admin-only because the block renders only for curated sources pending review, which only moderators see, and only on scroll (lazy load).

The orphans exist because the index-delete cascade (`process_index_delete_in_topic_links`) iterates links and re-saves the topic per deletion — one malformed record mid-loop aborted the cascade and stranded the rest.

## Changes (one per layer)

| File | Change |
|---|---|
| `static/js/Story.jsx` | `?.` guard on `Sefaria.index(...)` — missing index renders an empty category instead of crashing. Direct fix for the throw. |
| `sefaria/helper/topic.py` | `get_topic` drops ref links whose ref no longer resolves (`Ref()` raises `InputError`), so existing orphans never reach the client. Sheet links untouched. |
| `sefaria/model/topic.py` | Delete/rename cascades wrap each link in `try/except (InputError, DuplicateRecordError)` so one bad record can't strand the rest. Systemic failures (Mongo/Postgres down) still abort loudly, and the delete cascade logs a single `error` with failed/total counts if any links were left behind. |

## Verification

- **In-browser as moderator** on a real orphan (`Ohr Ne'erav, Introduction 61`): crash reproduced pre-fix; page renders cleanly with either the backend filter or the frontend guard alone.
- **Unit test** `topic_test.py::test_get_topic_omits_orphaned_ref_links`: inserts a valid + a dead-ref link directly into Mongo, asserts `get_topic` (via the same `all_links` branch the topic-page API uses) returns the valid one and drops the orphan. Fails without the filter.

## Out of scope

Bulk cleanup of orphans already in Mongo (they're now filtered on read) — reasonable follow-up. Same for `process_topic_delete`, which has the same abort-mid-loop shape.


sefaria/helper/tests/topic_test.py
```diff
@@ -213,3 +213,32 @@ def test_update_topic(some_topic):
 								 {"c": "תמר", "lang": "d", "disambiguation": "יהודה"}])
 	with pytest.raises(Exception):
 		topic.update_topic(some_topic, slug='abc')
+
+
+def test_get_topic_omits_orphaned_ref_links():
+	"""sc-38063: a RefTopicLink whose text was deleted from the library (its ref no longer
```
stevekaplan123: comment shouldn't refer to shortcut story

sefaria/model/topic.py
```diff
@@ -1179,15 +1179,35 @@ def process_index_title_change_in_topic_links(indx, **kwargs):
...
+    # topic is itself malformed and fails to save, the set loop would abort and leave the
+    # remaining orphaned links behind. Isolate each deletion so one bad record can't strand
+    # the rest. Only bad-record errors are swallowed; systemic failures still abort loudly.
+    total, failed = 0, 0
+    for link in RefTopicLinkSet({"ref": {"$regex": pattern}}):
+        total += 1
+        try:
+            link.delete()
+        except (InputError, DuplicateRecordError) as e:
+            failed += 1
```
stevekaplan123, line 1212: Are InputError and DuplicateRecordError the only errors that can get generated by a bad delete?

sefaria/model/topic.py
```diff
@@ -1179,15 +1179,35 @@ def process_index_title_change_in_topic_links(indx, **kwargs):
...
         o.ref = o.ref.replace(kwargs["old"], kwargs["new"], 1)
+        # Rewrite each matching link independently. A single malformed link (e.g. one
+        # whose topic no longer exists) must not abort the rename cascade and strand the
+        # remaining links. Only bad-record errors are swallowed; systemic failures still abort loudly.
         try:
             o.save()
-        except InputError:
-            logger.warning("Failed to convert ref data from: {} to {}".format(kwargs['old'], kwargs['new']))
+        except (InputError, DuplicateRecordError) as e:
+            logger.warning("Failed to convert ref topic link '{}' on topic '{}' from '{}' to '{}': {}".format(
```
stevekaplan123, line 1192: Are InputError and DuplicateRecordError the only errors generated by bad save?

sefaria/helper/topic.py
```diff
@@ -18,6 +18,25 @@
...
+    what make a topic page blank out for admins, so we omit them from the topic response.
+    Sheet links (is_sheet) reference sheets rather than library texts and are always kept.
+    """
+    if getattr(link, 'is_sheet', False):
+        return True
+    try:
+        Ref(link.ref)
+        return True
+    except InputError as e:
+        logger.warning("Omitting orphaned ref topic link '{}' on topic '{}': {}".format(
```
stevekaplan123, line 35: Check what errors a bad Ref parse can generated.  It may be that the only one is InputError

stevekaplan123 (changes requested):

sefaria/helper/tests/topic_test.py
dcschreiber: removed the sc-38063 reference

sefaria/model/topic.py
dcschreiber: well DuplicateRecordError is actually a subclass of InputError so I will remove that.
On the next run Claude caught additional classes and added them. I still can't be sure it's exhaustive, but the other option is a global catch.

sefaria/model/topic.py
dcschreiber: Also here I broadened the catch

sefaria/helper/topic.py
dcschreiber: Yes, it seems that is the only one (there are a bunch of sub-classes of it that are raised)

stevekaplan123 (approved):

PR #3485, opened 2026-07-06
Feature/sc 45455/extend elastic search to index new entities

stevekaplan123:
## Description
This implements the backend for entity search.  (1) Adds an entity_search_api for searching for books, topics, and authors. (2) Adds books, topics, and authors to ElasticSearch database.  (3) The new entities will get added to ElasticSearch via the cronjob (index_all calls the entity indexing), but there are also two scripts added to this PR that can be run to do the indexing specifically for entities.


helm-chart/sefaria/templates/configmap/local-settings.yaml
saengel, line 1: You ran this by @yodem and team?

reader/views.py
```diff
@@ -4793,6 +4793,45 @@ def search_wrapper_api(request, es6_compat=False):
         return jsonResponse({"error": "Error with connection to Elasticsearch. Total shards: {}, Shards successful: {}, Timed out: {}".format(response._shards.total, response._shards.successful, response.timed_out)}, callback=request.GET.get("callback", None))
     return jsonResponse({"error": "Unsupported HTTP method."}, callback=request.GET.get("callback", None))
 
+@csrf_exempt
+def entity_search_api(request):
+    """
+    Entity search endpoint powering the Topics / Authors / Books tabs.
+
+    GET /api/entity-search?q=<query>&type=<topic|author|book>
```
saengel: Let's add query params for all of the various sorts we need, and filters for books as well.

sefaria/model/topic.py
```diff
@@ -52,6 +52,16 @@ def get_title(self, lang):
     def get_url(self):
         pass
 
+    @abstractmethod
+    def is_category(self):
```
saengel, line 62: Maybe let's add a comment to explain that these are needed for the author-specific class inheritance, otherwise puzzling

scripts/reindex_entity_search.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+On-demand full reindex of the entity-search indices (`topic` and/or `book`).
```
saengel, line 4: This script seems OK, did we plug it into the existing cronjob yet? Am I missing it in this PR?

sefaria/helper/search.py
```diff
@@ -205,6 +214,265 @@ def make_filter(type, agg_type, agg_key):
...
+                },
+                "boost_mode": "multiply",
+            }
+        }
+
+    return search_obj[start:start + size]
+
+
+def _total_from_response(response):
+    total = response.hits.total
```
saengel, line 450: What will this return if there are no hits? `None` or 0?

sefaria/search.py
```diff
@@ -1012,6 +1160,249 @@ def index_public_notes():
...
+        doc['era'] = topic.get_property('era')
+        doc['birthYear'] = topic.get_property('birthYear')
+        doc['deathYear'] = topic.get_property('deathYear')
+        # Denormalize the titles of this author's works (EN + HE) so the author is
+        # searchable by a book they wrote (e.g. "Mishneh Torah" -> Maimonides).
+        authored_en, authored_he = [], []
+        for authored_index in IndexSet({"authors": slug}):
+            en = authored_index.get_title('en')
+            if en:
+                authored_en.append(en)
```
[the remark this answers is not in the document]
saengel: @stevekaplan123 does making this faster make sense to you also?

sefaria/search.py
```diff
@@ -1012,6 +1160,249 @@ def index_public_notes():
...
+            cache[slug] = list(dict.fromkeys(resolved))  # de-dup, preserve order
+        names += cache[slug]
+    return list(dict.fromkeys(names))
+
+
+def make_book_index_document(index, author_name_cache=None):
+    """
+    Build an Elasticsearch document for an Index (book) record for the `book` index.
+
+    The document id is the English title, so reindexing is idempotent.
```
saengel, line 1327: What if the book title changes? Does the reindex handle that? Would having a slug make more sense?

saengel (changes requested):
Almost there - some (mostly small) feedback

helm-chart/sefaria/templates/configmap/local-settings.yaml
stevekaplan123: No I don't think it's necessary as it's just adding local settings variables.  This follows the pattern we used previously with SEARCH_INDEX_NAME_TEXT and SEARCH_INDEX_NAME_SHEET

reader/views.py
stevekaplan123: I'm not sure what you mean here.  Can you explain more?

scripts/reindex_entity_search.py
stevekaplan123: Good catch.  Adding it into the cronjob...

sefaria/helper/search.py
stevekaplan123: With no hits it returns 0, not None.

```In elasticsearch-dsl 8.x (which this repo has — elasticsearch-dsl==8.0.0), response.hits.total is always present in a search response and is an AttrDict like {'value': 0, 'relation': 'eq'}. Elasticsearch includes that object even when nothing matched — an empty result set just means value is 0. ```

sefaria/search.py
stevekaplan123: That's a great point.  I think it would still be a problem if the book title changes and we use a slug, for the one week before the cronjob gets run again.  We should probably have a dependencies.py function that on book title change runs through ElasticSearch and modifies it.

stevekaplan123: I'm going to create a story for that

sefaria/search.py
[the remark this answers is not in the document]
stevekaplan123: Implementing

helm-chart/sefaria/templates/configmap/local-settings.yaml
saengel: Ok got it, but more broadly - in terms of the helm chart and cauldrons etc?

reader/views.py
saengel: We need to support sorting and filtering. 

We could theoretically do that client-side, maybe worthwhile to add parameters here to support server side sorts and filtering? 

/api/entity-search?q=<query>&type=<topic|author|book>**?sort=date_asc&filter="Tanakh"**

That type of thing...

saengel (changes requested):
Looking good! I still think we need to address support for sorts and filters by various types (acc to the data spec) - and additionally, I think we should add some tests.

PR #3491, opened 2026-07-08
Feature/sc 45462/setting up weblate

stevekaplan123:
## Description
Re-factors the way we handle translation in for most strings in Sefaria-Project, which are found in strings.js.  We now use the English as a key that will return the proper string based on the interface language.  Added playwright tests to make sure all strings in strings.js actually show up on the site.


static/js/TextColumnBanner.jsx
```diff
@@ -54,7 +54,7 @@ const TransLangPrefAcceptedBanner = () => {
         <TextColumnBanner>
             <InterfaceText>
                 <EnglishText> Thanks! We'll show you {lang} translations first when we have them. </EnglishText>
-                <HebrewText>תודה! כשנוכל, נציג לכם תרגומים בשפה ה<span className="bold">{Sefaria._(lang)}</span> כאשר אלו יהיו זמינים. </HebrewText>
+                <HebrewText>תודה! כשנוכל, נציג לכם תרגומים בשפה ה<span className="bold">{Sefaria.translateISOLanguageName(translation_language_preference_suggestion)}</span> כאשר אלו יהיו זמינים. </HebrewText>
```
yitzhakc, line 57: I don't understand why it changed this - this is the value of `lang`

static/js/TextColumnBanner.jsx
```diff
@@ -71,13 +71,13 @@ const TransLangPrefAskBanner = ({ setAccepted, setTranslationLanguagePreference
...
     const lang = Sefaria.translateISOLanguageCode(translation_language_preference_suggestion);
-    const buttons = [{text: "Yes", onClick: accept}, {text: "No", onClick: reject, sideEffect: "close" }];
+    const buttons = [{text: "Yes", textId: "text_column_banner.yes", onClick: accept}, {text: "No", textId: "text_column_banner.no", onClick: reject, sideEffect: "close" }];
 
     return (
         <TextColumnBanner buttons={buttons} onClose={reject}>
             <InterfaceText>
                 <EnglishText> Prefer to see <span className="bold"> {lang} </span> translations when available? </EnglishText>
-                <HebrewText>האם תעדיפו לראות תרגומים בשפה ה<span className="bold">{Sefaria._(lang)}</span> כאשר הם זמינים?</HebrewText>
+                <HebrewText>האם תעדיפו לראות תרגומים בשפה ה<span className="bold">{Sefaria.translateISOLanguageName(translation_language_preference_suggestion)}</span> כאשר הם זמינים?</HebrewText>
```
yitzhakc, line 80: Here too

static/js/TextColumnBanner.jsx
```diff
@@ -146,7 +147,7 @@ const TextColumnBannerButton = ({ button, closeBanner }) => {
     }
     return (
         <button tabIndex="0" className="yesNoButton" aria-label={button.text} onClick={onClick}>
-            <InterfaceText>{button.text}</InterfaceText>
+            <InterfaceText>{button.textId || button.text}</InterfaceText>
```
yitzhakc: Why wouldn't we just use `textId`?

yitzhakc (changes requested):

static/js/TextColumnBanner.jsx
stevekaplan123: The thing is this re-factor changes the `Sefaria._` function to expect key IDs rather than plain strings.  Since we're passing an ISO code like 'en' or 'he', we need to use the helper `translateISOLanguageName` to map the ISO code to the right i18n key and finally to the actual language name like "English" or "Hebrew".

static/js/TextColumnBanner.jsx
stevekaplan123: You're right.  Changing...

yitzhakc (approved):

PR #3492, opened 2026-07-08
Feature/sc 45363/migrate the page to the template

stevekaplan123:
## Description

This PR creates Django 'dedications' template for the William Davidson Talmud and Fleishman Hirsch dedications, which are now data rather than HTML.  It also changes the link to go to "/dedication/fleishman-hirsch-on-torah-in-english" not "/dedication/hirsch"


templates/static/dedication/dedication.html
```diff
@@ -4,17 +4,229 @@
 
 {% block meta_title %}{% meta_title %}{% if request.interfaceLang == "hebrew" %}{{ dedication.he_title|striptags }}{% else %}{{ dedication.en_title|striptags }}{% endif %}{% endmeta_title %}{% endblock %}
 
+{% block css %}
+    /* ------------------------------------------------------------------ *
+     * Dedication pages (DB-backed, editable in the Django admin).
+     * Page bodies live in Dedication.en_content / he_content; per-page
+     * styling is kept here (engineer-managed) and scoped by the wrapper id
+     * #dedication-<slug> that the template emits below.
```
YishaiGlasner, line 12: i don't understand this - if we have different hardcoded css for any dedication, why we want the html to be in the data?
it seems to me better to have it all as templates.

PR #3500, opened 2026-07-09
Zero Results Display for Search

saengel:
## Description
Null state rendering on a per-tab basis when searches return 0 results. 

## Code Changes
A new `static/js/NoSearchResults.jsx` component with associated CSS, and rendering on a per-tab basis in `static/js/SearchPage.jsx`. Once the backend has been wired up to populate the tabs, we will need to add a conditional for rendering the null state component only in cases of 0 results. 

## Notes
1. Still waiting on author copy, tracking in a separate card to not hold this back (lorem ipsum for now)
2. Instead of hardcoded copy, do we need to integrate with Weblate?
3. Hebrew was not yet addressed, waiting for a decision.


static/js/sheets/SheetsWithRefPage.jsx
```diff
@@ -281,11 +281,13 @@ const SheetsWithRefPage = ({srefs, searchState, updateSearchState, updateApplied
...
     };
-    const aiBadgeText = searchState?.sortType?.toLowerCase?.() === 'relevance' ? 'ai_info_tooltip.these_sheet_results_are_ranked_by_ai' : undefined;
-    return <SheetsWithRefLayout
+    const aiBadgeText = searchState?.sortType?.toLowerCase?.() === 'relevance'
+        ? 'ai_info_tooltip.these_sheet_results_are_ranked_by_ai' : undefined;
+    return <SearchPage
           key={"sheetsPage"}
           isQueryRunning={loading}
           sortTypeArray={sortTypeArray}
+          searchTopMsg="search_page.sheets_with"
```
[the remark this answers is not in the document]
stevekaplan123: This seems right.  Why did you switch from SheetsWithRefLayout to SearchPage?

static/js/NoSearchResults.jsx
```diff
@@ -0,0 +1,87 @@
...
+    captionKey:'search.null.authors.caption',
+    ctaHref:   '/people',
+  },
+  topics: {
+    h1Key:     'search.null.topics.h1',
+    bodyKey:   'search.null.topics.body',
+    buttonKey: 'search.null.topics.button',
+    captionKey:'search.null.topics.caption',
+    ctaHref:   '/topics',
+  },
```
stevekaplan123: You could make this generalized by doing "search.null.{type}.{keyType}", but generally we do want to avoid spelling out the key in the code in case it gets changed in Weblate.  However, I myself have added keys into the code and I'm a little unsure how strict we should be.  If there's an easy way to do this without the keys that would be ideal.

static/js/SearchPage.jsx
```diff
@@ -474,9 +467,11 @@ class SearchPage extends Component {
                 options={ENTITY_SORT_OPTIONS.authors}
                 sortType={this.state.entitySort.author}
                 onSortChange={(key) => this.setEntitySort('author', key)}
+                disabled={!this.getSortedEntityData('author')?.hits?.length}
```
stevekaplan123: What does `!this.getSortedEntityData('author')?.hits?.length` check?  Maybe it should be a function with a descriptive name as this line appears often.

stevekaplan123 (changes requested):

static/js/sheets/SheetsWithRefPage.jsx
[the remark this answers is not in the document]
saengel: A mistake with the overly complicated merge - my bad, thanks for catching - will revert!

static/js/NoSearchResults.jsx
stevekaplan123: Ignore my previous comment about putting the keys in the code.  I think that's fine as is.  But I still think that we could make this generalizable by doing "search.null.{type}.{keyType}".  What do you think?

saengel: Ok, addressed

stevekaplan123 (approved):

PR #3502, opened 2026-07-09
Fix Tali Midrash title getting small in linker

nsantacruz:
## Summary
- Host page CSS rules targeting `a` elements (e.g. a generic `font-size` rule) were overriding the wrapped text's inherited styling in the linker, causing linked ref titles (e.g. Midrash Tali) to render smaller than the surrounding text.
- Explicitly set `font`, `color`, `text-decoration`, and `letter-spacing` to `inherit` on generated `<a class="sefaria-ref">` tags so they always match the surrounding text regardless of the host page's own CSS.

## Test plan
- [ ] Verify Midrash Tali (and other) linked titles render at the correct size on a page with a generic `a { font-size: ... }` host CSS rule
- [ ] Confirm linked ref styling still matches surrounding text on pages without conflicting host CSS

### DIFF
--- a/static/js/linker.v3/main.js
+++ b/static/js/linker.v3/main.js
@@ -162,6 +162,13 @@ import pRetry, {AbortError} from 'p-retry';
         atag.target = "_blank";
         atag.textContent = text;
         atag.className = "sefaria-ref";
+        // force the wrapped text to render exactly as it did before wrapping, regardless of the
+        // host page's own CSS for `a` elements (e.g. a generic `a { font-size: ... }` rule would
+        // otherwise override the inherited size/family/color of the surrounding text)
+        atag.style.font = "inherit";
+        atag.style.color = "inherit";
+        atag.style.textDecoration = "inherit";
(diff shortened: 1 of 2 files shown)


static/js/linker.v3/main.js
```diff
@@ -162,6 +162,13 @@ import pRetry, {AbortError} from 'p-retry';
         atag.target = "_blank";
         atag.textContent = text;
         atag.className = "sefaria-ref";
+        // force the wrapped text to render exactly as it did before wrapping, regardless of the
+        // host page's own CSS for `a` elements (e.g. a generic `a { font-size: ... }` rule would
+        // otherwise override the inherited size/family/color of the surrounding text)
+        atag.style.font = "inherit";
```
yonadavGit: what about properties that aren't in the list?
perhaps we shouldn't worry about it but worth flagging i think

nsantacruz: @yonadavGit good point. I've made the fix a bit more general.

yonadavGit (approved):

PR #3503, opened 2026-07-09
chore(sc-45178): include app version in web analytics events

dcschreiber:
## Summary
- Resolves the web slice of [sc-45178](https://app.shortcut.com/sefaria/story/45178) / sc-45625 ("Include Version Metadata in Analytics Events").
- The deployed app version already exists as a semantic-release semver tag (the same one used as the Docker image tag), but it was never surfaced past the container — nothing in Django or the frontend knew what version was running.
- Plumbs it through: Helm sets `APP_VERSION` as a pod env var from the existing `.Values.web.containerImage.tag` → Django reads it into a new `APP_VERSION` setting → `base_props()` includes it as `appVersion` → the GA4 `gtag('config', ...)` call in `base.html` tags every analytics event with `app_version`, and Sentry's `release` field gets it too (free error-version correlation).
- Reuses the team's existing version identifier instead of introducing a new git-SHA scheme, since it's already what appears in deploy commits and Datadog labels.

## Test plan
- [x] Python syntax check on `sefaria/settings.py` / `reader/views.py`
- [x] Babel parse check on the edited JS/JSX files
- [x] Local smoke test (`manage.py runserver` + real browser via claude-in-chrome): confirmed `DJANGO_VARS.props.appVersion`, the GA4 `gtag` dataLayer push (`app_version` param), and no console errors
- [x] Cauldron smoke test (`sc45178-analytics`, since deleted): confirmed the real app code (Django + JS) deploys and runs correctly from a genuine PR-merge-commit image
- [x] CI green end-to-end (PyTest, Jest, Sandbox Deploy, `continuous-branch-protection`) after merging latest `master`, which also picked up an unrelated `gke-gcloud-auth-plugin` CI fix (#3506) that had been failing this PR's checks
- [ ] Confirm with Josh/Steve (comment thread on sc-45178) that a semver string satisfies the ask, since the ticket as written says "git commit"

**Known gap, not fixable from a branch:** the Helm chart line itself (`APP_VERSION` env var in `web.yaml`) couldn't be exercised in the cauldron, because cauldrons track whatever chart was last published from `master`/`preprod`/`prod` (chart publishing is decoupled from the branch's own app-image build) unless you pin a pre-released chart via `--chart`. Accepted as low-risk since it mirrors 3 already-proven uses of the same value in that same file — first real validation happens on merge + deploy, same as any other chart change here.


templates/base.html
```diff
@@ -275,6 +275,7 @@
               'user_id': DJANGO_VARS.props ? DJANGO_VARS.props._uid : null,
               'traffic_type': DJANGO_VARS.props && DJANGO_VARS.props._email.includes('sefaria.org') ? 'sefariaemail' : null,
               'site_lang': '{{ request.interfaceLang }}',
+              'app_version': DJANGO_VARS.props ? DJANGO_VARS.props.appVersion || null : null,
```
stevekaplan123: This is confusing syntax.  If this line is actually necessary, can you add a comment explaining what this is for?

stevekaplan123 (changes requested):

templates/base.html
stevekaplan123: Ask Claude why you can't just put 
``` 
DJANGO_VARS.props?.appVersion
```

stevekaplan123 (changes requested):
Maybe try DJANGO_VARS.props?.appVersion

templates/base.html
dcschreiber: it was for consistency. updating now

stevekaplan123 (approved):

PR #3509, opened 2026-07-10
Upgrade codebase to use Strapi v5 (take two)

relyks:
## Description
Upgrading to Strapi v5 involves breaking changes to its interface. This PR adjusts the codebase for compatibility.

**This is an update to https://github.com/Sefaria/Sefaria-Project/pull/3487 to fix the frontend crashing issue**

## Code Changes - v1
* Updated the GraphQL queries in all locations to use the new flattened schema structure and response format.
* Updated the code to preserve the existing behavior for moderators to be able to see drafts and published entries together for the products page. In Strapi v5, `publicationState` is removed and replaced by status, with values `draft` and `published`.

## What went wrong
- The frontend was upgraded to expect Strapi v5's new data format, but the live Strapi used in production is still on v4, which returns data in an older, differently-shaped format.
- The two formats nest data differently: v4 wraps lists like { data: [...] }, while v5 returns a plain [...] list.
- The sidebar ads code assumed the new v5 shape and called .forEach() directly on the data. When it got the old v4 shape (an object, not a list), that call threw an error. Because it wasn't wrapped in any error handling, the crash took down the entire React app (blank page), not just the sidebar ads.
- The old data reached the new code through the cache: it's shared and survives deploys, and its key was based only on the date. A response saved by the old code got handed to the new code, which couldn't read it.

## Code changes - v2

- Guard the sidebar ads against bad data: data is only looped over if it's actually a list (`Array.isArray`), so an unexpected shape is treated that are no available ads instead of crashing
- Wrap the ads logic in try/catch: if anything unexpected happens, the message is logged; a single bad payload can never blank the whole page again.
- Version the cache key (`v5`) — like versioning a non-backwards-compatible API, this guarantees the new code never reads a response cached by the old code
- Stop caching error responses — previously a failed Strapi query got cached for up to 7 days; now errors are returned but not stored, so one bad response can't poison the cache.
- Tests — updated the cache tests for the new key format and added one confirming error responses aren't cached

### Note
- Other content types (team, jobs, products) already had error handling, so they degraded safely to a "CMS cannot be reached" message. Only sidebar ads lacked it.
- A separate PR will be considered to add a step to clear the Strapi cache on deploy as extra safety


static/js/Promotions.jsx
```diff
@@ -14,76 +14,81 @@ const Promotions = () => {
...
-          // Add a separate ad if there's a Hebrew translation. There can't be an ad with only Hebrew
-          if (sidebarAd.localizations?.data?.length) {
-            const hebrewAttributes = sidebarAd.localizations.data[0].attributes;
-            const [buttonText, bodyText, buttonURL, title] = [
-              hebrewAttributes.buttonText,
-              hebrewAttributes.bodyText,
-              hebrewAttributes.buttonURL,
-              hebrewAttributes.title,
-            ];
+        // Only an array is iterable here. A stale/incompatible payload may nest the ads under a wrapper object (Strapi v4's { data: [...] })
```
yitzhakc, line 23: I'm happy to see we're checking for the correct format - but would it make sense to do a more thorough check in terms of structure? Ie to be 100% sure that the data matches the new format? Or alternatively, degrade gracefully when the format is not 100% matching..
Also, maybe add some JS tests, as you have examples of "old" data and "new" data so we can ensure that we are catching standard use cases.

static/js/StaticPages.jsx
```diff
@@ -2911,83 +2892,74 @@ const ProductsPage = memo(() => {
...
 
-    // GraphQL query to Strapi
+    // GraphQL query to Strapi.
+    // Regular users see only published products.
+    // Moderators see published products plus any products that exist only as drafts (not yet published),
+    // so they can preview upcoming content alongside what is currently live.
+    // When a product is already published, the published version is always shown — even to moderators —
+    // because the published version is the canonical one. Draft edits to live products are not surfaced here.
+    //
+    // WHY TWO QUERIES INSTEAD OF ONE:
```
yitzhakc, line 2902: Great comment.
Unfortunate, that they don't have a way to replicate the existing behavior..

yitzhakc (changes requested):

static/js/Promotions.jsx
```diff
@@ -14,76 +14,81 @@ const Promotions = () => {
...
+        // The wrong data type  must be treated as "no ads" rather than crashing on .forEach iterator
+        if (Array.isArray(sidebarAds)) {
+          sidebarAds.forEach((sidebarAd) => {
```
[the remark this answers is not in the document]
relyks: This doesn't matter, because there won't be anything already cached in the frontend

static/js/Promotions.jsx
relyks: Yeah, it does. You are totally right. However, I forgot to mention: I'm planning on making a future PR (Genevieve approved this) to add a lot more extensive automated tests (Jest and Playwright) and to make sure everything continues working in the future using some of the techniques I did for the newsletter sign-up page, like mocking the API to Strapi with HTTP archives. I want to refactor a lot of Strapi-related code eventually, but I need to make sure I don't break any existing functionality. 

I don't think checking the structure is necessary right now. I manually tested that it works locally and on a cauldron this time with Strapi v4 and v5 for all the content types. It will work once I upgrade the Strapi instance that production uses. To create the tests, I'd need to extract the code into a separate function file (which I've already started for the PRs that depend on this), so I can import it into Jest.

I added exception handling here, so any problems should degrade gracefully. My goal here was just to prevent any issues while Strapi v4 is still in use, until I manually upgrade. Are you ok with me adding a lot of tests as a separate PR?

yitzhakc: Yes, that's fine.

yitzhakc (approved):

PR #3511, opened 2026-07-12
SSO

YishaiGlasner:
## Summary

Replaces Sefaria's old server-rendered login/register/password-reset templates with a single React auth experience (`AuthPage`), and adds Google + Apple Sign-In (SSO) for web and mobile, built on `django-allauth`. Also includes a new sign-up funnel analytics pipeline, an account-settings UI update to reflect SSO-linked accounts, and a handful of unrelated fixes picked up along the way (SSR crash, proxy header handling, i18n).

## Backend: `django-allauth` integration + new `sso` app

- Added `django-allauth` (`allauth`, `allauth.account`, `allauth.socialaccount`, Google + Apple providers, `allauth.headless`) to `INSTALLED_APPS`, plus its authentication backend and account middleware (`sefaria/settings.py`).
- New `sso` app (`sso/adapters.py`, `sso/views.py`, `sso/urls.py`):
  - `SefariaAccountAdapter` — ties `emailusernames` (username == email) into allauth, and honors a `next` redirect stashed in a short-lived `sefaria_sso_next` cookie (needed because Google's redirect-mode SSO can't carry `next` as a query param — the `login_uri` must be an exact-match registered redirect URI).
  - `SefariaSocialAccountAdapter` — handles first-time social sign-up (creates the Mongo `UserProfile`, assigns slug, imports Gravatar, registers the user in Salesforce CRM) and email-collision linking (disables the password on an existing email/password account when a Google/Apple sign-in claims the same email, making it SSO-only going forward).
  - `import_gravatar` extracted out of `sefaria/views.py` into `sso/adapters.py` so both the SSO and classic email registration paths share it.
  - New endpoints under `sso/urls.py`:
    - `POST /api/auth/google/redirect` — Google One Tap redirect-mode callback (delegates to allauth's `LoginByTokenView`)
    - `POST /api/auth/google/mobile` / `POST /api/auth/apple/mobile` — native mobile (React Native) Google/Apple sign-in; verifies the provider ID token server-side and returns SimpleJWT `{access, refresh}` tokens
    - `POST /api/auth/apple/callback` — Apple Sign-In popup mode (web)
    - `POST /api/auth/login` — JSON email login for the new AuthPage
    - `POST /api/auth/password/reset` — JSON password-reset request
  - Mobile JWT endpoints are hardened: session is flushed after issuing tokens, Apple-provided names are sanitized, and JWKS fetch failures are reported to Sentry.
- `ClearSsoNextCookieMiddleware` — deletes the `sefaria_sso_next` cookie as soon as either SSO callback view responds (success or failure) so it can't be replayed by a later, unrelated attempt.
- `SECURE_PROXY_SSL_HEADER` now set so `request.is_secure()` (and allauth's OAuth2 callback URLs built from `request.build_absolute_uri()`) correctly report `https` behind the TLS-terminating ingress.
- `sefaria/forms.py` — registration now distinguishes "already registered via Google", "via Apple", or plain email/password when an email collision is hit.
- `sefaria/views.py`:
  - `CustomLoginView` / `CustomPasswordResetConfirmView` now render the SPA shell (`base.html`) instead of the old dedicated templates, and `CustomPasswordResetConfirmView` gained a JSON POST path (submit new password, or request a resend when the link has expired) so the flow works fully client-side.
  - `process_register_form` simplified now that Gravatar import lives in `sso/adapters.py`.
- `account_settings` view/template now surface `social_providers` (from `request.user.socialaccount_set`) so the account page shows "Google/Apple Sign-In" instead of an editable email field for SSO-only accounts, and the Google Drive connection copy was reworded for clarity.
- Deleted the old, now-unused templates: `registration/login.html`, `registration/register.html`, `registration/accounts.html`, and the long-dead `translate_campaign.html`.
- Removed the dead `/accounts` view.
- `LanguageSettingsMiddleware` refactored to extract interface-language resolution into a shared helper (used by both the "excluded path" short-circuit and the normal path), fixing interface language on excluded paths (e.g. auth API paths) instead of hardcoding English.
- Config/deploy: new `GOOGLE_SSO_CLIENT_ID` / `APPLE_SSO_*` settings (`local_settings_example.py`, Helm chart `local-settings-file.yaml`), `django-allauth` added to `requirements.txt`.

## Frontend: new `AuthPage` React experience (`static/js/auth/`)

- `AuthPage.jsx` — single state machine (`view` × `flow`) that swaps card content in place with no full page navigation, replacing the old `/login`, `/register`, and `/password/reset/confirm` server-rendered pages. Views: `choose`, `email` (login or register), `forgot`, `forgot-sent`, `reset`, `reset-expired`, `reset-success`.
- Supporting components: `ChooseView`, `LoginView`, `RegisterView`, `ForgotView`, `ResetView`, `ResetExpiredView`, `MessageView`, `ErrorBanner`, `AuthCard`, `Divider`, `LegalText`, `EmailInput`, `PasswordInput`, `ProviderButton`, `FormView`.
- `useSsoSignIn.jsx` — shared hook wiring up Google Sign-In (button + One Tap) and Apple Sign-In JS SDKs, exposing provider-ready flags and click triggers to the views above.
- `GoogleOneTap.jsx` — renders Google's One Tap prompt globally, mounted once in `ReaderApp`.
- `utils.js` — path/flow helpers (`pathToFlow`, `flowToPath`, `nextFromPath`, `isAuthPath`, `withNext`) shared between `AuthPage` and `ReaderApp`.
- New reusable primitives moved/added under `static/js/common/`: `Input.jsx`, `Captcha.jsx`.
- `ReaderApp.jsx` — now owns auth routing: recognizes `/login`, `/register`, and reset-confirm paths, tracks `showAuth`/`authPath`/`authSource` in component state and browser history, and renders `AuthPage` in place of the normal panel container when active. Also mounts `GoogleOneTap` and resumes any pending sign-up analytics attempt on mount.
- `Header.jsx` — login/sign-up links now route through the in-app `AuthPage` (via a new `AuthNavLink` that calls `openURL`) instead of full-page navigation, and the nav-bar sign-up button carries a `data-signup-source` marker for funnel attribution.
- New CSS: `auth.scss`/`auth.css` and `common-component.scss`/`common-component.css`, plus new icons (`google.svg`, `apple.svg`, `eye.svg`, `eye-off.svg`, `arrow-left.svg`, `info-error.svg`) and auth background images.
- `base.html` — exposes `window.SefariaAuth` / passes `googleClientId`, `appleClientId`, `recaptchaSiteKey`, and reset-link state (`authResetUid`, `authResetValid`) down to the client.
- `templates/elements/login_method_text.html` — small partial used by the updated account-settings page.

## Sign-up funnel analytics

- `signupAnalytics.js` / `useSignUpTracking.js` — new tracking layer for the sign-up funnel (method chosen, process started/ended, resuming an in-progress attempt across a redirect-based SSO round trip via `resumePendingSignUpAttempt`), replacing the old inline GA4 script that used to live in `register.html`.
- Test coverage added: `static/js/auth/tests/signupAnalytics.test.js`, `useSignUpTracking.test.js`, `utils.test.js`.

## i18n

- New `auth.*` string keys added to `static/js/sefaria/i18n/interface-context/en.json` and `he.json` (English + Hebrew) for every AuthPage view, replacing the old templates' hand-rolled `int-en`/`int-he` spans.

## Tests

- `sso/tests/adapters_test.py`, `sso/tests/views_test.py`, `sso/tests/middleware_test.py` — new backend coverage for the adapters, all `sso/views.py` endpoints, and `ClearSsoNextCookieMiddleware`.
- `reader/tests/password_reset_confirm_view_test.py` — coverage for the new JSON reset-confirm flow.
- `sefaria/tests/forms_test.py`, `sefaria/system/tests/interface_language_test.py` — coverage for the updated registration-form email-collision messaging and the refactored language middleware.
- `e2e-tests/` — Playwright specs and page objects (`loginPage.ts`, `signupPage.ts`, `mobileHamburgerPage.ts`) updated for the new `AuthPage` flow; added `User Menu/auth-page.spec.ts` and updated `mobile web/auth-flow.spec.ts`.

## Other fixes bundled into this branch

- Fixed a chatbot site-wide-banner SSR crash by computing its text at render time instead of module load (`static/js/SiteWideBanner.jsx` / `sitewidebanner`).
- Trust `X-Forwarded-Proto` behind the TLS-terminating proxy (see `SECURE_PROXY_SSL_HEADER` above).
- Google mobile SSO now honors `next` instead of always redirecting home; the `sefaria_sso_next` cookie is sent as `SameSite=None` and percent-decoded correctly before use.
- Percent-decoding fix for the `sefaria_sso_next` cookie value.
- Real interface language is now sent to Salesforce on SSO sign-up (previously hardcoded).

## Notable dependencies/config

- New Python dependency: `django-allauth[socialaccount]`.
- New settings: `GOOGLE_SSO_CLIENT_ID`, `APPLE_SSO_CLIENT_ID`, `APPLE_SSO_IOS_BUNDLE_ID`, `APPLE_SSO_TEAM_ID`, `APPLE_SSO_KEY_ID`, `APPLE_SSO_PRIVATE_KEY` — must be populated (via env vars in deployed environments) for Google/Apple sign-in to work.


sefaria/views.py
```diff
@@ -91,8 +91,15 @@ def get_context_data(self, **kwargs):
         context['renderStatic'] = True
         return context
 
-class CustomLoginView(StaticViewMixin, LoginView):
+class CustomLoginView(LoginView):
     authentication_form = SefariaLoginForm
+    template_name = 'base.html'
```
yodem, line 98: **P0 — this makes `/login` render a blank page.**

Dropping `StaticViewMixin` and pointing `template_name` at `base.html` means this view never goes through `render_template()` / `render_react_component()`. So neither `html` nor `renderStatic` is ever set in the context — and `base.html` only emits the `<div id="s2">` React mount point inside `{% if html and not renderStatic %}` / `{% elif renderStatic %}`.

I ran the branch against a real DB to confirm:

```
GET /login    -> 200, id="s2" occurrences: 0   <-- no React root
GET /register -> 200, id="s2" occurrences: 1
GET /texts    -> 200, id="s2" occurrences: 1
```

`inReaderApp: true` is still emitted, so `client.jsx` calls `ReactDOM.render(component, document.getElementById('s2'))` with `null` and throws. Any fresh load of `/login` (bookmark, refresh, external link, SSO redirect back) is a blank page.

`register()` below correctly calls `render_template(...)`, which is why `/register` works — that isolates the bug to this class. Suggested fix: route `CustomLoginView` through `render_template()` with `app_props` the same way `register()` does.

static/js/ReaderApp.jsx
```diff
@@ -119,11 +120,17 @@ class ReaderApp extends Component {
       translationLanguagePreference: props.translationLanguagePreference,
       editorSaveState: 'saved',
       notificationCount: props.notificationCount || 0,
+      showAuth:  ['/login', '/register'].includes(window.location.pathname),
```
yodem: **P0 — this breaks SSR for the entire site, not just the auth pages.**

This reads `window.location` **in the constructor**, unguarded. `master` has zero `window` access in this constructor — this is the first. But `ReaderApp` is server-rendered in Node: `node/server.js:93` does `ReactDOMServer.renderToString(ReaderApp(props))`.

I built the SSR bundle and confirmed:
- the compiled line `includes(window.location.pathname)` **is** present in the node-targeted `server-bundle.js`;
- there is **no** `window`/jsdom shim anywhere in `node/` or the bundle (`global.window =` / `jsdom` → 0 matches);
- `node -e "window.location.pathname"` → `ReferenceError: window is not defined`.

So `renderToString` throws on **every** SSR render. It won't 500 — `render_react_component` catches it and silently falls back to the loading placeholder — so this degrades quietly into "SSR is just off everywhere", which is easy to miss in review and painful for SEO/TTFB.

Fix: guard with `typeof window !== 'undefined'`, or better, derive `showAuth`/`authFlow`/`authNext` from `props.initialPath` / `props`, which are already passed in from Django and safe on both sides.

Separately, Copilot is right that the exact-match is wrong: `urls_shared.py` uses `re_path(fr'^login/?$')`, so `/login/` and `/register/` are genuinely reachable and won't match here.

static/js/auth/ChooseView.jsx
```diff
@@ -0,0 +1,210 @@
...
+  // click target over the custom Figma button after the async SDK script is ready.
+  useEffect(() => {
+    setGoogleReady(false);
+    if (!googleClientId) return;
+    const useRedirect = Sefaria.ssoUseRedirect();
+    if (useRedirect) {
+      Sefaria.ssoSetRedirectState(ssoRedirectState);
+    }
+    const stopWaiting = whenReady(
+      () => window.google?.accounts?.id && googleBtnRef.current,
```
yodem: **P0 — the Google and Apple SDKs are never loaded, so both provider buttons are permanently disabled.**

This polls for `window.google?.accounts?.id`, and the Apple effect polls for `window.AppleID?.auth`. But nothing on this branch ever loads those SDKs — searching the whole branch for `gsi/client`, `appleid.auth`, `accounts.google.com`, `appleid.cdn-apple` returns **zero** hits in any template or JS. The `base.html` diff only adds CSS.

So `whenReady` polls 80×100ms, gives up silently after 8s, `googleReady`/`appleReady` stay `false`, and both `ProviderButton`s render permanently `disabled`. The whole SSO surface of the PR is dead on arrival even before the missing `/api/auth/*` routes.

Needs the GIS + Apple JS `<script>` tags added to `base.html` (ideally only on the auth pages, and `async`/`defer`).

Minor, same file: `onSSOResult` omits the `X-CSRFToken` header that `EmailView` sends. Worth fixing now so it doesn't 403 the moment the backend lands.

static/css/auth.scss
```diff
@@ -0,0 +1,307 @@
...
+
+// Google GIS injects an iframe with inline width/height; !important overrides those
+// so the transparent overlay covers the full visual button and captures clicks correctly.
+.sefaria-provider-sdk-overlay {
+  position: absolute;
+  inset: 0;
+  z-index: 1;
+  opacity: 0.0001;
+  overflow: hidden;
+  pointer-events: none;
```
yodem: **Confirming Copilot here — this is real, and it's a second independent reason Google sign-in can't work.**

`pointer-events: none` on the wrapper is inherited by the GIS `<iframe>` inside it — the `> div, iframe` rule only overrides `width`/`height`/`margin`, and nothing re-enables pointer events (`pointer-events` appears exactly once in this file: here).

That matters because `ProviderButton` gives the Google button **no `onClick`** — per the comment in `ChooseView.jsx`, GIS has no programmatic trigger, so the invisible iframe overlay receiving the click is the *only* activation path. With pointer events off, nothing is clickable.

Note the comment two lines above says the overlay "captures clicks correctly" — which is exactly what this rule prevents. Looks like a late edit that was never re-tested.

static/js/auth/EmailView.jsx
```diff
@@ -0,0 +1,211 @@
...
+        const hasCaptchaError = !!(data?.captcha);
+        const nonCaptchaError = Object.keys(data || {}).some((key) => key !== '_auth' && key !== 'captcha');
+        setError(nonCaptchaError ? authError(data, message) : null);
+        if (hasCaptchaError) setCaptchaError(Sefaria._('Verify that you are not a robot'));
+        if (window.grecaptcha && captchaWidgetId.current !== null) {
+          try { window.grecaptcha.reset(captchaWidgetId.current); } catch (e2) { /* noop */ }
+          captchaToken.current = '';
+        }
+      } else {
+        const res = await fetch('/api/auth/login', {
```
yodem: **P0 — email login POSTs to a route that does not exist, and this PR deletes the page that currently handles login.**

There is no `/api/auth/login` in `sefaria/urls_shared.py` (searching the branch for `api/auth` returns nothing). `/api/login/` exists but is `TokenObtainPairView` — it mints JWTs and does not establish a Django session, so it isn't a drop-in.

The PR description says server endpoints are "not in this PR", but that framing only covers the *SSO callbacks*. This is ordinary email login, and `templates/registration/login.html` — which currently handles it — is deleted in this same PR. Net effect on merge: **nobody can log in.**

The blast radius goes past the login page: `e2e-tests/global-setup.ts` drives this exact form to capture storage state for every authenticated e2e profile, and explicitly hard-fails the whole suite if login doesn't complete. `e2e-tests/pages/loginPage.ts` also matches on `getByPlaceholder('Email Address')` and `getByRole('button', { name: 'Login' })`, whereas the new UI uses placeholder `you@example.com`, a button labelled "Log In", and hides the form behind a "Continue with Email" click. Those page objects need updating in this PR.

Simplest path that keeps this PR self-contained: POST form-encoded credentials to the existing `/login` (Django's `LoginView` already handles it and sets the session), the same way the register branch above reuses `/register` with `noredirect`.

static/js/auth/utils.js
```diff
@@ -0,0 +1,64 @@
...
+  if (!data || typeof data !== 'object') return null;
+  if (typeof data.error === 'string') return data.error;
+  for (const k of Object.keys(data)) {
+    if (k === '_auth') continue;
+    if (typeof data[k] === 'string') return data[k];
+  }
+  return null;
+}
+
+export function safeNext(next) {
```
yodem, line 41: **P1 — open redirect.** `next` is attacker-controlled (query string) and flows straight into `window.location.href` in both `EmailView` and `ChooseView`.

This regex only rejects a second literal forward slash, so it blocks `//evil.com` but **passes `/\evil.com`** — and both Chrome and Firefox normalize backslashes to forward slashes when parsing a URL, so that navigates to the protocol-relative `//evil.com`:

```js
safeNext("//evil.com")   // -> "/"            blocked
safeNext("/\\evil.com")  // -> "/\\evil.com"  PASSES -> lands on https://evil.com
```

So `sefaria.org/login?next=/%5Cevil.com` drops the user on an attacker's page immediately after a legitimate login — prime phishing setup, since the URL bar showed sefaria.org the whole way through.

The old flow POSTed `next` back to Django, which validated it server-side. Worth noting `url_has_allowed_host_and_scheme` is **already imported and used in `sefaria/views.py` (line 112)** — the correct pattern is right there. Either mirror it client-side (reject any `\`, then `new URL(next, location.origin).origin === location.origin`) or let the server own the post-auth redirect.

static/js/sefaria/strings.js
```diff
@@ -746,9 +746,48 @@ const Strings = {
...
+    "Send Reset Link": "שליחת קישור לאיפוס",
+    "Reset Link Sent": "קישור לעדכון סיסמה נשלח",
+    "Reset Password": "עדכון סיסמה",
+    "Forgot Password?": 'שכחת סיסמה?',
+    "Forgot password?": 'שכחת סיסמה?',
+    "Password": "סיסמה",
+    "First Name": "שם פרטי",
+    "Last Name": "שם משפחה",
+    "Don't have an account?": 'אין לך חשבון?',
+    "Already have an account?": 'יש לך חשבון?',
```
yodem: This key already exists on `master` and is used by the pre-existing sign-up modal in `Misc.jsx`. Since `_i18nInterfaceStrings` is one flat object literal, this later declaration silently wins — so merging this **changes the Hebrew text of that unrelated modal** as an unreviewed side effect.

Either drop the duplicate and reuse the existing string, or scope it under the `Auth` context block (the pattern already used correctly for "Sign Up"/"Log In").

While here: `"First Name"`/`"Last Name"` duplicate existing keys, `"Reset Password"` is unused, and the error strings routed through `authError()` can never match — Django has already translated those messages server-side by the time they reach `Sefaria._()`, which keys on English.

templates/base.html
```diff
@@ -115,6 +115,8 @@
     <link rel="stylesheet" href="{%  static 'css/themes/library-theme.css' %}">
     <link rel="stylesheet" href="{%  static 'css/themes/sheets-theme.css' %}">
     <link rel="stylesheet" href="{%  static 'css/common-components.css' %}">
+    <link rel="stylesheet" href="{%  static 'css/common-component.css' %}">
```
yodem, line 118: Two things here:

1. `common-component.css` (singular) sits directly above the pre-existing `common-components.css` (plural), and both are now loaded globally. Near-identical names one line apart is a maintenance trap — worth renaming to something unambiguous.
2. This is where the missing Google/Apple SDK `<script>` tags need to go (see my comment on `ChooseView.jsx`) — without them the SSO buttons never enable.

Also: `auth.css` is loaded on **every** page sitewide for a feature used on two. Worth gating behind the auth views if that's easy.

yodem (changes requested):
Reviewed the branch and **ran it locally** (real DB + built SSR bundle) rather than only reading the diff. Flagging as request-changes: as it stands the auth page is non-functional and it takes existing login down with it.

**Verified by running the branch**

| Path | HTTP | `id="s2"` mount point present? |
|---|---|---|
| `/login` | 200 | **0 — absent** |
| `/register` | 200 | 1 |
| `/texts` | 200 | 1 |

So a fresh load of `/login` has no React root at all (details inline on `sefaria/views.py`).

**The four blockers**
1. `/login` renders a blank page — `CustomLoginView` no longer goes through `render_template()`.
2. SSR is dead **site-wide** — `window` is read in the `ReaderApp` constructor, which runs in Node.
3. SSO can't work for three independent reasons — the Google/Apple SDK `<script>` tags are never loaded anywhere, `pointer-events: none` blocks the GIS overlay, and `/api/auth/*` doesn't exist.
4. Email login POSTs to a route that doesn't exist, while this PR deletes the `login.html` that currently handles it. This also breaks `e2e-tests/global-setup.ts`, which logs in through the real `/login` form to mint storage state for **every** authenticated profile and hard-fails the suite if it can't. No tests or page objects were updated, and no tests were added — CI is green only because the auth suite doesn't run there.

Plus a P1 open redirect (inline on `utils.js`).

**Credit where due:** RTL is handled properly (logical `inset-inline-*` properties throughout), the password toggle is a correct `type="button"` with proper `htmlFor`/`aria-invalid`/`role="alert"` wiring, and committing the compiled `.css`/`.css.map` does match existing repo convention. Copilot's `Divider.jsx` propTypes nit and its doc-comment comment are noise — safe to ignore.

static/css/auth.scss
```diff
@@ -0,0 +1,307 @@
...
+
+  > div,
+  iframe {
+    width: 100% !important;   // overrides Google SDK inline style
+    height: 100% !important;  // overrides Google SDK inline style
+    margin: 0 !important;     // overrides Google SDK inline style
+  }
+}
```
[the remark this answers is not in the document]
yodem: Confirmed — this is a genuine feature-killer, not a nit. `ProviderButton` gives the Google button **no `onClick`**; per the comment in `ChooseView.jsx`, GIS exposes no programmatic trigger, so the invisible iframe overlay receiving the click is the *only* activation path. `pointer-events` appears exactly once in `auth.scss` (this line), and the `> div, iframe` rule only overrides width/height/margin — nothing re-enables it. Also note the comment directly above claims the overlay "captures clicks correctly", which this rule prevents.

Worth flagging that even with this fixed, Google sign-in still can't work: the GIS SDK script is never loaded anywhere on the branch, and `/api/auth/google/callback` doesn't exist. Three independent blockers.

static/js/common/Input.jsx
```diff
@@ -0,0 +1,148 @@
...
+        <div className="sefaria-input-labelRow">
+          {label ? <label className="sefaria-input-label" htmlFor={inputId}>{label}</label> : <span />}
+          {trailingLink && (
+            <a
+              className="sefaria-input-trailingLink"
+              href={trailingLink.href}
+              onClick={trailingLink.onClick}
+            >
+              {trailingLink.text}
+            </a>
```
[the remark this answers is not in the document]
yodem: Confirmed. `EmailView` passes `passwordTrailingLink: { text, onClick }` with no `href`, so React renders a bare `<a onClick>` — not in the tab order, so keyboard users cannot reach "Forgot password?" at all.

Cleanest fix given the file already has the pattern: render a `<button type="button">` when `href` is absent, mirroring the reveal-password toggle a few lines down (which is correctly implemented).

static/js/ReaderApp.jsx
```diff
@@ -119,11 +120,17 @@ class ReaderApp extends Component {
...
+      authFlow:  window.location.pathname === '/register' ? 'register' : 'login',
+      authNext:  new URLSearchParams(window.location.search).get('next') || '/',
```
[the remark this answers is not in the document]
yodem: Confirmed — `urls_shared.py` uses `re_path(fr'^login/?$')` / `^register/?$`, so `/login/` and `/register/` really are reachable and won't match this exact-string check.

More urgent on this same line though: `window.location` is read **in the constructor**, which also runs under SSR in Node. I built the SSR bundle and confirmed the compiled line ships in the node-targeted `server-bundle.js`, there's no `window`/jsdom shim anywhere in `node/`, and `window` in Node throws `ReferenceError`. Since `node/server.js:93` does `renderToString(ReaderApp(props))`, this breaks SSR for **every page on the site**, not just auth — silently, because Django falls back to the loading placeholder rather than 500ing. Details in my inline review comment.

static/js/auth/EmailView.jsx
```diff
@@ -0,0 +1,211 @@
...
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json', 'X-CSRFToken': csrf },
+          body: JSON.stringify({ email: fields.email, password: fields.password }),
+        });
+        const data = await res.json().catch(() => ({}));
+        if (res.ok) { window.location.href = safeNext(next); return; }
+        setError(authError(data, 'Email and/or password are incorrect'));
```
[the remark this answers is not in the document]
yodem: Confirmed, and it's worse than a 404: this PR also **deletes** `templates/registration/login.html`, the page that currently handles login. So on merge, nobody can log in.

For the record `/api/login/` does exist but is `TokenObtainPairView` (JWT) — it doesn't establish a Django session, so it's not a drop-in substitute. It also breaks `e2e-tests/global-setup.ts`, which drives this form to mint storage state for every authenticated e2e profile and hard-fails the suite if login doesn't complete.

static/js/auth/ChooseView.jsx
```diff
@@ -0,0 +1,210 @@
...
+        try {
+          const config = {
+            client_id: googleClientId,
+            ux_mode: useRedirect ? 'redirect' : 'popup',
+          };
+          if (useRedirect) {
+            config.login_uri = `${window.location.origin}/auth/google/redirect`;
+          } else {
+            config.callback = (resp) => onSSOResult('/api/auth/google/callback', { credential: resp.credential });
+          }
```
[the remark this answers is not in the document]
yodem: Confirmed on the missing routes. Adding one you didn't catch: the Google Identity Services SDK script is **never loaded** anywhere on this branch (no `gsi/client` reference in any template or JS), so `window.google` never appears, `whenReady` times out after 8s, and the button stays permanently `disabled` — the flow dies before it ever reaches a callback URL. Same for Apple (`appleid.auth`).

static/js/auth/Divider.jsx
```diff
@@ -0,0 +1,15 @@
...
+ * Divider — a horizontal rule with centered text ("or" / "או"), used between the
+ * SSO buttons and the email button on the auth choose screen. Figma `Form Card`.
+ */
+const Divider = () => (
+  <div className="sefaria-divider" role="separator"><InterfaceText>or</InterfaceText></div>
+);
+
+Divider.propTypes = { children: PropTypes.node };
+
+export default Divider;
```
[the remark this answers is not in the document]
yodem: Non-issue — I'd skip this one. `PropTypes` is not unused (`Divider` is trivial, and an extraneous propTypes entry is harmless); this is lint noise, not a defect worth a round-trip on a PR with functional blockers.

sso/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    # Google One Tap redirect mode: Google POSTs credential + g_csrf_token here
+    path("auth/google/redirect", views.google_redirect),
+    # Apple Sign In popup mode: ChooseView.jsx POSTs id_token + name here
+    path("api/auth/apple/callback", views.apple_callback),
```
yitzhakc, line 11: I don't really understand why some urls are prefixed with api and some aren't. Let's keep it consistent?

sefaria/views.py
yitzhakc: Not sure why it commented this, you use `render_template` on line 100?

sso/views.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if not id_token:
+        return JsonResponse({"error": "id_token required"}, status=400)
+
+    adapter = get_social_adapter(request)
+    try:
+        provider = adapter.get_provider(request, "apple")
+        sociallogin = provider.verify_token(request, {"id_token": id_token})
+    except Exception as e:
+        logger.warning("Apple token verification failed", error=str(e))
```
yitzhakc: If this fails, I think we want to have that tracked somewhere beyond the logs.. Push to celery? Sentry? Slack?

sefaria/views.py
YishaiGlasner: @yodem i ignore all your (claude) comments for now - some of them are based on the fact that there was no server side, some claim that working things are not working.

sso/views.py
YishaiGlasner: i'll throw error to logger for some scenatios (and that will go to sentry)

yodem:
Raising this as a finding rather than a change — it's your branch and your call.

While adding `sso_only_account` to `api/login/` for the mobile app (#3580), I found **four** implementations of "which SSO providers is this account linked to" on this branch:

| | |
|---|---|
| `sso/views.py:email_login` | inline `socialaccount_set` query + `has_usable_password()` gate |
| `sefaria/forms.py:clean_email` | its own `SocialAccount.objects.filter(user=user)` query |
| `reader/views.py` (account settings) | `socialaccount_set.values_list('provider', flat=True)` |
| `static/js/auth/RegisterView.jsx` | string-matches `clean_email`'s English back into `{code, providers}` |

The last pair is the one worth a look. `clean_email` holds structured `SocialAccount` rows and flattens them into English prose:

```python
raise forms.ValidationError("This email address is already registered via Google Sign-In.")
```

and the client regex-reconstructs the structure that was just discarded:

```js
'This email address is already registered via Google Sign-In.': { code: 'sso_only_account', providers: ['google'] },
```

**The Sefaria mobile app now does the same string-match**, against the same sentences. So one English sentence is load-bearing API for two clients in two repos, with no test or type that would catch a copy edit — a wording change in `forms.py` breaks both silently.

Notably, **login already does this correctly** — `email_login` returns structured `_auth: {code, providers}` and `ErrorBanner` reads it. Register is the only path that round-trips through prose.

A possible shape, if you think it's worth it:

- one `linked_providers(user)` primitive behind all the backend call sites
- `clean_email` raises with `code='sso_only_account'` and the providers in `params`, so the register response can carry the same `_auth` shape login already returns
- keep the English message alongside it, so clients that predate `_auth` keep working
- keep `account_exists` (a plain password account) distinct — it must never report as `sso_only_account`, or the UI tells the user to click a Google button that can't help them
- both clients then read `_auth.code` and their English lookup tables go away

I had implemented this and have since removed it from #3580 — it was out of scope for a mobile fix and it's your branch's code. Happy to open it as a separate PR against `sso` if you'd like it, or to leave it entirely.

yodem (commented):
## 🤖 Automated review by Claude Code

**This review was generated by Claude (Claude Code), run by @yodem — not a human line-by-line read.** Treat it as a prioritised starting point: every finding cites file:line and was checked against source, but please verify before acting, especially the two items explicitly marked ⚠️ unverified.

Posted as **Comment**, not Request Changes — the merge decision is the team's.

### How this was produced

Seven parallel review agents over the full diff (95 files, +6324/−786) at `d2709af6`, plus the team wiki for prior decisions:

| Track | Focus |
|---|---|
| security-auditor | authn/authz, token verification, CSRF, redirects, secrets |
| code-quality-reviewer | correctness, error paths, transactions, module boundaries |
| frontend-ui-developer | AuthPage state machine, SDK loading, a11y, i18n, analytics |
| test-generator | coverage gaps, mock-vs-behaviour, actually ran the JS suite |
| CandleKeep → OWASP WSTG v4.2 | grounded the enumeration/session/redirect findings in a standard |
| general reviewer | plan alignment, production readiness, scope discipline |
| verification pass | re-derived every 🔴 firsthand against real allauth 65.14.3 source |

Wiki context used: `SSO Integration POC` (the 07-22 web-session/mobile-JWT decision, the 07-30 gap analysis) and `Sefaria Auth Service Decoupling`.

### Findings

**🔴 Blockers (4)** — open redirect via `?next=` (`ALLOWED_HOSTS=["*"]` makes allauth's `is_safe_url` a no-op); `pre_social_login` reimplementing `wipe_password` without its two safety gates; account-enumeration oracle on password reset; `simplejwt` CVE.

**🟡 Should fix (9)** — missing transaction in `save_user`; unpinned `django-allauth`; two WCAG 2.4.3 issues; silent SSO dead-end when SDKs are blocked; `authSource` lost on Back; no FedCM opt-in; Gravatar timeout; the i18n/string-match conflict.

### ✅ Verified correct — worth stating

Negative results matter, and the highest-risk area checked out:

- **ID-token verification is sound.** Traced into real allauth 65.14.3: signature via fresh HTTPS JWKS, `verify_iss` / `verify_aud` / `verify_exp` all enabled, plus a `jti` replay cache.
- **The Apple audience widening in `d2709af6` is safe** — it adds exactly two Sefaria-owned identifiers, not a wildcard.
- **CSRF posture is right** — mobile endpoints are `@csrf_exempt` but token-authed and session-flushed; cookie-based endpoints keep protection.
- **Session flush on JWT issuance is real and genuinely tested** (the tests establish a real session first, so they'd actually fail if it regressed).
- i18n key parity 45/45 en↔he · no secrets reach the browser · no dangling refs to the four deleted templates · the `LanguageSettingsMiddleware` refactor is behaviour-preserving.
- **78/78 JS auth tests pass** (actually executed). ⚠️ **Python tests were not run** — no Django available in the review environment. Not claimed as passing; needs CI.

### Architecture

The 2026-07-22 decision holds without drift: `_social_login_or_error()` is a single shared verification path, with thin wrappers differing only in session-vs-JWT handling. The `sso/` ↔ `sefaria/` coupling was raised as an extraction blocker but per `Sefaria Auth Service Decoupling` no extraction decision has been made and "leave it segmented in code" is an explicit live option — so it's a note, not a defect.

One process observation: the password-wipe (🔴 #2) is now the **third** time it's been flagged, after the 07-30 design-gap analysis. This PR adds the account-settings display but not a warning at the moment the password is destroyed. Worth an explicit accept-or-fix.

---

sefaria/urls_shared.py
```diff
@@ -16,6 +16,10 @@
 from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
 
 shared_patterns = [
+    path('_allauth/', include('allauth.headless.urls')),
+    path('accounts/', include('allauth.urls')),
```
yodem, line 20: 🤖 *Claude Code review*

### 🔴 Open redirect — the `next` guard is a no-op in production

Mounting the full allauth URL tree makes a pre-existing setting security-load-bearing:

```
helm .../local-settings-file.yaml:37   ALLOWED_HOSTS = ["*"]
        ↓
allauth/account/adapter.py:604         allowed_hosts = {get_host()} | set(ALLOWED_HOSTS)   ← "*"
allauth/account/adapter.py:618-619     if validate_host(parsed_host, allowed_hosts):
                                            allowed_hosts.add(parsed_host)   ← ANY host passes
        ↓
allauth/account/utils.py:42-45         redirect_to = get_request_param(request, "next")
                                       if not is_safe_url(redirect_to): redirect_to = None
```

Django's `validate_host()` treats `"*"` as matching any host, so `is_safe_url()` returns `True` for any external netloc. Because this line mounts allauth's classic views, `next` is attacker-supplied **directly via the URL** — no cookie-write primitive needed:

`https://www.sefaria.org/accounts/google/login/?next=https://evil.com`

...lands a freshly-authenticated user on an attacker page. Strong phishing primitive precisely because the login genuinely succeeded.

`sefaria/views.py:304` (added in this PR) already has the right pattern — `allowed_hosts={request.get_host()}` — which sidesteps `ALLOWED_HOSTS` entirely.

**Fix:** override `is_safe_url` in `SefariaAccountAdapter` with that same host-scoped check, and/or replace the `["*"]` wildcard with the domains already enumerated in `CSRF_TRUSTED_ORIGINS` immediately below it in the configmap.

⚠️ This is a **static trace** through allauth 65.14.3 source, not a live reproduction — worth a 30-second confirmation on a cauldron before acting.

sso/adapters.py
```diff
@@ -0,0 +1,152 @@
...
+        email = (sociallogin.user.email or "").lower()
+        if not email:
+            return
+        try:
+            existing_user = User.objects.get(email__iexact=email)
+            if existing_user.has_usable_password():
+                existing_user.set_unusable_password()
+                existing_user.save(update_fields=["password"])
+        except User.DoesNotExist:
+            pass
```
yodem: 🤖 *Claude Code review*

### 🔴 This reimplements `wipe_password()` and drops both of allauth's safety gates

allauth already ships exactly this behaviour, deliberately gated twice:

| Gate | allauth | this hook |
|---|---|---|
| provider email must be `verified` | `socialaccount/models.py:387` — `[e.email for e in self.email_addresses if e.verified]` | ❌ bare `User.objects.get(email__iexact=…)` |
| skip if local email already verified | `internal/flows/email_authentication.py` — `if address and address.verified: return` | ❌ wipes unconditionally |

Two consequences:

**1. Silent data loss (happens today, every time).** Any existing password user who signs in with Google permanently loses their password — *including* the case allauth explicitly declines to touch, whose own comment reads `# Verified email address, no reason to worry.`

**2. Targeted lockout.** If a provider returns `email_verified=false` (Google does for some Workspace-provisioned accounts), allauth refuses to link and `_social_login_or_error` correctly falls through to `social_signin_failed` — the attacker gains nothing. But this hook has *already* destroyed the victim's password. Denial of access to them, no access for anyone.

The data needed is right there: `providers/google/provider.py:89` populates `verified` from the `email_verified` claim, and `sociallogin.email_addresses` is in scope.

**Fix:** simplest is to delete this hook and let allauth's gated `wipe_password` do the job. If it must stay, gate on `{e.email.lower() for e in sociallogin.email_addresses if e.verified}` and add a test asserting an unverified-email collision leaves the password intact.

📎 Context: this is the third time this has been raised — the 2026-07-30 SSO design-gap analysis flagged the missing point-of-action warning, and this PR adds the account-settings display but not a warning at the moment the password is destroyed. Worth an explicit accept-or-fix rather than carrying it forward again.

sso/views.py
```diff
@@ -0,0 +1,255 @@
...
+    if u.has_usable_password() or not u.socialaccount_set.exists():
+        return None
+    providers = list(u.socialaccount_set.values_list("provider", flat=True))
+    return JsonResponse(
+        {
+            "error": "auth.generic_error",
+            "_auth": {"code": "sso_only_account", "providers": providers},
+        },
+        status=401,
+    )
```
yodem, line 52: 🤖 *Claude Code review*

### 🔴 Account-enumeration oracle (leaks existence *and* provider)

As used by `password_reset_api`, this produces three distinguishable outcomes:

| Input | Response |
|---|---|
| nonexistent address | `200 {}` |
| normal password account | `200 {}` |
| **SSO-only account** | **`401`** + `{"code":"sso_only_account","providers":["google"]}` |

An attacker POSTs a candidate list and learns which addresses have Sefaria accounts and which OAuth provider each uses. `grep -rn "ratelimit\|throttle" sso/ sefaria/settings.py` finds nothing, so it scales.

**OWASP WSTG v4.2, WSTG-IDNT-04 (p. 39)** names password-recovery responses as the canonical enumeration oracle and sets the bar explicitly: *"the application should respond with the same error message and length to the different incorrect requests."* WSTG-ATHN-09 (p. 50) adds *"Does the reset process allow user enumeration?"* — note the criterion is **any** differentiating signal: status, body, length, or timing.

This also regresses the PR's own stated contract — `sso/tests/views_test.py:332` asserts *"the API must not leak whether the address exists"*.

**Fix that keeps the UX:** always return `200 {}`, and move the "this account uses Google Sign-In" guidance into the **email** — an out-of-band channel reaches the legitimate owner without answering the attacker's question.

Same class, smaller blast radius: `sefaria/forms.py:80-86`.

sso/views.py
```diff
@@ -0,0 +1,255 @@
...
+
+from emailusernames.utils import user_exists, get_user
+from sefaria.forms import SefariaPasswordResetForm
+
+logger = structlog.get_logger(__name__)
+
+
+def _jwt_for_user(user):
+    refresh = RefreshToken.for_user(user)
+    return {"access": str(refresh.access_token), "refresh": str(refresh)}
```
yodem, line 26: 🤖 *Claude Code review*

### 🔴 `simplejwt==5.3.1` — CVE-2024-22513, newly exercised by this path

`RefreshToken.for_user()` performs no `user.is_active` check, so tokens minted here stay valid after an account is deactivated or banned, until natural expiry. With `ACCESS_TOKEN_LIFETIME = timedelta(days=1)` that's a 24-hour window in which a disabled account keeps working — and this PR is what puts that call on a live, user-reachable path (`google_mobile` / `apple_mobile`).

**Fix:** bump to `djangorestframework-simplejwt>=5.5.1` and re-run `sso/tests/views_test.py`.

⚠️ Advisory reported by tooling and **not independently verified** in this review — please confirm the affected/fixed ranges before acting.

sso/adapters.py
```diff
@@ -0,0 +1,152 @@
...
+                user.email,
+                first_name=user.first_name,
+                last_name=user.last_name,
+                lang=getattr(request, "LANGUAGE_CODE", "en"),
+                educator=False,
+            )
+        except Exception:
+            logger.error("SSO: failed to create CRM user", exc_info=True)
+
+        return user
```
yodem, line 185: 🤖 *Claude Code review*

### 🟡 No transaction — a Mongo failure orphans a live Postgres user

`super().save_user()` has already committed the Django `User` and `SocialAccount` rows. If `assign_slug()`, `join_invited_collections()`, or `p.save()` raises, that exception propagates with **no rollback** — there's no `ATOMIC_REQUESTS` in settings and allauth doesn't wrap adapter `save_user()` in an atomic block.

Result: an account that can authenticate via SSO but has no `UserProfile`, which will break anywhere the app assumes every authenticated user has one.

The email-registration path already gets this right — `sefaria/views.py:236-249` wraps `form.save()` and `p.save()` together in `with transaction.atomic():`. The SSO path should get the same protection (or delete the orphaned `User` and re-raise).

Only the CRM call is currently guarded, and that part is correct — the `try/except` + test coverage for a CRM outage is good.

sso/adapters.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Attempt to fetch the user's Gravatar and upload it to Google Storage.
+    Used by both SSO registration (sso/adapters.py) and email registration
+    (sefaria/views.py process_register_form) so both registration paths get
+    the same treatment.
+    """
+    email_hash = hashlib.md5(profile.email.lower().encode("utf-8")).hexdigest()
+    gravatar_url = f"https://www.gravatar.com/avatar/{email_hash}?d=404&s=250"
+    try:
+        with urllib.request.urlopen(gravatar_url) as r:
```
yodem: 🤖 *Claude Code review*

### 🟡 `urlopen` with no `timeout=`, now on two signup paths

No timeout is set, so a slow or hanging `gravatar.com` blocks the registration request. Pre-existing behaviour, but this refactor widens the blast radius: both `process_register_form` (email signup) and `SefariaSocialAccountAdapter.save_user` (every new SSO signup) now call it synchronously in the request path — and per the `save_user` comment above, it blocks while a Django `User` row is already committed and unprotected.

**Fix:** `urllib.request.urlopen(gravatar_url, timeout=5)`.

requirements.txt
```diff
@@ -89,6 +89,8 @@ django==6.0.4
     #   djangorestframework
     #   djangorestframework-simplejwt
 django-admin-sortable==2.3
+django-allauth[socialaccount]
```
yodem: 🤖 *Claude Code review*

### 🟡 Unpinned — the only auth-critical dependency without a version

Every other direct dependency in this pip-compile output is pinned (`django==6.0.4`, `djangorestframework-simplejwt==5.3.1`, …). This one isn't, which makes the build non-reproducible for the package now sitting on the authentication critical path — a routine `pip-compile` re-run could pull a new major with different adapter-hook semantics or changed security defaults, with no review.

Currently resolves to `65.14.3`.

**Fix:** pin it explicitly, e.g. `django-allauth[socialaccount]==65.14.3`.

static/js/auth/ProviderButton.jsx
```diff
@@ -0,0 +1,80 @@
...
+        id={id}
+        ref={trackingRef}
+        className={`sefaria-provider-button-shell${disabled ? ' is-disabled' : ''}`}
+        tabIndex={-1}
+      >
+        <div className="sefaria-provider-button" aria-hidden="true">
+          {content}
+        </div>
+      </div>
+    );
```
yodem, line 55: 🤖 *Claude Code review*

### 🟡 WCAG 2.4.3 (Focus Order) — tab order doesn't match visual order

When rendered with `trackingRef` (every Google button — `ChooseView.jsx:57-64`), this renders a `tabIndex={-1}` wrapper around `aria-hidden="true"` content: entirely non-interactive and invisible to assistive tech.

The **real** interactive control is `overlayNode` (`useSsoSignIn.jsx:323-327`) — a `position: fixed` div holding Google's iframe button, appended last in `AuthPage.jsx:146` and positioned visually via `getBoundingClientRect()`.

So visually the Google button sits at the top of the card, but in DOM/tab order it is the **last** focusable element on the page. A keyboard user tabs into an inert `aria-hidden` "Google" button first, then has to traverse Apple, the divider, "Continue with Email", and every `LegalText` link before reaching a Google control that actually works.

**Fix:** put the real overlay in DOM order at the button's position, or manage `tabindex` so focus order matches the visual order.

static/js/auth/AuthPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+    content = (
+      <ChooseView
+        flow={flow} switchFlow={switchFlow}
+        onEmailClick={onEmailClick}
+        googleReady={googleReady} appleReady={appleReady} ssoLoading={ssoLoading}
+        registerGoogleTarget={registerGoogleTarget} triggerApple={triggerApple}
+        setActiveErrorHandler={setActiveErrorHandler}
+      />
+    );
+  }
```
yodem, line 144: 🤖 *Claude Code review*

### 🟡 No focus management across view transitions

Grepping the whole `auth/` tree for `.focus(` / `autoFocus` returns nothing outside `role="alert"` containers. Each swap in this chain (`choose → email`, `email → forgot`, `forgot → forgot-sent`, …) is functionally a page navigation, but focus is never moved to the new heading or first field.

For screen-reader and keyboard users the content silently changes with no cue that anything happened, and focus can be left on a node that no longer exists.

**Fix:** on each `view` change, move focus to the new card's heading (`tabIndex={-1}` + `.focus()`), which also gives SR users an announcement of the new context.

static/js/auth/utils.js
```diff
@@ -0,0 +1,147 @@
...
+    if (check()) { cb(); return; }
+    if (--tries <= 0) return;
+    timer = setTimeout(tick, 100);
+  };
+  tick();
+  return () => {
+    cancelled = true;
+    if (timer) clearTimeout(timer);
+  };
+}
```
yodem, line 25: 🤖 *Claude Code review*

### 🟡 SSO dead-ends silently when the SDK is blocked

`whenReady` polls 80 × 100 ms = 8 s and then just stops — no error, no callback, no state change. `useSsoSignIn.jsx` leaves `googleReady`/`appleReady` at `false` forever, and `ProviderButton` renders `disabled` with no message.

Google and Apple identity SDKs are common ad-blocker/tracking-protection targets, and this is also the offline and flaky-network path. Those users see two permanently greyed-out buttons and no explanation — with no hint that "Continue with Email" is the working path.

**Fix:** surface a timeout state — hide the SSO buttons entirely, or show a message like `Google Sign-In couldn't load — continue with email below`.

static/js/ReaderApp.jsx
```diff
@@ -454,6 +471,11 @@ class ReaderApp extends Component {
 
   makeHistoryState() {
     // Returns an object with state, title and url params for the current state
+    if (this.state.showAuth) {
+      const { authPath } = this.state;
+      return { state: { showAuth: true, authPath, panels: [] }, url: authPath, title: document.title };
```
yodem: 🤖 *Claude Code review*

### 🟡 `authSource` survives Back navigation → wrong funnel attribution

`makeHistoryState` pushes `{ showAuth, authPath, panels }` and never includes `authSource`. `handlePopState` does `this.setState(newState)` — a **shallow merge** — so on Back, `authSource` is not reset and silently retains its previous value.

That value flows straight into analytics: `AuthPage.jsx:56` → `useSignUpTracking({ flow, source: authSource })` → `fireFlowStarted(flowId, source)`. A user who reaches `/register` by backing into it gets attributed to whatever CTA they used earlier.

Not user-visible, which is what makes it worth catching now — it silently corrupts the signup-funnel data this PR is being built to measure.

**Fix:** include `authSource` in `makeHistoryState`, or explicitly reset it in the popstate handler.

static/js/auth/GoogleOneTap.jsx
```diff
@@ -0,0 +1,83 @@
...
+    const initOneTap = () => {
+      timeoutId = setTimeout(() => {
+        if (hasInterruptiveUI()) { markShown(); return; }
+        const flowId = makeUuid();
+        const attemptId = makeUuid();
+        window.google.accounts.id.initialize({
+          client_id: googleClientId,
+          callback: (resp) => handleCredential(resp, flowId, attemptId),
+        });
+        window.google.accounts.id.prompt();
```
yodem: 🤖 *Claude Code review*

### 🟡 No FedCM opt-in + the auth-path guard is checked at the wrong time

**FedCM (the more urgent half).** `initialize()` is called without `use_fedcm_for_prompt: true`, and `renderButton` in `useSsoSignIn.jsx:198-203` has no `use_fedcm_for_button`. Google's migration guide makes the opt-in required and warns that prompt-UI-status APIs may stop functioning once FedCM becomes mandatory. Neither file references FedCM at all — this is a scheduled-breakage item, not a style nit.

**The path guard.** Line 18 checks `AUTH_PATHS.has(pathname)` once at effect-setup, and the deps array is `[googleClientId]` so it never re-runs on SPA navigation. The `setTimeout` here doesn't re-check the path before calling `prompt()`. If the user client-side-navigates to `/login` inside the 1200 ms window, One Tap fires on top of `AuthPage`.

To be fair to the implementation, I'd rate that second point **low-to-medium**, not high: the window is 1200 ms and `wasShown()` caps it at once per session. Still a one-line fix — re-check `window.location.pathname` inside the timeout callback.

sefaria/forms.py
```diff
@@ -76,7 +77,13 @@ def clean_email(self):
...
             user = get_user(email)
             if not user.groups.filter(name=SEED_GROUP).exists():
-                raise forms.ValidationError(_("A user with that email already exists."))
+                social = SocialAccount.objects.filter(user=user)
+                if social.filter(provider='google').exists():
+                    raise forms.ValidationError("This email address is already registered via Google Sign-In.")
+                elif social.filter(provider='apple').exists():
+                    raise forms.ValidationError("This email address is already registered via Apple Sign-In.")
+                else:
+                    raise forms.ValidationError("An account with this email address already exists.")
```
yodem: 🤖 *Claude Code review*

### 🟡 Two i18n pressures in conflict here

**(a)** These three messages lost the `_()` wrapper the message they replaced had (`_("A user with that email already exists.")`), so they'll never translate for Hebrew-interface users.

**(b)** They *can't* simply be wrapped, because `RegisterView.jsx:17-21` (`EMAIL_EXISTS_ERRORS`) pattern-matches on their **literal English text**. Adding `_()` would silently break that match for every non-English interface — and `sefaria/tests/forms_test.py` runs in the default locale, so it would still pass.

So the current state is load-bearing but undocumented, and the obvious "i18n consistency" cleanup is a trap with no test to catch it.

**Fix:** give these a stable error `code` and match on that — exactly what the adjacent `BACKEND_MESSAGES`/`required` mapping already does per its own comment at `RegisterView.jsx:24-25`. Then `_()` can be restored. Failing that, add a comment on both sides warning against localizing these three strings.

sso/views.py
YishaiGlasner: this is the product decision

sso/adapters.py
YishaiGlasner: the problem here is not we wipe password, but we don't logging the user in. this is the product decision (which can be questioned). i'm fixing it so unverified google login will work as verified.

sso/views.py
YishaiGlasner: Claude:
Traced this through the installed allauth version: _jwt_for_user is only reachable via _social_login_or_error, which calls complete_social_login. That routes through DefaultAccountAdapter.pre_login() (allauth/account/adapter.py:486), which checks if not user.is_active: return self.respond_user_inactive(...). flows/login.py:perform_login() returns that response immediately without calling adapter.login(), so Django session auth never fires for a deactivated user. Back in sso/views.py:80-81, _social_login_or_error checks request.user.is_authenticated and returns auth.social_signin_failed before _jwt_for_user is ever called.

So a deactivated account can't reach _jwt_for_user through this path — allauth's own pre_login hook already gates on is_active upstream. The CVE-2024-22513 concern doesn't apply here (we do have account deactivation via the spam-ban flow in sefaria/views.py:1414 / sefaria/model/user_profile.py:789, but this SSO path is already protected against it).

static/js/auth/utils.js
YishaiGlasner: i don't think, that needs fix. we can ask product what they want but it seems too edge case to address (when the current state is not so bad)

yodem (approved):

PR #3516, opened 2026-07-13
Feature/sc 45606/clean up elastic search when entities change

stevekaplan123:
## Description
This implements a number of dependencies functions that clean up and reindex elastic search when an entity is saved/renamed/deleted.  It also adds tests that handle each situation.

## Tests
1. Book rename. Create a small test book, put its segment docs in the fake ES, then rename the book. Verify: every doc under the old title got deleted, every segment got re-indexed under the new title, the book entry moved to its new id with the right breadcrumb path, and the delete happened before the re-add (the ordering the design depends on).
2. Book create/delete. Creating a book puts one doc in the book index (with its categories and author names filled in); deleting the book removes it.
3. Topic create/delete. Same, for topics — run three times, once for each topic class (Topic, PersonTopic, AuthorTopic). This matters because the notification system dispatches on the exact class, so each one needs its own subscription; if someone "simplifies" that, authors would silently stop syncing.
4. Category rename. Rename a category and verify its books' docs get their new path — and that the fix came only from the category hook (a spy proves the per-book save hook never fired, since the cascade suppresses it).
5. Version rename. Rename a version and check old docs deleted, new ones indexed — guards the refactoring the commit did to pre-existing code.
6. Failure tolerance. Make one segment fail to index mid-rename: the rest must still index and the failure must be logged, not raised. And deleting a topic that was never indexed must shrug off the "not found" error.


sefaria/helper/search.py
```diff
@@ -178,6 +182,8 @@ def normalize_linked_ref_filters(refs):
     """
     Expand raw refs into normalized segment refs for linked_refs search filters.
     """
+    from sefaria.model import Ref
```
saengel, line 185: Why is it necessary to move this import here?

sefaria/helper/search.py
```diff
@@ -594,3 +600,116 @@ def get_elasticsearch_client_for_indexer():
...
+# --------------------------------------------------------------------------- #
+
+def process_index_title_change_in_search(indx, **kwargs):
+    from sefaria.settings import SEARCH_INDEX_ON_SAVE
+    if SEARCH_INDEX_ON_SAVE:
+        from sefaria.model import library, text
+        from sefaria.search import delete_version, TextIndexer, get_new_and_current_index_names
+        search_index_name = get_new_and_current_index_names("text")['current']
+        old_title, new_title = kwargs.get("old"), kwargs.get("new")
+        text_index = library.get_index(new_title)
```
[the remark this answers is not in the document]
saengel: @stevekaplan123 this looks reasonable to me

sefaria/helper/search.py
stevekaplan123: I don't know why it chose to move imports but it seems harmless

stevekaplan123: I assume there may be a case where it was worried about circular imports.

saengel (approved):

PR #3522, opened 2026-07-14
Exact match toggle for sources

saengel:
## Description
Pulling the exact match checkbox out of SearchFilter into an independent component controlled by the SearchPage component, and creating a new component for the toggle and animation. 

## Code Changes
1. `static/css/s2.css` - Styling, including animation on the toggle
2. `static/js/SearchFilters.jsx` - Extracting existing exact match
3. `static/js/SearchPage.jsx` - Rendering the toggle with the exact match logic
4. `static/js/SearchToggle.jsx` - New search toggle component

### DIFF
--- a/static/css/s2.css
+++ b/static/css/s2.css
@@ -4043,7 +4043,7 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
   align-items: center;
   padding: 4px;
   background: #fafafa;
-  border: 1px solid #ededec;
+  border: 1px solid var(--lighter-grey);
   border-radius: 9999px;
   --english-font: var(--english-sans-serif-font-family);
   --hebrew-font: var(--hebrew-sans-serif-font-family);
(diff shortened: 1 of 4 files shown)


static/css/s2.css
```diff
@@ -4037,6 +4037,48 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
...
   border-bottom: 4px solid var(--medium-grey);
 }
+.searchToggle {
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+  padding: 4px;
+  background: #fafafa;
+  border: 1px solid #ededec;
+  border-radius: 9999px;
```
stevekaplan123, line 4047: it's my comment again: "colors should be variables"

static/css/s2.css
```diff
@@ -4037,6 +4037,48 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
...
+  background: none;
+  border: none;
+  border-radius: 9999px;
+  cursor: pointer;
+  font-size: 14px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: 18px;
+  color: #575757;
+  white-space: nowrap;
```
stevekaplan123, line 4075: again

stevekaplan123 (changes requested):

static/css/s2.css
saengel: I should be better about this by now! Thank you, just pushed a shift to variables for any colors already defined as variables in the codebase.

stevekaplan123 (approved):

PR #3524, opened 2026-07-14
Limit oversized semantic linked refs

yonadavGit:
Implemented linked-ref size control for semantic search.

  Oversized linked refs now use a 1,300-character threshold. If a linked ref’s full text exceeds that, the API reuses the original query embedding to
  retrieve the top 2 semantic chunks scoped to that ref, then returns those chunks as flattened linked-ref entries with parent/chunk metadata. This avoids
  sending very large linked-ref texts while preserving useful context.

  Also split semantic search into reusable get_query_embedding() and semantic_search_by_embedding() helpers, and added test coverage for the embedding reuse
  path.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -235,29 +235,60 @@ class KnnSearch(View):
             return ""
 
     @staticmethod
-    def _semantic_chunk_filters(base_filters, ref, ref_field):
+    def _semantic_chunk_filters(base_filters, ref_filter):
         filters = dict(base_filters or {})
-        filters[ref_field] = ref
+        filters.update(ref_filter)
         return filters
+
+    @staticmethod
+    def _segment_ref_filter_for_linked_ref(ref):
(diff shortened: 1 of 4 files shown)


api/views.py
```diff
@@ -221,13 +234,85 @@ def _ref_text(ref):
...
+        filters = dict(base_filters or {})
+        filters[ref_field] = ref
+        return filters
+
     @classmethod
-    def _serialize_linked_ref(cls, ref, include_text):
+    def _semantic_chunks_for_linked_ref(cls, ref, query_embedding, base_filters):
+        from semantic_search.search import semantic_search_by_embedding
+
+        for ref_field in ("chunked_from_ref", "ref"):
```
nsantacruz: should we prefer `ref` over `chunked_from_ref`? I'm actually not sure which is more likely to be the case.

PR #3535, opened 2026-07-19
Feature/sc 29017/add name of parsha to reader s aliya section

stevekaplan123:
## Description
Problem: With the Aliyot display option on, aliyah headers just say "Second", "Third", etc. — never which parasha they belong to. That's especially confusing in chapters that start mid-parasha, where no parasha header is on screen at all.

Fix: With Aliyot on, headers now read Bereshit: Second. With Aliyot off, nothing changes — still a single Bereshit header at the parasha start.

## Code Changes
[sefaria/model/text.py:459-489]— every aliyah marker now carries aliyah_en/he and parasha_en/he, so an aliyah knows its own parasha without needing the parasha-start segment. The Rishon (first aliyah) is no longer skipped outright; it's only kept out of the title list, so it can render as "Bereshit: First". The new fields are additive — existing readers of alt.en/he/whole see the same values.
[static/js/sefaria/util.js:13-43] — parashahHeader() moved from a TextRange method to a static Util method (so it can be unit-tested without rendering the component) and updated to build the combined string.
[static/js/TextRange.jsx:289]— calls Util.parashahHeader(...). Styling is unchanged.
Tests
[parashahHeader.test.js] — 4 unit tests covering aliyot on/off at the parasha start and mid-parasha.


sefaria/model/text.py
```diff
@@ -471,8 +471,20 @@ def get_trimmed_alt_structs_for_ref(self, oref) -> dict:
...
+                            # parasha start. Keep it out of the en/he title lists (that slot
+                            # already shows the parasha name), but still record its aliyah
+                            # metadata below so the reader can render "Parashat X: First".
+                            if i != 0:
+                                val["en"] += [n.sectionString([i + 1], "en", title=False)]
+                                val["he"] += [n.sectionString([i + 1], "he", title=False)]
+                            # Attach the aliyah and its containing parasha to every aliyah
+                            # entry, so the reader can show the parasha on each aliyah even in
+                            # chapters that begin mid-parasha (where no "whole" marker exists).
+                            val["aliyah_en"] = n.sectionString([i + 1], "en", title=False)
```
dcschreiber: _(Claude writing on Daniel's behalf)_

Deploy question rather than a code comment. The reader now requires `aliyah_en` to render an aliyah header: with Aliyot on and a payload that predates this change, `Util.parashahHeader` returns null and a mid-parasha aliyah header disappears instead of showing "Second" as it does today.

That only lasts as long as stale payloads do, so it does not warrant a fallback in the code. But if anything caches `/api/texts` responses, they will want clearing when this deploys. Do you know whether they are cached, and if so is that part of your rollout?

static/js/sefaria/util.js
```diff
@@ -10,6 +10,37 @@ var INBROWSER = (typeof document !== 'undefined');
...
+    static parashahHeader(data, segment, includeAliyot = false) {
+        if (!data) { return null; }
+        if ("alts" in data && data.alts.length && ((data.categories[1] == "Torah" && !data["isDependant"]) || data.categories[2] == "Onkelos")) {
+            const alt = segment.alt;
+            if (alt != null) {
+                if (includeAliyot && alt.aliyah_en) {
+                    return {
+                        "en": `${alt.parasha_en}: ${alt.aliyah_en}`,
+                        "he": `${alt.parasha_he}: ${alt.aliyah_he}`,
+                        "parashaTitle": false, // styled as an aliyah header (see .parashahHeader.aliyah)
```
dcschreiber: _(Claude writing on Daniel's behalf)_

Worth confirming with design, since it is a visible restyle rather than only new text. With Aliyot **on**, the header at a parasha start now returns `parashaTitle: false`, which gives it the `aliyah` class in `TextRange` instead of the full parasha-header styling. So Genesis 1:1 goes from a large "Bereshit" header to a smaller "Bereshit: First" one.

That follows from the description, and it is probably the point. Just flagging that it changes how the top of every parasha looks in that mode, not only what it says.

e2e-tests/library/parashah-headers.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+ *   - PAR-002: enabling Aliyot switches the headers to aliyah style, naming the parasha
+ *              ("Bereshit: First"), with more than one aliyah header in the chapter.
+ *   - PAR-003: the same, in Hebrew, on the .org.il domain ("בראשית: ראשון").
+ *
+ * Data: Genesis 1 opens Parashat Bereshit (Rishon at 1:1); Bereshit's later aliyot begin
+ * mid-chapter (Second at 2:4, Third at 2:19, ...). Verified against /api/texts/Genesis.N.
+ *
+ * NOTE: The combined "<parasha>: <aliyah>" string is produced by
+ * feature/sc-29017; PAR-002/PAR-003's content assertions only pass once that
+ * branch is deployed to the sandbox under test. PAR-001 passes today.
```
dcschreiber: _(Claude writing on Daniel's behalf)_

This NOTE is stale now that the tests live on the branch it is talking about. Read from `master` after the merge it will say the assertions do not pass yet, which is the opposite of true. Worth deleting.

Same line of thing above at line 5: the PURPOSE says PAR-002 checks for "more than one aliyah header in the chapter", but the test asserts `toBeGreaterThanOrEqual(1)` and the inline comment at line 61 explains there is exactly one. The three should agree.

Minor: line 63's count assertion is already implied by `expect(aliyahHeaders.first()).toBeVisible()` two lines up, so it can go.

dcschreiber (changes requested):
_(Claude writing on Daniel's behalf — these review comments were AI-generated and Daniel reviewed them before posting.)_

Reviewed. The approach is sound and the shape of the change is good: the backend fields are additive, so nothing that reads `alt.en`/`he`/`whole` today sees a difference, and pulling `parashahHeader` out of the component into `Util` makes it testable without rendering. CI is green and the unit tests cover the four cases that matter.

Nothing structural to change. One inline item, about two stale comments in the e2e spec.

dcschreiber (dismissed):
I'm going ahead and improving this because the change requests are minor.

PR #3540, opened 2026-07-22
fix: fixes banner issue with weblate breaking node

saengel:
## Description
The new weblate config was breaking something with banners which was causing Node to fail on startup, this fixes that.


saengel:
Already fixed

PR #3545, opened 2026-07-27
Feature/sc 45836/allow admins to delete bad links created by

nsantacruz:
This pull request introduces a comprehensive set of changes to support advanced admin and editor functionality for managing linker data and non-unique term structures, as well as improvements to citation resolution logic and new data export capabilities. The most important changes are summarized below.

**New API Endpoints and Admin/Editor Functionality:**

* Added a suite of staff-only API endpoints in `api/views.py` for managing linker citations, match templates, address types, node properties, and non-unique terms. These endpoints allow for creating, updating, and deleting linker-related data, and include robust JSON validation and error handling. (`api/views.py`) [[1]](diffhunk://#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R172-R258) [[2]](diffhunk://#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R521-R684)
* Introduced a `StaffRequiredMixin` to enforce staff-only access for these endpoints. (`api/views.py`)

**Citation Resolution and Linker Task Improvements:**

* Improved citation resolution logic in `sefaria/helper/linker/tasks.py` to correctly skip processing and writing links for citations that have been marked as deleted, ensuring data integrity and preventing unwanted link creation. (`sefaria/helper/linker/tasks.py`) [[1]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR167-R168) [[2]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR217-R218) [[3]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR234-R235) [[4]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR354-R356) [[5]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR381-R383) [[6]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR417-R419) [[7]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aR445-R447)
* Refactored how linker output and span data are handled and passed between functions, improving clarity and maintainability. (`sefaria/helper/linker/tasks.py`) [[1]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aL114-R114) [[2]](diffhunk://#diff-6ca2bd7b035940a24296b5c125b8855f08f70f9a87712b2e9dd9eb666f64e23aL125-R128)

**Indexing and Data Export:**

* Added a new Django management command `build_nonuniqueterm_index.py` for rebuilding the NonUniqueTerm-to-MatchTemplate usage index from scratch, ensuring the index can be re-warmed after deployments or schema changes. (`reader/management/commands/build_nonuniqueterm_index.py`)
* Added a script to export all Mishnaic and Talmudic people topics and their relationships to a CSV, supporting downstream data analysis. (`scripts/export_mishnaic_talmudic_people.py`)

**Reader UI and Navigation Updates:**

* Added the "LinkerAdmin" mode to the list of valid sidebar modes in the reader UI, and a new staff-only view for the linker editor page. (`reader/views.py`) [[1]](diffhunk://#diff-26c40bdd3eeaef442340754126fdbcb1ed9e451eb3bb31c02cb78053f5702eaaL521-R521) [[2]](diffhunk://#diff-26c40bdd3eeaef442340754126fdbcb1ed9e451eb3bb31c02cb78053f5702eaaR1385-R1391)

**Dependency and Import Updates:**

* Updated imports in `api/views.py` to include the new `linker_editor` and `linker_admin` modules, supporting the new API endpoints. (`api/views.py`)

These changes collectively provide a robust backend for advanced linker management and improve the reliability and maintainability of citation processing and term management.


reader/management/commands/build_nonuniqueterm_index.py
nsantacruz, line 1: remove this file. not necessary since index can auto-heal

scripts/export_mishnaic_talmudic_people.py
nsantacruz, line 1: remove this file

sefaria/helper/linker_index_converter.py
```diff
@@ -314,6 +314,21 @@ def add_all_dibur_hamatchils(self):
             self.add_dibur_hamatchil_to_index(index)
         db.dibur_hamatchils.bulk_write([InsertOne(d) for d in self._dhs_to_insert])
 
+    def rebuild_index_dibur_hamatchils(self, index) -> int:
+        """
+        Recompute and replace the dibur_hamatchils for a single index. Used by the
```
nsantacruz: make comment less verbose

sefaria/model/linker/tests/linker_entity_recognizer_test.py
nsantacruz, line 1: remove this file

sefaria/model/linker/linker_entity_recognizer.py
```diff
@@ -5,10 +5,27 @@
...
 import requests
 from ne_span import NESpan, NEDoc, NamedEntityType, RefPartType
+from sefaria.system.exceptions import InputError
 import structlog
 logger = structlog.get_logger(__name__)
 
 
+def get_linker_normalizer(lang: str) -> NormalizerComposer:
+    """
+    The normalizer the linker applies to input text server-side before recognition.
```
nsantacruz: make comment more succinct

sefaria/model/text.py
```diff
@@ -5576,6 +5587,18 @@ def refresh_index_record_in_cache(self, index_object, old_title = None):
         assert new_index, "No Index record found for {}: {}".format(index_object.__class__.__name__, index_object_title)
         self.add_index_record_to_cache(new_index, rebuild=True)
 
+    def refresh_non_unique_term_in_cache(self, slug: str):
+        """
+        Drop `slug` from NonUniqueTerm's process-level `.init()` cache so the next lookup
```
nsantacruz, line 5618: make comment more succinct

sefaria/model/text.py
```diff
@@ -5755,6 +5778,31 @@ def get_linker(self, lang: str, rebuild=False):
             linker = self.build_linker(lang)
         return linker
 
+    def rebuild_linker(self, langs=("en", "he")):
```
nsantacruz: i believe this function is never used. confirm and if so, delete.

sefaria/urls_shared.py
```diff
@@ -103,6 +104,23 @@
     re_path(r'^api/calendars/topics/parasha/?$', reader_views.parasha_data_api),
     re_path(r'^api/calendars/topics/holiday/?$', reader_views.seasonal_topic_api),
     path('api/name/<path:name>', reader_views.name_api),
+    # Linker editor (staff-only). Node key paths are dot-separated keys, e.g. "Berakhot.Intro".
```
nsantacruz: as per the convention in this file, break off all of these new linker paths (besides the admin one below) into there one section with newlines before and after and a comment explaining the section

static/js/ConnectionsPanel.jsx
```diff
@@ -1023,6 +1033,505 @@ WebPagesList.propTypes = {
   srefs: PropTypes.array.isRequired,
 };
 
+const LINKER_PART_COLORS = {
```
nsantacruz: move linker admin box and all supporting code to its own file

api/linker_admin_views.py
```diff
@@ -0,0 +1,248 @@
+from django.views import View
+
+from sefaria.client.util import jsonResponse
+from sefaria.system.exceptions import InputError
+from sefaria.helper import linker_editor
+from sefaria.helper import linker_admin
+from .views import _load_json_body
+
+
+class StaffRequiredMixin:
```
nsantacruz: move it to vanilla views

api/linker_admin_views.py
```diff
@@ -0,0 +1,248 @@
...
+        body, err = _load_json_body(request, empty_as_object=True)
+        if err:
+            return err
+        try:
+            return jsonResponse(handler(body), status=status)
+        except InputError as e:
+            return jsonResponse({"error": str(e)}, status=400)
+
+
+class LinkerAdminDeleteCitationView(LinkerAdminAPIView):
```
nsantacruz, line 22: What's the fundamental difference between all these views? Can we extract the differences and use a Mixin or some other abstraction to avoid repeating?

api/linker_admin_views.py
nsantacruz, line 1: Check if it's possible to only add the deleted marker on the LinkerOutput span and NOT the MarkedUpTextChunk span and just reference LinkerOutput before saving a new span.

sefaria/helper/linker/tasks.py
```diff
@@ -301,6 +306,13 @@ def _apply_ambiguous_resolution(payload: AmbiguousResolutionPayload, result: Opt
     if not citing_ref or not resolved_ref:
         return
 
+    query = {"ref": payload.ref, "versionTitle": payload.versionTitle, "language": payload.language}
```
nsantacruz, line 309: is this code duplicated above?

sefaria/helper/linker/tasks.py
```diff
@@ -346,6 +362,13 @@ def _apply_non_segment_resolution_with_record(payload: NonSegmentResolutionPaylo
...
```
nsantacruz, line 365: also here looks duplicated

sefaria/helper/linker/tasks.py
```diff
@@ -403,6 +428,13 @@ def _apply_ambiguous_resolution_with_record(payload: AmbiguousResolutionPayload,
...
```
nsantacruz, line 431: also here duplicated?

sefaria/helper/linker/tasks.py
```diff
@@ -526,6 +562,77 @@ def _update_linker_output_resolution_fields(payload: object, result: object) ->
...
+
+
+def _citation_was_deleted(
+    payload: object,
+    candidate_refs: list[Optional[str]],
+    linker_output: Optional[LinkerOutput] = None,
+    mutc: Optional[MarkedUpTextChunk] = None,
+) -> bool:
+    """
+    linker_output/mutc let a caller pass docs it already loaded for this payload's
```
nsantacruz, line 596: this doc isn't helpful. write one that explains WHAT the function does, not the optimization

sefaria/helper/linker_admin.py
yonadavGit, line 1: i think the separation of logic between `linke_admin` and `linker_editor` can perhaps be better comminuctad, maybe different file names to make the difference more obvioues to readers of the code

yonadavGit, line 1: I wonder if much of the logic here an in `linker_editor` could perhaps be moved into the relevant data models:
```
LinkerOutput
MarkedUpTextChunk
```

sefaria/helper/linker_editor.py
```diff
@@ -0,0 +1,659 @@
...
+# ---------------------------------------------------------------------------
+
+def _save_linker_metadata(index) -> None:
+    """
+    Persist linker-only index metadata without running the full Index save
+    dependency chain. MatchTemplates, addressTypes, and linker node properties do
+    not affect the TOC, version state, or title lists, but other web processes
+    still need their in-process Index cache refreshed.
+    """
+    index.save(override_dependencies=True)
```
yonadavGit, line 32: Worth thinking about - are we sure we want the linker-related information be part of `Index` data model?

sefaria/model/linker_editor_history.py
yonadavGit, line 1: what's the relationship between this and `sefaria/model/history.py`? seems they are completey separate, perhaps they should be united or at least share some core logic? also, the tracker is never norified for changed in `linker_history` it seems

sefaria/model/linker/nonuniqueterm_index.py
yonadavGit, line 1: I take it this is some sort of Django cache. why not good old `Library()` cache? or else, maybe good old `Library` cache should become Django cache?

sefaria/celery_setup/app.py
yonadavGit, line 1: if i understand it correctly, since the linker eidtor now enables frequent changes to many important data models and caches, celery workers need to have access to the most updataed data. 
is that correct, and how exactly do these 2 new functions solve the issue? do they solve the problem hermetically or just make data inconsisteny problems less likely?

api/linker_admin_views.py
nsantacruz: Investigated — I'd leave the dual-write as-is. It's not redundant duplication of logic, just cheap sync:

- `MarkedUpTextChunk._validate()` reads its own `deleted` flag directly (`sefaria/model/marked_up_text_chunk.py:95`) to skip the charRange/text consistency check on save. Dropping MUTC's copy would mean `_validate()` needs to query LinkerOutput on every MUTC save — including the ones `MarkedUpTextChunkGenerator`/`tracker.py` trigger on ordinary text edits, a hot path unrelated to this admin feature.
- Nothing outside the linker subsystem reads MarkedUpTextChunk at all (no reader-facing API exposes it), so there's no risk of the reader UI showing a stale citation — this is purely an internal-consistency concern.
- LinkerOutput-as-source-of-truth already exists where it matters: `_mutc_deleted_spans_from_linker_output` (`sefaria/helper/linker/tasks.py:738`) reconstructs MUTC's deleted spans purely from LinkerOutput when a segment is rerun/regenerated.
- `set_linker_citation_deleted` just doesn't go through that rerun path — it flips the flag in place on both docs directly via `_update_deleted_marker`. Making LinkerOutput sole source of truth here would add a cross-collection query to `_validate()` for no behavioral upside.

TL;DR: keeping both is cheaper and simpler than the alternative, and MUTC's copy is load-bearing for its own validation, not just a mirror.

sefaria/helper/linker_admin.py
nsantacruz: From Claude

This follows the repo-wide split documented in CLAUDE.md: sefaria/model/ holds schema + validation (thin), sefaria/helper/ holds business logic. LinkerOutput/MarkedUpTextChunk are AbstractMongoRecords like everything else in model/ — schema, _validate, _sanitize, no domain logic. Keeping citation-delete/rerun/parse logic in the helper matches how e.g. text.py's models stay thin while sefaria/helper/* carries the mutation logic. Happy to reconsider for a specific method if there's one that feels like it belongs on the model itself.

sefaria/helper/linker_editor.py
nsantacruz: This is a very big decision to make. I tend to agree that it shouldn't be on the index but it's very difficult to decouple (also, not directly related to this PR)

  From Claude:
  
  ▎ Worth flagging: this PR doesn't introduce that coupling — match_templates/addressTypes have been
  ▎ attributes on schema nodes (TitledTreeNode/SchemaNode in schema.py) since the linker itself was 
  ▎ built, well before the linker-editor. linker_editor.py only adds a staff-facing API to edit     
  ▎ data that already lived there; _save_linker_metadata just calls the existing index.save(). So   
  ▎ this is really "is the linker's existing coupling to Index correct," not a new decision this PR 
  ▎ is making — but agreed it's worth a separate conversation if you think that coupling should     
  ▎ change.

sefaria/model/linker_editor_history.py
nsantacruz: They're intentionally separate: LinkerEditorHistory is an append-only audit log of who changed what linker config (for replaying config changes across environments). sefaria/model/history.py + sefaria/tracker.py log content edits (text/links/etc.) for revert/edit and the general activity feed.

sefaria/model/linker/nonuniqueterm_index.py
nsantacruz: From Claude:

Not Django's cache framework — it's Sefaria's existing shared/Redis-backed cache (sefaria.system.cache), the same primitive other cross-pod state already uses. The reason it's not Library()'s in-process cache: Library() is per-process, so with N web pods each would need to independently rebuild the reverse index (a full library walk, thousands of nodes) and would drift out of sync with the others without wiring into the multiserver pubsub invalidation that Library() caches rely on. Using the shared cache means one copy, self-healing on read (see the module docstring), no new invalidation plumbing needed

sefaria/celery_setup/app.py
nsantacruz: From Claude:

Your understanding is correct. _reconnect_multiserver_coordinator (worker_process_init) gives each forked worker its own Redis subscription instead of sharing the parent's copy-on-write socket across siblings. _sync_multiserver_before_task (task_prerun) drains the multiserver channel right before a task runs — nothing else does this for Celery, since MultiServerEventListenerMiddleware only drains per-request in Django web processes, and workers have no request cycle. Honest answer on hermetic-vs-probabilistic: it's hermetic against staleness from a channel nobody ever drained — any invalidation already published before the task starts is guaranteed applied. It does not add a new ordering guarantee beyond what multiserver already provides web pods: a task enqueued concurrently with (or immediately after) the triggering edit's publish has the same race every multiserver consumer has. So: closes a real gap (workers previously never drained the channel at all), doesn't invent a stronger consistency model than the rest of the app already relies on.

PR #3546, opened 2026-07-27
feat: lazy load

saengel:
## Description
This PR implements the lazy loading/infinite scroll across all 4 tabs of the new search UI, with the strings coming from weblate

## Code Changes
- Infinite scroll hook is not isolated, and called where needed
- New texts
- Small styling changes
- Pagination params to the backend of the entity search

## Notes
- When testing, use the query `שלום` and `Rabbi`

### DIFF
--- a/reader/views.py
+++ b/reader/views.py
@@ -4904,12 +4904,12 @@ def entity_search_api(request):
         )
 
     try:
-        size = min(int(request.GET.get("size", 20)), 100)
+        size = max(1, min(int(request.GET.get("size", 20)), 100))
     except (TypeError, ValueError):
         size = 20
 
     try:
-        start = max(int(request.GET.get("start", 0)), 0)
+        start = max(0, min(int(request.GET.get("start", 0)), 10000 - size))
     except (TypeError, ValueError):
         start = 0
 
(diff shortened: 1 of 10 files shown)


stevekaplan123 (commented):
This implementation may be fine, but I'm wondering if there's any reason you didn't use the existing infinite scroll hooks in Hooks.jsx.  See useScrollToLoad, usePaginatedDisplay, and useIncrementalLoad.  I'm unsure if we should re-use them or just re-write them as you do here because I'm not sure if the specced behavior is fundamentally different or not.  What do you think?

saengel:
> This implementation may be fine, but I'm wondering if there's any reason you didn't use the existing infinite scroll hooks in Hooks.jsx. See useScrollToLoad, usePaginatedDisplay, and useIncrementalLoad. I'm unsure if we should re-use them or just re-write them as you do here because I'm not sure if the specced behavior is fundamentally different or not. What do you think?

After investigating with Claude, the fundamental behavior seems to be different (slightly) enough to warrant a new hook... an eventual hook refactor might help but that is out of scope for now

stevekaplan123 (approved):

PR #3553, opened 2026-07-28
chore: binary not count

stevekaplan123:
## Description
I looked into two facets of the bad search results when querying "talmud".  First, we should eventually include categories in search rather, but since there is not a product decision yet, I added research regarding categories to the arch doc.  Second, we should not count how many title variants match a query, but simply check whether there is a match.  That change is implemented using `alias_bag`


docs/arch_docs/search_improvements_arch.md
saengel, line 1: This is a great add, thank you!

saengel (approved):
Thank you!

PR #3561, opened 2026-07-29
chore: switch death and birth year

stevekaplan123:
## Description
Switch birth and death year based on spec.  Default to death year and fall back to birth year.


saengel (approved):
Kol hakavod for also updating the arch doc!

PR #3565, opened 2026-07-30
Fix bug with opening sources links

saengel:
## Description
This PR fixes a bug where returned sources card title links wouldn't resolve. 

## Code Changes
Passing the ref through to the search result card

### DIFF
--- a/static/js/SearchResultCard.jsx
+++ b/static/js/SearchResultCard.jsx
@@ -32,6 +32,7 @@ function SearchResultCard({
   icon,
   type = 'text',
   href,
+  tref,
   onResultClick,
   query,
   accentColor,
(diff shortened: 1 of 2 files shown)


stevekaplan123 (commented):
I'm wondering why we need both tref and href.  Is it possible we only need the tref?

saengel:
> I'm wondering why we need both tref and href. Is it possible we only need the tref?

We need href if we want to pass (for sources) url params to enable the text clicked to be highlighted when the text panel opens, see Claude's explanation below:

```href = URL for the anchor tag (supports cmd+click, version params, query
  highlight). tref = bare Sefaria ref passed to handleNavigationClick → 
  openPanel, which drives in-app panel navigation without a page reload.```

stevekaplan123 (approved):

PR #3572, opened 2026-08-03
chore(sc-46272): Library Assistant — ship the setting (phase 1)

dcschreiber:

Phase 1 of switching the Library Assistant from opt-in (beta whitelist) to opt-out. This PR rides a regular deploy and is **behaviorally inert** — the flip itself is a separate migration run on rollout day.

## Design

A new per-user setting `profiles.settings.library_assistant` becomes the single source of truth, read only through `sefaria/helper/library_assistant.py`:

- **key present** → obey it
- **key absent** → today's exact rule (whitelist row AND `profile.experiments`)

Until the migration writes the key, every user behaves exactly as today; the migration run is the launch, and rollback is unsetting the key. The key is deliberately **not** in `UserProfile`'s settings defaults — a `True` default would make an un-migrated opt-out read as on, and `update()`'s deep merge would persist that wrong value on their next save.

The one visible change: the account-settings toggle is un-gated from whitelist membership (the users the migration is about to switch on need an off switch from day one). It renders the helper's *effective* value, so a legacy-on user with no key sees On. Per review: the form submits `library_assistant` **only when the toggle was actually changed** — the rendered value is the effective one, so an unconditional submit would write the key onto untouched profiles and remove them from the migration cohort (an unrelated settings save would become a permanent opt-out).

## What ships

- The helper module (`SETTING_KEY`, `is_enabled`, `normalize`, `set_enabled`), with every read site routed through it: the script-tag context processor and `base_props`.
- Write paths coerce posted values (`/api/profile` and `/api/profile/sync` are public — a posted `"false"` must not read truthy).
- **Nothing is sent to the CRM.** The application code carries no Salesforce calls; the webhook infrastructure (`crm/tasks.py`) stays in place behind a deactivation flag — comms has no use for the opt-in signal once the assistant is on by default — and its tests prove both the no-op and, with the flag flipped, the preserved reference behavior.
- Registration writes the key `True` explicitly; the promo banner's Join button and `/enable-library-assistant` write the setting instead of enrolling in the whitelist.
- The migration and rollback scripts for the flip. Cohorts come from **Postgres `UserExperimentSettings` rows**, never from Mongo `experiments` (which is serialized `False` by default on every save, so it can't distinguish a choice from a default). Rows inherit their value, everyone else backfills to `True`; every write is archived; the rollback unsets only where the stored value still matches what the migration wrote.

## CI: `reader/tests` had never run

The pytest config lived at `sefaria/pytest.ini`, so nothing under `reader/tests` was ever collected. The config moves to the repo root and CI now runs `./sefaria ./reader`. This also lights up four previously-never-run test files (`password_reset_urls`, `social_image_api`, `static_pages`, `test_catchall_redirects`); if any fail for unrelated reasons, quarantine them in a follow-up rather than block this PR.

Running `reader/tests` for the first time exposed a hazard, fixed here: the tests operate on Mongo `profiles` keyed by the Django user id, Mongo is **not** swapped for a test database, and the public dump (restored on laptops and the CI sandbox alike) ships real profile documents at exactly the low ids a fresh test Postgres hands out. `reader/conftest.py` now provides `create_test_user()` — explicit ids above a 2,000,000,000 floor, where no real user exists — and `purge_test_profiles()`, which refuses to delete below the floor; `build/ci/cleanup_test_data.py` sweeps leftover synthetic profiles out of the shared sandbox Mongo.

**Convention going forward:** any reader test touching a user-keyed Mongo collection uses the factory and the guarded purge — never a default-sequence user id.


reader/views.py
```diff
@@ -382,11 +382,16 @@ def base_props(request):
         "chatbot_promo_session_length_seconds": remoteConfigCache.get(CHATBOT_PROMO_SESSION_LENGTH_SECONDS, default=30*60),
         'show_join_chatbot_banner': remoteConfigCache.get(SHOW_JOIN_CHATBOT_BANNER, default=False),
     }
-    if user_has_experiments(request.user):
-        chatbot_data["in_chatbot_experiment"] = True
-        if _is_user_in_experiment(request):
+    if request.user.is_authenticated:
+        if library_assistant.is_enabled(profile):
```
akiva10b, line 386: there is a method is_enabled_for_user which checks request.user.is_authenticated - seems like it is designed to be used instead of `request.user.is_authenticated`

reader/views.py
```diff
@@ -4289,6 +4306,9 @@ def profile_sync_api(request):
                 except ValueError as e:
                     logger.warning(f'profile_sync_api: {e}')
                     continue
+                if la_key in field_data:
```
akiva10b: why is the la_key outside of "settings" I would think it would just be part of settings

templates/account_settings.html
```diff
@@ -263,12 +263,9 @@ <h1>
                         reading_history: $("#readingHistory .toggleOption.on").attr("data-value") === "true" ? true : false,
                         translation_language_preference: $newTransLangPref,
                         translation_language_preference_suggested: true,
+                        library_assistant: $("#libraryAssistantSetting .toggleOption.on").attr("data-value") === "true",
```
akiva10b: this code will always submits the rendered library_assistant value, even when the user never touched that toggle.
Before migration, a non-whitelisted user sees “Off.” If they save an unrelated preference—language, reading history, email notifications—the page persists library_assistant: false. The migration subsequently skips them because it only processes profiles where the key is absent.
That converts an unrelated settings save into a permanent opt-out and corrupts the migration cohort.

If users or off by default that is fine - but what it means is that the order of operation should be to run the data migration first, and then push the code

akiva10b (commented):
overall looks good - left 2 comments I am not entirely sure about (didnt verify with AI), specifically regardign the location of the la_key, and 1 comment is that we clearly should add the profile field before deploying the code

reader/views.py
dcschreiber: _(Claude writing on Daniel's behalf)_ `is_enabled_for_user` builds its own `UserProfile`, so it is meant for call sites that only have a Django user (e.g. the `chatbot_user_token` context processor). `base_props` already has `profile` loaded a few lines up, so `is_enabled(profile)` reuses it instead of doing a second Mongo fetch. The `is_authenticated` check here also guards the `in_chatbot_experiment` line below, which needs `profile` (only defined for logged-in users).

reader/views.py
dcschreiber: _(Claude writing on Daniel's behalf)_ Fixed in 80259c098 — dropped the alias; the key is now referenced inline where the coercion happens, inside the settings branch.

templates/account_settings.html
dcschreiber: _(Claude writing on Daniel's behalf)_ Good catch — fixed in 28621be6d. The form now sends `library_assistant` only when the toggle differs from its rendered value, so an unrelated settings save no longer writes the key and the migration cohort stays intact. That keeps deploy-then-migrate safe: untouched profiles keep the key absent (migration still processes them), while a deliberate flip during the window is persisted and correctly skipped by the migration. So the planned order — deploy Sunday, migrate Monday — can stand, with the migration run remaining the launch moment.

akiva10b (approved):

PR #3575, opened 2026-08-03
fix: rename camelCase search i18n keys so Sefaria._() can resolve them

dcschreiber:

Master's Jest check has been red since the last Weblate sync (#3566) added two camelCase keys, `search.exactMatchToggle.allResults` and `search.exactMatchToggle.exactMatch`, to `i18n/interface/{en,he}.json`. The keyed-string ID regex (`Sefaria._keyedStringIdRegex`) only admits lowercase snake_case segments, so these keys can never resolve through `Sefaria._()` — `strings.test.js` is correctly flagging them as dead, and the failure blocks the Jest check on every open PR (e.g. #3572).

This renames them to `search.exact_match_toggle.all_results` / `search.exact_match_toggle.exact_match`, matching the convention every other key follows (e.g. `search_filters.exact_matches_only`). No code anywhere references the old names, so the rename is behaviorally free; the translations are preserved. A source-string rename may need to be mirrored in Weblate so the next sync doesn't reintroduce the camelCase names.

`strings.test.js` passes locally (13/13) with this change.


dcschreiber:

Closing as a duplicate of #3568, which predates this PR and makes the identical rename — it's confirmed and queued for merge instead.

PR #3578, opened 2026-08-04
chore: cosmetic treatment of LA promo banner ahead of Aug 10

dcschreiber:

Cosmetic refresh of the LA promo banner per [sc-46012](https://app.shortcut.com/sefaria/story/46012), ahead of Aug 10.

## What changed
- **Icon** — new double-star asset from [Figma](https://www.figma.com/design/Y31hDgxSjr0l1fcm0nNJSD/Library-Assistant-Wireframes?node-id=7008-1904) at `static/icons/ai-double-star.svg`, added via a new optional `icon` prop on `SiteWideBanner`. In Hebrew it sits right of the text and is deliberately **not** mirrored.
- **Color** — background `--mussar-purple` with white text; both CTAs use the existing `.button.white` modifier, with the banner overriding only the text color.
- **Copy** — new heading and subheading as keyed strings in `interface/{en,he}.json`; `Maybe later` → `Maybe Later`.
- **Help Center link removed** — the inline anchor is gone, along with the now-unreachable `promoLearnMoreUrls` prop and `.bannerSecondaryText a` rule that only existed to serve it.

## Color: why not the hex in the ticket
The ticket said kabbalah-purple `#653259`, but the codebase's `--kabbalah-purple` token is `#594176` — a different shade. Flagged on the story; Penina confirmed `--mussar-purple` (`#7C416F`), which already exists in `color-palette.css`. Referenced as the variable, not a hardcoded hex.

## Hebrew
Final — Penina updated the story description and `he.json` matches it verbatim (`שאלו את עוזר הספרייה` / `דרך חדשה לגלות תשובות מעוגנות במקורות היהדות.`). Buttons unchanged in Hebrew, per the ticket.

## Testing
- `strings.test.js` — 13/13 (en/he key parity, ID regex, no empty EN values)
- Isolated render against the real stylesheets, EN + HE × logged-in/out: background resolves to `rgb(124, 65, 111)`; both CTAs `rgb(124, 65, 111)` on white; icon computes `transform: none` under `direction: rtl`
- Visual QA on cauldron `sc-46012-la-promo` — pending, for Penina

## Note for reviewers
`SiteWideBanner`'s `learnMoreUrl` / `learnMoreText` props and the `.bannerLearnMore` rule are dead, but were **already** dead on `master` — no consumer has passed them. Left alone to keep this diff cosmetic; worth a follow-up. Same for the backend `chatbot_promo_learn_more_urls` (`reader/views.py:379`), now unread by the client.


static/css/s2.css
```diff
@@ -383,11 +383,16 @@ body.hasBannerMessage #s2.headerOnly {
   flex: 1;
   align-items: center;
   justify-content: space-between;
-  color: black;
-  border: 1px solid #d2dcff;
+  color: white;
   border-radius: 10px;
   padding: 20px;
-  background: #f0f7ff;
+  background: var(--mussar-purple);
```
YishaiGlasner: this coupling musar and baner. they have the same color, but i don't think they should share a var.

static/js/SiteWideBanner.jsx
```diff
@@ -104,6 +104,7 @@ const SiteWideBanner = ({
   enableBackoffDismissal,
   nudgeSchedule,
   promoSessionLengthSeconds,
+  icon,
```
YishaiGlasner: i think imgSrc will reflect better what this prop is

YishaiGlasner (changes requested):

static/js/SiteWideBanner.jsx
dcschreiber: _(Claude writing on Daniel's behalf)_

Agreed — done in `9ad73b0`. The prop holds a URL, so `imgSrc` says what it actually is; `icon` was naming the concept rather than the value.

Renamed at all four sites (destructure, render, propTypes, and the call site in `ChatbotExperimentBanner`). Left the `siteWideBannerIcon` CSS class alone since that one really is describing the visual role.

static/css/s2.css
dcschreiber: I asked Penina what to do in practice

YishaiGlasner (approved):

PR #3580, opened 2026-08-04
feat(sso): surface sso_only_account on api/login/ for the mobile app

yodem:
Targets `sso` (the head branch of #3511), not `master`. Stacks on it — no conflict.

## Why

The mobile app cannot tell a wrong password from an account that only has Google or Apple sign-in, so it has no way to tell the user to use the SSO buttons instead.

Web already handles this: #3511's `email_login` returns `401 {error, _auth: {code: 'sso_only_account', providers}}`. But the mobile app authenticates against a **different endpoint** — `api/login/`, which is stock SimpleJWT and returns the same generic *"no active account found"* for every failure. This closes only that gap.

## What changed

Three files.

- `sso/views.py` — `SSOAwareTokenObtainPairSerializer` delegates to SimpleJWT and decorates only the failure path; `MobileTokenObtainPairView` binds it. The SSO-only detection is extracted into `sso_only_account_info(email)`, which `email_login` now also calls so the two endpoints cannot drift.
- `sefaria/urls_shared.py` — `api/login/` points at the new view. `api/login/refresh/` untouched.
- `sso/tests/views_test.py` — four cases, one per row of the table below.

## Compatibility

The mobile app depends on this endpoint, so it is deliberately narrow:

| Path | Before | After |
|---|---|---|
| Success | `200 {access, refresh}` | identical |
| Wrong password | `401 {detail: "No active account…"}` | identical — original exception re-raised |
| Unknown user | `401 {detail: "No active account…"}` | identical |
| SSO-only account | `401 {detail: "No active account…"}` | `401 {error, _auth: {code, providers}}` |

DRF emits a dict-valued `detail` as the response body verbatim, so the new body is not nested under `detail` — verified against the installed DRF 3.15.2 and asserted in the tests.

## Please decide this before merging

**This extends account enumeration to `api/login/`.** An unauthenticated caller can now learn from the response body whether an email is registered and which providers are linked. Neither this endpoint nor `email_login` is throttled — there are no `DEFAULT_THROTTLE_CLASSES` anywhere.

The same exposure already ships on the web endpoint via #3511, so this doubles the surface rather than introducing a new class of leak, and it is inherent to showing a provider-specific message at all. But it is a tradeoff, not an oversight. The alternative is returning `sso_only_account` with no `providers` list and having the client show generic copy.

A timing side channel already exists independently: `check_password()` short-circuits for unusable passwords, so SSO-only accounts were already measurably faster to reject.

## Tests

`sso/tests/views_test.py` 28/28, full `sso/tests/` 47/47, against a real Postgres + Mongo.

## Note

An earlier revision of this PR also refactored the shared provider-lookup duplication across `clean_email`, `reader/views.py` and the web register client. That was out of scope for a mobile fix and has been removed — it is raised as a comment on #3511 instead, since it is that branch's code and its owner's call.

### DIFF
--- a/sefaria/urls_shared.py
+++ b/sefaria/urls_shared.py
@@ -13,7 +13,8 @@ import powered_by.views as powered_by_views
 from sefaria.heapdump import heapdump_view
 from sefaria.site.urls import site_urlpatterns
 from django.contrib.staticfiles.urls import staticfiles_urlpatterns
-from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
+from rest_framework_simplejwt.views import TokenRefreshView
+from sso.views import MobileTokenObtainPairView
 
 shared_patterns = [
     path('_allauth/', include('allauth.headless.urls')),
@@ -31,7 +32,7 @@ shared_patterns = [
         name='password_reset_complete'),
     re_path(fr'password/reset/done/$', sefaria_views.CustomPasswordResetDoneView.as_view(), name='password_reset_done'),
 
-    re_path(fr'api/login/$', TokenObtainPairView.as_view(), name='token_obtain_pair'),
+    re_path(fr'api/login/$', MobileTokenObtainPairView.as_view(), name='token_obtain_pair'),
     re_path(fr'api/login/refresh/$', TokenRefreshView.as_view(), name='token_refresh'),
 
     re_path(r'^saved/?$', reader_views.saved_content),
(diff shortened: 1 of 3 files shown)


yodem:
Closing as not currently needed. Nothing in Sefaria-Mobile consumes `sso_only_account` — the code is inert until the mobile client adds the corresponding UX. The branch `fix/sso-only-account-api-login` is left in place, so this can be reopened if that work is picked up.

PR #3581, opened 2026-08-04
fix(explore): keep hovered link attached when raising it to the front

EliezerIsrael:
## Problem

The `/explore` Link Explorer degraded in current Chrome: hovered link lines stay highlighted after the pointer leaves, clicking a link does nothing, and the "Click to explore" tooltip gets stuck in the upper-left corner of the visualization.

## Cause

Nothing changed on our side — Chrome 144 shipped ["boundary event dispatch tracks node removal"](https://developer.chrome.com/release-notes/144) (an Interop 2025 item, matching the UI Events spec and other engines). Re-inserting the element under the pointer now counts as a DOM removal, after which the browser stops dispatching `mouseout`/`mousemove`/`click` to that element.

`moveToFront()` in `static/js/explore.js` does exactly that: each link's `mouseover` handler calls `this.parentNode.appendChild(this)` to repaint the hovered path on top. From that instant the path receives no further events, which produces all three symptoms:

- **Stuck highlights** — `mouseout` never fires, so `.classed("active", false)` never runs
- **Dead clicks** — `click` is never dispatched to the re-inserted path
- **Stuck tooltip** — `mouseover.tooltip` shows it, but the `mousemove.tooltip` handler that positions it and the `mouseout.tooltip` handler that hides it never fire, so it sits at its default (0,0) position permanently

Verified empirically on production with native event listeners: hovering links produced `mouseover` events but zero `mouseout`/`click` events.

## Fix

When the node being raised is the one currently hovered, achieve the same paint order without detaching it: move its later siblings behind it. Non-hovered nodes (e.g. the link paths raised when hovering a book bar) keep the original cheap `appendChild` path.

Verified by injecting the exact patched function into the live page: highlights clear correctly on mouse-out, the tooltip hides, and clicking a link drills down as expected.

## Notes

- Requires a webpack rebuild of the `explore` bundle on deploy (no template/Python changes).
- Other D3 visualizations may use the same re-append-on-hover idiom and be broken the same way in Chrome 144+; worth a follow-up grep.


EliezerIsrael:
Tested on cauldron.  Still live at xplr.cauldron.sefaria.org

akiva10b (approved):

PR #3585, opened 2026-08-05
test(sc-46273): Library Assistant opt-out test infrastructure

dcschreiber:

Test coverage for the Library Assistant opt-out setting.

## What's here

**Browser suite** — `e2e-tests/library-assistant-setting/`, 10 specs. Covers the cohorts (explicit on, explicit off, logged out), the settings-toggle round trip, and `/enable-library-assistant`. The oracle checks both enablement gates — the chatbot `<script>` tag and the React prop — since either one alone can be wrong while the other looks fine. Banner behaviour is deliberately not asserted (the shared harness suppresses banners as a flake source; banner logic belongs in Jest pure-helper tests — see the suite README). Every assertion is valid both before and after #3579 — the two PRs are independent and merge in either order.

**`reader/tests/register_library_assistant_test.py`** — 6 tests, run in CI. Verifies registration writes the setting key. Intercepts `UserProfile.save` rather than writing to Mongo, because registration ids land in the low range where a restored dump holds real profiles.

**`scripts/dev/seed_library_assistant_e2e_users.py`** — seeds one account per cohort at deterministic synthetic ids (≥ `SYNTHETIC_USER_ID_FLOOR`). The `la-*` Playwright projects only exist when its gitignored manifest (`e2e-tests/.la-e2e-users.json`) does, so unseeded machines and CI skip them instead of failing.

## Changes to shared e2e infrastructure

Not LA-specific — if something unrelated changes behaviour, look here.

1. `e2e-tests/moduleUrls.ts` is now the single derivation of module URLs (previously duplicated in `playwright.config.ts` and `constants.ts` with a "keep in sync" comment). Verified byte-identical output for all four host shapes.
2. `.env` no longer overrides exported shell variables. Before this, `SANDBOX_URL=http://localhost:8000 npx playwright test` silently ran against production.
3. `global-setup.ts` skips the shared auth setup when only `la-*` projects are selected; unchanged behaviour otherwise.

`harness.ts` also refuses live-site hostnames outright.

## Running it

```bash
python scripts/dev/seed_library_assistant_e2e_users.py --reset
SANDBOX_URL=http://localhost:8000 npx playwright test --project=la-setting
python scripts/dev/seed_library_assistant_e2e_users.py --teardown
```

## Verification

- pytest: 57 passed / 0 failed across the four LA suites.
- Playwright: all 10 specs collect under `la-setup` + `la-setting` (the listed count scales with seeded accounts); the full config collects clean across all 48 files. `tsc --noEmit` clean, including `--noUnusedLocals`.


dcschreiber:

## First staging run of this suite — what it found

This suite had never been run against a remote target; every prior result was `localhost`. It has now been run against staging (2026-08-06), and the run is worth reporting in full, because **it failed 13 of 19 tests without a single product defect among them.** Every failure traced to test infrastructure or to the environment being different from what this PR's own README assumed.

Two commits here fix that: `62b590e8d` (the login helper) and `09302df9d` (the runbook).

### Phase 1 is fine

Verified on staging directly, driving real HTTP logins as the seeded cohort accounts:

| cohort | expected (pre) | actual | |
|---|---|---|---|
| `beta_opt_in` | on | on | ✅ |
| `beta_opt_out` | off | off | ✅ |
| `never_chose` | off | off | ✅ |
| `explicit_on` | on | on | ✅ |
| `explicit_off` | off | off | ✅ |

`beta_opt_out` — the person who joined the beta and switched the assistant off, the one the flip must not disturb — is correct on staging. The settings page renders `#libraryAssistantSetting` for an authenticated cohort account. Locally: 114 pytest, 27 Jest, 19/19 browser, all green.

One caveat stated plainly: the staging check read `chatbot_enabled` from the page props, which is **one of the two independent enablement calls**. The `<script>`-tag gate (`chatbot_user_token`) was not verified remotely. That is exactly the weakness §2 of the README warns about, and my substitute oracle inherited it.

### Defect 1 — the login helper reported success when the login had not happened

The important one. `logIn` waited only for the URL to leave `/login`. A rejected submission satisfies that. So does a cross-domain redirect.

So `la-setup` reported all seven cohorts logged in, and then thirteen tests failed as though the Library Assistant had regressed — while the browser was simply anonymous. **An infrastructure fault wearing the costume of a product bug**, produced by the suite whose job is to tell those apart.

It now asserts the browser is still on the expected origin *and* that the server rendered `Sefaria._uid`, which happens only for an authenticated request. Both failures name their cause.

Two notes for reviewers:
- `/api/profile` has no GET route despite the name — it 404s. I tried it first and the new assertion caught me, which is a fair advertisement for it.
- The browser-side cause of the original failure is **still unexplained**: identical credentials authenticate fine over `curl`. Two theories were tested and disproved — capacity (identical failures at `LA_WORKERS=2`) and the `-il` geo-redirect (identical failures with `SANDBOX_URL` pointed there). This fix makes the failure loud and correct; it does not claim to have root-caused it.

### Defect 2 — the README was wrong about staging's capacity

It said staging "is Postgres behind several web pods and is not the bottleneck a dev server is, so leave `LA_WORKERS` unset."

Staging runs **one** web replica (`rollout.argoproj.io/sefariastaging-web`). Following that advice drove Cloudflare into returning **504 Gateway time-out** pages, which surface as ordinary assertion failures. `LA_WORKERS=2` for staging.

### Defect 3 — the documented seeding round trip hangs

```
kubectl exec -i $POD -- … --manifest-stdout > e2e-tests/.la-e2e-users.json
```

`-i` holds stdin open after the script exits, so this hangs until something kills it — and the shell has already truncated the target, so you end up with an empty manifest *and* your previous local one destroyed. Replaced with a `kubectl cp` sequence that works, with the password passed via a file rather than `env VAR=` in the argv, since exec arguments are recorded in GKE Cloud Audit Logs and shell history.

### The environment, now documented in-repo (README §4a)

Several of these contradict what the file previously asserted:

| | |
|---|---|
| Cluster | **development** (`gke_development-205018_…`), namespace `default` — *not* the production cluster |
| Pod | `sefariastaging-web-<hash>`, one replica |
| Postgres | db `sefariastaging`, host `postgres-18` |
| Mongo | db `sefaria-sefariastaging`, `mongo.default:27017` |
| **Weekly wipe** | `sefariastaging-sync-{postgres,mongo}-production-data`, both `0 2 * * 0` — **both databases restored from production every Sunday 02:00** |

Two things that turn out to be safe by construction, so nobody re-worries them: production is a different cluster with `production-*` pod names and no developer `pods/exec`, so seeding it by accident is impossible; and CI's `cleanup_test_data.py` range-delete cannot reach staging, because `SEFARIA_DB` is suffixed per `deployEnv` and it resolves `db` from its own pod.

Also worth knowing: a request from Israel is served the `-il` domain, and `kubectl` IAM denials are reported against whichever account `gcloud` has active — a personal account produces errors that read exactly like a missing grant.

### Housekeeping

This branch is **behind master**, so the diff re-carries #3584's already-merged migration-script changes (master does have the distinct-id fix). Worth rebasing before review so the diff shows only what this PR actually adds.

PR #3592, opened 2026-08-09
Migration fix for PBS cauldron

saengel:
## Migration
For some reason, `prod` and `local` are aligned but dev env has a weird field on the powered_by postgres. This migration fixes the dev env to be aligned with prod and local - and will be trivial when run on prod (makes no changes). 

## Script
Script which ingests PBS projects from a CSV, can be taken out.

### DIFF
--- a/powered_by/migrations/0002_fix_jsonfield_columns.py
+++ b/powered_by/migrations/0002_fix_jsonfield_columns.py
@@ -1,0 +1,38 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    """
+    0001_initial was edited in-place from ArrayField to JSONField after it had
+    already been applied on some databases, leaving those columns as
+    character varying[] instead of jsonb. This migration corrects them.
+    The DO block is a no-op on databases where the columns are already jsonb.
+    """
+
+    dependencies = [
+        ("powered_by", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="""
+                DO $$
+                BEGIN
+                    IF EXISTS (
+                        SELECT 1 FROM information_schema.columns
+                        WHERE table_name = 'powered_by_project'
+                          AND column_name = 'sefaria_tools_used'
+                          AND data_type != 'jsonb'
+                    ) THEN
+                        ALTER TABLE powered_by_project
+                            ALTER COLUMN sefaria_tools_used TYPE jsonb
+                            USING to_jsonb(sefaria_tools_used);
+                        ALTER TABLE powered_by_project
+                            ALTER COLUMN tags TYPE jsonb
+                            USING to_jsonb(tags);
+                    END IF;
+                END $$;
+            """,
+            reverse_sql=migrations.RunSQL.noop,
+        ),
+    ]
--- a/powered_by/scripts/pbs_ingest_script.py
+++ b/powered_by/scripts/pbs_ingest_script.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+"""
+One-time import: upsert Powered by Sefaria projects from a CSV export
+(matching the powered_by.models.Project field names) into the Project table.
+Rows are matched on project_link, so re-running is safe -- existing rows are
+updated in place rather than duplicated.
+
+Usage:
+    python powered_by/scripts/import_pbs.py path/to/pbs.csv
+"""
+import csv
+import os
+import sys
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+if REPO_ROOT not in sys.path:
+    sys.path.insert(0, REPO_ROOT)
+
+import django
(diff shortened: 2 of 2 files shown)


yodem:
**[claude]** — automated review, posted from @yodem's account.

**Worth fixing:** the guard checks only `sefaria_tools_used`, but the `THEN` branch alters both columns (`powered_by/migrations/0002_fix_jsonfield_columns.py:22-38`). On a database where `tags` is still `character varying[]` but `sefaria_tools_used` is already `jsonb`, the whole block is skipped and `tags` is never fixed. The "no-op on prod" claim holds only for `sefaria_tools_used`'s state. Gating each `ALTER` on its own `IF EXISTS` makes it correct regardless of how the two columns drifted.

Minor:

- `reverse_sql=migrations.RunSQL.noop` (line 42) makes this reversible in name only — rolling back to `0001` won't restore the array type. Fine for a corrective migration, just don't rely on it.
- The script's usage line references `import_pbs.py` (`powered_by/scripts/pbs_ingest_script.py:14`) but the file is `pbs_ingest_script.py`.
- `SUBMISSION_SOURCE_MAP[...]` (`pbs_ingest_script.py:95`) raises `KeyError` on an unmapped `submission_source` and drops the whole row, including its valid fields — a default plus a warning would be safer.
- One-off scripts elsewhere in the repo live in the single top-level `scripts/` directory; `powered_by/scripts/` is a new pattern. Since you note the script can be taken out, dropping it seems cleaner than establishing that.

powered_by/migrations/0002_fix_jsonfield_columns.py
```diff
@@ -0,0 +1,38 @@
...
+
+    operations = [
+        migrations.RunSQL(
+            sql="""
+                DO $$
+                BEGIN
+                    IF EXISTS (
+                        SELECT 1 FROM information_schema.columns
+                        WHERE table_name = 'powered_by_project'
+                          AND column_name = 'sefaria_tools_used'
```
yodem, line 24: **[claude]** The guard tests only `sefaria_tools_used`, but the `THEN` branch alters both this column and `tags` (lines 27-32). If a database has `tags` still `character varying[]` while `sefaria_tools_used` is already `jsonb`, the whole block is skipped and `tags` is never corrected.

Since the stated cause is an in-place edit to `0001_initial`, the two columns can have drifted independently on any database that applied it mid-flight — so the "no-op on prod" claim is only verified for `sefaria_tools_used`. Gating each `ALTER` on its own `IF EXISTS` makes this correct regardless of how they drifted.

powered_by/migrations/0002_fix_jsonfield_columns.py
```diff
@@ -0,0 +1,38 @@
...
+                        ALTER TABLE powered_by_project
+                            ALTER COLUMN sefaria_tools_used TYPE jsonb
+                            USING to_jsonb(sefaria_tools_used);
+                        ALTER TABLE powered_by_project
+                            ALTER COLUMN tags TYPE jsonb
+                            USING to_jsonb(tags);
+                    END IF;
+                END $$;
+            """,
+            reverse_sql=migrations.RunSQL.noop,
```
yodem, line 36: **[claude]** `noop` marks this reversible in Django's bookkeeping, but rolling back to `0001` won't restore the array type. Fine for a corrective migration — worth a one-line comment so nobody reads `migrate powered_by 0001` as an undo.

powered_by/scripts/pbs_ingest_script.py
```diff
@@ -0,0 +1,121 @@
...
+        for line_num, row in enumerate(reader, start=2):
+            project_link = (row.get("project_link") or "").strip()
+            if not project_link:
+                print(f"line {line_num}: skipping, no project_link", file=sys.stderr)
+                error_count += 1
+                continue
+
+            try:
+                defaults = {field_name: (row.get(field_name) or "").strip() for field_name in DIRECT_FIELDS}
+                defaults["submission_source"] = SUBMISSION_SOURCE_MAP[(row.get("submission_source") or "").strip().lower()]
```
yodem, line 95: **[claude]** An unmapped `submission_source` raises `KeyError` here, which `except Exception` (line 107) catches and `continue`s past — so the entire row is dropped, including its valid fields. The printed message for a bare `KeyError` is just the missing key, which makes it hard to tell a bad source value from a real failure. `.get(..., <default>)` plus a warning would keep the row.

powered_by/scripts/pbs_ingest_script.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
+"""
+One-time import: upsert Powered by Sefaria projects from a CSV export
+(matching the powered_by.models.Project field names) into the Project table.
+Rows are matched on project_link, so re-running is safe -- existing rows are
+updated in place rather than duplicated.
+
+Usage:
+    python powered_by/scripts/import_pbs.py path/to/pbs.csv
```
yodem, line 9: **[claude]** Usage line says `import_pbs.py`; the file is `pbs_ingest_script.py`.

PR #3612, opened 2026-08-12
feat(sso): surface sso_only_account on the mobile JWT login endpoint

yodem:
## Summary

Mobile's email/password login hits `api/login/`, which is stock `django-rest-framework-simplejwt` `TokenObtainPairView`. On bad credentials it returns a generic 401 `{"detail": "No active account found with the given credentials"}` -- there is no way to tell the iOS/Android app that the account exists but only has Google or Apple sign-in linked, so a user with an SSO-only account who tries their email/password just sees a failed login with no explanation.

The web session login path already solves this: `sso.views.email_login` (the JSON login behind `POST api/auth/login`) calls `_sso_only_account_error(email)` and, when the account is SSO-only, returns
```json
{"error": "auth.generic_error", "_auth": {"code": "sso_only_account", "providers": ["google", "apple"]}}
```
at 401. That helper was not wired into the JWT path mobile actually uses.

This PR wires it in via `MobileTokenObtainPairView`, a thin `TokenObtainPairView` subclass in `sso/views.py` that overrides `post()` to catch simplejwt's `AuthenticationFailed` and, only then, consult `_sso_only_account_error(email)`:
- if it matches, return that response (same body/shape web already returns -- reused verbatim, not reimplemented)
- otherwise re-raise, so a genuine wrong password or unknown account still gets simplejwt's stock 401 unchanged

`sefaria/urls_shared.py`'s `api/login/` route now points at `MobileTokenObtainPairView` instead of `TokenObtainPairView` directly. **The URL is unchanged** -- apps already in the wild that POST to `api/login/` pick up the richer error immediately, with no store release required. A new endpoint would only help clients built after the change ships.

Import direction: `sefaria/urls_shared.py` already imports `sso.urls` (`include('sso.urls')`), so `sso` -> `sefaria` is the existing dependency direction (`sso/views.py` imports `sefaria.forms`). Importing `sso.views.MobileTokenObtainPairView` from `sefaria/urls_shared.py` follows the same direction and introduces no cycle.

Mobile side: once this ships, `Sefaria.api.login`'s error handler can check `_auth.code === 'sso_only_account'` and use `_auth.providers` to show "This email is registered via Google Sign-In" (etc.) instead of a generic failed-login message.

Related: #3609 (which introduced `_sso_only_account_error` and wired it into `password_reset_api` and `email_login`, but not the JWT endpoint -- this PR closes that gap).

Story: sc-46556

## Test plan
- [x] `sso/tests/views_test.py::MobileLoginTest` (new): valid login still returns `{access, refresh}` at 200; wrong password on a normal account returns stock simplejwt 401 (`detail`, no `_auth`); unknown username returns stock 401; SSO-only account returns `_auth.code == 'sso_only_account'` with `providers` and no tokens.
- [x] Ran the full `sso/tests/views_test.py` suite locally (31 passed) to confirm no regressions to the existing SSO/email/social-login tests.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01JMEqissBzfRtQuukMijY4u

### DIFF
--- a/sefaria/urls_shared.py
+++ b/sefaria/urls_shared.py
@@ -13,7 +13,8 @@ import powered_by.views as powered_by_views
 from sefaria.heapdump import heapdump_view
 from sefaria.site.urls import site_urlpatterns
 from django.contrib.staticfiles.urls import staticfiles_urlpatterns
-from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
+from rest_framework_simplejwt.views import TokenRefreshView
+from sso.views import MobileTokenObtainPairView
 
 shared_patterns = [
     path('_allauth/', include('allauth.headless.urls')),
@@ -31,7 +32,7 @@ shared_patterns = [
         name='password_reset_complete'),
     re_path(fr'password/reset/done/$', sefaria_views.CustomPasswordResetDoneView.as_view(), name='password_reset_done'),
 
-    re_path(fr'api/login/$', TokenObtainPairView.as_view(), name='token_obtain_pair'),
+    re_path(fr'api/login/$', MobileTokenObtainPairView.as_view(), name='token_obtain_pair'),
     re_path(fr'api/login/refresh/$', TokenRefreshView.as_view(), name='token_refresh'),
 
     re_path(r'^saved/?$', reader_views.saved_content),
--- a/sso/tests/views_test.py
+++ b/sso/tests/views_test.py
@@ -369,6 +369,61 @@ class AppleMobileTest(TestCase):
         self.assertFalse(User.objects.get(pk=user.pk).has_usable_password())
 
 
+class MobileLoginTest(TestCase):
+    """api/login/ -- the SimpleJWT endpoint the mobile app posts email/password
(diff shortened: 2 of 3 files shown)


yodem:
Superseded by #3614.

Both PRs implement the same feature (surfacing `sso_only_account` on `api/login/` for the mobile app). #3614 takes a different approach:

- It subclasses the serializer (`SSOAwareTokenObtainPairSerializer` via `serializer_class`), simplejwt's documented extension point, rather than catching `AuthenticationFailed` in `post()` and returning a raw `JsonResponse` from a DRF view — which bypasses DRF's exception handler and content negotiation.
- It extracts a single shared `sso_only_account_info(email)` helper that `email_login` and `password_reset_api` also call, so the web and mobile detection paths can't drift apart.
- It carries forward the account-enumeration disclosure from the original implementation (#3580) as an explicit open note for reviewers, rather than leaving it unstated.

Closing this one in favor of #3614. Thanks for the work here — nothing personal about the approach, just consolidating on the serializer-based version.

PR #3614, opened 2026-08-12
feat(sso): surface sso_only_account on the mobile JWT login endpoint

yodem:
Two backend fixes the Sefaria-Mobile app needs for its auth screens. Consumer: Sefaria-Mobile#217.

## What this does

- **`api/login/` can now say *why* a login failed.** It was stock SimpleJWT, so an account that only has Google/Apple linked returned the same generic "no active account" as a wrong password. The app had no way to tell the user to use the SSO buttons. It now returns the same `401 {error, _auth: {code: 'sso_only_account', providers}}` that `email_login` already gives web.
- **`password_reset_api` is now `@csrf_exempt`.** It wasn't, so the app's forgot-password screen got a 403 with an HTML body on every attempt. `google_mobile` and `apple_mobile` already carry the decorator; this one was missed.
- **One shared `sso_only_account_info(email)` helper**, called by `email_login`, `password_reset_api` and the new serializer — so web and mobile detection can't drift.
- **Four queries collapsed to two** in that helper (`user_exists` + `get_user` repeated the same lookup). Flagged on #3580's review, never addressed. Semantics unchanged.

## Verified

```
POST /api/auth/password/reset  ->  403 text/html "CSRF verification failed"   (before)
POST /api/auth/google/mobile   ->  400 {"error": "auth.social_signin_failed"} (exempt, reaches the view)
```

## Why a serializer, not a view override

Subclasses via `serializer_class` — SimpleJWT's documented extension point. #3612 did the same job by catching `AuthenticationFailed` in `post()` and returning a raw `JsonResponse`, which bypasses DRF's exception handler and content negotiation. That PR is closed in favour of this one.

## History

- Supersedes **#3580**, closed 2026-08-06 as "not currently needed — nothing in Sefaria-Mobile consumes `sso_only_account`". That condition is now met.
- Replaces **#3612**.
- Related: **#3609**.

## ⚠️ Needs a decision before merge

This extends **account enumeration** to `api/login/` — an unauthenticated caller can learn whether an email is registered and which providers it uses. `password_reset_api` already leaks the same. Neither endpoint is throttled; the project has no `DEFAULT_THROTTLE_CLASSES` anywhere. The exposure already ships on `email_login`, so this widens an existing surface rather than opening a new one — but it's a deliberate tradeoff, not an oversight. Throttling is worth its own story.

Separately: middleware-level rejections return **HTML**, so the app can only ever show a generic error for them. A JSON error handler for the mobile-facing auth endpoints is worth doing on its own.

## Tests

- `sso/tests/views_test.py` — 31 passing, including `MobileLoginTest` (valid login / wrong password / unknown user / SSO-only account)
- `password_reset_api` and `email_login` keep identical status codes and bodies after the refactor

### DIFF
--- a/sso/views.py
+++ b/sso/views.py
@@ -43,6 +43,14 @@ def sso_only_account_info(email):
     One user lookup plus one social-account fetch: has_usable_password()
     reads a field already on the fetched user, so it needs no query of
     its own.
+
+    That password check is defence-in-depth rather than a live branch: an
+    account with both a usable password and a linked provider is not
+    reachable today, since linking wipes the password (adapters.py's
+    "SSO always wins on an email collision"). It is kept because it is
+    free and because that is a product decision, not an invariant -- if it
+    is ever revisited, this check is what stops a mistyped password being
+    reported as "your account is Google-only".
     """
     try:
         user = get_user(email)
--- a/sso/views.py
+++ b/sso/views.py
@@ -10,12 +10,16 @@ from allauth.socialaccount.providers.google.views import (
     login_by_token as google_login_by_token,
 )
 from django.contrib.auth import authenticate, login as auth_login
+from django.contrib.auth.models import User
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt, ensure_csrf_cookie
 from django.views.decorators.http import require_POST
+from rest_framework import exceptions
+from rest_framework_simplejwt.serializers import TokenObtainPairSerializer
 from rest_framework_simplejwt.tokens import RefreshToken
+from rest_framework_simplejwt.views import TokenObtainPairView
 
-from emailusernames.utils import user_exists, get_user
+from emailusernames.utils import get_user
 from sefaria.forms import SefariaPasswordResetForm
 
 logger = structlog.get_logger(__name__)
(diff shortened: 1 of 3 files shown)


sso/views.py
```diff
@@ -26,29 +30,80 @@ def _jwt_for_user(user):
...
+    One user lookup plus one social-account fetch: has_usable_password()
+    reads a field already on the fetched user, so it needs no query of
+    its own.
+    """
+    try:
+        user = get_user(email)
+    except User.DoesNotExist:
+        return None
+    if user.has_usable_password():
+        return None
```
YishaiGlasner, line 60: isn't this redundant?

yodem: Not redundant — it is the only thing distinguishing "wrong password" from "no password at all".

We only reach this function *after* the credential check has already failed. At that point two very different users look identical:

1. Has a password **and** a linked Google account, and just typo'd it.
2. Has no usable password and can only sign in through Google.

`return providers or None` on the last line cannot tell them apart — it only knows whether `socialaccount_set` is non-empty, which is true for both. Without the `has_usable_password()` guard, user 1 mistypes their password once and gets told their account only works with Google Sign-In, with no way forward.

The guard is also free: `has_usable_password()` reads the `password` field already loaded by the `get_user()` above it, so it costs no extra query.

Your question did expose a real gap, though: the existing `test_wrong_password_returns_generic_error_unchanged` used an account with **no** linked provider, so it passed with or without the guard — nothing in the suite actually pinned this behaviour down. Added `test_wrong_password_with_linked_social_account_is_not_sso_only` in `141b1aa48`, which fails if the guard is removed.

YishaiGlasner: i mean redundant for it can go on and find no providers, so it will be `[] or None` which means the same None

yodem: That holds for a password user with **no** linked provider — `[]` falls through to `None` exactly as you say, and that case really is redundant. It breaks for a password user **with** a linked provider, where the query returns `['google']`, which is truthy, so `providers or None` returns it and the API reports `sso_only_account` for someone who has a perfectly good password and merely mistyped it.

I checked rather than argued it — removed the guard and ran the wrong-password tests:

```
FAILED  test_wrong_password_with_linked_social_account_is_not_sso_only
        assertIn("detail", data)   # got _auth.code == sso_only_account
PASSED  test_wrong_password_returns_generic_error_unchanged
```

The one that keeps passing is the pre-existing test — its user has no `SocialAccount`, which is why the suite never caught this and why your reading looks right from the code alone. The new test is the one that pins it. Guard restored; 33/33 green.

Such accounts are real: sign up with email/password, later link Google; or the reverse, then set a password. `sso/adapters.py:139` already branches on `has_usable_password()` for exactly that population.

Fair to say the docstring undersold this. It says "or None if the account can (also) sign in with a password" without spelling out that the *only* thing standing between those users and a wrong error message is this line. Happy to tighten it if you think that would have saved you the round trip.

YishaiGlasner: but if the account is linked to google (or applle), the password is not usable

yodem: You're right, and my example was wrong. I went looking for a path that produces a usable password alongside a linked provider and there isn't one:

- **Linking wipes it.** `adapters.py` `pre_social_login` — `if existing_user.has_usable_password(): set_unusable_password()` before `connect()`, and allauth's own `is_existing` path does the same wipe. That is the documented "SSO always wins on an email collision" decision.
- **The reset API refuses.** `password_reset_api` returns `sso_only_account` before ever calling `form.save()`.
- **The web reset form refuses too.** `SefariaPasswordResetForm` only overrides the email field, so it inherits Django's `PasswordResetForm.get_users()`, which filters on `has_usable_password()` — an SSO-only user never receives a link.
- The only other `set_password` is `forms.py:95`, on new-user registration.

So the state my test constructs is one the application cannot reach, and I built it by going around the adapter. Sorry for the confident wrong answer — the test failing without the guard proved the guard does something, not that anything was calling for it.

I have kept it in `841105c69` but stopped it pretending to be a live branch. The docstring and the test now say plainly that it is defence-in-depth and unreachable today. The reason I would rather keep than delete: "SSO always wins" is a *product decision* recorded in `adapters.py`, not an invariant — it is the kind of thing that gets revisited, and if it is, this line is what stops a mistyped password being reported as "your account is Google-only". It costs no query.

Happy to drop both the check and the test if you would rather not carry code for a state that cannot happen — your call, and I do not feel strongly.

YishaiGlasner: i don't see the scenario where has_usable_password remain but one can have google account with usable password. all the idea here is that if you use sso you can't use the password. if this changes, all the logic will be changed

PR #3615, opened 2026-08-12
feat(Sheets): open in new tab if not sheets

stevekaplan123:
## Description
Links in sheets will open in the current window if they are links to other sheets or to voices.  Otherwise we open them in a new tab.


static/js/sheets/Sheet.jsx
```diff
@@ -51,7 +51,21 @@ class Sheet extends Component {
       if (!href) { return; }
       const moduleTarget = target.getAttribute('data-target-module');
       const fullUrl = Sefaria.util.fullURL(href, moduleTarget); // Ignores moduleTarget if it's null
-      Sefaria.util.openInNewTab(fullUrl);
+      let url;
+      try {
+        url = new URL(fullUrl, window.location.href); // second arg resolves relative paths against the current page
+      } catch (err) {
+        Sefaria.util.openInNewTab(fullUrl);
```
dcschreiber: I think your indentation is off.

static/js/sheets/Sheet.jsx
```diff
@@ -51,7 +51,21 @@ class Sheet extends Component {
...
+        return;
+      }
+      const isSheetLink = /^\/sheets\/\d+/.test(url.pathname);
```
dcschreiber: Could be worth creating a function for this or seeing if we have an existing one.

dcschreiber:
_(Claude writing on Daniel's behalf — these review comments were AI-generated and Daniel reviewed them before posting. Posted as one comment because a pending draft review blocks inline comments.)_

The approach looks sound and the fallbacks (try/catch around `new URL`, defaulting to `openInNewTab`) are careful. Three comments on `static/js/sheets/Sheet.jsx`, ordered by importance:

**1. `isSheetLink` checks only the pathname, not the hostname** (line 61)

A link to `https://someothersite.com/sheets/42` would match the regex and navigate the current window to the external site. There's an existing helper that does exactly the right check against all of our module domains in both languages:

```js
const isSheetLink = Sefaria.isSefariaURL(url) && /^\/sheets\/\d+/.test(url.pathname);
```

**2. Modifier clicks are overridden** (line 65)

Because `e.preventDefault()` fires unconditionally, a cmd/ctrl-click (or shift-click) on a sheet/Voices link now navigates the *current* tab, contradicting the user's explicit "open in new tab" gesture. An early return before `preventDefault()` lets the browser handle it:

```js
if (e.metaKey || e.ctrlKey || e.shiftKey) { return; }
```

(The old code hijacked everything into a new tab too, but it never forced a same-tab navigation, so this branch makes it newly noticeable.)

**3. The Voices hostname comparison is interface-language-scoped** (line 62)

Two edge cases, since `getModuleURL` resolves via `domainModules[currentInterfaceLang]`:

- A link to the Hebrew Voices domain clicked while in the English interface won't match, and opens a new tab instead of staying put.

Comparing against `getDomainHostnames()` (or both languages' Voices hostnames) would cover both. Fine to leave as-is if the mapping is guaranteed populated.

dcschreiber (changes requested):
These comments are small, but Claude should have left one or two comments that should actually be fixed.

static/js/sheets/Sheet.jsx
stevekaplan123: It's only one line so I think it's fine as is.  Could be something to address eventually if we do re-use this logic elsewhere

stevekaplan123:
> _(Claude writing on Daniel's behalf — these review comments were AI-generated and Daniel reviewed them before posting. Posted as one comment because a pending draft review blocks inline comments.)_
> 
> The approach looks sound and the fallbacks (try/catch around `new URL`, defaulting to `openInNewTab`) are careful. Three comments on `static/js/sheets/Sheet.jsx`, ordered by importance:
> 
> **1. `isSheetLink` checks only the pathname, not the hostname** (line 61)
> 
> A link to `https://someothersite.com/sheets/42` would match the regex and navigate the current window to the external site. There's an existing helper that does exactly the right check against all of our module domains in both languages:
> 
> ```js
> const isSheetLink = Sefaria.isSefariaURL(url) && /^\/sheets\/\d+/.test(url.pathname);
> ```
> 
> **2. Modifier clicks are overridden** (line 65)
> 
> Because `e.preventDefault()` fires unconditionally, a cmd/ctrl-click (or shift-click) on a sheet/Voices link now navigates the _current_ tab, contradicting the user's explicit "open in new tab" gesture. An early return before `preventDefault()` lets the browser handle it:
> 
> ```js
> if (e.metaKey || e.ctrlKey || e.shiftKey) { return; }
> ```
> 
> (The old code hijacked everything into a new tab too, but it never forced a same-tab navigation, so this branch makes it newly noticeable.)
> 
> **3. The Voices hostname comparison is interface-language-scoped** (line 62)
> 
> Two edge cases, since `getModuleURL` resolves via `domainModules[currentInterfaceLang]`:
> 
> * A link to the Hebrew Voices domain clicked while in the English interface won't match, and opens a new tab instead of staying put.
> 
> Comparing against `getDomainHostnames()` (or both languages' Voices hostnames) would cover both. Fine to leave as-is if the mapping is guaranteed populated.

Great points, thanks.  I addressed each of these three issues.

static/js/sefaria/sefaria.js
```diff
@@ -520,16 +520,57 @@ Sefaria = extend(Sefaria, {
...
+    const langModules = Sefaria.domainModules?.[Sefaria._getShortInterfaceLang()] || {};
+    return Sefaria._hostnamesFromModuleURLs(Object.values(langModules));
+  },
+  getAllHostnames: function() {
+    // Every module, every interface language -- i.e. every hostname this deploy answers on.
+    // example -> Set { "www.sefaria.org", "voices.sefaria.org", "www.sefaria.org.il", "voices.sefaria.org.il" }
+    const allModuleUrls = Object.values(Sefaria.domainModules || {}).flatMap(langModules => Object.values(langModules || {}));
+    return Sefaria._hostnamesFromModuleURLs(allModuleUrls);
+  },
+  getCurrentModuleHostnames: function(module) {
```
dcschreiber, line 566: _(Claude writing on Daniel's behalf)_

The name reads as the opposite of what the function does. `getCurrentLangHostnames` returns the *current* language, so `getCurrentModuleHostnames` sounds like it returns the *current* module — but it takes a module as an argument and deliberately spans every language.

`getModuleHostnames(module)` would say it plainly, and the contrast with `getCurrentLangHostnames` then does real work: one is scoped to what the reader is using, the other is not scoped at all.

static/js/sheets/Sheet.jsx
```diff
@@ -45,12 +45,30 @@ class Sheet extends Component {
...
     const target = e.target.closest('a');
-    if (target) {
-      e.preventDefault();
-      const href = target.getAttribute('href');
-      if (!href) { return; }
-      const moduleTarget = target.getAttribute('data-target-module');
-      const fullUrl = Sefaria.util.fullURL(href, moduleTarget); // Ignores moduleTarget if it's null
+    if (!target) { return; }
+    // Cmd/ctrl/shift-click is the user explicitly asking for a new tab or window -- let the browser do its thing.
+    if (e.metaKey || e.ctrlKey || e.shiftKey) { return; }
```
dcschreiber, line 50: _(Claude writing on Daniel's behalf)_

No change needed, just so you know it is belt-and-braces: `ReaderApp` already installs `handleInAppClickWithModifiers` as a **capture**-phase listener on `document`, and it calls `stopImmediatePropagation()` on any modifier click. Since `document` sits above React's root container, that fires first and the event never reaches this handler.

So this line is unreachable for `Sheet` as rendered from `ReaderPanel`. Harmless, and arguably worth keeping in case the component is ever mounted outside `ReaderApp`. Flagging it only so nobody later reads it as the thing that makes cmd-click work.

dcschreiber (changes requested):
_(Claude writing on Daniel's behalf — these review comments were AI-generated and Daniel reviewed them before posting.)_

Re-reviewed. The three earlier points are properly fixed: the hostname check is real now, both languages' Voices domains match, and splitting `getDomainHostnames` into three getters is a clean way to get there. Both old callers were updated, so nothing else reads the old helper.

One thing I want to flag as checked rather than as a problem: this PR newly wires `handleClick` into reader mode, so a link inside sheet content no longer opens a new panel and does a full page navigation instead. I went looking for the reasoning and found it in sc-46102 — the stacking, uncloseable panels are the bug being fixed, and Mickey's acceptance criteria ask for exactly this. So no change wanted there.

Two small items inline, one of which I would like changed before merge.

dcschreiber (approved):

PR #3620, opened 2026-08-13
chore: fall back to enDesc/heDesc

stevekaplan123:
## Description
Prefer short description and fall back to long description


saengel (approved):
Thanks!

PR #3637, opened 2026-08-18
fix(backup): make pg_dump verification actually compare row counts

yodem:
## Summary

The `=== Verifying Dump Contents ===` block in `helm-chart/sefaria/templates/configmap/create-pg-dumps.yaml` was tautologically green: it printed `SUCCESS: All row counts match within tolerance!` while comparing nothing.

**Root cause**: `pg_restore --list` output lines look like:

```
218; 1259 16485 TABLE DATA public auth_user sefaria
```

The script extracted `$6` (`public`, the schema) instead of `$7` (`auth_user`, the table). Every loop iteration then queried a nonexistent relation named `public`. The psql/pg_restore errors were swallowed by `2>/dev/null`, so `DB_COUNT` and `DUMP_COUNT` both collapsed to `0`, `DIFF` was always `0`, `STATUS` was always `OK`, `MISMATCH` stayed `0` across all 30 tables, and the script declared success.

**Evidence**: production run `production-postgresbackup-manual-1787039791888` logged 30 identical rows reading `public | 0 | 0 | 0 | OK` and a final `TOTAL | 0 | 0 | 0` — yet still printed SUCCESS. The dump itself was fine (a correct 37,486,432-byte file was produced); only the verification check was blind. This check exists specifically to catch silent backup regressions, and as written it could not have caught any.

## Changes (all inside the `create-dumps.sh` script body; connection/templating and pg_dump flags untouched)

1. `awk '{print $6}'` → `awk '{print $7}'` to pull the actual table name from `pg_restore --list`.
2. The DB_COUNT (`psql`) and DUMP_COUNT (`pg_restore --data-only`) calls no longer silently coerce a failure into `0`. A failed count is captured, echoed, and marked as a verification failure (`DB QUERY FAILED` / `DUMP READ FAILED`), incrementing `MISMATCH`.
3. Added a guard right after `DUMP_TABLES` is computed: if it's empty, the script prints an error and exits `1` instead of falling through to a false "SUCCESS".
4. Added a tautology guard before the final summary: if `TOTAL_DUMP` or `TOTAL_DB` is `0`, the verification is meaningless — the script prints an explicit error and counts it as a mismatch. This is the check that would have caught the present bug.
5. The previously commented-out
   ```bash
   # if [[ $MISMATCH -gt 0 ]]; then
   #   exit 1
   # fi
   ```
   is now live, so a real mismatch fails the Job instead of printing a warning under a green checkmark. The existing propagation of a nonzero `pg_dump` exit code (`exit $DUMP_EXIT_CODE`) is preserved.

`PG_HOST`/`PG_DB` and the `{{ if .Values.postgres.host }}...{{ else }}${DATABASES_HOST}{{ end }}` templating, and all `pg_dump` flags, are unchanged.

## Test plan

- [x] Extracted the rendered `create-dumps.sh` body and ran `bash -n` — parses cleanly.
- [x] Verified the awk field fix: `echo '218; 1259 16485 TABLE DATA public auth_user sefaria' | awk '{print $7}'` → `auth_user` (table name), vs `awk '{print $6}'` → `public` (schema, the bug).
- [x] `helm template sefaria helm-chart/sefaria -f <prod values extracted from envs/prod/helmrelease.yaml> -s templates/configmap/create-pg-dumps.yaml` renders successfully with the new logic intact.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

https://claude.ai/code/session_01TfZU5AMmB6D9gtNos3jnQn

### DIFF
--- a/helm-chart/sefaria/templates/configmap/create-pg-dumps.yaml
+++ b/helm-chart/sefaria/templates/configmap/create-pg-dumps.yaml
@@ -140,15 +140,25 @@ data:
         continue
       fi
 
-      # Get DB count (live query)
+      # Get DB count (live query). Note: "set -e" is in effect and these are bare
+      # command substitutions, so the "|| VAR=$?" form is required -- without it a
+      # failing psql/pg_restore aborts the script before the error branch below is
+      # ever reached. stderr goes to a file so it cannot be counted as dump rows.
+      DB_COUNT_STATUS=0
       DB_COUNT_OUTPUT=$(psql -h "$PG_HOST" -p "$PG_PORT" -U "$PG_USER" -d "$PG_DB" -t -c \
-        "SELECT COUNT(*) FROM \"$TABLE\";" 2>&1)
-      DB_COUNT_STATUS=$?
+        "SELECT COUNT(*) FROM \"$TABLE\";" 2>"$DATADIR/verify_db_err.txt") || DB_COUNT_STATUS=$?
+      if [[ $DB_COUNT_STATUS -ne 0 ]]; then
+        DB_COUNT_OUTPUT=$(cat "$DATADIR/verify_db_err.txt")
+      fi
       DB_COUNT=$(echo "$DB_COUNT_OUTPUT" | xargs)
 
       # Get dump count by extracting and counting rows
-      DUMP_COUNT_OUTPUT=$(pg_restore --data-only --table="$TABLE" "$DUMP_FILE" 2>&1)
-      DUMP_COUNT_STATUS=$?
+      DUMP_COUNT_STATUS=0
+      DUMP_COUNT_OUTPUT=$(pg_restore --data-only --table="$TABLE" "$DUMP_FILE" \
+        2>"$DATADIR/verify_dump_err.txt") || DUMP_COUNT_STATUS=$?
+      if [[ $DUMP_COUNT_STATUS -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


yodem:
Follow-up commit \`1eb5b1d2b\` fixes a latent bug in the first commit that a syntax check would not catch.

The error branches added in \`0b7caf36a\` were unreachable. The script runs under \`set -e\`, and

```bash
DB_COUNT_OUTPUT=$(psql ... 2>&1)
DB_COUNT_STATUS=$?
```

is a *bare* command substitution, not a pipeline — so a failing `psql` aborts the script at the assignment and `DB_COUNT_STATUS=$?` never executes. The job would die with no diagnostic instead of printing which table failed.

This is the original bug's mechanism inverted, which is worth stating explicitly: the old code captured through a **pipeline** ending in `xargs`, whose exit status is always 0 — so `set -e` never fired, the empty result silently became `0`, and the check went green. Same `set -e`, opposite failure, both invisible.

Fixed with the `|| VAR=$?` form (suppresses errexit) and by sending stderr to a file rather than merging via `2>&1` — merged stderr flowed into the `wc -l` line count and would have inflated the dump row count on any `pg_restore` warning.

Verified: handler now reached (`status=127`, stderr captured) where the previous form exited before it; `bash -n` clean on the extracted script; `helm template` renders against `envs/prod/helmrelease.yaml` values.

PR #3646, opened 2026-08-20
POST endpoint for PBS

ekoslow1-creator:
## Description
POST request for sefaria.org/api/powered-by which is also specifically compatible with Formstack submission source

## Code Changes
Updates to sefaria/powered_by/views.py


powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+# --- Formstack payload translation -------------------------------------------
+#
+# Field IDs below are from the live "Powered by Sefaria" Formstack form. The
+# webhook isn't wired up yet, so the exact payload shape (Field<ID> vs bare ID
+# keys, checkbox values as a list vs a comma-joined string) is an assumption
+# based on Formstack's documented options; _formstack_checkbox_values() and
+# _formstack_field() tolerate the reasonable variants. Once the webhook is
+# live, sanity-check a real payload against these assumptions.
+
+FORMSTACK_FIRST_NAME_FIELD = "179244240"
```
saengel, line 50: Why are these fields separated out? Shouldn't they all live in the same dict?

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+    "196457843": "is_developer",
+    "196457992": "vibe_coded",
+    "179245509": "consent_to_display",
+    "191150099": "has_pbs_logo",
+}
+
+# "Which Sefaria data or tools did you use?", "Which categories of endpoints
+# did you utilize?", and the 11 conditional "Which specific endpoints did you
+# use?" fields all merge into Project.sefaria_tools_used.
+FORMSTACK_TOOLS_USED_FIELDS = (
```
saengel, line 81: This makes sense to keep separate

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+    cleaned, error = clean_and_default_post_body(body)
+    if error:
+        return jsonResponse({"error": error}, status=400)
+
+    project_link = cleaned.pop("project_link")
+    # Apply defaults only when creating a new project (not on partial updates).
+    if not Project.objects.filter(project_link=project_link).exists():
+        cleaned.setdefault("submission_source", SubmissionSource.FORMSTACK)
+        cleaned.setdefault("submission_date", timezone.now())
+    # TEMP LOCAL TESTING ONLY (elza, 2026-08-20): is_published force-False on
```
saengel: This comment needs to be cleaned up

powered_by/tests/powered_by_api_test.py
```diff
@@ -103,3 +103,477 @@ def test_staff_sees_all_projects_with_private_fields(client, projects):
     for project in data:
         for field in PRIVATE_FIELDS:
             assert field in project
+
+
+# --- clean_and_default_post_body() -------------------------------------------
```
saengel, line 108: Do all the tests pass?

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+FORMSTACK_TOOLS_USED_FIELDS = (
+    "196457970", "196602042",
+    "196602151", "196602402", "196602409", "196602439", "196602489",
+    "196602566", "196602632", "196602641", "196602679", "196602688", "196602699",
+)
+
+
+def _formstack_field(body, field_id):
+    """Formstack fields may arrive as "Field<id>" or bare "<id>" keys."""
+    return body.get(f"Field{field_id}", body.get(field_id))
```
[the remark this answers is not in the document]
saengel: This is easy and logical, just accept lowercase versions as well.

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+    # DO NOT COMMIT. Revert before pushing.
+    #
+    # else:
+    #     # Any edit to an existing project un-publishes it, regardless of prior
+    #     # state or anything in the request body, so a staff member must review
+    #     # and re-publish. project_link is a public field (visible via GET), so
+    #     # without this an anonymous caller could deface a live project by
+    #     # POSTing an "update" and have it stay published immediately.
+    #     cleaned["is_published"] = False
+    cleaned.setdefault("is_published", True)
```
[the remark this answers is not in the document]
saengel: For now, we want new submissions to bypass the editorial gate. That gate will be developed at a later point and is being scoped separately. 

We definitely don't want to overwrite though, perhaps for now via the API we should limit it to only being able to create "new" projects - any project changes must be made manually via the admin for now.

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+
+    project, created = Project.objects.update_or_create(
+        project_link=project_link,
+        defaults=cleaned,
+    )
```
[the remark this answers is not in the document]
saengel: Let's use an ID instead.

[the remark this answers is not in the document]
saengel: This is resolved by us removing the ability to upsert for now.

saengel (changes requested):
Still needs some polishing, great progress in the right direction.

saengel:
Looks like Formstack can accept some kind of key passed in the headers for webhooks (see: https://help.formstack.com/hc/en-us/articles/44592107329683-Webhooks) let's also add a key of sorts to avoid any un-authenticated POST issues.

powered_by/views.py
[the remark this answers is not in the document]
yitzhakc: Is there a story for this?

yitzhakc (changes requested):
Overall looks good. I would ideally want a way for us to be notified if we're getting invalid submissions from  Formstack in a situation where their fields or structure changes, but that can be a separate story.

powered_by/views.py
```diff
@@ -1,22 +1,298 @@
...
+                parsed = None
+            if parsed is None:
+                return None, "submission_date must be a valid ISO 8601 datetime"
+            value = parsed
+
+        if field == "creator_email" and value:
+            try:
+                validate_email(value)
+            except DjangoValidationError:
+                return None, "creator_email must be a valid email address"
```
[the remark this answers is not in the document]
ekoslow1-creator: change         if field == "creator_email" and value: --> drop "and value"

powered_by/tests/powered_by_api_test.py
ekoslow1-creator: The tests pass

powered_by/views.py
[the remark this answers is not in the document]
ekoslow1-creator: yes

powered_by/views.py
```diff
@@ -1,22 +1,310 @@
...
+    sefaria/local_settings.py / env, never in source control.
+    """
+    expected = getattr(settings, "FORMSTACK_HANDSHAKE_KEY", None)
+    provided = body.get("HandshakeKey") if isinstance(body, dict) else None
+    return bool(expected) and isinstance(provided, str) and hmac.compare_digest(provided, expected)
+
+
+def _powered_by_post(request):
+    try:
+        body = json.loads(request.body)
```
yodem: **Unauthenticated 500 — parse runs before auth.** `json.loads` is on line 282; the handshake check is on line 286. Deeply-nested JSON raises `RecursionError`, which is not in the `except` on line 283. Verified in this repo's venv:

```
payload = "["*10000 + "]"*10000     # ~20 KB
json.loads(payload) -> RecursionError
issubclass(RecursionError, (json.JSONDecodeError, UnicodeDecodeError)) -> False
issubclass(RecursionError, ValueError) -> False
```

`catch_error_as_json` doesn't catch it either, so a 20 KB anonymous POST returns 500, repeatably, with no rate limiting in front of it.

Cheapest fix is to widen the except; the more principled one is to authenticate before deserialising, which isn't possible while the credential lives *inside* the JSON. Worth noting as a real cost of the handshake-in-body design.

powered_by/views.py
```diff
@@ -1,22 +1,310 @@
...
+
+
+REQUIRED_FIELDS = ("project_name", "project_link")
+
+WRITABLE_FIELDS = (
+    "submission_date", "submission_source", "sefaria_tools_used", "tech_used_raw",
+    "technical_experience", "vibe_coded", "project_why", "project_name", "project_link",
+    "project_source_code", "project_reach", "project_desc", "project_category",
+    "image_url", "has_pbs_logo", "consent_to_display", "creator", "creator_email",
+    "is_developer", "job_title", "found_sefaria", "submitter", "salesforce_id", "notes",
```
yodem: `submitter`, `salesforce_id` and `notes` are writable here, but `powered_by/models.py:43-46` declares all three in `PRIVATE_FIELDS` — "PII / internal staff metadata." They're stripped on read and settable on write.

The risk is trust rather than data loss: a forged `notes` reads in the admin as if a colleague already vetted the submission, and `salesforce_id` is a soft FK something may cross-reference. `is_published`/`featured`/`tags`/`status` are correctly excluded.

Suggest dropping `submitter` and `salesforce_id` from `WRITABLE_FIELDS`.

powered_by/views.py
```diff
@@ -1,22 +1,310 @@
...
+# did you utilize?", and the 11 conditional "Which specific endpoints did you
+# use?" fields all merge into Project.sefaria_tools_used.
+FORMSTACK_TOOLS_USED_FIELDS = (
+    "196457970", "196602042",
+    "196602151", "196602402", "196602409", "196602439", "196602489",
+    "196602566", "196602632", "196602641", "196602679", "196602688", "196602699",
+)
+
+
+def _formstack_field(body, field_id):
```
yodem, line 88: Worth settling before the webhook goes live: per [Formstack's docs](https://help.formstack.com/hc/en-us/articles/44592535914387-WebHook-Submit-Actions), *"Post using field names"* is the **default** and field IDs are an opt-in checkbox ([`postDataFieldKeys`](https://developers.formstack.com/reference/getwebhook-1): `field_names | field_ids | api_friendly_field_names | ...`).

Left on the default, keys arrive as labels (`"Project Name"`), this lookup finds nothing, `project_link` is missing, and the whole submission 400s — every real submission dropped.

Not a code change: a Formstack config setting plus one captured real payload to test the map against.

sefaria/local_settings_example.py
```diff
@@ -358,6 +358,8 @@
 WEBHOOK_USERNAME = os.getenv("WEBHOOK_USERNAME")
 WEBHOOK_PASSWORD = os.getenv("WEBHOOK_PASSWORD")
 
+FORMSTACK_HANDSHAKE_KEY = os.getenv("FORMSTACK_HANDSHAKE_KEY")
```
yodem: **Blocker — this setting never reaches any k8s environment, so every POST 401s on cauldron, staging and prod.**

`sefaria/settings.py:362-370` imports `sefaria.local_settings`, which in k8s is rendered from `helm-chart/sefaria/templates/configmap/local-settings-file.yaml` and mounted with `subPath: local_settings.py` (`rollout/web.yaml:210-213`). That template has no FORMSTACK line, so `getattr(...)` at views.py:275 returns `None` and line 286 returns 401 — **even once the cluster secret exists**, because nothing turns the env var into a Django setting.

Fix: add the `os.getenv` line to `local-settings-file.yaml` beside `WEBHOOK_USERNAME` (~line 399). Cluster secrets are staged in Sefaria/infrastructure#671 (dev) and #672 (prod, draft).

powered_by/tests/powered_by_api_test.py
```diff
@@ -103,3 +103,485 @@ def test_staff_sees_all_projects_with_private_fields(client, projects):
...
+def test_post_invalid_json_returns_400(client):
+    response = client.post("/api/powered-by", data="not json", content_type="application/json")
+    assert response.status_code == 400
+    assert "JSON" in response.json()["error"]
+
+
+# --- view: HandshakeKey requirement --------------------------------------------
+
+@pytest.mark.django_db
+def test_post_missing_handshake_key_returns_401(client):
```
yodem: These tests never run in CI, so a green `pytest-job` here says nothing about them.

`build/ci/createJobFromRollout.sh:27` runs `pytest ... ./sefaria ./sso ./reader` — no `./powered_by` — and `pytest.ini`'s `python_files` has no `powered_by` pattern, so the file is excluded twice. The job *is* triggered; it just collects nothing from here.

Pre-existing, but this PR adds ~470 lines of security-relevant tests into the blind spot. Fix: add `./powered_by` to the pytest args and `powered_by/tests/*_test.py` to `python_files`.

yodem (commented):
Independent review pass on the handshake implementation, run with multi-agent verification and every claim re-checked against commit `3e20344c` before posting. Findings below are only the ones not already covered by Copilot or @saengel, and each was executed or grepped rather than reasoned about.

**Two things I checked and am NOT reporting, because they turned out to be wrong:**
- *Non-string `image_url` / `project_source_code` causes a 500.* Refuted — Django 6.0.4's `URLValidator.__call__` opens with `if not isinstance(value, str) ... raise ValidationError`, so it's caught at views.py:206-208 and returns a clean 400. (This was true of the `creator_email` path before 3e20344c fixed it; it was never true of the URL fields.)
- *The handshake key gets captured by request logging.* Refuted for this deployment — `sefaria/system/logging.py:decompose_request_info` binds only `requestUrl` and `requestMethod`, no body, and no Sentry/APM body capture is configured. The generic risk of a body-carried secret stands, but there is no live exposure here.

The design call itself (handshake key rather than the HMAC signature Formstack also offers) is defensible: the credential is fail-closed, compared with `compare_digest`, and auth sits inside the POST branch so the public GET is untouched. The findings below are about the edges around it, not the choice.

yodem:
## Recommendation: switch the auth to Formstack's HMAC signature

I should have raised this in my review above instead of reviewing around it — apologies. The design [we agreed on](https://app.shortcut.com/sefaria/story/46160) was Formstack's **HMAC signature**, and this PR implements the **handshake key** instead. They're two separate Formstack features ([`hmacSecret` + `customHmacHeader` vs `sharedSecret`](https://developers.formstack.com/reference/getwebhook-1)), and `git grep -i "x-fs\|signature" -- powered_by/` on this branch returns nothing.

Both are shared-secret schemes, so this isn't "simple vs. enterprise." The difference is **where the secret travels**:

| | Handshake key (as built) | HMAC signature |
|---|---|---|
| Secret on the wire | every request, cleartext | never — only a digest |
| A stolen credential lets an attacker… | forge **unlimited new** submissions | replay **only the one** payload it signed |
| Payload tampering | undetectable | detected |
| Authenticate before parsing | impossible by design | yes — verify raw bytes first |
| Lines of code | ~4 | ~8 |

The second row is the crux: a handshake key is a bearer secret, so anyone who sees one delivery can mint submissions indefinitely. A signature is bound to the bytes it signed.

**It also removes two real bugs rather than patching them.** I originally filed both as inline comments and have deleted them, because switching makes them vanish:

1. `json.loads` currently runs *before* the auth check (unavoidable — the credential is inside the JSON). A ~20 KB deeply-nested body raises `RecursionError`, which isn't in the `except` clause and isn't caught by `catch_error_as_json` → **unauthenticated 500**, verified end-to-end with `django.test.Client`. HMAC verifies `request.body` bytes first, so an unsigned request never reaches the parser.
2. `hmac.compare_digest` on `str` requires ASCII on both sides; a `HandshakeKey` of `"café"` raises `TypeError` → another **unauthenticated 500** (also verified). Comparing hex digests makes that unreachable.

### Sketch

```python
def _formstack_signature_ok(request):
    secret = getattr(settings, "POWERED_BY_FORMSTACK_HMAC_SECRET", None)
    sent = request.headers.get("X-FS-Signature", "")
    if not secret or not sent:
        return False
    # request.body = the exact bytes received; never re-serialise a parsed dict
    digest = hmac.new(secret.encode(), request.body, hashlib.sha256).hexdigest()
    return hmac.compare_digest(f"sha256={digest}", sent)


def _powered_by_post(request):
    if not _formstack_signature_ok(request):          # BEFORE json.loads
        return jsonResponse({"error": "Unauthorized"}, status=401)
    try:
    ...
```

Formstack side: set **HMAC Key** (not Shared Secret) and keep Payload Format = **JSON** — with `urlencoded`, touching `request.POST` first consumes the stream and breaks verification. Cluster secrets are already staged under this name in Sefaria/infrastructure#671 (dev) and #672 (prod, draft), so they merge unchanged.

### What this does *not* solve

Formstack signs the body with **no timestamp**, so a captured signed request stays replayable indefinitely. Idempotency on the Formstack submission ID is still needed either way — which is the same change as @saengel's "let's use an ID instead."

Happy to push this as a commit on the branch if that's easier than reworking it.

ekoslow1-creator:
Changed from Handshake to HMAC key. One thing worth flagging: the infra tickets (Sefaria/infrastructure#671 dev, #672 prod) were staged for a FORMSTACK_HANDSHAKE_KEY secret — those need to be updated to provision POWERED_BY_FORMSTACK_HMAC_SECRET instead, or the endpoint will 401 in those environments the same way as before.

powered_by/views.py
yitzhakc: Did we check this with Elise? @ekoslow1-creator

yitzhakc (changes requested):

PR #3648, opened 2026-08-21
Feature/sc 46385/skip bad record follow up

stevekaplan123:
## Description
This expands upon a previous attempt to start the server or run `/admin/reset/cache` or `/admin/reset/toc` without failing due to bad data.  The paradigm there was to fail gracefully and alert us on Slack of the problem.  The first implementation to handle bad data deliberately kept the "this is a bad record" exception list narrow:
BAD_RECORD_EXCEPTIONS = (InputError, KeyError, ValueError, IndexError)
AttributeError and TypeError were left out on purpose. The worry was reasonable: those two usually mean a code bug (someone renamed a method, someone called a function with the wrong arguments), not bad data. If the guards caught them, a bad refactor would stop being a loud crash at boot and become a warning printed once per record — while the site quietly came up with half a library.  However, not including AttributeError and TypeError meant that the majority of bad data cases are never caught. The basic idea is this: Bad data and a broken refactor raise the same exception classes, but what actually differs is volume: one corrupt topic fails once; a renamed attribute fails on every record.

## Code Changes
First widen BAD_RECORD_EXCEPTIONS to include AttributeError and TypeError.
Then, add two breakers in [sefaria/helper/skip_tracking.py] to handle the cases when it's really the code that's causing the problem:

1. Signature breaker (10 identical errors, primary detector). A code change will generate the same error message every time, whereas real data corruption produces varied messages. Ten identical messages at one site therefore means broken code, and the build aborts. 
2. Volume breaker (500 skips at one site). Catches mass corruption too varied to repeat — e.g. a half-finished migration damaging many records in many different ways, which the signature breaker is structurally blind to. It's a "this library is too damaged to serve" stop, which is why it's set high at 500.
Note: Both counters are keyed per site (pathway + operation), so unrelated small problems in different parts of the build don't add up into a spurious abort. 

If the breaker itself fails, it raises a new error: BuildDegradationError, which means it won't get accidentally counted as a BAD_RECORD_EXCEPTION.  The breaker posts the summary itself before aborting.
The pathway calls are wrapped in try/finally — [reader/startup.py], and reset_cache / rebuild_toc in [sefaria/views.py] — so any abort, breaker or otherwise, still posts. Because the breaker also resets the log, the finally clause finds it empty and doesn't double-post.


sefaria/helper/skip_tracking.py
```diff
@@ -79,18 +123,73 @@ def get_skip_counts():
...
+            signature_key = (pathway, operation, error_type, detail)
+            _skip_signature_counts[signature_key] += 1
+            signature_count = _skip_signature_counts[signature_key]
+
+    # Checked outside the lock: plain int comparisons against local copies, and the caller
+    # must not be holding the lock when it responds.
+    if SIGNATURE_BREAKER_THRESHOLD and signature_count >= SIGNATURE_BREAKER_THRESHOLD:
+        return "signature"
+    if VOLUME_BREAKER_THRESHOLD and group_count >= VOLUME_BREAKER_THRESHOLD:
+        return "volume"
```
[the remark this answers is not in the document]
stevekaplan123: Correct that the counters are module-global and that two overlapping builds share them, but I don't think either named harm survives contact with the thresholds.

**Spurious trip.** For the signature breaker to fire falsely, two builds would have to contribute 10 *byte-identical* messages at the same site. Per the design note above, a message only repeats identically when it is record-independent — i.e. when the code is broken. So a combined trip is still diagnosing broken code correctly; it just gets there sooner. For the volume breaker you'd need 500 combined skips at one site, meaning each build independently skipped hundreds of records — that library is unservable regardless of which run owns which skip. Neither trip is actually spurious.

**A real systemic failure never reaching the threshold.** True that build B's `signal_and_reset_skip_counts()` zeroes build A's counters mid-run. But a systemically broken A raises the same error on every remaining record, so it re-climbs to 10 almost immediately. Escaping would require another build to finish and reset inside each of those short windows, repeatedly, until A runs out of records.

What's left is real but cosmetic: a Slack summary can carry the other build's skips or get split across two messages. Scoping state to a build invocation would mean threading a build id through every guard call site in the codebase — a large, invasive change for a message-attribution wart. The exposure is also small: startup runs before the process serves requests, so the only genuine overlap is two staff hitting `/admin/reset/cache` at the same moment.

Leaving as-is deliberately.

sefaria/helper/tests/audit_skip_bad_record_test.py
```diff
@@ -0,0 +1,865 @@
...
+                        {"_id": ObjectId(entry["_id"])}).deleted_count
+                except Exception:
+                    continue
+        os.remove(JOURNAL)
+
+    for collection in sorted(collections):
+        for field in NAME_FIELDS:
+            for prefix in NAME_PREFIXES:
+                total += db[collection].delete_many(
+                    {field: {"$regex": "^" + prefix}}).deleted_count
```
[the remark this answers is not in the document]
stevekaplan123: The literal reading is right and the SAFETY bullet is overstated — I'll reword it. But the by-name sweep stays, for three reasons:

1. `assert_local_mongo()` hard-exits unless Mongo is localhost, with no override flag. This can only ever run against a developer's disposable local DB.
2. `ZZAudit` / `zzaudit-` were chosen precisely because nothing in the Sefaria corpus starts with them.
3. Restricting cleanup to journalled ids removes the only backstop for the case it exists to cover — a lost journal. And a per-document ownership marker is the alternative that was already ruled out: several models reject unrecognised keys, so a `_skip_audit` field makes `Topic.load()` raise "loaded with unhandled key(s)" and the marker corrupts the very record under test. That's documented at the `JOURNAL` definition.

So the mismatch is in the prose, not the behaviour. Changing the bullet to: *"Never mutates or deletes real data. Everything it writes is synthetic; cleanup deletes by journalled id, falling back to a `ZZAudit*` / `zzaudit-*` name sweep on a local DB if the journal is lost."*

scripts/audit_skip_bad_record.py
```diff
@@ -0,0 +1,937 @@
...
+
+
+def cleanup():
+    """Delete everything inserted this run. Safe to call repeatedly."""
+    while _inserted:
+        collection, _id = _inserted.pop()
+        try:
+            db[collection].delete_one({"_id": _id})
+        except Exception as e:  # never let cleanup mask the real failure
+            print("  ! cleanup failed for {}/{}: {}".format(collection, _id, e))
```
yitzhakc: I'd recommend using a logging mechanism for the script, and not just random print statements everywhere

scripts/audit_skip_bad_record.py
```diff
@@ -0,0 +1,937 @@
...
+    doc.update(overrides)
+    return doc
+
+
+CASES = [
+    # -- category.py:223  TocTree vstate record -----------------------------------
+    case("S1", "TocTree vstate record", "vstate doc with no `title` field",
+         "vstate", {"first_section_ref": "Audit 1", "flags": {}},
+         _toc_tree, CAUGHT, "KeyError"),
+    case("S1", "TocTree vstate record", "`flags` is a string instead of a dict",
```
yitzhakc: What is `site`? What is the meaning of `S1`? My human eyes couldn't find any reference..

yitzhakc (changes requested):

scripts/audit_skip_bad_record.py
stevekaplan123: Added logging

scripts/audit_skip_bad_record.py
stevekaplan123: A "guard site" is one `with skip_bad_record(...)` or `log_skip(...)` block in the codebase -- one specific place that can swallow a bad record. Each one gets a short id so it can be selected and grouped:
```
        S1 .. S20   the twenty guard sites under audit, numbered in the order they appear
                    in CASES / LOG_SKIP_CASES below.
        B1, B2      the two breaker cases. These are not guard sites: they seed systemic
                    damage and assert the build ABORTS rather than skips.
```

These ids are invented by this script and exist nowhere else. What identifies the real code is the `operation` string sitting next to the id in each case (e.g. S1 is "TocTree vstate record", the guard sefaria/model/category.py:223), and that string is also what skip_tracking records at runtime and prints in its Slack summaries. The id is just a stable handle for it, since several cases share one site -- one guard is usually worth able to be corrupted in several ways.

sefaria/helper/tests/audit_skip_bad_record_test.py
```diff
@@ -0,0 +1,1016 @@
+# -*- coding: utf-8 -*-
```
yitzhakc, line 1: This might belong under `tests`, and not in the top-level `scripts` directory

yitzhakc (changes requested):

sefaria/helper/skip_tracking.py
```diff
@@ -49,6 +57,42 @@
 # Per (pathway, operation): total skips seen this build (including ones not stored verbatim).
 _skip_group_counts = defaultdict(int)
 
+# Per (pathway, operation, error_type, detail): how many times ONE error signature repeated.
+# Only guard-caught skips are counted here — see the breaker notes below.
+_skip_signature_counts = defaultdict(int)
+
+# --- Breakers -------------------------------------------------------------------------
```
yodem: I think its too many words for docstring, if you want to store info maybe use the wiki

yodem:
**[claude]** — automated review, posted from @yodem's account.

**Worth fixing — breaker counters outlive the build they're scoped to.** `_skip_group_counts` / `_skip_signature_counts` (`sefaria/helper/skip_tracking.py:52-63`) are process-global and cleared only by `signal_and_reset_skip_counts()`, which has exactly three call sites: `reader/startup.py:59`, `sefaria/views.py:806`, `sefaria/views.py:941`.

But the same guarded rebuild path is also entered directly from `add_new_topic_api` (`reader/views.py:3628`), `delete_topic` (`reader/views.py:3645`), and `sefaria/helper/topic.py:1267` — all of which call `library.get_topic_toc(rebuild=True)` without ever resetting. Skips from ordinary single-topic admin edits therefore accumulate into the same global keys across unrelated requests, so a breaker can trip on aggregated noise rather than on one degraded build. Neither view has a `try/finally`, so `BuildDegradationError` surfaces as a raw 500 to a staff member editing one topic.

Copilot flagged this same reset-scope gap on `skip_tracking.py:158` in the 8/28 review; I couldn't find follow-up on it, and it's still reachable in the current diff.

The rest holds up well — the nested-guard-escape fix (`BuildDegradationError` raised `from e`, `skip_tracking.py:363-365`) and `TestBreakers` genuinely exercise the masking case (a repeated identical error aborting mid-loop and escaping enclosing guards), not just the happy path.

**Minor:** the signature breaker treats identical error text as evidence of a code bug, but `AttributeError` text is derived from type + attribute, not the record — uniform data corruption (one field dropped across many distinct real documents) produces the same signature. Aborting is still right; the "indicates broken code" wording (`skip_tracking.py:181`) would just misdiagnose the cause.

yodem:
**[claude]** — correction to the line numbers above: the three `signal_and_reset_skip_counts()` call sites were cited from `master`, and this branch shifts them. On the PR head they are `reader/startup.py:64`, `sefaria/views.py:810`, and `sefaria/views.py:947`. The `reader/views.py:3628,3645` anchors are unchanged, and the finding itself is unaffected.

sefaria/helper/skip_tracking.py
```diff
@@ -49,6 +57,42 @@
 # Per (pathway, operation): total skips seen this build (including ones not stored verbatim).
 _skip_group_counts = defaultdict(int)
 
+# Per (pathway, operation, error_type, detail): how many times ONE error signature repeated.
+# Only guard-caught skips are counted here — see the breaker notes below.
+_skip_signature_counts = defaultdict(int)
```
yodem: **[claude]** These counters are module-global and live for the life of the process, but `signal_and_reset_skip_counts()` is called from only three places — `reader/startup.py:64`, `sefaria/views.py:810`, `sefaria/views.py:947`.

The same guarded rebuild is also entered from `add_new_topic_api` (`reader/views.py:3628`), `delete_topic` (`reader/views.py:3645`), and `rebuild_topic_toc` (`sefaria/helper/topic.py:1267`), none of which reset. Per the note at lines 76-84, one bad topic record produces a byte-identical signature on every rebuild — so ten unrelated staff edits accumulate to `SIGNATURE_BREAKER_THRESHOLD` and the trip surfaces as an unhandled `BuildDegradationError` in whichever admin request happens to cross it, rather than in a build.

Copilot raised the same reset-scope gap on 8/28; I couldn't find follow-up on it.

sefaria/views.py
```diff
@@ -937,8 +941,10 @@ def delete_orphaned_counts(request):
 
 @staff_member_required
 def rebuild_toc(request):
-    model.library.rebuild_toc()
-    signal_and_reset_skip_counts("reset_toc")
+    try:
+        model.library.rebuild_toc()
+    finally:
+        signal_and_reset_skip_counts("reset_toc")
```
yodem: **[claude]** This `try/finally` is what makes the monitored pathways safe — counters reset even when the rebuild raises. The topic-admin entry points (`reader/views.py:3628,3645`) call `library.get_topic_toc(rebuild=True)` with no equivalent, which is the other half of the note on `skip_tracking.py`.

sefaria/helper/skip_tracking.py
stevekaplan123: Confirmed — and this is worse than the overlapping-builds version I dismissed on 8/28, because it defeats the argument I closed that one with.

I argued there that a spurious signature trip can't happen, since "a message only repeats identically when it is record-independent, i.e. when the code is broken." That inference holds only within ONE pass over the records, where each record is visited once: 10 identical messages then means 10 *different* records produced the same message. Across repeated rebuilds it collapses — the same corrupt topic, walked ten times, gives the same message ten times because it is the same record, not because the message is record-independent.

So your scenario is a genuine false positive: one bad topic + ten routine staff edits → `BuildDegradationError` out of an admin request, reporting broken code when nothing is broken, and after `t.save()` has already run.

The gap was also wider than the three sites named — about ten in-process entry points reach guarded builds without resetting, including `init_shared_cache(rebuild=True)` from `base_props`, `refresh_all_states()`, and the eight `library.rebuild()` calls in `helper/schema.py`.

Fixed in two parts.

**1. The signature breaker now counts DISTINCT records, not occurrences,** so it measures its actual premise: "N *different* records gave the same message, so the message cannot depend on the record." One record re-skipped N times can never trip it, regardless of anyone's reset discipline; a rename still trips at 10 because every record is distinct. Records skipped without an identifier fall back to occurrence counting, and the volume backstop still counts every skip, so mass re-skipping of one record is still caught as degradation.

**2. `build_pathway(name)` resets the counters on ENTRY** and posts the summary on exit — reset-on-entry being the load-bearing half, since an entry point that forgets to summarize can then no longer poison the next build.

It lives on the build METHODS rather than their callers, because a caller can forget and a method can't: `rebuild`, `rebuild_toc`, `init_shared_cache`, the three autocompleter builders and `build_linker` wrap whole; `get_toc_tree`, `get_topic_toc` and `get_topic_toc_category_mapping` wrap only their building branch, since their cache-hit path is a request hot path. That covers `refresh_all_states`, the schema helpers and the maintenance scripts without touching any of them.

Blocks nest per-thread with the outermost owning the reset and the report, so the only callers still wrapping are those running a *sequence* of builds, where the wrap buys grouping rather than coverage: `init_library_cache` (→ "startup"), the three topic-admin views (→ "topic_admin"), and `rebuild_auto_completer`. The `reset_cache` and `rebuild_toc` views no longer wrap anything — the single method each one calls covers it.

One deliberate trade-off: because the reset is on entry, two genuinely overlapping builds in one worker means the second clears the first's counters rather than mixing into them. Same trade-off as the pre-existing shared-counter overlap, and strictly better than what it replaces, where an unmonitored rebuild's skips persisted into every later build.

11 new unit tests in `skip_tracking_test.py` (34 pass). One change to `audit_skip_bad_record_test.py` worth knowing about: `run_case()` now opens its own `build_pathway` around the trigger and reads the skip log inside it. The outermost block clears the log on exit, so reading it after a bare builder call sees nothing and every caught record classifies as a misleading NO_EFFECT — and a bare builder call is no longer how anything runs in production anyway. Audit is 39 passed / 1 skipped.

sefaria/views.py
stevekaplan123: Right — and the `try/finally` turned out to be the wrong home for the guarantee, since it only fires for whoever remembers to write it. It's gone from this view entirely: `build_pathway` now lives inside `library.rebuild_toc()` itself, so this call site and every other one (the maintenance scripts, `refresh_all_states`, the schema helpers) is covered without the caller doing anything.

Same for the topic-admin entry points you name, via `get_topic_toc` / `get_topic_toc_category_mapping` wrapping their own building branch — plus a wrap on the views themselves, not for coverage but so one staff click reports once instead of four times.

`build_pathway` also resets on entry rather than only summarizing on exit, which is what actually closes the hole. Details on the other thread.

sefaria/helper/skip_tracking.py
```diff
@@ -176,8 +204,38 @@ def signal_and_reset_skip_counts(pathway):
...
+
+    Usage:
+        with build_pathway("rebuild_toc"):
+            ...
+    """
+    depth = getattr(_pathway_depth, "value", 0)
+    _pathway_depth.value = depth + 1
+    try:
+        if depth == 0:
+            reset_skip_counts()
```
yodem, line 229: **[claude]** Reset-on-entry is the right mechanism, but it converts the old failure mode (skips leak forward into the next build) into a new one (skips are silently discarded) for any guarded builder that runs at depth 0 without a `build_pathway`. Three of those are left:

- `_build_index_maps()` — called at **import time** from `sefaria/model/__init__.py:54`. Guards at `text.py:5058` and `text.py:5068`, both `level="error"`.
- `build_term_mappings()` — reached from `Library.__init__` via `get_simple_term_mapping()` (`text.py:5044`). Guard at `text.py:5776`.
- `_build_topic_mapping()` — reached from `get_topic_mapping()` (`text.py:5837`). Guard at `text.py:5853`.

The first one bites at every boot. `init_library_cache()` does `from sefaria.model.text import library` (`reader/startup.py:26`), which executes `sefaria/model/__init__.py` and with it `library._build_index_maps()` — before `with build_pathway("startup")` is entered. Any corrupt `Index` skipped there is recorded, and then thrown away by the reset on entry. On master the trailing `signal_and_reset_skip_counts("startup")` picked it up, because it reported whatever was in the log regardless of when it landed.

Net effect: a corrupt `Index` record at boot — the `level="error"` case, arguably the highest-value alert in this whole feature — now posts nothing to #engineering-signal. The breakers still work (counters start empty at process start, so a rename during `_build_index_maps` still aborts); it's the reporting that goes dark.

The docstring right above already states the fix: *"Goes on the build METHOD, not its callers — a caller can forget."* These three are build methods that didn't get one. Wrapping them nests harmlessly under `rebuild()`, and makes the import-time call self-reporting rather than dropped.

sefaria/helper/skip_tracking.py
```diff
@@ -1,63 +1,69 @@
...
-#
-# Not reentrant, and deliberately never held across the Slack post — see the comment in
-# signal_and_reset_skip_counts().
+# Per (pathway, operation, error_type, detail): the set of DISTINCT records that produced ONE
+# error signature. Distinct records, not occurrences: the breaker's premise is "N different
+# records gave the same message, so the message does not depend on the record" — counting
+# occurrences would also trip on ONE bad record re-skipped by N successive rebuilds, which is
+# bad data, not broken code. Records without an identifier get a unique placeholder, so they
+# stay one-per-occurrence as before.
+_skip_signature_records = defaultdict(set)
```
yodem, line 50: **[claude]** Distinct-record counting is the right premise, but it promotes `record=` from a display label into the dedup key that gates the breaker — and at one existing site that key can't distinguish anything, so the signature breaker is now structurally unreachable there.

`sefaria/model/category.py:464`:

```python
record=getattr(n, "title", None) or n.__class__.__name__
```

`TocNode.__init__` deletes the raw title attribute after deserialization (`category.py:448`: `delattr(self, self.title_attrs[lang])`), and `TocCategory`'s `title_attrs["en"]` is `"category"`, not `"title"`. So `getattr(n, "title", None)` is `None` for every node and the record is always the class name — one of `TocCategory` / `TocTextIndex` / `TocCollectionNode`. The distinct set at that site tops out at 3 against a threshold of 10.

So a rename inside `serialize()` is now caught only by the volume backstop at 500, not by the signature breaker. That's the recursive site Copilot flagged on 8/21 as the reason `BuildDegradationError` exists at all, so it's an unlucky one to lose. `n.primary_title("en")` or `n.full_path` would restore it, and would also make the Slack summary name the node instead of its class.

Milder version of the same thing wherever the id falls back to a shared sentinel — `category.py:246,272`, `text.py:5058,5068,5776,5853,6119`, `autospell.py:588` all use `getattr(x, ..., "<unknown>")`. Fine while the id attribute is intact, but the case where they all collapse into one bucket is precisely the case the breaker exists for: the attribute being read for the id is the one the refactor broke. Consider having a missing id fall back to the unique placeholder (`next(_no_record_seq)`) rather than to a shared string — that's already the behavior you chose for `record=None`, and it's the conservative direction: worst case it counts occurrences, which is what this code did last week.

Neither case is covered by `TestSignatureBreakerCountsDistinctRecords` — a test where every skip passes the same non-`None` record and asserts no trip would pin it.

yodem (commented):
**[claude]** Second pass. The two things I raised on 8/31 are genuinely fixed, and fixed at the right level — distinct-record counting makes the signature breaker measure its own premise instead of a proxy for it, and putting `build_pathway` on the build methods rather than their callers is the version that can't be forgotten by the next entry point someone adds. The reset-on-entry argument is right: it's the half that actually closes the hole.

Two findings, both narrow, both a line or two to fix. Neither is a rethink.

1. Reset-on-entry drops skips from the three guarded builders that still run at depth 0 with no `build_pathway` — including the import-time `_build_index_maps()`, whose guards are the `level="error"` ones. On master those reached the startup summary; now they're wiped. Details inline.
2. `record=` is now the dedup key rather than a display label, and at `TocTree.serialize` it evaluates to a class name for every node, so the signature breaker is structurally unreachable there. Details inline.

One non-blocking note, no action needed. `get_toc_tree` / `get_topic_toc` / `get_topic_toc_category_mapping` wrap their building branch, which means a request thread that misses the shared cache now enters `build_pathway` at depth 0. After a cache flush under load, several request threads can do that at once: each resets the shared counters and each posts its own partial summary, and any of them will clobber a concurrent `/admin/reset/cache`. Same shared-state trade-off you already called out, just now reachable from ordinary traffic rather than only from admin actions. Given the alternative is threading a build id through every guard call site, leaving it is the right call — worth a line in the wiki page so the next person reading a split summary doesn't go hunting.

Happy to approve once 1 and 2 are addressed.

stevekaplan123:
Both findings confirmed and fixed. Thanks — the second one in particular was not visible from reading the code.

## 1. Skips dropped at depth 0 — `e25d660de6`

Confirmed, including the mechanism: `init_library_cache()` does `from sefaria.model.text import library`, which runs `sefaria/model/__init__.py` and its trailing `library._build_index_maps()` while `reader/startup.py` is still executing its own imports — so those `level="error"` skips are always recorded before `build_pathway("startup")` can open, and always cleared by its reset.

The list was wider than three. Two more guarded builders run at depth 0 with no pathway:

- `build_virtual_books()` — reached from `get_virtual_books()` in `sefaria/system/context_processors.py:117`, i.e. any request that finds the shared cache cold.
- `rebuild_linker_resolvers()` — guards in `linker/category_resolver.py:55,59`; the Celery task in `helper/linker/tasks.py:1060` calls it bare, while `build_linker` right next to it is wrapped.

All five now open their own `build_pathway`, per the rule in the docstring. The calls from `rebuild()` nest at depth 1 and stay part of the `rebuild` summary.

**Deliberately not wrapped: `all_index_records()`.** It also guards at depth 0, but it is an accessor over already-built maps that `reader/views.py` calls per request, so a pathway there would post a summary on every request touching a corrupt index. It is not orphaned in practice — it reports on both paths that matter: `TocTree` construction (`category.py:263`, under `get_toc_tree`) and `get_index_forest()` → `_build_ref_resolver()` → `build_linker` / `rebuild_linker_resolvers`. Its docstring rationale is "one corrupt index can't take down the boot," and the boot path is exactly where it still reports.

Covered by `TestGuardedBuildMethodsOpenTheirOwnPathway` in `skip_tracking_test.py`, which asserts each of the five wraps itself and records the `all_index_records` exclusion. It reads `text.py` via `ast` rather than instantiating `Library`, which would hit Mongo and break that file's DB-free contract.

## 2. `record=` as the dedup key — `da752ee79e`

Confirmed, and measured against the real TOC rather than reasoned about — 7,846 nodes walked:

| `record=` expression | distinct values | cost per serialize pass |
|---|---|---|
| `getattr(n, "title", None) or n.__class__.__name__` | **3** | 0.6 ms |
| `n.primary_title("en")` | 6,940 | 2.8 ms |
| `"/".join(n.full_path)` | 7,846 | 11.5 ms |

Zero of the 7,846 has a truthy `title`. Three distinct values against a threshold of 10, exactly as you said.

**One correction to the suggested fix.** `record=` is an *argument* to `skip_bad_record(...)`, so it is evaluated before the guard is entered — anything it raises escapes uncaught. `n.primary_title("en")` resolves to `self.title_group.primary_title(lang)` (`schema.py:194`), which raises `AttributeError` when `title_group` is broken — the exact corruption this guard's own comment says it catches. `full_path` is worse, since it does that per ancestor. Either one bare would crash the build the guard exists to protect.

So the fix is a `toc_node_id()` helper in `category.py` that computes `"/".join(node.full_path)` inside a `try/except`, returning `None` on failure. `full_path` over `primary_title` on the numbers above: fully unique instead of 6,940 (80 titles are shared — "Torah" by 28 nodes), for ~9 ms more on a path that runs for seconds, and the summary now names `Tanakh/Torah/Genesis` instead of `TocCategory`.

**Shared sentinels:** all eight `getattr(x, attr, "<unknown>")` are now `getattr(x, attr, None)`. No change to `skip_tracking` was needed — `None` already earns the unique placeholder at line 111 and already renders as `<unknown>` at line 239, so summaries are byte-identical and only the dedup behavior changes, degrading to occurrence counting.

**On the test point** — `TestSignatureBreakerCountsDistinctRecords::test_one_record_skipped_repeatedly_never_trips` already does 50 skips of the same non-`None` record and asserts no trip, so the `skip_tracking` mechanism was pinned. The gap was on the other side: nothing checked that the call sites hand it ids that distinguish records. That is what the new `Test_Toc_Node_Id` in `category_test.py` covers — reverting `toc_node_id` to the old expression makes it fail with `assert 3 > 10`.

yodem (approved):

PR #3649, opened 2026-08-21
Feature/sc 46788/modify abstractmongoset to have a method

stevekaplan123:
## Description
This builds on previous work to handle bad data on server start and reset of cache.  The startup guards wrap the body of each loop in startup/reset of cache, but AbstractMongoSet._read_records() builds every record in one pass the moment the set is touched:

for i in IndexSet():                    # ← all 6,584 Index objects built HERE
    with skip_bad_record(...):          # ← guard only starts HERE
So one bad document aborts the whole set and the loop body never runs even once.

## Code Changes
The new function, AbstractMongoSet.with_skip_guard(), stores a guard on the set and returns self; The existing `_read_records()` is now modified so that it guards each record as it's constructed. Applied at the five guarded build sites that iterate a set.  The solution is to do this:

```
for i in IndexSet().with_skip_guard(skip_bad_record, "reset_cache,startup",
                                    "_build_index_maps index record", level="error"):
    with skip_bad_record(...):   # still needed — guards USING the record, not building it
```
 Note: `_read_records()` is called implicitly when iterated and is called in 9 other ways (array(), len(), contents(), …). 
I didn't use an __init__ flag, because pathway/operation are required, not a bare flag: they group the Slack summary and key the skip-tracking breakers.

Note 2: Two new hooks: _instantiate_record() and _post_read_records() are both called from `_read_records` and they make record-building overridable. LexiconEntrySet and TopicSet override these methods so that they can modify the way records are loaded and do post-processing.


sefaria/model/abstract.py
```diff
@@ -337,12 +351,79 @@ def __getitem__(self, item):
...
+        record — `_read_records()` below materializes every record the moment the set is first
+        touched, including by the `for rec in SomeSet():` clause itself, so a single malformed
+        document aborts the set before the loop body runs even once, and no `with` block inside
+        that body can catch it at any exception breadth. The inner guard covers USING the record.
+
+        Why the guard lives on the set rather than on one method: `_read_records()` is reached
+        from __iter__, __getitem__, __len__, array(), count(), contents(), remove(), update(),
+        save() and delete(). A method that guards only its own call protects the `for x in set`
+        shape and nothing else, so rewriting a loop as `set.array()` would silently drop the
+        protection.
```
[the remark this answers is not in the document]
stevekaplan123: Good catch.  Addressing in the comment

stevekaplan123:
> ### 🟡 Changes recommended
> Topic subclass casting remains outside the per-record guard and can still abort startup on malformed data.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> Pull request overview
> Adds per-record construction guards to Mongo sets so malformed documents do not abort cache and startup builds.
> 
> **Changes:**
> 
> * Adds guarded record instantiation and extensible materialization hooks.
> * Applies guards to index, term, topic, category, and lexicon builds.
> * Adds regression and bad-record audit coverage.
> 
> File summaries
> File	Description
> `sefaria/model/abstract.py`	Implements guarded set materialization and hooks.
> `sefaria/model/topic.py`	Moves topic subclass casting into a post-read hook.
> `sefaria/model/lexicon.py`	Uses hooks for polymorphic entries and sorting.
> `sefaria/model/text.py`	Guards cache-building sets.
> `sefaria/model/category.py`	Guards TOC record construction.
> `sefaria/model/autospell.py`	Guards lexicon trie construction.
> `sefaria/model/tests/abstract_test.py`	Tests guard and hook behavior.
> `sefaria/helper/tests/audit_skip_bad_record_test.py`	Updates corruption audit expectations.
> Review details
> ### Suppressed comments (1)
> **sefaria/model/topic.py:906**
> 
> * This post-processing still runs after the per-record guard in `_read_records`. A malformed topic whose `subclass` is not in `subclass_map` is instantiated successfully, then raises `KeyError` here and aborts `_build_topic_mapping`, so the new startup protection remains bypassed for that bad-data shape. Make subclass casting part of `_instantiate_record` so the guard can skip that individual document.
> 
> ```
>     def _post_read_records(self):
>         for rec in self.records:
>             if getattr(rec, 'subclass', False):
>                 Subclass = globals()[self.recordClass.subclass_map[rec.subclass]]
>                 rec.__class__ = Subclass  # cast to relevant subclass
> ```
> 
> * **Files reviewed:** 8/8 changed files
> * **Comments generated:** 1
> * **Review effort level:** Balanced
> 
> 💡 [Add a `code-review` agent skill](/Sefaria/Sefaria-Project/new/master?filename=.github/skills/code-review/SKILL.md) or configure MCP servers for context-aware, tailored reviews. [Learn more in the docs.](https://docs.github.com/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review?tool=webui#mcp-servers-and-agent-skills)

Good catch.  Addressing the _post_read_records issue in the next commit.

sefaria/model/category.py
```diff
@@ -246,8 +246,13 @@ def __init__(self, lib=None, mobile=False):
...
-        # Build Category object tree from stored Category objects
-        for c in CategorySet(sort=[("depth", 1)]):
+        # Build Category object tree from stored Category objects. Two guards, covering two
+        # different failures: with_skip_guard() covers CONSTRUCTING each Category — a bad
+        # `sharedTitle` raises out of _process_terms during _set_derived_attributes, while the
+        # set is materializing, so it would otherwise abort the whole TOC build before
+        # _add_category ran even once — and _add_category's own guard covers USING it.
+        for c in CategorySet(sort=[("depth", 1)]).with_skip_guard(
+                skip_bad_record, "reset_toc,startup", "TocTree category record"):
             self._add_category(c)
```
yodem, line 256: [claude] **One dropped parent category can now abort startup via the signature breaker.**

`CategorySet` is sorted by `depth`, so parents build before children. When the new construction guard drops a category, every direct sub-category then fails in `_add_category` (category.py:370-374) at `self._path_hash[tuple(cat.path[:-1])]`. The `KeyError`'s `str(e)` is the *parent's* path tuple — identical for every sibling — while `record='/'.join(cat.path)` is distinct per sibling.

`_note_skip` (skip_tracking.py:102-111) keys `_skip_signature_records` on `(pathway, operation, error_type, detail)` and dedupes by `repr(record)`. Ten or more direct sub-categories therefore reach `SIGNATURE_BREAKER_THRESHOLD = 10`, `skip_bad_record` raises `BuildDegradationError`, and since that is deliberately outside `BAD_RECORD_EXCEPTIONS` it escapes `TocTree.__init__` and aborts the build. Mid-tier categories under `Tanakh` or `Talmud/Bavli` routinely have more than ten children.

The breaker's premise — "one message across ten records means broken code, not bad data" — is inverted here: a single bad parent manufactures ten identical messages. Note this path is *new*. Before this PR a category that failed construction aborted the set loudly, so the cascade could not occur; the PR converts a loud abort into a silent drop that then trips the breaker.

Below the threshold the failure mode is still poor: a whole subtree vanishes from the TOC while the summary reports one skipped category.

sefaria/model/category.py
```diff
@@ -246,8 +246,13 @@ def __init__(self, lib=None, mobile=False):
                     "enComplete": bool(vs.get("flags", {}).get("enComplete", False)),
                 }
 
-        # Build Category object tree from stored Category objects
-        for c in CategorySet(sort=[("depth", 1)]):
+        # Build Category object tree from stored Category objects. Two guards, covering two
+        # different failures: with_skip_guard() covers CONSTRUCTING each Category — a bad
```
yodem, line 250: [claude] **The comment's motivating example cannot occur under prod settings.**

`_process_terms` (schema.py:213) carries `@conditional_graceful_exception()`, which (decorators.py:117-122) swallows the exception whenever `FAIL_GRACEFULLY` is true. Prod and preprod both set it true (`envs/prod/helmrelease.yaml:269`, `envs/preprod/helmrelease.yaml:250`, chart default `helm-chart/sefaria/values.yaml:576`); only pytest (`local_settings_example.py:360`) and CI integration (`build/ci/integration-values.yaml:66`) force it false.

So on a prod server a category with `sharedTitle: "ZZNoSuchTerm"` has `_process_terms` log and return `None`, the `Category` constructs fine with a default empty `title_group`, and this guard never fires — the record fails later at the `TocTree.serialize node` guard instead.

The guard is not wrong to exist, but the rationale written into the comment is a test-configuration artifact. Under prod settings the remaining reachable construction triggers on `Category` are the narrow `change_key_name` paths (category.py:75-84), since `TitleGroup.load` (schema.py:70-72) does no validation. Worth rewording so the next reader does not rely on an example that does not happen where it matters.

sefaria/model/text.py
```diff
@@ -5391,7 +5396,8 @@ def _build_topic_mapping(self):
             from .topic import Topic, TopicSet
             # One topic with a missing/corrupt title_group must not abort startup.
             self._topic_mapping = {}
-            for t in TopicSet():
+            for t in TopicSet().with_skip_guard(skip_bad_record, "reset_cache,startup",
```
yodem, line 5399: [claude] **An unclosed instance of the exact gap this PR exists to fix: `text.py:4808`.**

`get_topic_toc_json_recursive` does:

```python
ts = TopicSet({"isTopLevelDisplay": True})     # line 4808 — materializes unguarded
for t in ts:
    with skip_bad_record("reset_toc,startup", "get_topic_toc_json_recursive top-level topic", ...):
```

`TopicSet` is precisely the class whose `_read_records` raises `KeyError` during construction when a topic's `subclass` is absent from `subclass_map` — proven by this PR's own `test_an_unknown_subclass_aborts_an_unguarded_topic_set`. `_build_topic_mapping` (this line) got `with_skip_guard`; that site, on the same `reset_toc,startup` pathway, did not.

One topic with `isTopLevelDisplay: True` and a bad `subclass` still aborts the entire topic ToC before the inner guard runs once. The comment at 4810 already says this site needs per-record protection; the inner guard just cannot deliver it.

Suggested: `.with_skip_guard(skip_bad_record, "reset_toc,startup", "get_topic_toc_json_recursive top-level topic")`.

sefaria/helper/tests/audit_skip_bad_record_test.py
```diff
@@ -501,19 +505,27 @@ def _topic_doc(slug, **overrides):
...
-         trigger=_toc_tree_serialize, expect=PROPAGATED, error_type="IndexError", outside_guard=True,
-         note="IndexError IS in BAD_RECORD_EXCEPTIONS, but it is raised during TocTree "
-              "__init__/_sort, outside every with-block -- see `escaped at` column"),
+         trigger=_toc_tree_serialize, expect=WRONG_SITE,
+         note="a missing term raises out of _process_terms during _set_derived_attributes, "
+              "i.e. while CategorySet() instantiates the record, so the category is dropped "
+              "before any tree exists to serialize. TocTree's construction guard "
+              "(category.py:254) catches it -- a skip is recorded, at a different site than "
+              "this one. Was PROPAGATED before with_skip_guard() existed. "
+              "COVERAGE GAP: this is the only case at the serialize site, so that guard is "
```
yodem, line 514: [claude] **This coverage was removed on the strength of pytest-only behavior, and the claim in the note is not true on prod.**

The note says the `TocTree.serialize node` guard is now unexercised because "No such corruption is known yet: Category validates eagerly, at load time." That eager validation only happens because pytest sets `FAIL_GRACEFULLY = False`. With `FAIL_GRACEFULLY=True` (prod/preprod) `_process_terms` swallows, the `Category` builds, and this exact corruption *does* still reach the serialize site.

Net effect: a broken `sharedTitle` in prod now ships with zero audit coverage at the site that will actually catch it, and the harness records a site attribution that does not match the deployed configuration.

`WRONG_SITE` is defined at line 17 as "a skip was recorded, but by some other guard than the one under test" — so flipping to it genuinely does leave the serialize guard with no cases. Either keep a case parameterized on `FAIL_GRACEFULLY`, or mark the site explicitly uncovered so a later audit does not read green as tested.

sefaria/model/category.py
```diff
@@ -260,7 +265,13 @@ def __init__(self, lib=None, mobile=False):
         # Place Indexes. Wrap each index so one malformed record (empty categories,
         # bad base_text_titles, broken schema, etc.) logs and is skipped rather than
         # aborting the whole TOC build and preventing server startup.
-        indx_set = self._library.all_index_records() if self._library else text.IndexSet()
+        # The library branch is already covered: all_index_records() returns a plain list built
+        # from _index_map, which _build_index_maps guards as it materializes. The IndexSet()
+        # fallback materializes here, so it needs its own construction guard.
+        indx_set = (self._library.all_index_records() if self._library
+                    else text.IndexSet().with_skip_guard(
```
yodem, line 272: [claude] **`CollectionSet` at category.py:300 has the same shape and was left unguarded.**

Within this same `TocTree.__init__`, `CategorySet` (254) and this `IndexSet` fallback both got `with_skip_guard`, but 30 lines further down:

```python
collections = collection.CollectionSet({"toc": {"$exists": True}, "listed": True, "slug": {"$exists": True}})
for c in collections:
    with skip_bad_record("reset_toc,startup", "TocTree collection", record=..., level="error"):
```

still materializes unguarded and guards only the body, at `level="error"`. Lower blast radius than the others — `Collection` defines no `_set_derived_attributes`, so construction raises less readily — but it is the same structural gap, and leaving it will read as deliberate to the next person deciding whether a site needs the guard.

Either add it for consistency or leave a one-line note saying why this set does not need it.

sefaria/model/tests/abstract_test.py
```diff
@@ -199,3 +201,209 @@ def test_copy(self):
...
 
 
+class _FakeGuard(object):
+    """Stands in for skip_bad_record: records what it swallowed."""
+
+    def __init__(self, exceptions=Exception):
+        self.exceptions = exceptions
+        self.skipped = []
+
+    def __call__(self, pathway, operation, record=None, level="warning"):
```
yodem, line 211: [claude] **`_FakeGuard` records neither `pathway` nor `level`, so nothing asserts they are forwarded.**

`__call__` accepts both and `skipped` stores only `(operation, record, type(e).__name__)`. Two production call sites depend on `level="error"` — `text.py:4595` and `category.py:272` — and that level is what makes the end-of-build Slack summary error-severity rather than warning (`signal_and_reset_skip_counts`, via `_skip_saw_error`). If `with_skip_guard`'s forwarding of `level` or `pathway` broke tomorrow, every test in this file still passes.

Cheap fix: append `pathway` and `level` to the `skipped` tuple and assert them in one existing test.

Related, same section: `test_record_is_identified_from_the_raw_document` (line 300) covers `title`, `slug`, `name`, `_id` and the fallback, but not `lastPath`, `headword` or `path` — which are exactly the fields added for Category and Lexicon documents — and does not pin field *precedence*. Both gaps are already written up in `wiki/meta/decision-2026-09-09-build-records-guard-boundary.md` under "Known test gaps"; they are still open at `63f9d4f4e`.

sefaria/model/abstract.py
```diff
@@ -337,12 +351,80 @@ def __getitem__(self, item):
...
+        for raw in self.raw_records:
+            if self._skip_guard is None:
+                self.records.append(instantiate(raw))
+            else:
+                guard, pathway, operation, level = self._skip_guard
+                record = None
+                with guard(pathway, operation,
+                           record=_raw_record_identifier(raw), level=level):
+                    record = instantiate(raw)
+                if record is not None:
```
yodem, line 421: [claude] **`record is not None` conflates "the guard fired" with "`instantiate` legitimately returned `None`".**

A future `instantiate` that returns `None` for a valid document has that record dropped with no `_note_skip` call, no breaker increment, no Slack line and no log entry — the "silently having no protection is the worst failure mode" outcome the docstring just above warns about.

Latent, not live: `recordClass(attrs=…)`, `LexiconEntrySubClassMapping.instance_from_record_factory` and `TopicSet`'s closure all return an object unconditionally today. But the contract is enforced only by convention. A local `_MISSING = object()` sentinel, or a boolean flag set on the last line inside the `with` block, makes it unambiguous.

yodem (commented):
## [claude] Review of `with_skip_guard()` / `_build_records()`

Reviewed at `63f9d4f4e` against `master`, with the branch checked out locally. Cross-checked against `wiki/meta/decision-2026-09-09-build-records-guard-boundary.md`, which was written against this branch and records the design rationale and the alternatives rejected.

The core mechanism is sound and the diagnosis is right: `_read_records()` materializes the whole set before the `for` clause yields, so a guard on the loop body can never protect construction at any exception breadth. Routing every subclass through one `_build_records()` is the right shape, and moving `TopicSet`'s subclass cast *inside* `instantiate` is the load-bearing detail — a second pass would let that `KeyError` escape the guard. Both overriding subclasses (`topic.py:902`, `lexicon.py:553`) were migrated correctly; no caller was left with a hand-rolled unguarded loop.

Seven inline comments. The two that would change my merge vote:

- **category.py** — a single dropped parent category now cascades into ten-plus identical-signature `_add_category` skips, trips the signature breaker, and aborts startup. This path is created by the PR: the loud abort it replaces could not cascade.
- **text.py:4808** — `TopicSet({"isTopLevelDisplay": True})` is still unguarded at construction, which is the precise failure this PR's own regression test demonstrates.

Two further comments concern `FAIL_GRACEFULLY`: the `CategorySet` guard's stated motivating example is swallowed by `@conditional_graceful_exception()` on prod, and an audit case was retired on the basis of that same pytest-only behavior.

**Not verified:** I could not execute the suite locally (`ModuleNotFoundError: No module named 'allauth'`; no Mongo). Every finding above is from reading source and settings, not from a failing test.

**Housekeeping, no anchor line:** the PR description documents `_instantiate_record()` and `_post_read_records()` as the mechanism. The branch has neither — per the wiki page, that narrow-hook variant existed briefly mid-branch and was removed. The description also lists `count()` among the paths that materialize the set; `count()` returns `self.max` if set and otherwise falls through to `count_documents()`, never reaching `_read_records()`. The docstrings in the code get both of these right; only the description is stale.

sefaria/model/category.py
stevekaplan123: Agreed on the facts: with `FAIL_GRACEFULLY=True` (prod/preprod) `_process_terms` swallows, the Category builds with an empty `title_group`, and this guard's stated example never fires there. The guard stays since it is still correct, but the comment's rationale is a test-settings artifact.

Scoping decision: the `FAIL_GRACEFULLY` interaction (this comment and the audit case below) will be handled in a separate branch rather than in this PR, which stays focused on the construction-guard mechanism.

One pointer for that follow-up, unverified: after the swallow, `get_primary_title()` returns `None`, and `TocTree._sort()` sorts categories without an `order` attr by title (`_explicit_order_and_title`). Comparing `None` against a str there would raise `TypeError` outside any guard, so on prod this record may abort the TOC build in `_sort` rather than reaching the serialize guard.

sefaria/helper/tests/audit_skip_bad_record_test.py
stevekaplan123: Deferring this to a separate branch, together with the `FAIL_GRACEFULLY` point on the `CategorySet` guard above. That branch will revisit this case (parameterizing on `FAIL_GRACEFULLY`, or marking the serialize site explicitly uncovered) rather than leaving it as `WRONG_SITE`. Not addressed in this PR.

sefaria/model/category.py
stevekaplan123: Fixed in cb80a868f0. Confirmed against local data first: twelve parents have ten-plus direct children (`Halakhah/Mishneh Torah/Commentary` alone has 54), so the cascade was reachable.

`_add_category` now checks for the parent explicitly and reports a missing one through `log_skip` rather than letting the `KeyError` raise inside `skip_bad_record`. `log_skip` counts toward the volume backstop only, so N siblings under one dropped parent are N soft skips, not N identical signatures. The subtree still drops, but each child is logged individually and the summary shows the real count.

Regression test: `Test_Toc_Add_Category_Missing_Parent` in `category_test.py` builds `SIGNATURE_BREAKER_THRESHOLD + 2` children under a missing parent and asserts no `BuildDegradationError`. The S5 audit case for a missing parent is annotated accordingly (it now reports as a log_skip, not a caught KeyError).

sefaria/model/text.py
stevekaplan123: Fixed in cb80a868f0. `get_topic_toc_json_recursive` now builds the top-level set with `.with_skip_guard(skip_bad_record, "reset_toc,startup", "get_topic_toc_json_recursive top-level topic")`, the same pathway and operation the loop body already used, so a construction skip and a use skip land in the same summary group. The S9 audit cases still pass.

sefaria/model/category.py
stevekaplan123: Added in cb80a868f0. `CollectionSet` in `TocTree.__init__` now carries `.with_skip_guard(..., "TocTree collection", level="error")`, matching the category and index loops above it, with a comment noting the two-guard shape. Agreed it is lower risk than the others, but consistency wins here so the next reader does not infer that the omission was deliberate. The S4 audit cases still pass.

sefaria/model/tests/abstract_test.py
stevekaplan123: Fixed in cb80a868f0. `_FakeGuard` now records every guard entry as `(pathway, operation, record, level)` in a separate `calls` list, so the existing `skipped` assertions are untouched. Two new tests: `test_pathway_and_level_are_forwarded_to_the_guard` asserts both values arrive unchanged with `level="error"`, and `test_level_defaults_to_warning` pins the default.

The identifier-field gaps you mention (`lastPath`, `headword`, `path`, and precedence) are not covered in this commit.

PR #3662, opened 2026-08-26
perf(deps): defer sefaria.helper.search import to first use

yodem:
## Why

`sefaria/model/dependencies.py` is imported while `sefaria.model.__init__` is still running. Its module-level import:

```python
from sefaria.helper.search import (process_index_save_in_book_search, ...)
```

pulls `sefaria/helper/search.py` — and transitively `sefaria/search.py` and the Elasticsearch client — into **every worker boot**, whether or not anything is ever indexed.

Datadog's Continuous Profiler shows worker startup is already dominated by module-import work (`marshal.loads`, `compile`, `__build_class__`, `stat`) under `Arbiter.spawn_worker → post_fork`, and CPU roughly doubles (108 → 218 ms/min) during periods when pods are cycling. Every avoidable import at boot is paid again on every restart.

Note the inconsistency this fixes: **the handlers already defer their own imports** — e.g. `process_topic_save_in_topic_search` does `from sefaria.search import index_topic_doc` *inside* the function body. Only the top-level import in `dependencies.py` was eager.

## What changed

Each of the eight ES cascade handlers is wrapped in a thunk that imports on first call. One file, no behaviour change — only *when* the import happens.

**Subscription order is unchanged**, which matters: the file carries explicit ordering comments (`must stay last in this block`, `must stay subscribed after process_index_title_change_in_versions`, `must stay subscribed after text.rebuild_library_after_category_change`). All nine `subscribe()` calls keep their exact positions.

## How to test on preprod

**The measurement is independent of any incident hypothesis** — it either reduces boot cost or it doesn't:

```bash
# inside a preprod web pod, or any container built from this branch
time python -c "import sefaria.model"
```

Run it on `preprod` before merge and on this branch after. Expect the ES/search import chain to drop out of the boot path.

To see it directly in the profiler: **APM → Profiles → Explorer**, `service:web`, metric **CPU**, and **turn off "My Code"** (library frames are hidden by default and that is where the import work lives). Compare a window containing worker starts before and after.

Functional check — the subscriptions must still fire:

1. Save a Topic on preprod and confirm the corresponding `PUT /topic-a/_doc/<slug>` still reaches Elasticsearch (visible in APM as an `elasticsearch` span on `service:web`).
2. Rename an Index title and confirm the search-index sync still runs.
3. Confirm no import errors at boot: `kubectl -n preprod logs <pod> -c web | head -50`.

**Rollback:** revert the commit. Single file, no schema or config impact.

## Scope

This is a **boot-cost** improvement, not a fix for the ~2× per-request regression seen since `v6.111.0-prod.2`. It matters most while pods are cycling frequently — which is itself addressed by #3661 (liveness probe). Worth landing on its own merits regardless of what caused the 2×.

Related: #3661 (liveness `failureThreshold`), and the investigation artifact for full context.

https://claude.ai/code/session_01DbbDpzvvxiRi2U9J9wssxi

### DIFF
--- a/sefaria/model/dependencies.py
+++ b/sefaria/model/dependencies.py
@@ -6,18 +6,57 @@ from . import abstract, link, note, history, schema, text, layer, version_state,
 
 from .abstract import subscribe, cascade, cascade_to_list, cascade_delete, cascade_delete_to_list
 # ES cascade handlers that keep the search indices in sync with model changes.
-# sefaria.helper.search deliberately avoids importing sefaria.model at module level,
-# so importing it here (while sefaria.model's __init__ is still running) is safe.
-from sefaria.helper.search import (
-    process_index_title_change_in_search,
-    process_index_title_change_in_book_search,
-    process_index_save_in_book_search,
-    process_index_delete_in_book_search,
-    process_topic_save_in_topic_search,
-    process_topic_delete_in_topic_search,
-    process_category_path_change_in_book_search,
-    process_category_change_in_category_search,
-)
+#
+# These are wrapped in thunks rather than imported at module level. `dependencies` is
+# imported while sefaria.model's __init__ is still running, so a module-level import of
+# sefaria.helper.search pulls that module -- and transitively sefaria.search and the
+# Elasticsearch client -- into *every* worker boot, whether or not anything is ever
+# indexed. The handlers already defer their own imports (`from sefaria.search import
+# index_topic_doc` inside the function body), so deferring this one makes the module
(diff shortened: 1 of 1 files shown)


yitzhakc (dismissed):

yodem:
Superseded by https://github.com/Sefaria/Sefaria-Project/pull/3663 — same commit, re-cut onto `preprod` under a `hotfix/*` branch so it passes the promotion gate (`preprod` only accepts `master`, `prod`, or `hotfix/*`).

PR #3668, opened 2026-08-27
hotfix: revert #3661/#3663, raise Varnish memory headroom, add backend health probe

yodem:
## Root cause

Per `wiki/incidents/prod-oomkill-cascade-2026-08-26.md` (§ "Update 2026-08-27"): a spoofed-UA bot campaign hammering `api/related` and `api/strapi/graphql-cache` at 10–25× baseline, plus Varnish's own pod restarting 9 times during the incident (confirmed via `kubernetes.containers.restarts` + raw kubelet events), flushing its cache each time. Neither #3661 nor #3663 touches either mechanism.

## Reverts

- **#3661** (liveness `failureThreshold: 7`): only changed which failure mode fires (liveness-kill → slower cgroup-OOMKill). Doesn't fix anything by itself — reverted to reduce moving parts now that real fixes are in flight.
- **#3663** (defer search import): a real optimization, but never implicated in any finding. Reverted for the same reason.

Both via `git revert -m 1` against `origin/preprod` (neither commit is on `master` — they shipped as direct hotfixes). Clean, no conflicts.

## Helm changes

- **Varnish backend health probe** — was missing entirely. Varnish now stops routing to a backend already failing `/healthz` instead of waiting out `first_byte_timeout` on every connection.
- **Fixed `first_byte_timeout`**: hardcoded to `900s` directly in the VCL, ignoring the Helm value entirely. Now templated. ⚠️ **Risk**: once live, this drops the effective prod timeout from 900s to prod's real tuning value (60s) — watch closely through preprod before promoting.
- **Varnish memory bump** (25Mi/300Mi/250m → 700Mi/1024Mi/512m) — **chart default only**. Correction: prod/preprod already override this to 17Gi/18Gi/14.4GB malloc (unchanged since March) — ~3.5Gi of real headroom, not ~50Mi. **This has no effect on prod/preprod**; it only helps staging/dev/cauldrons, which inherit the thin default. The Varnish restart cause is still unresolved for prod.

## Follow-up — not in this PR

A Cloudflare WAF rule on the bot campaign's actual fingerprint (paths + near-zero static-asset ratio) is the real fix on the traffic side — existing rules are IP/country-based and aren't catching 60K+ distinct IPs. Tracked separately; see `wiki/concepts/Cloudflare Edge and WAF.md`.

Full evidence: `wiki/incidents/prod-oomkill-cascade-2026-08-26.md`.

### DIFF
--- a/helm-chart/sefaria/templates/configmap/varnish-config.yaml
+++ b/helm-chart/sefaria/templates/configmap/varnish-config.yaml
@@ -13,10 +13,19 @@ data:
   sefaria.vcl: |-
     vcl 4.0;
 
+    probe web_probe {
+        .url = "/healthz";
+        .timeout = 30s;
+        .interval = 10s;
+        .window = 5;
+        .threshold = 2;
+    }
+
     backend default {
         .host = "web-{{ $.Values.deployEnv }}-{{ $v }}";
         .port = "80";
-        .first_byte_timeout = 900s;
+        .first_byte_timeout = {{ $.Values.varnish.tuning.first_byte_timeout }}s;
+        .probe = web_probe;
     }
 
     sub vcl_recv {
--- a/helm-chart/sefaria/templates/rollout/web.yaml
+++ b/helm-chart/sefaria/templates/rollout/web.yaml
@@ -198,17 +198,6 @@ spec:
           # initialDelaySeconds: 120
           periodSeconds: 60
           timeoutSeconds: 60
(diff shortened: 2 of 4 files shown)


BrendanGalloway (approved):

yodem:
Superseded by #3680 — same content, retargeted to `master` instead of shipping as a direct-to-preprod hotfix, so it goes through the normal master → preprod → prod pipeline.

PR #3669, opened 2026-08-27
Deprecate binary en/he TextChunk for real-language TextChunk

YishaiGlasner:
*Naming note: the new real-language class was originally called `TextRange` and renamed to `TextChunk` partway through this work, with the old binary class becoming `LegacyTextChunk`. This description uses the final names throughout — `TextChunk` always means the new, real-language class; `LegacyTextChunk` always means the old binary en/he one.*

`LegacyTextChunk` was built on a binary en/he assumption: a version's real language (Yiddish, French, Judeo-Arabic, etc.) got forced into an `"en"`/`"he"` bucket by direction, so a German translation stored as `language="en"` would be returned to callers as if it were literally English. This PR does two things.

## 1. Gives the new `TextChunk` save support and wires it end-to-end, including the editor

`TextChunk.save()` mirrors `LegacyTextChunk.save()`'s architecture directly — it inherits `AbstractTextRecord` and reuses `_validate`, `_sanitize`, `_trim_ending_whitespace`, `_check_available_text_pre_save`, `_update_link_language_availability` unchanged. `sefaria/tracker.py`'s `modify_text` — the shared entry point for every text save — now constructs the new `TextChunk` (`actual_lang`/`direction` threaded through from its own new `direction` kwarg) instead of the old one; this is where the save path actually moved. What's different, and why:

- **New constructor params, `direction` and `actual_lang`**, alongside `lang`/`vtitle`. `lang` (`languageFamilyName`) can now be *derived* from `actual_lang` via the existing `LANGUAGE_CODES` map, so a caller creating a brand-new version only needs to supply the real ISO code an editor picked — not compute both independently.
- **`direction` is a new lookup/save dimension**, and also the source for the legacy `language` field (`"he" if direction == "rtl" else "en"`) when a brand-new `Version` is created, since that field still has to satisfy the old binary schema underneath.
- **What "saveable" means changed.** `LegacyTextChunk` was saveable whenever `lang` and `vtitle` were both given, gated further by `fallback_on_default_version`/`exclude_copyrighted` to guard its implicit/fuzzy vtitle-guessing mode. The new `TextChunk` never has that fuzzy mode — it's saveable whenever `vtitle` is given plus *either* `direction` or `lang` (not both required), with no fallback/guessing params. (`exclude_copyrighted` was also confirmed dead on `LegacyTextChunk` itself — never passed `True` anywhere — and dropped from `Ref.text()`'s public signature entirely; grepped both this branch and master for any caller passing it, none found.)
- **`Version.pkeys` changed from `["title", "versionTitle"]` to `["title", "direction", "versionTitle"]`.** This isn't a hypothetical concern: the live data already has real collisions on the old key. We found 60 `(title, versionTitle)` pairs that currently exist as *both* an `ltr` and an `rtl` version — e.g. "Rashi on Psalms" / "Wikisource Mikraot Gedolot" exists in both directions. Whether that data pattern itself is desirable is a separate question this PR doesn't touch — we didn't create or change it, we only changed `pkeys` to correctly reflect the uniqueness boundary the data already has, instead of one that was already silently violated.
- **`TextChunk._validate()` is narrower than `LegacyTextChunk._validate()`** — it only checks that the posted text's depth matches the ref, dropping the old spanning-ref/range-length validation branches entirely. This is safe, not an oversight: `_saveable` excludes `is_range()` refs outright, so a spanning or range ref can never reach `save()` on the new class in the first place — there's nothing left for those branches to validate.
- **The `[xx]`-versionTitle-suffix convention is kept, not retired** — `db.history` and some legacy version-grouping code still key on the coarse `(language, versionTitle)` pair, so a German version forced into the `language="en"` bucket needs the suffix to avoid colliding with an actually-English version sharing that title. It's now generated automatically on save instead of something an editor had to type manually. Because the save response doesn't otherwise expose the possibly-auto-suffixed title, `TextChunk.save()` sets `self.vtitle` from the saved `Version`'s real title before returning, and the save API echoes it back in its JSON response — without this, a caller holding the pre-suffix title (e.g. a second save in the same session, or a script doing sequential per-segment saves) could create a colliding duplicate version.
- **Editor (`editor.js`/`edit_text.html`):** the hardcoded `<select>` with two options (en/he) is replaced with an ISO-code picker driven by `Sefaria.ISOMap` (same source used elsewhere in the client). There's no separate direction control: direction is derived from `Sefaria.ISOMap[lang].defaultDirection` for a brand-new version, and left completely alone when editing an existing one, so a version whose direction was deliberately saved as an exception to its language's default (a handful of real transliterated Arabic/Persian versions) is never silently overwritten. Adding/editing a "Sefaria Community Translation" hides the language picker and forces English, since that identity is fixed by definition. The `/edit/<ref>/<lang>/<version>` URL and its client-side builder (`sefaria/urls_library.py`'s route regex, `ConnectionsPanel.jsx`'s edit-link construction) both moved from a 2-letter `\w\w` (en/he only) segment to a full `languageFamilyName` string, to carry a real language through the URL instead of a binary code.

## 2. Renames the classes, moves the legacy ones out, and migrates every live caller to the new `TextChunk`

This starts with the rename itself — `TextRange` → `TextChunk`, old `TextChunk` → `LegacyTextChunk` — and physically relocating `LegacyTextChunk`, `VirtualTextChunk` (its delegate for virtual/dictionary nodes), and `TextFamily` into a new `sefaria/model/legacy_text.py`, no longer exported from `sefaria/model/__init__.py`.

**`TextFamily` is worth calling out specifically** — it was the bridge/assembly class that built the full legacy response shape (text plus commentary, links, and version metadata bundled together) for the old v1 texts API and everything that rendered off of it. It was a heavily-used object across the old GET path, not a minor helper. It's kept around for the legacy v1 `texts_api` GET (external third-party consumers) — see the class docstring on `LegacyTextChunk` for the full breakdown of what else still technically routes to it and why. `reader/views.py`'s `social_image_api` is fully migrated *off* `TextFamily` entirely (`ref.padded_ref().text(direction=...)`) — it no longer needs it at all. (`sefaria/image_generator.py` is unrelated to this migration — its only change in this PR is folding an existing ternary into the shared `direction`/`lang` helper described below; it never used `TextChunk`/`TextFamily`.)

With the rename and relocation in place, every live caller of the old class is migrated. Three categories:

**Migrated to `direction=` — deliberately preserving the old bucket-based behavior**, for call sites whose output is implicitly a closed "he"/"en" pair or that are legacy-keyed and shouldn't change behavior: `client/wrapper.py`'s `get_links()`, `linker.py`'s `_get_ref_text_by_lang_for_linker`, `marked_up_text_chunk.py`, `marked_up_text_chunk_generator.py`, `disambiguator.py`, `sourcesheets/views.py`'s `get_correct_text_from_source_obj`, `reader/views.py`'s `social_image_api`, `history.py`'s `text_at_revision` (matches `db.history`'s own legacy `language`-field keying), `search.py`'s `index_ref`, `user_profile.py`'s recently-read text preview.

**Migrated to real `actualLanguage` matching — deliberately closing the leak**, where the caller's intent was genuinely "real English" or "real Hebrew," not "whatever's in the direction bucket": `Ref.word_count()` (no live callers found anywhere, migrated anyway rather than leaving a redundant legacy-only path around). `api/views.py`'s `KnnSearch._ref_text` also belongs here, with a wrinkle worth flagging explicitly: **its diff against master is empty.** The call site reads `Ref(ref).text(lang="he")`, textually identical to the line on master — but `lang=` means something different now (exact `actualLanguage` match, not the old positional binary bucket), so its behavior differs from master even though nothing shows up in a file diff.

**Moved away from language-bucketed logic entirely — behavior unchanged:** `_build_links_internal` (`sefaria/helper/link.py`) no longer touches `TextChunk`/`TextFamily` at all — its `TextFamily(...).contents()`-based non-emptiness check is replaced with `Ref.is_empty(vstate=vstate)`/`get_state_ja(vstate=vstate)`/`get_subrefs_count(...)`, none of which take a language argument. This is equivalent to the old behavior, but not because the new check is language-agnostic — it isn't. `VersionState`'s `_all` mask (`sefaria/model/version_state.py:219`) is literally `_en + _he`, built from the same legacy `Version.language` field `TextFamily` read from. Both the old and new checks are equally blind to real non-en/he-languaged versions; this migration just stops routing that (pre-existing, unchanged) blind spot through legacy field names. Two structural helpers in `sefaria/helper/schema.py` (used when converting between simple/complex text structures) similarly construct the new `TextChunk` directly from an already-loaded `Version`'s own `lang`/`direction`/`vtitle` — no bucket decision needed there at all, since the version is already known.

**Completely removed:** the alt-structure text preview (`SchemaNode.as_index_contents()`'s `wholeRefPreview`/`refsPreview` generation, `expand_ref`) — confirmed dead, its only consumer was the pre-React `templates/js/headers.js`, itself gated behind an already-unused flag.

**Marked as legacy, not removed (beyond `TextFamily` above):** it's for the v1 API — see the `LegacyTextChunk` class docstring (`sefaria/model/legacy_text.py`) for what else still technically routes to it and why none of that is a real reason to keep the class around. The legacy v1 `texts_api` POST now derives a sensible `direction` from `language` when a caller doesn't send it, so older API clients keep working without needing to be updated themselves. `LegacyTextChunk`'s own internal availability-check helpers construct `LegacyTextChunk` directly rather than going through the new `Ref.text()`, so they keep matching the whole legacy bucket rather than one exact language.

**Won't work, for a different reason than it might look like:** four scripts (`find_replace.py`, `fix_rashi_double_quotes.py`, `remove_marks_from_tyt.py`, `replace_stuma_and_ptuha.py`) *were* updated — not because they construct `TextChunk` directly, but because `sefaria/helper/text.py`'s `find_and_replace_in_text()`/`modify_text_by_function()` had their positional signature changed (`lang` → `direction`) as real Phase 1/2 migration work, and these scripts call those functions directly. Without updating them, they'd silently pass `"he"` where `"rtl"` is now expected. Separately, roughly 20 *other* one-off analysis scripts under `scripts/` still construct `TextChunk(oref, "he"/"en", vtitle)` directly, using the pre-rename 3-positional-argument convention — since `TextChunk` now refers to the new class, the literal `"he"`/`"en"` is read as `languageFamilyName` and matches nothing. These were left as-is: none are wired into any scheduled job, only invoked by hand, so updating them was judged not worth the review burden until someone actually needs to run one.

**Deviation from the original plan worth flagging:** `sefaria/helper/text.py`'s `split_text_section` was planned for deletion (confirmed zero external callers — only its own recursive self-call). It was migrated instead of deleted. Still zero callers as of this PR — a candidate for actual removal as a follow-up.

## Bugs fixed (pre-existing, not introduced by this PR)

- **`merge_texts()` had two bugs**, both already reachable in production through `LegacyTextChunk._choose_version_by_lang` (used throughout the app for any multi-version merge): depth>2 merges flattened the source-attribution list across the *entire* node, losing which position each source belonged to (the code's own comment already flagged this: "the mapping of source names to segments is lost for merged texts of depth > 2"); and empty positions defaulted to attributing the first candidate version instead of being left unattributed, fabricating version credit for content that version never actually had. Fixed by preserving nesting through the merge and using an explicit "no attribution" sentinel for empty positions. `export.py`'s merge visitor (used for exporting a merged text) was updated to match the corrected, now-nested shape.
- **The v3 texts API reported a version as "merged from multiple versions"** (`TextRequestAdapter._append_version`, read by the client as `currentVersion.merged = !!(currentVersion.sources)`) whenever 2+ candidate versions existed for a language, even when the requested version alone was already complete for that ref — surfacing a misleading "merged" badge on an ordinary single-version read. Fixed to only report it on genuine multi-version attribution.

## Performance

- `TextRequestAdapter._add_ref_data_to_return_obj` now shares a single `VersionState` fetch to compute next/prev section links, instead of two independent uncached DB round-trips. This isn't editor-specific — `_add_ref_data_to_return_obj` is shared by `edit_text`, the `/api/v3/texts` endpoint, and the main reader's own SSR panel loading (`make_panel_dict`), so this benefits ordinary reader page loads too, not just the editor.

## Testing

- New regression tests, all synthetic fixtures (no dependency on real book data that could change): `merge_texts()`'s positional-correspondence/no-fake-attribution fix (`chunk_test.py`), the version-title-suffix save fix (`chunk_test.py`), the v3 API's false-merge-flag fix — both the no-false-positive and still-flags-genuine-merges cases (new `text_request_adapter_test.py`), `get_links()`'s version-metadata attribution (`links_test.py`), and `edit_text`'s response shape — `isPrimary`/`isEdited` flags and position metadata (`reader/tests.py`).
- **`TextChunk` read/save coverage parity with `LegacyTextChunk`** (`chunk_test.py`, 6 new tests, `LegacyTextChunk`'s own tests untouched): `LegacyTextChunk` had comprehensive read-path coverage (verse/chapter/range/spanning refs, depth-1, nested commentary-style addressing) and an extensive save-path test (blank writes, extending beyond current extent, writing within extent, HTML sanitization, whole-chapter overwrite with blank trimming, depth-3 saves) — but that save-path test only ever exercised `LegacyTextChunk.save()`, which is no longer the live save path. Added equivalent coverage for the new `TextChunk`, all synthetic fixtures.
- Also updated three test files that built fixtures via `LegacyTextChunk` even though the production code they exercise was already migrated to `TextChunk` (`sefaria/tests/search.py`'s `test_make_text_index_document`, `search_dependencies_test.py`'s book fixture, a `linker_test.py` validation helper) — consistency, not a behavior change.
- **`resolve_default_version()` guarded against an explicit `priority=None`** (`text_request_adapter.py`) — `Version._normalize()` can set `priority` to `None` on a `float()` parse failure, not just leave it missing; `getattr(v, 'priority', 0)`'s default only covers the latter, so `max()` could raise `TypeError` comparing `None` to a number. Pre-existing on master (carried over unchanged when this logic was extracted into the new helper), fixed here since we were already in this function.
- Full existing suite run clean: `chunk_test.py`, `text_test.py`, `schema_test.py`, `auto_linking_test.py`, `linker_test.py`, `links_test.py`, `social_image_api_test.py`, `search_dependencies_test.py` (20 tests against the real indexing pipeline via a fake ES client) — no regressions.
- Manually checked already:
  - `/api/texts` and `/api/v3/texts` for a real ref, `/api/links?with_text=1` on a heavily-commented ref, `/api/img-gen` for en/he (viewed the rendered image).
  - A full delete+regenerate of every auto-generated Genesis/Rashi-on-Genesis link (2018 links, confirmed byte-identical before vs. after).
  - "Gur Aryeh on Shemot" (depth≥3, two English versions) — merged reading in the reader and merged-text export/download both work correctly.
  - Adding a source to a sheet with a specific version selected (`sourcesheets/views.py`'s `get_correct_text_from_source_obj`).

## Merging with master

This branch was rebased against a moving `master` partway through review. One real semantic conflict, in `Ref.is_text_fully_available()`: master had independently added a third branch (handling refs to whole branching/structural nodes, e.g. a Sifra parsha, via `_aggregate_structure_state`) that didn't exist when this branch diverged. Resolved by keeping this PR's `direction=`-based migration for the segment/section-level branch (also dropping a `try/except NoVersionFoundError` that the new `TextChunk` doesn't need — confirmed no code path reaches it, since the new class returns empty text gracefully instead of raising) while preserving master's new structural-node branch intact. Verified all three branches work post-merge against real refs (a plain segment/section ref, a simple book-level `JaggedArrayNode` ref, and a complex structural ref like "Pesach Haggadah").

Two master-authored tests needed updates for compatibility with this PR's real-language semantics — not regressions in their own original intent, confirmed by checking both against `origin/prod` directly (prod predates this PR's changes, and both pass there):
- `modtools_test.py::test_rename_collision_different_language_family_allowed` — asserted a literal versionTitle string for a non-en/he version that this PR's auto-suffix mechanism now rewrites on save (by design, to prevent exactly the collision the test was checking couldn't happen). Updated to assert the real, suffixed behavior.
- `text_test.py::test_index_rename_migrates_versions` — its fixture inserts `Version` docs via raw `db.texts.insert_many()`, bypassing `Version._normalize()` and therefore `actualLanguage`/`languageFamilyName`, which `Ref.text(lang=...)` now matches on instead of the legacy binary `language` field. Added both fields to the fixture.

Three other CI failures reported during review are unrelated pre-existing breakage, confirmed independently of this branch (checked against a clean `master`/`prod` checkout) and intentionally left untouched: `linker_editor_test.py::test_usage_index_surgical_add_remove` (new, unreleased feature, never had a clean run), and `search_dependencies_test.py`'s `test_hooks_noop_when_search_index_on_save_off`/`test_category_hook_empty_new_path_guard` (a naming mismatch between a test merged 2026-08-19 and an unrelated `dependencies.py` refactor merged one week later, already broken on `prod`).

## What should be manually checked before merge

- **Saving/editing through the UI** (existing English/Hebrew versions, a new non-en/he version, the "Add Translation" and "Original Translation" flows) — we've verified this via unit tests and targeted manual checks, but hope the content team can also run through real editing sessions before this ships, since that's the surface with the most user-facing change.
- **The linker widget** (`linker.v2.js`-embedded pages) — spot-check that it still returns text for a few refs, including one needing merge/gap-fill across versions.
- **Topics/LLM pipeline** — `topic_prompt.py` consumes `get_links(with_text=True)` and calls `.text()` directly; worth confirming topic-prompt generation still produces correct source text.


sefaria/model/legacy_text.py
```diff
@@ -0,0 +1,865 @@
...
+        :param strict: if True, throws error if len(ind_list) != len(ref_list). o/w truncates longer array to length of shorter
+        :return: list[(Ref, Match, str)] - list of tuples. each tuple has a segment ref, match object for the match, and text for the segment
+        """
+        ref_list = self.nonempty_segment_refs()
+        text_list = [x for x in self.ja().flatten_to_array() if len(x) > 0]
+        if len(text_list) != len(ref_list):
+            if strict:
+                raise ValueError("The number of refs doesn't match the number of starting words. len(refs)={} len(inds)={}".format(len(ref_list),len(ind_list)))
+            else:
+                print("Warning: The number of refs doesn't match the number of starting words. len(refs)={} len(inds)={} {}".format(len(ref_list),len(ind_list),str(self._oref)))
```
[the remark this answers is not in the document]
YishaiGlasner: pre-existing. moving to legacy is not the time to fix it.

sefaria/model/linker/tests/linker_test.py
```diff
@@ -927,7 +927,7 @@ def test_linker_output_validate(resolver_data, is_ambiguous):
     spans = _extract_debug_spans(doc)
     for span in spans:
         assert span['ambiguous'] == is_ambiguous
-    validation_text = TextChunk(Ref("Genesis 1:1"), lang="en", vtitle="Tanakh: The Holy Scriptures, published by JPS").text
+    validation_text = LegacyTextChunk(Ref("Genesis 1:1"), lang="en", vtitle="Tanakh: The Holy Scriptures, published by JPS").text
```
nsantacruz: You can use the new TextChunk here. No reason not to.

sefaria/model/tests/chunk_test.py
nsantacruz, line 1: why not use new TextChunk here?

sefaria/tests/search.py
```diff
@@ -12,7 +13,7 @@ def test_make_text_index_document():
     version = [v for v in oref.versionset() if v.versionTitle == vtitle][0]
     lang = version.language
     priority = version.priority
-    content = TextChunk(oref, lang, vtitle=vtitle).ja().flatten_to_string()
+    content = LegacyTextChunk(oref, lang, vtitle=vtitle).ja().flatten_to_string()
```
nsantacruz: why use LegacyTextChunk here?

sefaria/model/tests/chunk_test.py
YishaiGlasner: the legacy is still in use for api v1. i'm adding new tests for the new one

stevekaplan123 (commented):
Can you add to the PR a description of why we need LegacyTextChunk?  I see that it's used in gardens, but I find its existence confusing.

YishaiGlasner:
> Can you add to the PR a description of why we need LegacyTextChunk? I see that it's used in gardens, but I find its existence confusing.

@stevekaplan123 added comment to LegacyTextChnk and edoted desc

stevekaplan123:
### Real ISO language codes leaking into the legacy `en`/`he`-keyed subsystems

Three related problems, all at the seam where the new real-language code hands off to code that still keys on the binary `en`/`he` bucket. They share a root cause, so I've grouped them.

---

#### 1. `direction` is guessed wrong on save, which creates a duplicate `Version` instead of updating

[`reader/views.py:1867`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/reader/views.py#L1867) and [`:1874`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/reader/views.py#L1874):

```python
direction = t.get("direction") or get_direction_from_legacy_lang(t["language"])
```

`get_direction_from_legacy_lang` only knows `"he"` → `rtl`; everything else → `ltr`. But `t["language"]` is now a real ISO code, so `yi`, `ar`, `fa`, `jrb` all resolve to `"ltr"`.

That fallback isn't a rare path — it's the *normal* path when editing an existing version. `readNewVersion` deliberately omits `direction` in that case ([`editor.js:4185-4187`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/static/js/s1/editor.js#L4185-L4187), guarded on `sjs.editing.direction`, which is set from the edited version at [`editor.js:3676`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/static/js/s1/editor.js#L3676)).

Repro: edit an existing Yiddish version and save.
- Server computes `direction="ltr"`.
- `TextChunk(oref, actual_lang="yi", vtitle=..., direction="ltr")` matches nothing — the stored version is `rtl`.
- `_compute_saveable` is satisfied (vtitle + direction present), so the save proceeds **as a new version**.
- The duplicate-title guard queries with the same wrong `direction`, so it doesn't catch the collision either.

Result: a second `Version` row with the same `versionTitle`, and the text on the site is unchanged. No error surfaced to the user.

Cheapest fix is either to have `readNewVersion` always send `direction`, or to only derive it server-side when `language` is literally `"en"`/`"he"` and otherwise read it off the existing version.

---

#### 2. `modify_text` forwards the real ISO code into `db.history` and the search queue

[`sefaria/tracker.py:47`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/sefaria/tracker.py#L47) and [`:49`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/sefaria/tracker.py#L49):

```python
_post_modify_changed_segments(user, action, oref, lang, vtitle, old_text, text, version_id, skip_links=skip_links, **kwargs)
count_and_index(oref, lang, vtitle, to_count=count_after)
```

`lang` here is the real ISO code, but both destinations are still `en`/`he`-keyed:

- `post_modify_text` → `model.log_text(user, action, oref, lang, vtitle, ...)` writes a history row with `language: "de"`. `text_at_revision` and `split_text_section` query history by the legacy bucket, so that row is invisible to them — no revert, no diff, no attribution for that edit.
- `count_and_index` → `IndexQueue({"lang": lang, ...})` → `TextIndexer.index_ref`, which resolves a direction from the legacy bucket. For Yiddish it resolves `ltr`, finds no version, and indexes nothing. The segment silently never appears in search.

Both are silent: the edit itself succeeds and looks fine.

Suggestion: translate to the legacy bucket at this call site (derive it from `chunk`'s direction / `actual_lang`) rather than passing `lang` through unchanged.

---

#### 3. After the auto-suffix rename, `tracker` logs a version title that doesn't exist

Same two lines as above, different failure. `Version._normalize()` renames a new non-`en`/`he` version from `"Foo"` to `"Foo [de]"` ([`text.py:1439`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/sefaria/model/text.py#L1439)), but `modify_text` keeps using its own pre-save `vtitle` local for both the history row and the `IndexQueue` entry. So both name `"Foo"` while the stored version is `"Foo [de]"` — a history row and a search-index job pointing at a version that was never saved under that name.

Fires on every newly created non-`en`/`he` version. Fix looks like one line: use `chunk.vtitle` after `chunk.save()` instead of the `vtitle` argument.

stevekaplan123:
### `/add/<book>` builds a broken POST url and can't save

[`reader/views.py:1554`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/reader/views.py#L1554):

```python
fetch_oref = oref.padded_ref().context_ref()
adapter = TextRequestAdapter(fetch_oref, versions_params, fill_in_missing_segments=False)
text = adapter.get_versions_for_query()
text['sections'] = oref.sections[:]
text['toSections'] = oref.toSections[:]
```

The fetch correctly uses the **padded** ref, but `sections`/`toSections` are taken from the original **unpadded** `oref`. The old `TextFamily` path populated these from the padded ref, so this is a behavior change.

For a book-level ref like `/add/Genesis`, `oref.sections` is `[]`. That value reaches the client as `sjs.editing.sections`, and `readNewVersion` builds the post url from it ([`editor.js:4159-4161`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/static/js/s1/editor.js#L4159-L4161)):

```js
version.postUrl = sjs.editing.book.replace(/ /g, "_");
for (var i = 0; i < sjs.editing.sectionNames.length - 1; i++) {
    version.postUrl += "." + sjs.editing.sections[i];
}
```

Genesis has depth 2, so the loop runs once with `sections[0] === undefined`, producing `postUrl = "Genesis.undefined"` and a POST to `/api/texts/Genesis.undefined`. The save fails.

Worth noting `/add/<book>` isn't a hand-typed url — editor.js generates it itself, so this is on a normal user path.

Same unpadded/padded mismatch also feeds [`buildEditor`'s title construction at `editor.js:3511-3512`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/static/js/s1/editor.js#L3511-L3512), which indexes `sections[i]` over the same range.

Fix is presumably to derive these from `fetch_oref`-consistent padding, i.e. use the padded ref for `sections`/`toSections` as `TextFamily` did.

stevekaplan123:
### The `[lang]` auto-suffix runs on *every* `Version.save()`, not just on creation

[`sefaria/model/text.py:1435-1439`](https://github.com/Sefaria/Sefaria-Project/blob/99b08b1653784af12a86c13923af7a9ec2cc1583/sefaria/model/text.py#L1435-L1439):

```python
if self.actualLanguage not in ("en", "he") and versionTitle and not re.search(rf"\[{re.escape(self.actualLanguage)}\]$", versionTitle):
    self.versionTitle = f"{versionTitle} [{self.actualLanguage}]"
```

This lives in `_normalize()`, which runs on every save — not only when a version is created. So any unrelated save of an existing unsuffixed non-`en`/`he` version silently renames it.

Concrete case: someone updates a Yiddish version's license or `versionSource` through `update_version_metadata`. No one touched the title, but `_normalize` appends `" [yi]"` on the way out. And because `versionTitle` is part of the version's primary key, that rename isn't cosmetic — it fires the full pkey-change cascade in `dependencies.py`: whole-book Elasticsearch reindex, history rewrite, notification updates.

Two problems with that:

1. **Surprise renames.** A metadata-only edit changes the version's displayed title, and any external reference to the old title (links, sheets, saved urls, scripts) now points at nothing.
2. **Cost.** A one-field metadata save turns into a book-wide reindex plus a history rewrite.

There's also existing data to consider: any non-`en`/`he` version currently in the database without the bracket suffix will get renamed the first time anything saves it, for any reason.

If the intent is "new versions get a suffix," this probably belongs on the creation path rather than in `_normalize()` — or at minimum gated on the version not already existing.

reader/views.py
```diff
@@ -1861,17 +1866,20 @@ def _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=comment
...
+            # non-Hebrew rtl language.
+            direction = t.get("direction") or (get_direction_from_legacy_lang(t["language"]) if t["language"] in ("en", "he") else None)
             chunk = tracker.modify_text(apikey["uid"], oref, t["versionTitle"], t["language"], t["text"], t["versionSource"], direction=direction, method="API", skip_links=skip_links, count_after=count_after)
             return jsonResponse({"status": "ok", "versionTitle": chunk.vtitle})
         else:
             @csrf_protect
             def protected_post(request):
                 t = json.loads(j)
-                direction = t.get("direction") or get_direction_from_legacy_lang(t["language"])
+                direction = t.get("direction") or (get_direction_from_legacy_lang(t["language"]) if t["language"] in ("en", "he") else None)
```
stevekaplan123, line 1884: Isn't language always en or he?  It's actualLanguage that differs?

reader/views.py
```diff
@@ -1861,17 +1866,20 @@ def _get_text(oref, versionEn=versionEn, versionHe=versionHe, commentary=comment
...
-            # predate the `direction` field (e.g. scripts/pull_text_from_server.py) can still be
-            # served correctly by deriving it, rather than failing on a missing new version.
-            direction = t.get("direction") or get_direction_from_legacy_lang(t["language"])
+            # Only derive direction from the legacy en/he bucket when `language` really is that
+            # bucket (callers that predate the `direction` field, e.g.
+            # scripts/pull_text_from_server.py). A caller sending a real ISO code (e.g. our own
+            # editor, or any future caller) must send direction itself -- guessing one from a
+            # real language via the binary en/he mapping would silently be wrong for any
+            # non-Hebrew rtl language.
+            direction = t.get("direction") or (get_direction_from_legacy_lang(t["language"]) if t["language"] in ("en", "he") else None)
```
stevekaplan123, line 1877: Isn't language always en or he? It's actualLanguage that differs?

stevekaplan123 (commented):
Added two questions.  Once you answer I'll be ready to approve

reader/views.py
YishaiGlasner: language here comes from the request. the view is designed on a way it will be ok also if the client sent wrong request.

reader/views.py
YishaiGlasner: same

stevekaplan123:
One more issue:

Your PR description says:
```
What "saveable" means changed. LegacyTextChunk was saveable whenever lang and vtitle were both given, gated further by fallback_on_default_version/exclude_copyrighted to guard its implicit/fuzzy vtitle-guessing mode. The new TextChunk never has that fuzzy mode — it's saveable whenever vtitle is given plus either direction or lang (not both required), with no fallback/guessing params. (exclude_copyrighted was also confirmed dead on LegacyTextChunk itself — never passed True anywhere — and dropped from Ref.text()'s public signature entirely; grepped both this branch and master for any caller passing it, none found.)
Version.pkeys changed from ["title", "versionTitle"] to ["title", "direction", "versionTitle"]. This isn't a hypothetical concern: the live data already has real collisions on the old key. We found 60 (title, versionTitle) pairs that currently exist as both an ltr and an rtl version — e.g. "Rashi on Psalms" / "Wikisource Mikraot Gedolot" exists in both directions. Whether that data pattern itself is desirable is a separate question this PR doesn't touch — we didn't create or change it, we only changed pkeys to correctly reflect the uniqueness boundary the data already has, instead of one that was already silently violated.
```

However, how can TextChunk be saveable with just vtitle + (direction or lang) if Version.pkeys is now ["title", "direction", "versionTitle"].  Shouldn't TextChunks be saveable only if given vtitle, direction, AND lang?

YishaiGlasner:
> One more issue:
> 
> Your PR description says:
> 
> ```
> What "saveable" means changed. LegacyTextChunk was saveable whenever lang and vtitle were both given, gated further by fallback_on_default_version/exclude_copyrighted to guard its implicit/fuzzy vtitle-guessing mode. The new TextChunk never has that fuzzy mode — it's saveable whenever vtitle is given plus either direction or lang (not both required), with no fallback/guessing params. (exclude_copyrighted was also confirmed dead on LegacyTextChunk itself — never passed True anywhere — and dropped from Ref.text()'s public signature entirely; grepped both this branch and master for any caller passing it, none found.)
> Version.pkeys changed from ["title", "versionTitle"] to ["title", "direction", "versionTitle"]. This isn't a hypothetical concern: the live data already has real collisions on the old key. We found 60 (title, versionTitle) pairs that currently exist as both an ltr and an rtl version — e.g. "Rashi on Psalms" / "Wikisource Mikraot Gedolot" exists in both directions. Whether that data pattern itself is desirable is a separate question this PR doesn't touch — we didn't create or change it, we only changed pkeys to correctly reflect the uniqueness boundary the data already has, instead of one that was already silently violated.
> ```
> 
> However, how can TextChunk be saveable with just vtitle + (direction or lang) if Version.pkeys is now ["title", "direction", "versionTitle"]. Shouldn't TextChunks be saveable only if given vtitle, direction, AND lang?

vtitle+direction - we have all pkeys.
vtitle + lang - either:
1. no existing versions (new) - fails in assertion, line 1985.
2. 1 existing version - great, go ahead and save it. that allows you to fund the version in more natural way (language rather than direction).
3. more than 1 version (this should be rare and probably doesn't exist today - it means same language [and vtitle!] with different directions, like ladino in hebrew or latin letters) - line 1984 tries to get the version and raises in line 1838.

stevekaplan123 (dismissed):

yitzhakc (approved):

PR #3673, opened 2026-08-30
fix(search): restore start-date priority in comp_start_date (sc-46790)

YishaiGlasner:
## Summary
- Fixes [sc-46790](https://app.shortcut.com/sefaria/story/46790): search results sorted by chronology were ordering by end date instead of start date, and refs with a start/end year of 0 were incorrectly falling back to the default (3000).
- [PR #3095](https://github.com/Sefaria/Sefaria-Project/pull/3095) rewrote the comp-date logic as `getattr(tp, 'end', None) or getattr(tp, 'start', 3000)`, which unintentionally flipped priority from start→end to end→start, and used `or` — so a legitimate year of `0` (falsy) was treated as missing and replaced with `3000`.
- `make_book_index_document` (the Books entity index) had the identical regression, mirroring the same buggy expression. Extracted a shared `_comp_date_from_time_period()` helper and use it in both the text index (`TextIndexer`) and the book index builders so they can't drift out of sync again.
- Restores the original intent everywhere: prefer `start`, fall back to `end`, default to `3000` (or `None` for the sparse book-index field) only when both are actually `None` — via explicit `is not None` checks rather than truthiness.

## Test plan
- [ ] Confirm search sorted by chronology now orders by start date (e.g. Genesis before Psalms) rather than end date
- [ ] Confirm refs with a start/end year of `0` are no longer pushed to the end of chronological results
- [ ] Reindex the Books entity index — existing `compDate` values baked in under the old logic will need a reindex to pick up the fix

### DIFF
--- a/sefaria/search.py
+++ b/sefaria/search.py
@@ -1217,7 +1217,9 @@ class TextIndexer(object):
         indexed_categories = get_search_categories(oref, categories)
 
         tp = cls.best_time_period
-        comp_start_date = int(getattr(tp, 'end', None) or getattr(tp, 'start', 3000)) # If there is no end/start date, use 3000 which make it appear at the end of the results
+        start = getattr(tp, 'start', None)
+        end = getattr(tp, 'end', None)
+        comp_start_date = int(start if start is not None else (end if end is not None else 3000))
         ref_data = RefData().load({"ref": tref})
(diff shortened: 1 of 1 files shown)


yodem (approved):

yitzhakc:
Force merge

PR #3687, opened 2026-09-01
Sidebar ads: target by page type alongside keywords

relyks:
### Description

Sidebar ads are the small promotional boxes that appear in the sidebar of many pages. Until now, they could only be targeted by **keywords**, which describe *what* the reader is viewing, such as a book, category, or topic. Keywords don't distinguish between page types. For example, a Talmud book's table of contents and the Talmud category browse page both produce the keyword `talmud`, so an editor couldn't target one without also targeting the other.

This PR adds a **Page Type** dropdown to sidebar ads in Strapi. An ad must match the selected page type in addition to the existing keyword, language, date, and audience checks. Include-keyword matching stays the same, but the rules for empty keywords, exclusions, and combined include/exclude keywords were deliberately revised during review.

The work has two main parts:

1. **Replace recorded network traffic with code-built test payloads.**
   
   The existing tests replayed `.har` files containing recorded Strapi responses. The replay tool matches the exact request body, so adding even one field to the GraphQL query would break all 14 recordings.
   
   Each recorded scenario now uses the existing payload factory. The payloads were generated by comparing each recording with the factory's defaults, and each rebuilt payload was mechanically verified against its recording after generation. This comparison is not repeated on each test run. A rule in the header of `strapi.scenario-payloads.js` prohibits changing the frozen recorded values. The `.har` files remain in the repository for reference.
   
   The payload factory defines the test schema. Its field list must be updated alongside the GraphQL query, and setting an undeclared field in a spec throws. A browser test checks that the client's actual query requests the expected fields, and a separate self-consistency test checks that the builders emit exactly the fields the factory declares.

2. **Add page-type targeting and related sidebar ad changes.**
   
   The 17 page types are defined in `static/js/sefaria/pageTypes.js`:
   
   `all_pages` `homepage` `category_toc` `book_toc` `topic_page` `author_page` `portal_page` `topic_category_toc` `topics_landing` `all_topics` `calendars` `translations` `collection_page` `public_collections` `user_library` `notifications` `voices_home`
   
   Most page types are determined from the reader's panel state, including the open menu and selected categories. Topic pages require fetched data to distinguish between a regular topic, an author, and a sponsor portal. Once that data loads, `TopicPage`, which fetches it, passes the resolved type to the ad component through the new `pageTypeOverride` prop.
   
   Six pages that previously had no ad slot now have one: the topics landing page, Voices home, notifications, translations, collection pages, and the public collections directory. Portal pages also get an ad slot below the sponsor's content. These additions give the dropdown values corresponding places to render.
   
   **Deploy the production frontend first, then deploy the updated Strapi instance.** There is no fallback for older Strapi schemas. During the gap between deployments, missing fields cause GraphQL to reject the entire promotions query, so no sidebar ads, banners, or modals render for clients using the new query. The console error includes `LIKELY SCHEMA MISMATCH`. This temporary gap is intentional and preferable to serving a fallback response that could bypass targeting or be cached for other visitors.
   
   The keyword rules also changed. An empty keywords field now means "no keyword restriction." Exclusion-only keywords match only pages that have keywords, so they can no longer be used to mean "show everywhere."
   
   The background field changed from the `hasBlueBackground` checkbox to **`sidebarAdBackgroundColor`**, a hex color field like the banner background field. With no color, the ad uses its default styling. Dark colors use white title and body text, with a white button whose text matches the ad's background color. Entering `#004E5F` reproduces the previous blue styling exactly. Very light colors apply only a background tint and keep the default dark text.
   
   **Sidebar ad bodies now render Markdown**, using the same rendering path as banners and modals. Editors can include bold text, links, images, and line breaks instead of displaying literal Markdown syntax such as `**` and `[]()`. Body images appear centered on their own line and scale to fit the sidebar. The CSS is scoped so it doesn't affect the button's inline icon. Titles remain plain text, like banner and modal headers.

### What could warrant closer review

1. **Deploy production first, then Strapi. Promotions will temporarily stop rendering while the schemas differ.**
   
   The Strapi deployment needs two Side Bar Ad content-type changes: a `pageType` dropdown with 17 values and a default of `all_pages`, and a `sidebarAdBackgroundColor` color field replacing the `hasBlueBackground` checkbox.
   
   If either new field is missing, GraphQL rejects the entire promotions query. For clients using the new query, sidebar ads, banners, and modals won't render until the updated Strapi schema is deployed. The console error includes `LIKELY SCHEMA MISMATCH`.
   
   An earlier implementation retried a failed request once with the old query. That fallback was removed because the shared cache key didn't include the query body. A transient browser error could therefore cause a response without `pageType` to be cached for up to seven days, turning restricted campaigns into site-wide campaigns.
   
   The cache key now includes a hash of the query body. This also separates responses requested by older browser bundles from those requested by current clients. A pytest checks that separation, and the browser test checks that exactly one query is sent. The request count distinguishes "no fallback" from "a fallback that also failed."

2. **Keyword rules changed, and campaigns using exclusions to mean "everywhere" need an update.**
   
   Some targetable pages, including the homepage, calendars, and notifications, don't produce keywords. Previously, an exclusion-only ad such as `!social-issues` matched any page without that keyword, including pages with no keywords. Editors could also use a value such as `!nowhere` to make an ad appear everywhere.
   
   The revised rules are:
   
   - An empty keywords field means "no keyword restriction."
   - Exclusion-only ads match only pages that have keywords and don't contain an excluded keyword.
   - Ads with both includes and exclusions must satisfy both conditions. Previously, satisfying either was enough.
   
   Campaigns using an exclusion such as `!nowhere` to mean "everywhere" should use an empty keywords field. Campaigns with meaningful exclusions also need review because they will no longer appear on pages without keywords.

3. **Text-library collection browse pages and category pages intentionally share `category_toc`.**
   
   A collection browse page such as `/texts/.../Covenant and Conversation` and a category page such as `/texts/Tanakh/Torah` may look different. Still, both are backed by plain `Category` records and rendered by the same component. The client doesn't distinguish between them, so both use `category_toc`. These are separate from the collection pages targeted by `collection_page`.
   
   To target a specific text-library collection, combine `category_toc` with its keyword, such as `covenant and conversation`. Only commas split the keyword list, so spaces within a keyword are preserved. Unit and browser tests cover this behavior.
   
   The keyword format also depends on the page family. Text-library pages, including categories and books, use lowercased display names with spaces, such as `covenant and conversation`. Topic pages use slugs with hyphens, such as `adin-steinsaltz`. Matching is exact, so editors targeting a topic should use the slug from its URL.

4. **Portal pages match only `portal_page`, even when the topic is also an author.**
   
   Both current portals, Rabbi Sacks and Rabbi Steinsaltz, are also authors. However, `portal_slug` belongs to the base Topic model, so that future portals could represent something else, such as an organization.
   
   Classification checks for a portal first and returns only `portal_page`. An `author_page` campaign therefore won't appear on a sponsor portal. Editors must explicitly target portal pages.

5. **The factory defines the test schema; recordings are no longer checked on each run.**
   
   After generation, each rebuilt scenario payload was mechanically verified against its original recording. Tests no longer read or compare against the `.har` files.
   
   This removes the need to reconcile future query fields with old snapshots, but also removes an ongoing automated check against the recorded Strapi responses. The replacement checks have a narrower scope:
   
   - The factory rejects undeclared fields, and its field list must be updated alongside the GraphQL query.
   - A browser test reads the intercepted request body and checks that the client's actual query requests the expected fields.
   - The payload builders are checked against the factory's declared fields.
   - The scenario file's never-edit rule preserves the recorded values that were verified after generation.
   
   The initial comparison established equivalence with the recordings. Preserving those values now depends on following the file's rule, rather than rerunning a fidelity comparison.

6. **Page types are combined across all open panels.**
   
   A `book_toc` ad can match if any open panel contains a book's table of contents. This follows the existing keyword behavior, which also combines values across panels, so both targeting mechanisms use the same scope.

7. **The server cache key includes a hash of the query body.**
   
   The cache version changed from v5 to v7: v6 added `pageType`, and v7 replaced `hasBlueBackground` with `sidebarAdBackgroundColor`.
   
   The query-body hash gives different queries separate cache entries. This matters because browsers can continue running older JavaScript bundles during deployment. An older query can no longer populate the cache entry used by newer clients with a response that omits newer fields.
   
   Error responses were already excluded from caching. A new pytest checks that requests with the same dates but different query bodies use separate cache entries.

### Key Design Decisions

- **Page-type targeting is additive; the keyword changes are intentional.** An existing ad document without a `pageType` is treated as `all_pages`, so a missing value doesn't restrict the ad. This is separate from a Strapi schema that lacks the field entirely, which causes the query to fail. The revised keyword rules live in pure helpers with a Jest truth table: empty keywords are unrestricted, exclusions apply only to pages with keywords, and mixed include/exclude rules must both pass.

- **Unknown page types fail closed.** If Strapi returns an unrecognized value, such as a typo, the ad appears nowhere. This prevents an invalid targeting value from becoming a site-wide campaign. A campaign that doesn't appear can be investigated without first displaying it to unintended audiences.

- **Topic classification waits for the data.** The panel classifier returns `null` for a specific topic because it can't yet distinguish between a regular topic, an author, and a portal. `TopicPage` already waits for the topic data before mounting the ad component. An ad therefore won't appear under a provisional classification and then disappear when the data arrives; it may simply appear a moment later.

- **The portal sidebar passes `portal_page` directly.** That component only renders on portal pages, so it already has enough information to supply the type without passing additional data through the component hierarchy.

- **The test migration was verified once, and the recorded values are frozen.** Rebuilt payloads were generated by comparing the recordings against the factory's defaults, then mechanically checked for equality after generation. The scenario file's never-edit rule preserves those values. There is no fidelity comparison against recordings on subsequent test runs.

### Code Changes

| File                                                                                                                                               | Purpose                                                                                                                                                                                                                                                                                                                                                                                                           |
| -------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `static/js/sefaria/pageTypes.js` **(new)**                                                                                                         | Defines the 17 page types and the pure classification and matching helpers: `classifyPanel`, `classifyPanels`, `topicPageTypeOf`, `normalizePageType`, and `adMatchesPageTypes`. Topic classification checks portals before authors.                                                                                                                                                                              |
| `static/js/ReaderApp.jsx`                                                                                                                          | Adds the active page types (`pageTypes`) to the ad-targeting context.                                                                                                                                                                                                                                                                                                                                             |
| `static/js/Promotions.jsx`                                                                                                                         | Accepts `pageTypeOverride`, adds page-type and revised keyword checks to ad matching, and renders ad bodies as Markdown through the same `InterfaceText` path used by banners and modals. Titles remain plain text. Removes the `Sefaria._inAppAds` cache across mounts, which had masked an earlier effect-dependency bug. The processing effect rebuilds ads on each mount.                                     |
| `static/js/sefaria/sidebarAds.js`                                                                                                                  | Copies `pageType`, defaulting to `all_pages`, and `sidebarAdBackgroundColor` onto each ad. Implements the revised keyword rules through `parseKeywords` and `adMatchesKeywords`. Uses `isLightBackground` to choose between the full colored treatment and a light background tint.                                                                                                                               |
| `static/js/context.js`                                                                                                                             | Requests the new sidebar ad fields in the GraphQL query and reports schema-related fetch failures with `LIKELY SCHEMA MISMATCH`. Does not retry with a legacy query.                                                                                                                                                                                                                                              |
| `static/js/TopicPage.jsx`                                                                                                                          | Passes the resolved page type after topic data loads. Adds a portal ad slot below the sponsor's modules.                                                                                                                                                                                                                                                                                                          |
| `static/js/NavSidebar.jsx`                                                                                                                         | Forwards `pageTypeOverride` through the Promo module's existing props mechanism.                                                                                                                                                                                                                                                                                                                                  |
| `TopicsLandingPage.jsx`, `SheetsHomePage.jsx`, `NotificationsPanel.jsx`, `TranslationsPage.jsx`, `CollectionPage.jsx`, `PublicCollectionsPage.jsx` | Add the previously missing ad slots through one-line additions.                                                                                                                                                                                                                                                                                                                                                   |
| `sefaria/views.py`                                                                                                                                 | Bumps the cache version from v5 to v7: v6 adds `pageType`, and v7 replaces `hasBlueBackground` with `sidebarAdBackgroundColor`. Adds a query-body hash to the cache key.                                                                                                                                                                                                                                          |
| `static/css/s2.css`                                                                                                                                | Centers ad-body images on their own line through `.sidebarPromo p img`. Replaces the `.blue` rules with the `.colored` treatment, with the background color applied inline from the ad's configured color.                                                                                                                                                                                                        |
| `e2e-tests/tests/strapi.scenario-payloads.js` **(new)**                                                                                            | Provides code-built copies of all 14 recordings, attached to their test scenarios. Each was verified against its recording after generation. The file header prohibits changes to the frozen recorded values.                                                                                                                                                                                                     |
| `e2e-tests/tests/strapi-payload-contract.spec.js`                                                                                                  | Checks that payload builders emit exactly the fields declared by the factory. Reads no recordings and runs without a server.                                                                                                                                                                                                                                                                                      |
| `e2e-tests/support/strapi-payload-factory.js`                                                                                                      | Adds the `pageType` default and updates the empty-keyword default. Documents the factory as the test schema, which must be updated alongside the GraphQL query.                                                                                                                                                                                                                                                   |
| `e2e-tests/support/strapi-har-fixture.js`                                                                                                          | Retires the fixture in place. No tests import it, but it remains as documentation of how the recordings were made.                                                                                                                                                                                                                                                                                                |
| 14 × `strapi-{banner,modal,sidebar-ad}*.spec.js`                                                                                                   | Replace HAR replay with code-built payloads. Assertions are unchanged, and comments describing the recording workflow are updated.                                                                                                                                                                                                                                                                                |
| `e2e-tests/tests/strapi-sidebar-ad-page-type.spec.js` **(new)**                                                                                    | Tests page-type targeting, classification, navigation, and schema mismatch behavior in the browser, as described below.                                                                                                                                                                                                                                                                                           |
| `e2e-tests/tests/strapi-sidebar-ad-background.spec.js` **(new)**                                                                                   | Tests three background states on the live stack: a dark color uses the full colored treatment with the configured color, a very light color applies only a tint with default text styling, and no color produces no inline background.                                                                                                                                                                            |
| `e2e-tests/tests/strapi-sidebar-ad-markdown.spec.js` **(new)**                                                                                     | Checks that body Markdown renders as elements, including bold text, links, and an image that loads (`naturalWidth > 0`) and appears centered on its own line. Checks that title Markdown remains literal text.                                                                                                                                                                                                    |
| `e2e-tests/support/seed_topic_pools.py` **(new)**                                                                                                  | Adds a typed script to populate the local Postgres topic-pool table. Imports pool names from `PoolType` so enum renames carry through, and validates each slug against Mongo before writing. Without populated topic pools, `/topics/<slug>` pages return 404 in a fresh local setup. The docstring explains the required server restart and why the existing migration and admin workflows don't fit this setup. |
| `static/js/sefaria/tests/pageTypes.test.js` **(new)**, `sidebarAds.test.js`                                                                        | Cover page classification, page-type matching, keyword rules, and background handling, as described below.                                                                                                                                                                                                                                                                                                        |
| `e2e-tests/tests/README.md`, `e2e-tests/CLAUDE.md`, `strapi.fixtures.js`                                                                           | Update documentation for the testing model based entirely on code-built Strapi payloads.                                                                                                                                                                                                                                                                                                                          |

### Tests

The browser tests use the suite's existing deterministic setup. Strapi responses are served in the browser from generated payloads, the clock is fixed, and a guard fails any test whose page never requested Strapi data.

<details>
<summary><strong>Unit: pageTypes.test.js + sidebarAds.test.js additions</strong></summary>

- **`classifyPanel`** has a test for each branch: homepage versus category navigation, both book TOC menu spellings, the three topic-menu cases (category list, specific topic, and landing page), and each simple menu mapping. Reader panels, search, profile, admin, and missing input return `null`. The specific-topic test explains why `TopicPage` supplies the final classification.

- **`topicPageTypeOf`** checks that portal classification takes precedence when a topic is also an author, authors are identified by `subclass`, ordinary topics are classified correctly, and missing data doesn't throw. The tests also explain why `indexes.length` isn't used: an author can have no cataloged works, as some records in the local database do.

- **`normalizePageType`** converts null, undefined, and empty values to `all_pages`. Known values stay unchanged. Unknown strings pass through unchanged so the matching step can reject them.

- **`adMatchesPageTypes`** checks that `all_pages` matches everywhere, including pages with no classified type; specific types match only when active; and unknown values never match.

- **`sidebarAds.test.js`** checks that the trigger's `pageType` defaults correctly for old documents and preserves an explicit Strapi value. It also covers spaces within multi-word keywords and mapping of `sidebarAdBackgroundColor`, including null. The `isLightBackground` truth table checks the boundary between very light and dark colors: light colors retain default styling, `#004E5F` and other dark colors receive the full treatment, and invalid input uses the readable default. The keyword truth table covers empty, include-only, exclusion-only, and mixed rules.

</details>

<details>
<summary><strong>Browser: strapi-sidebar-ad-page-type.spec.js</strong></summary>

- **Each specific page type is tested on a real page.** Coverage includes the homepage, a real category, a real book's table of contents, seeded topic and author pages, a topic category, topics landing, the A–Z list, calendars, translations, collection pages, the public collections directory, Voices home on the Voices module host, both portals, notifications, and the user library. These tests exercise classification, the ad slot, and matching together.

- **Both portals are tested because their sidebar configurations differ.** The Sacks portal has one sponsor block, while Steinsaltz has all four. Together, they check that the ad slot works with either sponsor configuration.

- **Topic, author, and portal targeting are exclusive.** With all three ad types in one payload, a portal shows only the portal ad, an author page shows only the author ad, and a regular topic page shows only the topic ad.

- **Boundary cases cover missing and invalid values as well as page mismatches.** An ad with a null `pageType`, representing an older document, still appears. A `book_toc` ad doesn't appear on a category page, a `homepage` ad doesn't appear on a topic category page, and an unknown type doesn't match. Every negative assertion first checks that the payload arrived and the page rendered, so missing data or a failed page load can't produce a false pass.

- **Page type and keywords work together.** A `category_toc` ad with the keyword `covenant and conversation` appears on that collection's real browse page but not on `/texts/Tanakh/Torah`. This checks that keywords narrow targeting within a page type.

- **Topic ads don't flash before classification.** The test holds the topic API response while the page loads. No ads render while the response is pending because the ad component hasn't mounted. Once the response is released, only the correctly typed ad appears.

- **Schema mismatches stop all promotions without a retry.** The test simulates an older Strapi schema by returning GraphQL errors inside an HTTP 200 response. It checks that the page renders, the console includes `LIKELY SCHEMA MISMATCH`, no sidebar ad, banner, or modal appears, and exactly one query is sent. The request count distinguishes no retry from a retry that also fails.

- **The actual GraphQL query requests the targeting field.** The synthetic route matches only the URL and returns all fields regardless of what the client requested. A separate assertion therefore inspects the intercepted POST body for `pageType`, with `keywords` and `startTime` as additional checks. Without this assertion, removing `pageType` from the production query could let synthetic tests pass while real ads lose their targeting.

- **Targeting updates during client-side navigation.** A homepage-targeted ad appears on `/texts`, disappears when the reader clicks into a category, and returns on browser back. The Strapi request counter checks that the navigation didn't cause a full page reload. The other tests start with a fresh page load; this case checks that matching works during in-app navigation.

- **Authenticated pages use the stored login session.** Notifications are tested on `/notifications`, and `user_library` is tested on `/saved`, the actual saved-content route. The similar-looking `/texts/saved` route is a category-path navigation page.

</details>

### Testing Patterns

**Visibility tests use real pages and data.** Tests visit actual database content, including a real multi-word category, seeded topic pages, both portal documents, and the real Voices host. They don't substitute mocked page shells, so failures caused by actual page state can be caught.

**Missing prerequisites produce explicit skips.** If topic pools aren't seeded or the configured Voices host is unreachable, the affected tests skip with a message explaining the fix. These skips identify coverage that hasn’t run, rather than letting a test pass without exercising the intended behavior.

**Schema failure and the absence of a fallback are checked separately.** The old-schema test checks the console message, the absence of all promotions, and the request count. Exactly one request confirms that the client didn't retry with a legacy query.

**Fixture migration was verified once, then frozen.** Each rebuilt payload was mechanically compared with its original recording after generation. The scenario file’s never-edit rule preserves those recorded values, so subsequent test runs or future query changes don't need to be compared against the old recordings.


static/js/sefaria/pageTypes.js
```diff
@@ -0,0 +1,156 @@
...
+// "WHAT is the reader looking at" but not "what KIND of page are they on": a Talmud book TOC and
+// the Talmud category browse page both emit the keyword "talmud", and nothing distinguished them.
+// This module adds that second axis. An ad in Strapi may carry a single `pageType` string; an ad
+// only renders when the current page's classified type matches (Promotions.getCurrentMatchingAds
+// ANDs this gate with all the existing ones — keywords, language, dates, audience — which stay
+// exactly as they were).
+//
+// The values are STRINGS, not Symbols, because they cross the wire: each one is an option of the
+// Strapi enumeration field, chosen by a content editor. This frozen map is the single place the
+// vocabulary is defined on the client; the Strapi content type must offer the same strings.
```
yitzhakc, line 14: Is there a place where this is validated? Do we fail gracefully if the Strapi content type doesn't match?

static/js/sefaria/pageTypes.js
```diff
@@ -0,0 +1,156 @@
...
+  if (rawPageType && !KNOWN_PAGE_TYPES.has(rawPageType)) {
+    console.warn(`${UNKNOWN_PAGE_TYPE_LOG} "${rawPageType}"`);
+  }
+  return rawPageType || PAGE_TYPE.ALL_PAGES;
+};
+
+// The menuOpen values that map 1:1 onto a page type, with no further discrimination needed.
+// menuOpen names come from ReaderPanel.jsx's render branches; see classifyPanel for the values
+// that need more than a lookup.
+const MENU_TO_PAGE_TYPE = Object.freeze({
```
yitzhakc, line 79: What about search results page (`/search`)? Is there guidance or documentation on when/what to add here?

yitzhakc (changes requested):

static/js/sefaria/pageTypes.js
relyks: Nope. I essentially made the product decision to include the specific page types that I did. Good call to mention the search page, because I didn't consider it. It's an excellent place for sidebar ads, but it is unclear which parts of the search page context we should use and how to place them alongside the filter list. I asked about it, but we need to discuss it. I'll start a Shortcut story for it. We can address it in a follow-up PR.

PR #3688, opened 2026-09-01
Bug/sc 46770/new sso accounts don t automatically get the la

stevekaplan123:
## Description
New Google/Apple signups don't get the Library Assistant — they see the "Try it" promo banner instead.

The preference is one value on the profile: settings.library_assistant.  A missing key doesn't mean "on by default", it means "hasn't chosen yet" — which is the state that shows the banner.  Only email/password signup wrote it (process_register_form); SSO signup (sso/adapters.py::save_user) and lazy profile creation did not.

Fix: write the key once, where every path goes through — UserProfile.__init__, whenever no stored profile document exists. Email signup, all four SSO endpoints, lazy creation, and any future path inherit it. 

## Code Changes
sefaria/model/user_profile.py — __init__ writes the key when no stored doc exists (the choke point).
sefaria/views.py — removed the now-redundant per-path write.
sefaria/constants/model.py — new LIBRARY_ASSISTANT_SETTING_KEY, re-exported by sefaria/helper/library_assistant.py as SETTING_KEY. The model needs the key's name, and importing the helper to get it would be a circular import (that helper imports UserProfile) — so the string lives in a constants file both can read. No caller changes.
reader/conftest.py + 2 test files — new make_keyless_profile(). Creating a profile now always writes the key, so a pre-migration "legacy" user can only be modelled by creating the doc and unsetting it. Also adds ProfileCreationTest.

## Notes
This is deliberately not adding a settings default: update() merges the stored doc over the defaults and save() persists the result, so a default would flip existing keyless profiles to on. The new write fires only when there is no document, so existing profiles are untouched.  Linking SSO to an account that already exists never reaches save_user, so those users keep their preference too.


reader/conftest.py
```diff
@@ -39,6 +41,16 @@ def purge_test_profiles(*users):
     db.profiles.delete_many({"id": {"$in": ids}})
 
 
+def make_keyless_profile(user):
```
YishaiGlasner: the name tells something general, but this is only about LA. i'd suggest make_legacy_profile or make_la_keyless_profile

YishaiGlasner (changes requested):

reader/conftest.py
stevekaplan123: Good catch — renamed to `make_profile_missing_la_setting` in 8f339e5.

Went with spelling out the key rather than `make_legacy_profile` because `reader/conftest.py` is imported by every reader test, not just the LA ones, so "legacy" would still leave the next reader guessing which key is missing. The docstring spells out "Library Assistant" right underneath, so the `la` abbreviation gets decoded on the spot.

yitzhakc (approved):

PR #3694, opened 2026-09-03
fix(search): load more results when scrolling on mobile web

stevekaplan123:
## What's broken

On mobile web, scrolling to the bottom of search results does nothing. No "Loading more
results..." message appears and no additional results load. On desktop the same search loads
more results correctly.

The same thing is broken on Topics, Collections, Profiles, and reading history in mobile web.

## Why

A `scroll` event only fires on an element that actually has its own scrollbar. If you attach a
scroll listener to a plain `<div>` that grows to fit its content, that listener is never called —
no error, no warning.  That is what happens on mobile. PR #3285 made the **whole page** scroll on mobile so
phone browsers can hide their address bar, which means the inner `.content` panel stopped
scrolling:

```css
/* static/css/s2.css */
.readerApp.singlePanel .readerNavMenu .content,
.readerApp.singlePanel .homeFeedWrapper .content { overflow-y: visible; }
```

Search, Topics, Collections and Profiles all render inside `.readerNavMenu .content`. Every one
of them had a scroll listener bound to that div, and every one of those listeners went dead on
mobile.

There were **three** separate hand-written copies of this listener, so the bug existed in three
places at once:

| File | Used by |
|---|---|
| `InfiniteScroll.jsx` | Search: Sources tab + Books / Authors / Topics tabs |
| `Hooks.jsx` → `useScrollToLoad` | Reading history |
| `Hooks.jsx` → `usePaginatedDisplay` | `FilterableList`: Topics, Collections, Profiles |

## The fix

All three now call one shared hook, `useScrollNearBottom` in `static/js/Hooks.jsx`. It works out
which element is really scrolling and listens to that — the container if it scrolls, otherwise
the page itself.

### Why not just check `Sefaria.multiPanel`?

The obvious one-line fix — *"if mobile, watch the page instead"* — is wrong, and would have
broken a feature that currently works.

The CSS rule above only applies to two selectors. It does **not** apply to the connections-panel
component that holds **"Search in this Text"**. On mobile that component is a fixed-height panel with
its very own scrollbar:

```css
.readerApp.singlePanel .textList { position: fixed; height: 54vh; }
.textList .texts            { height: 100%; overflow-y: scroll; }
```

So on mobile, one search list scrolls the page and another scrolls its own container.  Therefore, **whether
something scrolls is a property of the element, not of the screen size**, so checking `multiPanel` would have broken **"Search in this Text"**.

### The check that determines whether the hook should fire

We check both its CSS values that mean that scrolling is possible and we check the size of the container to make sure that it actually contains overflowing elements.
```
// `scroll` always scrolls, `auto` and `overlay` only when the content does not fit, and
// `visible` never does.
const OVERFLOW_VALUES_THAT_CAN_SCROLL = ['scroll', 'auto', 'overlay'];

// Both halves are needed, because `auto` is a condition rather than a promise: it means "scroll
// only if the content does not fit". A container can report `auto` and still never scroll.
const isScrollingElement = (el) => {
  if (!el) { return false; }

  // Does the CSS permit a scrollbar? Not enough on its own -- `overflow-y: visible` beside an
  // `overflow-x` that is not visible computes to `auto`, which is what .noOverflowX does to the
  // topic and profile containers.
  const {overflowY} = window.getComputedStyle(el);
  const overflowAllowsScrolling = OVERFLOW_VALUES_THAT_CAN_SCROLL.includes(overflowY);

  // Is the box actually hiding anything? On singlePanel these containers grow to their full
  // content height, so nothing is hidden, no scrollbar appears, and no scroll event ever fires.
  const contentIsTallerThanBox = el.scrollHeight > el.clientHeight;

  return overflowAllowsScrolling && contentIsTallerThanBox;
};
```

### DIFF
--- a/static/js/Hooks.jsx
+++ b/static/js/Hooks.jsx
@@ -58,13 +58,26 @@ function useDebounce(value, delay) {
  * scroller: on singlePanel the document scrolls instead, though not for every
  * container, so it is decided per element rather than per breakpoint. */
 
-// Overflow alone is not enough: `overflow-y: visible` beside `overflow-x: hidden` computes to
-// `auto` (.noOverflowX does this to TopicPage), so real overflow has to be checked too.
+// `scroll` always scrolls, `auto` and `overlay` only when the content does not fit, and
+// `visible` never does.
+const OVERFLOW_VALUES_THAT_CAN_SCROLL = ['scroll', 'auto', 'overlay'];
+
+// Both halves are needed, because `auto` is a condition rather than a promise: it means "scroll
+// only if the content does not fit". A container can report `auto` and still never scroll.
 const isScrollingElement = (el) => {
   if (!el) { return false; }
+
+  // Does the CSS permit a scrollbar? Not enough on its own -- `overflow-y: visible` beside an
+  // `overflow-x` that is not visible computes to `auto`, which is what .noOverflowX does to the
+  // topic and profile containers.
   const {overflowY} = window.getComputedStyle(el);
-  if (!['scroll', 'auto', 'overlay'].includes(overflowY)) { return false; }
-  return el.scrollHeight > el.clientHeight;
+  const overflowAllowsScrolling = OVERFLOW_VALUES_THAT_CAN_SCROLL.includes(overflowY);
+
+  // Is the box actually hiding anything? On singlePanel these containers grow to their full
+  // content height, so nothing is hidden, no scrollbar appears, and no scroll event ever fires.
+  const contentIsTallerThanBox = el.scrollHeight > el.clientHeight;
+
+  return overflowAllowsScrolling && contentIsTallerThanBox;
 };
 
 // A null scroller means the document scrolls.
--- a/static/js/Hooks.jsx
+++ b/static/js/Hooks.jsx
@@ -51,6 +51,59 @@ function useDebounce(value, delay) {
   return debouncedValue;
 }
 
+
+/* Shared infinite-scroll trigger. Everything that loads on scroll binds through
+ * here, because getting the scroller wrong fails silently -- a listener on a non-scrolling
+ * element simply never fires. The container a component renders into is not always the
+ * scroller: on singlePanel the document scrolls instead, though not for every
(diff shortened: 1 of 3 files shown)


static/js/Hooks.jsx
```diff
@@ -97,25 +163,13 @@ function useScrollToLoad({scrollableRef, url, setter, itemsPreLoaded = 0, pageSi
...
-      }
-    };
-
-    scrollable.addEventListener('scroll', handleScroll);
-    return () => scrollable.removeEventListener('scroll', handleScroll);
-  }, [loadMore]);
+  // `loadMore` guards itself against overlapping and past-the-end calls.
+  useScrollNearBottom({
+    getScrollCandidate: () => scrollableRef.current,
+    margin: 600,
```
YishaiGlasner: why 600?

static/js/Hooks.jsx
```diff
@@ -51,6 +51,72 @@ function useDebounce(value, delay) {
...
+
+// Both halves are needed, because `auto` is a condition rather than a promise: it means "scroll
+// only if the content does not fit". A container can report `auto` and still never scroll.
+const isScrollingElement = (el) => {
+  if (!el) { return false; }
+
+  // Does the CSS permit a scrollbar? Not enough on its own -- `overflow-y: visible` beside an
+  // `overflow-x` that is not visible computes to `auto`, which is what .noOverflowX does to the
+  // topic and profile containers.
+  const {overflowY} = window.getComputedStyle(el);
```
YishaiGlasner, line 73: `isScrollingElement()` calls `window.getComputedStyle(el)` on every `scroll` event fired by any mounted `useScrollNearBottom` consumer (search results, topic/collection/profile lists). The old per-component handlers only read already-resolved geometry properties (`scrollTop`/`clientHeight`/`scrollHeight`) — none of them called `getComputedStyle`. Forcing a computed-style resolution inside the passive scroll handler, on every tick, on the mobile devices this PR specifically targets, is new per-event cost that wasn't there before.

PR #3695, opened 2026-09-06
chore: scroll should not trigger press

stevekaplan123:
## Description
This PR introduces several UI improvements to the search results page in mobile web.  The main one is that when users scroll, they first have to click, but we don't want any click behavior (i.e. the card gets highlighted upon being clicked) to trigger if they're scrolling. 

## Code Changes
The solution uses timers:
Two timers in [SearchResultCard.jsx:19-84]:

`showTimer` — delays turning the highlight on (100ms). Starts on touchstart. If the finger moves >10px first, the timer is cancelled and the card never lights up. That's what stops scrolls from highlighting cards.

`hideTimer` — delays turning the highlight off (150ms floor). Set on touchend. Without it, a fast tap would show and hide the highlight in a few milliseconds — it'd be invisible. This timer keeps the highlight on screen for at least 150ms from when it appeared.

The edge case they cover together: if the finger lifts before 100ms, then showTimer hasn't fired yet, so `onEnd` shows the highlight immediately, then `hideTimer` holds it 150ms.  

The two timers pull in opposite directions on purpose: showTimer says don't light up too eagerly (that's the scroll fix), and hideTimer says once you have lit up, stay long enough to be seen (that's the fast-tap fix).


static/css/s2.css
```diff
@@ -4784,15 +4784,27 @@ body .ui-autocomplete.dictionary-toc-autocomplete .ui-menu-item a.ui-state-focus
...
   cursor: pointer;
+  /* Mobile browsers paint their own grey flash over any tappable element. It reads as the
+     card changing background color on tap, which isn't in the spec — the pressed shadow
+     below is the only feedback we want. */
+  -webkit-tap-highlight-color: transparent;
 }
-.searchResultCard:hover {
+/* Hover and pressed are one and the same state in the spec: a darker shadow, no change of
+   background. `.is-pressed` is set by usePressState in SearchResultCard.jsx for taps only. */
+.searchResultCard.is-pressed {
```
YishaiGlasner, line 4794: Re: the CSS changes here — worth calling out for reviewers who might read `.searchResultCard.is-pressed` and `:hover` as two different visual treatments being merged into one by accident.

They're not accidentally the same — the spec calls for hover (mouse) and pressed (touch tap) to look identical: same box-shadow, no background-color change. Previously `:hover` gave a shadow and `.is-pressed` gave a grey background — two different effects for what the design treats as one state. This PR unifies the visual style and keeps two separate *triggers* for it:
- `.is-pressed` — set by JS (`usePressState` in `SearchResultCard.jsx`) for touch only.
- `:hover` — native CSS, now wrapped in `@media (hover: hover)` so touch devices (which lack a real hover concept but still fire `:hover` and can get it stuck on the last-tapped element) never match this rule.

So mouse users get the shadow via `:hover`, touch users get the identical shadow via `.is-pressed`, and neither path can leak into the other device type.

— Claude

stevekaplan123: I don't understand this comment.  Are you saying I shouldn't be doing it this way?

YishaiGlasner: No objection to the approach — sorry, that landed wrong. I meant it as a note for other reviewers, not a critique of your code.

The worry I was heading off: someone skimming the diff might think `:hover` giving a shadow and `.is-pressed` giving a grey background got accidentally collapsed into one rule. My comment was just flagging that the collapse is intentional and correct — per the spec, hover and pressed are supposed to look identical (same shadow, no background change), so unifying them into one visual style with two separate triggers (`.is-pressed` for touch, `:hover` for real hover devices) is exactly the right move here. Nothing to change.

YishaiGlasner: no. this is a comment for other reviewers

YishaiGlasner (approved):
Sorry, for not approving it previous time

PR #3714, opened 2026-09-09
Fix multiserver cache staleness across web pods (sc-47135)

nsantacruz:
## Summary

Diagnosed two linker-editor bugs reported in prod (multiple web pods, multiple task pods):

- Saving a linker-editor edit (e.g. a match template) and reloading the page sometimes showed the change, sometimes didn't.
- Investigating that led to a fork-safety bug affecting *any* multiserver cache propagation, not just the linker editor.

## Root causes

1. `library`'s cache is per-process, and the only live cross-pod propagation (`ServerCoordinator` over Redis pub/sub) is drained by web pods only every 20th HTTP request (`MultiServerEventListenerMiddleware`). A reload landing on a different pod/worker than the one that saved can be stale until that worker happens to hit its 20th request.
2. Worse: gunicorn's `on_starting` hook opens the multiserver Redis pub/sub connection **once, in the master process, before any worker forks** (`init_library_cache()` → `server_coordinator.connect()`). `fork()` duplicates that connection's file descriptor into every worker, so if `gunicornWorkerCount > 1`, workers in the same pod compete over one shared TCP stream — an incoming `msync` message is delivered to only whichever worker's `recv()` wins, never the others. This mirrors a hazard already fixed for Celery (`worker_process_init` in `sefaria/celery_setup/app.py`) but never applied to gunicorn's own `post_fork` hook.

## Fix

- **`sefaria/system/multiserver/coordinator.py`**: `ServerCoordinator.start_background_listener()` / `_listener_loop()` — a persistent daemon thread per worker that blocks on `pubsub.listen()` and applies events within milliseconds instead of waiting on the polling window. Self-healing on any connection drop (same backoff every other caller uses). A lock now guards `_process_message`'s method dispatch so the listener thread and a middleware-triggered `sync()` can't interleave a cache mutation.
- **`helm-chart/.../gunicorn.yaml`**: `post_fork` now reconnects the coordinator fresh per worker (unconditional — closes the fork-safety bug regardless of worker count) and optionally starts the listener thread, gated by a new `MULTISERVER_BACKGROUND_LISTENER_ENABLED` flag (off by default, wired end-to-end through `values.yaml` → `local-settings.yaml` → `local-settings-file.yaml` → `settings.py`) so it can be canaried per-environment without a code change.
- `MultiServerEventListenerMiddleware` is left running unconditionally as a zero-cost fallback: if the listener thread ever dies, a pod just reverts to today's every-20-request behavior instead of going stale forever with no signal.
- New tests: `sefaria/system/tests/multiserver_coordinator_test.py` (listener processes a message + reconnects on drop, backs off while disconnected, dispatch is serialized under the lock, `start_background_listener()` is idempotent).

## Out of scope (documented, not fixed here)

Also found during this investigation, tracked separately: missing `publish_event` on category-delete/topic-editor, a stuck lock in `admin/rebuild/shared-cache`, a per-process thundering-herd cooldown in `nonuniqueterm_index.py`, and RefResolver-rebuild task hardening (Celery `task_acks_late`/time limits, frontend poll timeout).

## Verification

- `pytest sefaria/system/tests/` — 62 passed, no regressions.
- `helm template` — all touched chart templates render and the generated `gunicorn.conf.py`/`local_settings.py` compile as valid Python.
- Manual verification still needed: canary `MULTISERVER_BACKGROUND_LISTENER_ENABLED=True` on one pod/environment and confirm a linker-editor edit propagates within milliseconds instead of the 20-request window (see plan verification section for the exact repro steps).


helm-chart/sefaria/templates/configmap/gunicorn.yaml
```diff
@@ -39,6 +39,23 @@ data:
         {{- if .Values.instrumentation.enabled }}
         from opentelemetry.instrumentation.auto_instrumentation import sitecustomize
         {{- end }}
+        # on_starting() (master, pre-fork) already called server_coordinator.connect() once via
```
nsantacruz: make less verbose. don't reference celery worker_process_init

helm-chart/sefaria/templates/configmap/local-settings-file.yaml
```diff
@@ -145,6 +145,7 @@ data:
 
     # Multiserver
     MULTISERVER_ENABLED = bool(os.getenv("MULTISERVER_ENABLED", default="True").lower() == "true")
+    MULTISERVER_BACKGROUND_LISTENER_ENABLED = bool(os.getenv("MULTISERVER_BACKGROUND_LISTENER_ENABLED", default="False").lower() == "true")
```
nsantacruz: default to true

helm-chart/sefaria/values.yaml
```diff
@@ -572,6 +572,11 @@ localSettings:
   DEFAULT_FROM_EMAIL: "Sefaria <hello@sefaria.org>"
   SERVER_EMAIL: "dev@sefaria.org"
   MULTISERVER_ENABLED: "True"
+  # Kill switch for the background multiserver listener thread on web pods (see
```
nsantacruz, line 575: change default to true. remove "Off by default so it
  can be canaried per-environment without a code change; MultiServerEventListenerMiddleware's
  every-20-request poll keeps working regardless of this flag."

sefaria/system/multiserver/coordinator.py
```diff
@@ -83,6 +92,58 @@ def sync(self):
         self._process_message(msg)
         self.sync()  # While there are still live messages, keep processing them.
 
+    def start_background_listener(self):
+        """
+        Start a persistent daemon thread that blocks on a dedicated multiserver pubsub
```
nsantacruz: make more succinct

sefaria/system/multiserver/coordinator.py
```diff
@@ -83,6 +92,58 @@ def sync(self):
...
+        if getattr(self, "_listener_thread", None) and self._listener_thread.is_alive():
+            return
+        self._listener_thread = threading.Thread(
+            target=self._listener_loop, name="multiserver-listener", daemon=True
+        )
+        self._listener_thread.start()
+
+    def _listener_loop(self):
+        """
+        Blocking loop: waits on the dedicated listener pubsub connection and applies each event
```
nsantacruz: make more succinct

sefaria/settings.py
```diff
@@ -359,6 +359,11 @@ def get_static_url():
 SEARCH_INDEX_NAME_BOOK = 'book'
 SEARCH_INDEX_NAME_CATEGORY = 'category'
 
+# Same reasoning as the SEARCH_INDEX_NAME_* defaults above: a local_settings.py that predates
+# this flag must still boot, and MULTISERVER_ENABLED-gated code (e.g. gunicorn's post_fork hook)
+# imports this by name. See sefaria/system/multiserver/coordinator.py:start_background_listener.
+MULTISERVER_BACKGROUND_LISTENER_ENABLED = False
```
nsantacruz: why is this in setting.py? Shouldn't it only be in local_settings?

sefaria/system/multiserver/messaging.py
```diff
@@ -40,8 +37,57 @@ def connect(self):
             self.redis_client = None
             self.pubsub = None
 
-    def _pop_subscription_msg(self):
-        m = self.pubsub.get_message()
+    def connect_listener(self):
+        """
+        (Re)build a second, independent connection dedicated to a long-lived blocking
```
nsantacruz: make much more succinct

akiva10b (commented):
One malformed message drops a healthy connection for 30s

In _process_message, obj = locals()[data["obj"]] and method = getattr(obj, data["method"]) sit outside the try/except. An unknown obj/method name — a mismatched app version publishing, say — raises KeyError/AttributeError straight up into _listener_loop's except Exception, which nulls out _listener_redis_client/_listener_pubsub and sleeps RECONNECT_BACKOFF_SECONDS. A message-shaped problem tears down a perfectly good socket and blinds the pod for 30 seconds. Those two lines belong inside the try.

PR #3723, opened 2026-09-15
fix(search): route sheet indexing through the admin index queue; drain the queue; fix reindex retry

yodem:
## Why
A published sheet (id 747331, created 2026-09-08) never showed up in Voices sheet search. It turned out search hasn't picked up **any** new or edited sheet since the 2026-09-04 weekly rebuild. I traced it in prod.

### Root cause (verified in prod)
Web pods authenticate to Elasticsearch as **`sefaria-viewer`, role `viewer`**. From a web pod, `_security/user/_has_privileges` returns:
```
sheet-b: read=true  create_doc=false  index=false  write=false
text-b:  read=true  create_doc=false  index=false  write=false
```
So every on-save `index_sheet` / `delete_sheet` from the web path has been rejected and swallowed. The rejection was caught in `index_sheet`, or by a bare `except:` in `save_sheet`. Sheets only ever reached search through the **weekly full rebuild**, which runs with `elastic-admin`. When that rebuild failed on 2026-09-11 (see the incident below), search stopped updating entirely.

Text edits never had this problem. The web pods queue them in `index_queue`, and the `index-from-queue` CronJob writes them to ES with `elastic-admin` credentials. Sheets never used that path.

## Changes
1. **Sheets go through the index queue** (`sefaria/search.py`, `sheets.py`, `views.py`, `utils/user.py`, `model/queue.py`)
   - Web code calls `add_sheets_to_index_queue(ids)`, one bulk `$setOnInsert` upsert per sheet that also bumps a `generation` token. The CronJob calls `sync_sheet_in_index`: it indexes the sheet if it is public and has an owner, and deletes the doc otherwise.
   - The decision uses the sheet's state when the job runs. So a publish followed by an unpublish ends up in the right state, and the queue's unique `(lang, version, ref)` index collapses repeat saves into one record. The consumer deletes a record only if its `generation` is unchanged, so a save that lands mid-sync is not lost.
   - Covers publish, edit, unpublish, `delete_sheet_by_id`, spam-dashboard deletes, spammer quarantine (`purge_spammer_account_data` unlists all of a spammer's sheets), and account deletion. The last three never removed sheets from search before.
   - `clear_index_queue()` at reindex finalize now keeps sheet records, so saves queued after the catch-up query survive.
   - Web keeps read-only ES access, so no credentials change. Search updates arrive within one CronJob cycle (every 7 minutes) instead of instantly.
2. **The queue actually drains, in bounded batches.** Records are fetched 100 at a time with a fresh query per batch, so a 25k-record drain can't hit Mongo's 10-minute idle-cursor timeout, and a failing record isn't refetched in a loop.
   Previously: `index_from_queue` called `Collection.remove`, which **pymongo 4 removed** (we upgraded in afd8142f8, 2025-05). It raised after every successful write, so no record ever left the queue and every run redid the whole backlog. It now uses `delete_one`, and keeps a record only when its write failed.
3. **`index_sheet`: `create` → `index`.** `create` fails with a 409 when the doc already exists, so edits to an indexed sheet were dropped.
4. **No more auto-created indices.** `index_from_queue` looks up the live index per item, so a run that spans an alias swap doesn't write to the name finalize just deleted. `reindex_init` also refuses to reuse an existing index that lacks our analysis settings, and recreates it instead. Prod evidence: `text-a` has a **dynamic mapping** (no analyzers, `path` as `text`) and was created 2026-09-06T02:43:13Z, seconds after the Sep 4 finalize deleted the old `text-a`. The Sep 11 run reused it and grew it to 39 GB on one node. (This corrects commit 6c406c72f's message, which wrongly retracted the auto-create explanation.)
5. **Orchestrator retry.** The stale-job poll used `read_namespaced_job`, but the Role grants `get` only on `jobs/status`. Every retry after a failed run died with a 403 within seconds (2026-09-11, attempt 2). It now polls with `read_namespaced_job_status`, and no RBAC change is needed.
6. **Errors are reported, not swallowed.** The Mongo lookups in `index_sheet` moved inside its `try`, failures log at error level with the exception, and `delete_sheet` returns success or failure so the queue can retry.

## Separate prod incident (not fixable in code, needs ops)
- **2026-09-12 02:17 UTC:** `elasticsearch-1b-es-default-2`'s data disk hit **40G/40G (100%)** during the Sep 11 reindex. It hit the flood-stage watermark, lost mastership and left the cluster.
- **Now:** `es-0` (master) and `es-1` run as a 2-node cluster at 37% disk. It is **red** only because shard 0 of `text-a` (the auto-created, dynamically mapped index, 39.18 GB) had its sole copy on es-2. Volume fix: Sefaria/infrastructure#680. `text-b` and `sheet-b` are fully allocated on es-0 and es-1.
- **Why clients still fail:** es-2 is still a **ready** endpoint of `elasticsearch-1b-es-http`, because ECK's readiness probe doesn't check cluster membership. Any client whose keep-alive connection lands on es-2 gets `503 no master` on writes. The `index-from-queue` job has been stuck on it for 3 days.

## Tests
- New tests use a pymongo-4-shaped fake collection with no `remove`. A `MagicMock` is how the drained-queue bug stayed hidden.
- Also new: sheet sync (index/delete/retry), `delete_sheet` result, index-sheet lookup failure, orchestrator stale-job helper, and a static guard that web modules never call `index_sheet`/`delete_sheet`.
- All new behavior tests **fail on master** and pass here.
- `search_test.py`, `reindex_orchestrator_test.py`, `reindex_pipeline_test.py` and `tracker_test.py`: **127 passed**.
- **Codex review** (read-only) found 3 major issues: a lost update when a save lands mid-sync, `clear_index_queue` dropping sheet records, and the spammer quarantine leaking into search. All are fixed in 0713f1376, each with a test that fails before the fix.
- **Final Codex pre-merge review:** one major finding, the long-lived queue cursor timing out during the backlog drain. Fixed in 1b85dcfca with a regression test. **128 passed** locally.
- `reader/tests.py::PagesTest::test_new_sheet` couldn't run locally (the worktree has no webpack build). It renders a page and doesn't touch this path; CI covers it.

## Rollout notes
- **Existing backlog:** prod `index_queue` currently holds **25,441** text records (23,206 from the new *Tanaj Español* import) that never drained. The first post-deploy run processes all of them. Fix ES first (below), or every write stuck on es-2 waits 60s × 4 retries.
- **Missed sheets:** **434** public sheets were modified after the last good rebuild, and they are not in the queue. Backfill them once with `index_sheets_by_timestamp("2026-09-04T18:00")` using admin credentials. Pod clocks and sheet timestamps are UTC-3.
- **Enqueue mechanics:** repeat saves while a sheet is already queued are a no-op, because enqueueing is an upsert on the queue's unique key. Spam and account purges enqueue all their sheets in one bulk write.
- **Remaining web writes:** the index/version rename cascades in `sefaria/helper/search.py` / `model/dependencies.py` also write from web with viewer credentials. They're rare admin operations; follow-up.


sefaria/views.py
```diff
@@ -1392,13 +1392,13 @@ def delete_sheet_by_id(request):
...
-                search.delete_sheet(es_index_name, id)
-            except NewConnectionError as e:
-                logger.warn("Failed to connect to elastic search server on sheet delete.")
-            except AuthorizationException as e:
-                logger.warn("Failed to connect to elastic search server on sheet delete.")
+            # Queued rather than deleted directly: web pods only have read access to ES. The
+            # sheet is already gone from Mongo, so the queue consumer removes its search doc.
+            if SEARCH_INDEX_ON_SAVE:
+                try:
+                    search.add_sheet_to_index_queue(id)
```
stevekaplan123: I wonder if there could be a better name for the function `add_sheet_to_index_queue` , since we're actually going to delete the sheet from ElasticSearch, not add it.

sefaria/search.py
```diff
@@ -2534,13 +2644,24 @@ def reindex_init(type, debug=False):
...
-        if doc_count > 0:
+        managed = _index_has_managed_settings(new_index)
+        if doc_count > 0 and managed:
             logger.info(
                 f"reindex_init reusing in-progress index - type: {type}, new_index: {new_index}, doc_count: {doc_count}"
             )
             set_index_bulk_load_settings(new_index)
             return names
-        logger.info(f"reindex_init recreating empty index - type: {type}, new_index: {new_index}")
+        if doc_count > 0:
```
stevekaplan123, line 2652: It's confusing to me that `doc_count > 0` is checked twice.  Seems like the if conditionals could be structured differently

sefaria/search.py
```diff
@@ -146,10 +147,16 @@ def delete_version(index, version, lang, old_title=None):
...
+    """Remove sheet `id` from the index. Returns True when the doc is gone (including
+    when it was never indexed), False when the delete failed and should be retried."""
     try:
         es_client.delete(index=index_name, id=id)
+        return True
+    except NotFoundError:
+        return True
     except Exception as e:
-        logger.error(f"deleting sheet {id}")
+        logger.error(f"Failed to delete sheet {id} from {index_name}: {type(e).__name__}: {e}")
```
stevekaplan123, line 163: What is NotFoundError?  Does it mean that there is no sheet?  If so, if the sheet can't be deleted because it doesn't exist, shouldn't that be logged?

scripts/scheduled/reindex_orchestrator.py
```diff
@@ -256,6 +256,38 @@ def run_barrier_loop(
         sleep_fn(poll_interval_seconds)
 
 
+def delete_stale_shard_job(batch, job_name, namespace, api_exception_cls,
+                           attempts=30, sleep_fn=time.sleep, log_fn=None):
+    """Delete a prior shard Job and wait until it is actually gone.
+
+    Returns True once the Job is confirmed absent (or never existed), False if it still
```
stevekaplan123, line 263: Why return true if the job never existed?  Doesn't that indicate that we're doing something wrong trying to delete something that's never existed?

stevekaplan123 (changes requested):

sefaria/views.py
stevekaplan123: Maybe process_sheet_in_index_queue?  Or is that also misleading?

yodem: Good call, renamed in cb2319001: `add_sheet_to_index_queue` → `queue_sheet_sync`, and `add_sheets_to_index_queue` → `queue_sheets_sync`. I didn't go with `process_sheet_in_index_queue` because "process" reads like the consumer side (`index_from_queue`), and this function only records that a sheet needs syncing. Whether it gets indexed or deleted is decided later from the sheet's state at that point. Happy to rename again if you prefer something else.

sefaria/search.py
yodem: Agreed, restructured in cb2319001. There's now one `doc_count > 0` branch: if the index has docs and our analysis settings, reuse it. If it has docs without them (auto-created), log an error and fall through. The empty case logs and falls through too, and both go to a single `create_index(..., force=True)`. No behavior change.

sefaria/search.py
yodem: It means Elasticsearch has no document with that id in that index. For us that's a sheet that was never indexed or was already removed. Since the goal (the doc being gone) already holds, it still returns True so the queue doesn't retry forever. You're right that it shouldn't be silent, though: in cb2319001 it now logs `Sheet {id} had no document in {index}; nothing to delete`, and the docstring explains it.

scripts/scheduled/reindex_orchestrator.py
yodem: A 404 here is the normal case, not a mistake. The shard Job has `ttlSecondsAfterFinished: 86400`, so Kubernetes usually deletes it before the next weekly run. The orchestrator deletes unconditionally before recreating instead of checking first, which also avoids a check-then-delete race. The True means "no stale Job is left to block the create". In cb2319001 that path logs `No stale shard job to delete; it never existed, or ttlSecondsAfterFinished already removed it`, and the docstring says so explicitly.

stevekaplan123 (approved):

